PR #65, opened 2021-05-20
Add headless embeddable mode

### LINKED ISSUE #63: Headless version for embedding
Make an embeddable version of the frontend with the following changes:

- [x] When the url contains a particular query parameter:
	- [x] Remove logo (for now)
	- [x] Reduce header font size, padding, and remove all caps
	- [x] Remove footer

obulat:
Fixes #63 
This PR adds headless embeddable mode to the application so that we can embed it into an iframe. 

### Embedded mode

The default mode `embedded` mode has smaller header without logo, and no footer. You can also  explicitly set the `embedded` mode by using a `embedded=true` parameter in the URL http://localhost:8443/?embedded=true . A router middleware checks for this parameter, and saves any changes to the vuex store. 

Screen Shot 2021-05-20 at 1 41 38 PM

---

### Standalone mode

The standalone mode, with a larger header, logo and footer, can be activated through the `embedded=false` parameter in the URL: http://localhost:8443/?embedded=false when running locally.

Screen Shot 2021-05-20 at 1 44 26 PM


zackkrida (changes requested):
I like the middleware + vuex approach, works well with the existing data model. 

Can you switch the h1 in the hero in embedded mode to just say "Openverse"?  I feel like the existing text is redundant anyway 😆

dhruvkb (commented):
We discussed this on Slack but I just wanted to formalize it as a comment. When running in embedded mode, a [message should be posted](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) to the `parentWindow` containing the page height so that the embedder can set the proper height on the `iframe`.

A simple object containing a height property should be sufficient for the `message` argument in `postMessage`.

PR #69, opened 2021-05-26
Update to new Google Forms

### LINKED ISSUE #66: Make sure all reporting forms work
We need to make sure all content reporting forms work. If we need to create new google forms or anything, please feel free to do so and just make sure my a8c email address has ownership access to the google sheets/forms/whatever.

dhruvkb:
Fixes #66.

This PR updates the the URLs to the Google forms.

### DIFF
--- a/src/components/ImageDetails/ReuseSurvey.vue
+++ b/src/components/ImageDetails/ReuseSurvey.vue
@@ -20,6 +20,10 @@ import {
   DETAIL_PAGE_EVENTS,
 } from '~/store-modules/usage-data-analytics-types'
 
+const reuseForm =
+  'https://docs.google.com/forms/d/e/1FAIpQLSegPUDIUj_odzclJhhWRfPumSfbHtXDVDCHqRfFl7ZS8cMn2g/viewform'
+const imageLinkEntry = '2039681354'
+
 export default {
   name: 'ReuseSurvey',
   props: ['image'],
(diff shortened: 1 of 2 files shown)


src/components/ImageDetails/ReuseSurvey.vue
```diff
@@ -29,7 +33,7 @@ export default {
   computed: {
     formLink() {
       const location = this.location
-      return `https://docs.google.com/forms/d/e/1FAIpQLSeSApxNMup8Ujt-8Vjv53ngltzhJeaHspMykHCD8VKQ39yXAA/viewform?usp=pp_url&entry.1690035721=${location}`
+      return `${reuseForm}?entry.${imageLinkEntry}=${location}`
```
obulat, line 36: What was the `usp=pp_url` used for? I assume it's not necessary anymore.

obulat (approved):
Great attention to details !

src/components/ImageDetails/ReuseSurvey.vue
dhruvkb: It seems to me that the `usp` query param usually indicates [where the request came from](https://stackoverflow.com/a/41904286). In that regard, it is very similar to UTM tags. 

Anyway, it didn't have any effect on the form so I dropped it.

krysal (approved):

PR #58, opened 2021-05-07
Update frontend copy

dhruvkb:
This PR
- removes deprecated pages (Support & Feedback) and their localizations
- adds details about the CC &rarr; WP migration
- adds a missing full-stop

PR #73, opened 2021-05-27
Restore Feedback page

### LINKED ISSUE #72: Bring back the `/feedback` page
I think collecting feedback from users is important, and now that we have new forms created we should use the same procedure to bring back the feedback page, currently live at https://search.creativecommons.org/feedback

As @dhruvkb did previously we'll want to rename any CC ~> Openverse in the forms.

dhruvkb:
Fixes #72.

This PR restores the Feedback page that was removed in #58, with new feedback forms and de-duplicated code.

### DIFF
--- a/src/components/EmbeddedNavSection.vue
+++ b/src/components/EmbeddedNavSection.vue
@@ -75,6 +75,9 @@
             <NuxtLink class="navbar-item" :to="localePath('/meta-search')">
               {{ $t('header.meta-search') }}
             </NuxtLink>
+            <NuxtLink class="navbar-item" :to="localePath('/feedback')">
+              {{ $t('header.feedback') }}
+            </NuxtLink>
             <a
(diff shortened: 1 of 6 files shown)


dhruvkb:
Suggestions:
image

Bugs:
image

panchovm (approved):

krysal (commented):
This looks good! It's only missing the menu entry in the embedded mode. In the `EmbeddedNavSection` component I think.

obulat:
I've added the changes for the EmbeddedNavSection and the embedded page to keep it the same as other embedded mode pages.

obulat (approved):

krysal (approved):

zackkrida:
So much collaboration haha! Glad you all worked on this together 😊

zackkrida (approved):
Some nice small fixes in here too. Thanks @dhruvkb

PR #74, opened 2021-06-01
Replace node-sass with dart-sass

### LINKED ISSUE #70: Switch to Dart Sass from the deprecated node-sass
Most of our styles are written in Sass. The `node-sass` package that is used to compile them into CSS is deprecated, so we should switch to Dart Sass implementation.
There is an official migrator script that helps to make the necessary changes: https://sass-lang.com/documentation/cli/migrator .

obulat:
Fixes #70 

This PR replaces the deprecated node-sass with the current implementation of sass: Dart SASS. 

There have been several deprecation notices that were fixed, mainly:
- the slash in Dart Sass is used only for path operations, and the mathematical division now uses `math.div()`
- Dart Sass no longer uses `~` tilda symbol for imports. The import paths have been replaced with relative imports.


dhruvkb (approved):
LGTM! I was going to recommend installing `fibers` but I discovered that it has been obsoleted and that the Sass team is working to [remove that dependency](https://github.com/sass/dart-sass/issues/868#issuecomment-821798612).

zackkrida (approved):
Cool, I didn't know about `math.div()`

PR #81, opened 2021-06-23
Add `urlChange` message and unify message format

obulat:
This PR improves the communication of the app with outer window in embedded mode.

When used in embedded mode inside an iframe, the app needs to set correct iframe height and display the correct URL in the address bar. This is achieved using messages sent from the app to the outer window. This PR specifically adds the message on every URL change.

Currently, there are two kinds of messages, which have unified format:
1. URL change, which sends the relative URL eg. `{type: 'urlChange', value: '/about' `}
2. Resize, which sends the height that iframe requires, in px, eg. `{type: 'resize', value: 1000 }`

### DIFF
--- a/src/utils/sendMessage.js
+++ b/src/utils/sendMessage.js
@@ -11,6 +11,7 @@ const TARGET_ORIGIN = '*'
 export const sendWindowMessage = (message) => {
   window.parent.postMessage(
     {
+      debug: config.dev,
       type: message.type,
       value: message.value,
     },
--- a/src/utils/sendMessage.js
+++ b/src/utils/sendMessage.js
@@ -1,24 +1,16 @@
 // TODO: set correct targetOrigin of the parent window
 const TARGET_ORIGIN = '*'
-const config = {
-  dev: false,
-}
+
 /**
  * Send message to the outer window. Can only be called on client-side
  * because it uses `window` object.
  * @typedef {Object} message
+ * @property {boolean} debug - whether more verbose debug output should be used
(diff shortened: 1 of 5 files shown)


src/utils/sendMessage.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Send message to the outer window. Can only be called on client-side
+ * because it uses `window` object.
+ * @typedef {Object} message
+ * @property {'resize'|'changeUrl'} type - event that triggers the message
+ * @property {Object} value - the value of event
+ */
+export const sendWindowMessage = (message) => {
+  window.parent.postMessage(
```
dhruvkb, line 12: I think we should add a check to see if it's running in an `iframe`:
```js
if (window.parent !== window) {
  ...
}
```

src/utils/sendMessage.js
```diff
@@ -0,0 +1,19 @@
...
+    {
+      type: message.type,
+      value: message.value,
+    },
```
dhruvkb, line 16: Can add another property `debug` for verbose logging on the Openverse theme side.
```suggestion
    {
      debug: config.dev,
      type: message.type,
      value: message.value,
    },
```

dhruvkb (dismissed):
Looks good to me. There can be some minor improvements.

dhruvkb (approved):

zackkrida (approved):
So cool. I think the '*' target origin is fine—we aren't dealing with many security concerns here.

PR #86, opened 2021-06-30
Implement transitional styles 

### LINKED ISSUE #88: Implement the transitional style into mockups
## Description
For this first release, we need to implement the transitional style in all pages to match WordPress's style. This mockup set includes the following tickets:

- #34 
- #87 
- #85 

### Mockups (transitional style)
All mockups are available [in this Figma file](https://www.figma.com/file/BlUTIXRHLOq5mRITpDpi3E/Transitional-styles?node-id=0%3A1&viewport=74%2C344%2C0.09312169253826141).

### Brand symbol
The brand symbol is available as both standalone and with openverse name in [this Figma file](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-brand?node-id=1%3A8&viewport=468%2C208%2C1.7461538314819336).

## Additional context
The style on these mockups does not have the final brand definitions, except for the brand symbol. Other brand elements will be added in the audio integration release (#76 )

zackkrida:
Fixes #88 

> Preview deployment: https://openverse.zack.cat/

This implements the transitional styles for the Openverse launch. When implementing these, it became very clear to me that the spacing and font sizes are already horribly out of sync with the designs. We will fix this when we redesign the app entirely. 

Here, I'd like the goal to be: "Is this good enough for going live?" I hope the answer is yes 😄 

- Add new Openverse logo
- New fonts + font styles
- Introduce new blue and gray colors; add borders
- Switch to new style for Meta Search (centered, under image results instead of in modal)
- Font size, color changes, and spacing improvements


dhruvkb (dismissed):
Loving the new look.

zackkrida:
The only real work left is to adjust font sizes and weights where needed. I love these new designs so much.

src/components/EmbeddedNavSection.vue
```diff
@@ -1,5 +1,12 @@
 <template>
   <nav :aria-label="$t('header.aria.primary')" class="navbar embedded">
+    <NuxtLink to="/" style="align-self: center; line-height: 0">
+      <img
+        alt="Openverse logo mark"
+        src="~/assets/logo.svg?data"
+        style="height: 24px; padding-right: 24px; width: auto"
+      />
```
obulat, line 8:
```suggestion
     <img
        alt="Openverse logo mark"
        src="~/assets/logo.svg?data"
        style="padding-right: 24px"
        width="160"
        height="24"
     />
```

obulat: Setting width and height for images is supposed to prevent layout shifts. The logo actually doesn't cause any, but still...

src/components/NavSection.vue
```diff
@@ -2,7 +2,13 @@
   <nav :aria-label="$t('header.aria.primary')" class="navbar">
     <div class="navbar-brand has-color-white">
       <!-- eslint-disable @intlify/vue-i18n/no-raw-text -->
-      <NuxtLink class="logo" :to="localePath('/')">Openverse</NuxtLink>
+      <NuxtLink class="logo" :to="localePath('/')">
+        <img
+          alt="Openverse logo mark"
+          src="~/assets/logo.svg?data"
+          style="height: 24px; padding-right: 24px; width: auto"
+        />
```
obulat, line 10:
```suggestion
        <img
          alt="Openverse logo mark"
          src="~/assets/logo.svg?data"
          style="padding-right: 24px"
          width="160"
          height="24"
        />
```

src/components/SearchTypeTabs.vue
```diff
@@ -1,23 +1,21 @@
...
-          :aria-controls="'tab-' + type"
-          role="tab"
-          :class="tabClass(type, 'tab')"
-        >
-          {{ capitalize(type) }}
-        </NuxtLink>
-      </div>
-    </section>
-  </div>
+  <section class="tabs">
```
obulat, line 2: So nice to get rid of extra divs!

src/styles/components/navbar.sass
```diff
@@ -32,21 +32,29 @@ $navbar-breakpoint: $tablet
...
-      max-width: $wp-max-width
-      margin-right: auto
-      margin-left: auto
+    max-width: $wp-max-width
+    margin-right: auto
+    margin-left: auto
 
 .navbar-brand
   align-items: stretch
+  justify-content: space
```
obulat, line 57: Typo. Should it be `space-around` or `space-between`?

obulat (commented):
Such exciting changes!!! 🎉 

I've added some nitpicking comments inline :) And the hamburger menu seems broken: 
Screen Shot 2021-07-02 at 9 14 29 AM

src/styles/components/navbar.sass
zackkrida: apparently it isn't even necessary, so i'll simply remove the rule!

src/assets/fonts.css
```diff
@@ -1,74 +1,49 @@
...
-    url('~assets/fonts/JetBrainsMono-Bold.woff2') format('woff2'),
-    url('~assets/fonts/JetBrainsMono-Bold.woff') format('woff'),
-    url('~assets/fonts/JetBrainsMono-Bold.otf') format('opentype');
-}
-@font-face {
-  font-family: 'Source Sans Pro';
+  font-family: 'Inter';
   font-style: normal;
-  font-weight: 400;
+  font-weight: 500;
```
dhruvkb, line 4: @panchovm this looks too bold to me. I thought 400 is usually the baseline for body text?

dhruvkb (approved):
This looks quite good (except for the bolder body weight)!

obulat (approved):
The site looks really nice, clean and beautiful!
I've added two tiny changes in a separate PR, but other than that this is wonderful 💯

src/assets/fonts.css
panchovm: I noticed that we have 400 and 500 for `.has-text-weight-medium` and `.has-text-weight-normal` classes. We should have different weights.

dhruvkb: Based on [this reference from MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping), normal should be 400 and medium should be 500.

panchovm: I think I missed the point. Let me start again: I agree with `400` as normal for default body text, and `500` as medium for `.has-text-weight-medium` text. Are you agree with that too?

PR #91, opened 2021-07-13
Make filter display vertically centered

obulat:
Currently, the search filter tags are aligned to top vertically.
Screen Shot 2021-07-13 at 12 00 48 PM

However, in the design mockups they are center-aligned. This PR centers the filter tags vertically, and removes unnecessary spans from around the tag buttons:

Screen Shot 2021-07-13 at 11 56 12 AM

### DIFF
--- a/src/components/Filters/FilterBlock.vue
+++ b/src/components/Filters/FilterBlock.vue
@@ -1,7 +1,7 @@
 <template>
   <button
     v-if="filterType === 'searchBy'"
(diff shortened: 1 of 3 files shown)


krysal (approved):
Looks good, the truly centered elements and the reduction of extras (`<span>`) is appreciated!

dhruvkb (approved):
Looks much better 👍

PR #94, opened 2021-07-15
Make store the only source of truth for filter data

### LINKED ISSUE #93: [Bug] Filters from URL are not saved to store
## Description

Currently, when you open a URL with a query that contains filters (ie. `openverse/search/?q=cat&license=cc0&license_type=commercial&source=animaldiversity` the search uses those filters, but the filter data is not saved to the store. This means that the filter list does not show those filters as applied.

## Reproduction

1. Go to `openverse/search/?q=cat&license=cc0&license_type=commercial&source=animaldiversity`
2. See error.

## Expectation

The filters from URL should be saved to the vuex store so that they are available for later navigation, and show the same values in Filter display (tags) and Filter list (left sidebar or popup on mobile)

## Environment

 - Device: 
 - OS: 
 - Browser: 
 - Version: 
 - Other info: 

## Resolution

obulat:
Fixes #93 

Ensure that the filter data from URL query is saved in the store.
Simplify filter display.

### DIFF
--- a/src/components/Filters/FilterBlock.vue
+++ b/src/components/Filters/FilterBlock.vue
@@ -1,25 +1,9 @@
 <template>
   <button
-    v-if="filterType === 'searchBy'"
-    class="filter-block button tiny tag"
-    :aria-label="label + 'filter'"
-  >
-    <span>{{ $props.label }}</span>
-    <span
-      :aria-label="$t('browse-page.aria.remove-filter')"
-      class="close margin-left-small"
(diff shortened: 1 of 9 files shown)


krysal (approved):
This is excellent! Fixes some extra problems also ⭐️

zackkrida (approved):
Excellent

zackkrida:
I'm going to merge this in light of @obulat's AFK.

PR #95, opened 2021-07-16
Change target branch to run CI tests

krysal:
Related to WordPress/Openverse#32.

The `ci` workflow was targeting the old `master` and `develop` branches, this PR updates it to the `main` branch in order to restore tests execution for each new pull request.

Also, it seems to me that the sentry workflow currently is not being used so I moved it to the `workflows-disabled` folder.


dhruvkb (approved):
Yay! CI!

PR #98, opened 2021-07-27
Install and scaffold Tailwind

dhruvkb:
This PR installs Tailwind, and configures it based on the tokens of the project.


zackkrida (approved):
Shame jit mode isn't working. Probably related to postcss and sass compilation? Not sure, but in any case this works well and the purge rules seem correct. Nice work

dhruvkb:
Yeah, my experience JIT with PostCSS 7 was quite buggy as well. Maybe the next major upgrade with Vue 3 will allow us to turn on JIT 🤞.

zackkrida (approved):

krysal (approved):

PR #96, opened 2021-07-22
Create a page for the web extension

dhruvkb:
Fixes #87.

This PR creates the extension page inside the Openverse frontend.

PR #99, opened 2021-07-28
Replace Bulma helpers with Tailwind utilities

### LINKED ISSUE #97: [Feature] Replace utility classes with Tailwind
## Problem
We have a bunch of custom utility classes defined in the code that are
- self-developed, so non-standard prefixes etc.
- inadequately documented, so it's a lot of guesswork
- not comprehensive, so leaving more to be desired

## Description
These self developed classes can be replaced with [Tailwind](https://tailwindcss.com) to give us a more comprehensive set of utility classes with documentation while allowing us to remove a ton of code.

The Tailwind configuration must be consistent with our design system (both the current system in place as well as Figma).

## Implementation

dhruvkb:
Fixes #97

This PR replaces [Bulma helpers](https://bulma.io/documentation/helpers/) with their Tailwind counterparts. After confirming that none of Bulma's helpers are in use, we can chuck the directory altogether.


dhruvkb:
There could be some variations due to specificity issues (Bulma's deeply nested classes in components mostly win over Tailwind 😢) but the PR seems to be in a good place. The specificity issues will likely be a recurring issue during the framework transition stage.

dhruvkb:
This PR will need updating when #96 is merged.

dhruvkb:
This is looking pretty good now, open for reviews!

zackkrida:
Awesome, I'm excited to explore the code here

zackkrida (approved):
This is really great, and actually makes me think we could have the old .scss removed _before_ we implement the finalized UI. That would actually be quite nice, as it would give us more of a 'clean slate' before heavily modifying and creating new components.

dhruvkb:
I agree, it'll be awesome to be able to create the new components purely in Tailwind.

Issue #105, opened 2021-07-30
[Audio] Sidebar filters


zackkrida:
Conditionally show relevant audio filters in the sidebar when switching to the `/search/audio` view.
We need to determine the full list of filters here:

- Use
- Licenses
- Category (music / sound effect / podcast)
- Source
- Genre (maybe?, via text input or something? too many to list in the sidebar)

We may want to choose a subset to display in our initial release, rather than all of our ideal items

PR #107, opened 2021-08-02
Filter refactor

obulat:
Related to #105 

This PR refactors `FilterChecklist` to remove unnecessary repetition.
It also adds `filter-store` getters for applied filter tags so that the UI gets a list of unified objects for the filter (instead of some filter types that are arrays, some objects and some strings.

This will prepare the code base for adding the media type  specific filters.


zackkrida (approved):
Thank you for these significant improvements, also smaller improvements to html semantics and code quality.

krysal (approved):
Great improvement!

PR #110, opened 2021-08-02
Add a PR template to the repository

dhruvkb:
Targets WordPress/openverse#35

This PR adds a PR template to the repository.


obulat (approved):

zackkrida (approved):
I agree with @krysal that I don't think we need the "or delete this section entirely." language (on any of the comments IMO) but this is great!

dhruvkb:
Those comments are generally to help people without prior experience with contribution. We came across lots of people at CC who didn't know what to put in PR descriptions.

PR #117, opened 2021-08-10
Audio filters

obulat:
## Fixes

Fixes #105 by @zackkrida

## Description

This PR further refactors and simplifies handling of Filters, adds Audio to the side filters, and makes it easier to add other media type filters in the future.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [ ] My pull request targets the *default* branch of the repository (`main` or `master`).
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #120, opened 2021-08-11
Add correct meta video filters to the new filter method

zackkrida:
Related to #105, builds on work in #117

This PR adds the correctly-displayed filters for video, as a way of testing the new filter style PR. 
I also removed some unnecessary spreads, and fixed a bug in the approach for removing the 'mature' filter in the `getMediaTypeFilters` function, which was incorrectly removing an array item if the mature filter wasn't present.

### DIFF
--- a/src/constants/media.js
+++ b/src/constants/media.js
@@ -3,6 +3,8 @@ export const IMAGE = 'image'
 export const VIDEO = 'video'
 export const ALL_MEDIA = 'all'
 
+// Media types
 export const mediaTypes = [AUDIO, IMAGE]
-export const supportedMediaTypes = [AUDIO, IMAGE]
+// Media types which support custom filters
+export const supportedMediaTypes = [VIDEO, IMAGE, VIDEO]
 export const allMediaTypes = [ALL_MEDIA, IMAGE, AUDIO, VIDEO]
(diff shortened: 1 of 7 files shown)


src/store-modules/filter-store.js
```diff
@@ -244,9 +247,12 @@ function setQuery(state) {
...
 
 function getMediaTypeFilters(state, { mediaType, includeMature = false }) {
-  let filterKeys = [...mediaFilterKeys[mediaType]]
+  // eslint-disable-next-line no-debugger
+  debugger
+
+  let filterKeys = mediaFilterKeys[mediaType]
   if (!includeMature) {
-    filterKeys.splice(filterKeys.indexOf('mature'), 1)
+    filterKeys = filterKeys.filter((filterKey) => filterKey !== 'mature')
```
zackkrida, line 255: this should explain the original bug: 

CleanShot 2021-08-11 at 16 41 39@2x

src/constants/media.js
```diff
@@ -3,6 +3,8 @@ export const IMAGE = 'image'
 export const VIDEO = 'video'
 export const ALL_MEDIA = 'all'
 
+// Media types
 export const mediaTypes = [AUDIO, IMAGE]
-export const supportedMediaTypes = [AUDIO, IMAGE]
+// Media types which support custom filters
+export const supportedMediaTypes = [VIDEO, IMAGE, VIDEO]
```
obulat, line 9:
```suggestion
export const supportedMediaTypes = [AUDIO, IMAGE, VIDEO]
```

PR #132, opened 2021-08-16
Fix link to an old issue

obulat:
Tiny change: an issue in the `creativecommons/cccatalog-frontend` was mistakenly changed to a non-existent issue in the `WordPress/openverse-frontend` repository. Changing it back. I scanned the repository, and it doesn't seem to have any other changes like this.


dhruvkb (approved):
Nice catch!

krysal (approved):

PR #133, opened 2021-08-17
Audio track + waveform A11y

zackkrida:
## Changes

- Adds the proper ARIA roles to the waveform to make it function as a slider
- Improves play/pause button a11y with a text label and hiding the icon from screen readers
- Adds left + right arrow key seeking + shift key modification (1s seek w/o modifier, 15s seek w/ shift)
- Adds play/pause toggling to the waveform
	- I logically wanted to apply this to the entire Audio Track, but doing so means double events on the play/pause button and makes pressing enter on the track title to navigate to the single view toggle the player state

## Demo vid

https://user-images.githubusercontent.com/6351754/130304731-40d18745-7fa9-4a42-9242-450c06346820.mp4

## Prior art

- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role
- https://deezer.io/towards-an-accessible-player-b38431bb5721
- https://blog.atrera.com/javascript/accessibility/2019/06/28/create-a-custom-audio-player-with-accessibility.html
- https://sarahmhigley.com/writing/playing-with-state/

### DIFF
--- a/src/components/AudioTrack/AudioTrack.vue
+++ b/src/components/AudioTrack/AudioTrack.vue
@@ -17,7 +17,7 @@
       {{ audio.category }}
     </div>
 
-    <div class="flex flex-row gap-2">
+    <div class="interactive-section flex flex-row gap-2">
       <PlayPause
         v-if="isCompact"
         class="flex-shrink-0"
@@ -26,7 +26,6 @@
         @toggle="setPlayerState"
       />
       <div
-        class="waveform-section"
         @keypress.enter="setPlayerState(!isPlaying)"
         @keypress.space="setPlayerState(!isPlaying)"
       >
--- a/src/components/AudioTrack/AudioTrack.vue
+++ b/src/components/AudioTrack/AudioTrack.vue
@@ -4,7 +4,11 @@
     aria-label="$t('audio-track.aria-label')"
     role="region"
   >
-    <div class="waveform-section">
+    <div
+      class="waveform-section"
+      @keypress.enter="setPlayerState(!isPlaying)"
+      @keypress.space="setPlayerState(!isPlaying)"
+    >
       <Waveform
         class="h-30 w-full"
         :is-ready="isReady"
--- a/src/components/AudioTrack/AudioTrack.vue
+++ b/src/components/AudioTrack/AudioTrack.vue
@@ -1,6 +1,6 @@
 <template>
-  <div class="audio-track">
-    <div class="waveform-section">
+  <div class="audio-track" aria-label="Audio Player" role="region">
+    <div class="waveform-section bg-dark-charcoal-04">
       <Waveform
         class="h-30 w-full"
         :is-ready="isReady"
@@ -8,10 +8,13 @@
         :duration="duration"
         :peaks="audio.peaks"
         @seeked="setPosition"
+        @jumpSeekedForward="jumpPositionForward"
+        @jumpSeekedBackward="jumpPositionBackward"
       />
     </div>
     <div class="info-section flex flex-row gap-6">
       <PlayPause
+        :aria-controls="ariaIdentifier"
         class="self-start flex-shrink-0"
         :is-playing="isPlaying"
         :disabled="!isReady"
@@ -39,6 +42,7 @@
     <!-- eslint-disable vuejs-accessibility/media-has-caption -->
     <audio
       v-show="false"
+      :id="ariaIdentifier"
       ref="audio"
       controls
       :src="audio.url"
@@ -78,6 +82,8 @@ export default {
     player: null, // HTMLAudioElement
     currentTime: 0,
     duration: 0,
+    /** Amount in seconds to jump seek **/
+    seekJumpDuration: 15,
 
     isReady: false,
     isPlaying: false,
@@ -93,6 +99,13 @@ export default {
       const date = new Date(0)
       date.setSeconds(seconds)
       return date.toISOString().substr(11, 8).replace(/^00:/, '')
+    },
+    /**
+     * A unique string to represent the audio player in aria properties
+     * @returns {string} the id of the audio file prefixed with 'audio-'
+     */
+    ariaIdentifier() {
+      return `audio-${this.audio.id}`
     },
   },
   mounted() {
@@ -119,6 +132,14 @@ export default {
         this.player.currentTime = this.player.duration * percentage
         this.updateTime()
       }
+    },
+    jumpPositionForward() {
+      this.player.currentTime = this.player.currentTime + this.seekJumpDuration
+      this.updateTime()
+    },
+    jumpPositionBackward() {
+      this.player.currentTime = this.player.currentTime - this.seekJumpDuration
+      this.updateTime()
     },
     async setPlayerState(isPlaying) {
       if (isPlaying) {
--- a/src/components/AudioTrack/AudioTrack.vue
+++ b/src/components/AudioTrack/AudioTrack.vue
@@ -1,6 +1,10 @@
 <template>
-  <div class="audio-track" aria-label="Audio Player" role="region">
-    <div class="waveform-section bg-dark-charcoal-04">
+  <div
+    class="audio-track"
+    aria-label="$t('audio-track.aria-label')"
+    role="region"
+  >
+    <div class="waveform-section">
       <Waveform
         class="h-30 w-full"
         :is-ready="isReady"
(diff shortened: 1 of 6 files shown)


src/components/AudioTrack/AudioTrack.vue
```diff
@@ -1,17 +1,20 @@
 <template>
-  <div class="audio-track">
-    <div class="waveform-section">
+  <div class="audio-track" aria-label="Audio Player" role="region">
+    <div class="waveform-section bg-dark-charcoal-04">
```
dhruvkb, line 3: The `bg-charcoal-04` is not needed because it's already applied on the `Waveform` component.

src/components/AudioTrack/AudioTrack.vue
```diff
@@ -78,6 +82,8 @@ export default {
     player: null, // HTMLAudioElement
     currentTime: 0,
     duration: 0,
+    /** Amount in seconds to jump seek **/
+    seekJumpDuration: 15,
```
dhruvkb, line 86: I think this should be determined as a percentage of the track length, clamped between a minimum and maximum number of seconds. A lot of tracks from freesound are very short and I suppose audiobooks, if and when we add them, will be very long.

This is more of a UX question so maybe @panchovm can help us with the user perspective.

src/components/AudioTrack/Waveform.vue
```diff
@@ -203,6 +213,14 @@ export default {
...
       this.$emit('seeked', this.getPositionPercentage(event))
     },
+    seekJumpBackward() {
+      this.clearSeekProgress()
+      this.$emit('jumpSeekedBackward')
+    },
+    seekJumpForward() {
+      this.clearSeekProgress()
+      this.$emit('jumpSeekedForward')
+    },
```
dhruvkb, line 223: To maintain parity with mouse-based seek, maybe we could emit the same `seeked` event with the position percentage calculated in a different way?

Pseudocode:
```js
this.$emit('seeked', this.percentage + percentageDelta)
```

src/components/AudioTrack/AudioTrack.vue
```diff
@@ -94,6 +100,13 @@ export default {
       date.setSeconds(seconds)
       return date.toISOString().substr(11, 8).replace(/^00:/, '')
     },
+    /**
+     * A unique string to represent the audio player in aria properties
+     * @returns {string} the id of the audio file prefixed with 'audio-'
+     */
+    ariaIdentifier() {
+      return `audio-${this.audio.id}`
```
dhruvkb, line 108: Thoughts on using the `identifier` field on the `audio` object?

dhruvkb (approved):
This is incredible work on a11y (especially the slider role on the `Waveform`). I'm leaving some non-blocking suggestions.

dhruvkb:
@zackkrida I'm incredibly sorry I merged #137 without realising it would cause a bunch on conflicts here.

src/components/AudioTrack/AudioTrack.vue
sarayourfriend: Based on the behavior I've seen on other players, it should just jump one second when using the arrow keys. If we wanted to implement a special behavior for power users to seek over larger sections of time than one second we could add a "shift" modifier to accomplish that, but I think the default behavior should just be second by second (otherwise we're making it impossible for keyboard users to seek to a specific second).

src/components/AudioTrack/AudioTrack.vue
```diff
@@ -1,17 +1,20 @@
 <template>
-  <div class="audio-track">
-    <div class="waveform-section">
+  <div class="audio-track" aria-label="Audio Player" role="region">
```
sarayourfriend, line 2: Should the `aria-label` be translated?

src/components/AudioTrack/Waveform.vue
```diff
@@ -2,15 +2,25 @@
...
     ref="waveform"
     class="waveform relative bg-dark-charcoal-04 overflow-x-hidden"
+    tabIndex="0"
+    role="slider"
+    aria-label="audio seek bar"
+    aria-orientation="horizontal"
+    aria-valuemin="0"
+    :aria-valuemax="duration"
+    :aria-valuenow="currentTime"
+    :aria-valuetext="`${timeFmt(currentTime)} seconds`"
```
sarayourfriend, line 12: Ditto here with translation?

src/components/AudioTrack/AudioTrack.vue
panchovm: Agree with @sarayourfriend it sounds good exploring other options for power users and specific audio categories. I'm thinking of podcasts.

zackkrida:
> @zackkrida I'm incredibly sorry I merged #137 without realising it would cause a bunch on conflicts here.

i don't mind!

src/components/AudioTrack/AudioTrack.vue
zackkrida: I think let's do 1s, folks can hold down to do more, and we could do a shift modifier in the future that does a % of the track instead.

zackkrida: could also always add dedicated `<- 30s` `30s ->` buttons to the player ui in the future!

zackkrida: Currently set to 1s, 15s with modifier.

src/components/AudioTrack/AudioTrack.vue
```diff
@@ -13,28 +17,35 @@
       {{ audio.category }}
     </div>
 
-    <div class="waveform-section flex flex-row gap-2">
+    <div class="flex flex-row gap-2">
```
dhruvkb, line 20:
```suggestion
    <div class="interactive-section flex flex-row gap-2">
```

src/components/AudioTrack/AudioTrack.vue
```diff
@@ -13,28 +17,35 @@
...
-        :current-time="currentTime"
-        :duration="duration"
-        :peaks="audio.peaks"
-        :show-duration="isCompact"
-        @seeked="setPosition"
-      />
+      <div
+        class="waveform-section"
+        @keypress.enter="setPlayerState(!isPlaying)"
+        @keypress.space="setPlayerState(!isPlaying)"
```
dhruvkb, line 31: Why not add the `@keypress` events on `Waveform` directly? We could omit the extra div.

dhruvkb (approved):
Some more non-blocking suggestions:

src/components/AudioTrack/AudioTrack.vue
```diff
@@ -13,28 +17,35 @@
...
-        :class="isCompact ? 'h-20' : 'h-30'"
-        :is-ready="isReady"
-        :current-time="currentTime"
-        :duration="duration"
-        :peaks="audio.peaks"
-        :show-duration="isCompact"
-        @seeked="setPosition"
-      />
+      <div
+        class="waveform-section"
```
dhruvkb, line 29: This class should not be a 'section', as the section is the level of div above it

- `info-section`
- `interactive-section`
  - &lt;this div&gt;
    - Waveform
- `info-section`

dhruvkb (commented):
I missed this change.

src/components/AudioTrack/AudioTrack.vue
zackkrida: Vue doesn't pass native events down to custom components without the `.native` modifier, which is advised against and removed in Vue 3.

https://vuejs.org/v2/guide/components-custom-events.html#Binding-Native-Events-to-Components

zackkrida:
good feedback, thanks @dhruvkb

src/components/AudioTrack/AudioTrack.vue
dhruvkb: Thanks for the docs link and sorry for the incorrect suggestion!

dhruvkb (approved):

sarayourfriend (approved):
This LGTM so far. I'm unable to test using VoiceOver in Safari due to the example audio file being an `ogg` file. We should change it to something that has cross browser support instead so that we can test with VO and Safari (the most popular screen reader combo on macOS). Because of that I'm not able to confirm whether the `aria-value*` values are working as expected. We might need to do extra work to make sure that the correct values are announced, but I'm not sure.

src/components/AudioTrack/Waveform.vue
```diff
@@ -203,6 +233,16 @@ export default {
...
+    seekJump(event) {
+      this.clearSeekProgress()
+      const { key, shiftKey } = event
+      const amount = shiftKey
+        ? this.modifiedSeekJumpPercentage
+        : this.seekJumpPercentage
+      const seekPercentage =
+        this.percentage + (key.includes('Left') ? -amount : amount)
+      this.$emit('seeked', seekPercentage)
+    },
```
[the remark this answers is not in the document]
zackkrida: Created a PR regarding the `ogg` file: https://github.com/WordPress/openverse-frontend/issues/144

zackkrida:
I'll merge this, and we can make any a11y enhancements on the audio branch as needed.

PR #138, opened 2021-08-19
Fix navbar keyboard accessibility

### LINKED ISSUE #116: [Bug] Impossible to focus nav bar items using keyboard navigation
## Description
Using only keyboard navigation, it is not possible to focus the nav bar items. They're not tabbable elements so the browser skips over them and right onto the search box.

## Reproduction

1. Open the frontend
2. Using the `tab` button, attempt to focus the nav bar dropdowns like the About and Resources dropdowns.

## Expectation
The dropdowns and their individual items should be focusable using keyboard navigation.

## Environment

This isn't environment specific. I've confirmed it with and without VoiceOver in Safari and Firefox on macOS.

## Additional context
I'm not sure why this is happening as the nav elements are all just plain anchor elements, but something is causing them to not be focusable.

## Resolution

sarayourfriend:
## Fixes

Fixes #116  by @sarayourfriend 

## Description
Adds a Dropdown component based on [Vuetensils' Dropdown](https://raw.githubusercontent.com/AustinGil/Vuetensils/production/src/components/VDropdown/VDropdown.vue) to make the navbar keyboard accessible.

## Technical details
Heavily modifies the Vuetensils' Dropdown to accomodate a couple of things:

0. Make the keyboard navigation behave the same as WordPress.org's navbar (i.e., no arrow key navigation, just tabbing through each item in order)
1. Adds focus management to each child by making an `onFocus` prop available through the default slot. This `onFocus` prop should be used in the `@focus` for each dropdown item. This makes it possible to tab "backwards" from a subsequent navbar item. For example, in the real example of the Openverse dropdown if you <kbd>Shift + Tab</kbd> from the Web Extension navbar item, you should focus into the last item of the Resources dropdown, not the Resources dropdown itself.
2. Makes the dropdown container always rendered into the DOM by using the visually hidden styles, lifted from `@wordpress/components`. This makes the dropdown items always appear in the DOM, making them keyboard focusable while hiding them from sighted users.
3. Use Bulma classes to fit the existing implementation. This is a drop-in replacement for the existing implementation, style wise, but adds keyboard accessibility (there should be no visual differences between this and the existing implementation).

## Tests
If you're on macOS, ensure that you've enabled keyboard navigation in the accessibility options following the instructions here: https://www.a11yproject.com/posts/2017-12-29-macos-browser-keyboard-navigation/

After restarting your browser, you should be able to tab through the navbar items including the dropdown items. Compare this to the live site and you'll see that it was not previously possible.

There are also unit tests to ensure that the focus management works (by teseting the presence of the `visible` class on the dropdown wrapper).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/Dropdown.vue
+++ b/src/components/Dropdown.vue
@@ -2,18 +2,21 @@
   <!-- Copied and modified from Vuetensils to match the existing bulma implementation -->
   <!-- https://github.com/AustinGil/vuetensils/blob/production/src/components/VDropdown/VDropdown.vue -->
   <div
-    :class="['navbar-item has-dropdown is-hoverable', classes.root]"
+    class="navbar-item has-dropdown is-hoverable"
     @mouseenter="isHovered = true"
     @mouseleave="isHovered = false"
     @focus="isFocused = true"
     @blur="isFocused = false"
     @focusout="onFocusout"
   >
-    <a
+    <div
       :aria-expanded="!!isHovered || !!isFocused"
       aria-haspopup="true"
-      :class="['navbar-link is-arrowless', classes.trigger]"
-      href="#"
+      class="navbar-link is-arrowless"
+      role="button"
+      tabindex="0"
+      @keydown.enter="isFocused = !isFocused"
+      @keydown.space.prevent="isFocused = !isFocused"
       @click="isFocused = !isFocused"
       @focus="isFocused = true"
     >
@@ -21,7 +24,7 @@
         {{ text }}
         <i class="icon caret-down" />
       </slot>
-    </a>
+    </div>
 
     <div
       class="navbar-dropdown"
@@ -46,34 +49,11 @@ const Dropdown = {
   name: 'Dropdown',
   props: {
     /**
-     * The toggle button text.
+     * The trigger element text.
      */
     text: {
       type: String,
       default: '',
-    },
-    /**
-     * Where the content should be placed in relation to the button.
-     *
-     * Options: 'bottom', 'top'
-     */
-    position: {
-      type: String,
-      default: 'bottom',
-      validator(value) {
-        return ['top', 'bottom'].includes(value)
-      },
-    },
-    /**
-     * The transition name.
-     */
-    transition: {
-      type: String,
-      default: '',
-    },
-    classes: {
-      type: Object,
-      default: () => ({}),
     },
   },
   data: () => ({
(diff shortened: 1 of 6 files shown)


src/components/Dropdown.vue
```diff
@@ -0,0 +1,110 @@
...
+  <div
+    class="navbar-item has-dropdown is-hoverable"
+    @mouseenter="isHovered = true"
+    @mouseleave="isHovered = false"
+    @focus="isFocused = true"
+    @blur="isFocused = false"
+    @focusout="onFocusout"
+  >
+    <div
+      :aria-expanded="!!isHovered || !!isFocused"
```
obulat, line 13: Can isHovered ever be not boolean? Do we really need the !! in front of it?

src/components/Dropdown.vue
```diff
@@ -0,0 +1,110 @@
...
+const Dropdown = {
+  name: 'Dropdown',
+  props: {
+    /**
+     * The trigger element text.
+     */
+    text: {
+      type: String,
+      default: '',
+    },
```
obulat, line 57: This could be 'required' since we lways have text in the dropdown trigger.

src/components/Dropdown.vue
```diff
@@ -0,0 +1,110 @@
...
+      this.isFocused = true
+    },
+  },
+}
+
+export default Dropdown
+</script>
+
+<style lang="scss">
+.navbar-dropdown {
```
obulat, line 93: Can we change this in the original scss navbar file insidethe styles folder?

obulat (approved):
This is fantastic, Sara. I have left some tiny comments and questions inline. So great to improve a11y on such a vital part of the site!

src/components/Dropdown.vue
sarayourfriend: I think you're right 👍

sarayourfriend: Done in 25b141e

src/components/Dropdown.vue
sarayourfriend: Fixed in 25b141e

src/components/Dropdown.vue
sarayourfriend: Done in d0c6ebe

src/components/Dropdown.vue
sarayourfriend: Done in 25b141e

zackkrida (approved):
LGTM! I appreciate the jest `--watch` script as well.

PR #156, opened 2021-08-24
Flip timestamp to prevent overflow

### LINKED ISSUE #150: [Feature] Waveform timestamp is cut off at the start of the audio track
## Problem
The timestamp is not visible at the start of the audio track.

## Description
The timestamp gets cut off at the beginning of the audio:

Captura de Tela 2021-08-23 às 13 21 38

(Note the cut off "2")

We should make it so that the timestamp "flips" sides at the start of the audio track until there's enough space for it to correctly appear to the left hand side.

Alternatively we could move the timestamp to the top of the waveform, above it, so that it isn't cut off by the bounding box of the waveform (but this would be a pretty different implementation and making it line up correctly would be hard I think).

## Implementation

dhruvkb:
## Fixes

Fixes #150 by @sarayourfriend 

## Description

This PR flips the timestamp when there isn't adequate space to display it in it's regular direction.

## Technical details

The flip effect was due to a `transform: translateX(-100%)` rule, which is now disabled unless there is enough space to display the timestamp on the left side.

## Screenshots

https://user-images.githubusercontent.com/16580576/130617445-b9345ca4-d246-4cd6-aca6-50c00832aafc.mov

There is a small bit of a jump involved because of the changing progress bar width and changing timestamp width being larger and smaller than each other.

## Checklist

- [ ] My pull request targets the *default* branch of the repository (`main` or `master`).

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/AudioTrack/Waveform.vue
+++ b/src/components/AudioTrack/Waveform.vue
@@ -51,21 +51,29 @@
     <!-- Timestamps -->
     <template v-if="isReady">
       <div
-        class="progress absolute top-1 font-bold text-sm bg-yellow z-10 px-1 transform -translate-x-full pointer-events-none"
+        ref="progressTimestampEl"
+        class="progress timestamp z-10 bg-yellow transform"
+        :class="[
+          ...(isProgressTimestampCutoff
+            ? ['bg-dark-charcoal-04-opaque']
+            : ['bg-yellow', '-translate-x-full']),
+        ]"
         :style="{ '--progress-time-left': `${progressBarWidth}px` }"
       >
         {{ timeFmt(progressTimestamp) }}
       </div>
       <div
         v-if="seekFrac"
-        class="seek absolute top-1 font-bold text-sm px-1 transform -translate-x-full pointer-events-none"
+        ref="seekTimestampEl"
+        class="seek timestamp transform"
+        :class="{ '-translate-x-full': !isSeekTimestampCutoff }"
         :style="{ '--seek-time-left': `${seekBarWidth}px` }"
       >
         {{ timeFmt(seekTimestamp) }}
       </div>
       <div
         v-if="showDuration"
-        class="duration absolute top-1 right-0 px-1 font-bold text-sm pointer-events-none"
+        class="duration timestamp right-0 bg-dark-charcoal-04-opaque"
       >
         {{ timeFmt(duration) }}
       </div>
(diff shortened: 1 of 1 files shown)


src/components/AudioTrack/Waveform.vue
```diff
@@ -343,6 +365,10 @@ export default {
 </script>
 
 <style scoped lang="css">
+.timestamp {
+  @apply absolute top-1 font-bold text-sm px-1 pointer-events-none;
+}
```
sarayourfriend, line 370: Is `@apply` safe to use? My understanding is that the proposal for it has been abandoned?

sarayourfriend (commented):
I wonder if we could add some arbitrary value to the offsetWidth of the timestamp element to make it not stutter like it is right now?

src/components/AudioTrack/Waveform.vue
dhruvkb: Doesn't seem to be abandoned from the [docs](https://tailwindcss.com/docs/extracting-components#extracting-component-classes-with-apply) where they propose it as a simpler alternative to creating a JS-framework component. Please share the link to the proposal if you have it?

dhruvkb:
@sarayourfriend the bit of shaking should be fixed by 975fbd1.

src/components/AudioTrack/Waveform.vue
sarayourfriend: Here it says it is abandoned: https://tabatkins.github.io/specs/css-apply-rule/

But it also looks like tailwind uses it a lot so I'm confused about the status of it.

sarayourfriend: Oh wait, is tailwind's `@apply` different from the CSS standard `@apply`? How does that work? Does tailwind take a pre-processor?

dhruvkb: Ah yes @sarayourfriend, the CSS draft proposal for `@apply` has been abandoned but PostCSS (which Tailwind uses) seems to work around it. We should be fine for now as this support extends into PostCSS 8 as well.

sarayourfriend: Oh okay, `PostCSS` is so confusing to me, as it just works on regular `css` files instead of adopting a different format 🙄 Oh well, it makes sense to me to continue to use it if we're sticking with Tailwind.

sarayourfriend (approved):
Perfect! 🚀

src/components/AudioTrack/Waveform.vue
dhruvkb: Sometimes PostCSS files have the `.pcss` extension but I've rarely seen it being used.

krysal (approved):
Cool to see the current timestamp from the beginning!

I still see a little jump, the magic number where is gone in my case (with Chrome) seems to be `+ 5`. Anyway, the original problem is solved so not blocking this, thanks!

Issue #118, opened 2021-08-11
Audio Functionality


zackkrida:
## Completed

- [x] #570
- [x] #146
- [x] #228
- [x] #119
- [x] #213 
- [x] #145
- [x] #149
- [x] #102 
- [x] #147
- [x] #148
- [x] #150 
- [x] #103 
- [x] #144
- [x] #105 
- [x] #104

zackkrida:
Audio functionality is fully implemented. Any remaining bugs can be triaged on the `main` branch.

PR #173, opened 2021-08-30
Refactor search-store to use generic media

obulat:
## Fixes

Related to #118 by @zackkrida

## Description

This PR refactors the search store so that it becomes possible to add other media types to it.

## Technical details

I tried adding type information for API responses. It will probably need to be changed/updated/removed?

## Checklist

- [ ] My pull request targets the *default* branch of the repository (`main` or `master`).

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/data/AudioService.js
```diff
@@ -20,27 +47,34 @@ const AudioService = {
   /**
    * Retrieve audio details by Id number.
    * SSR-called
+   * @param params
```
sarayourfriend, line 50: If this requires the `id` should we expand the definition to convey that?

```suggestion
   * @param {object} params
   * @param {string | number} params.id
```

src/data/AudioService.js
```diff
@@ -20,27 +47,34 @@ const AudioService = {
...
 
     return config.dev
-      ? Promise.resolve(sampleAudioResponses.detail)
+      ? Promise.resolve({ data: sampleAudioResponses.detail })
       : ApiService.get('audios', params.id)
   },
 
+  /**
+   * Retrieve related media
+   * @param params
```
sarayourfriend, line 67: Ditto here re: types

src/data/types.js
```diff
@@ -0,0 +1,14 @@
...
+ * @property {number} page_count
+ * @property {number} page_size
+ * @property {MediaResult[]} results
+ */
+
+/**
+ * @typedef {Object} APIResponse
+ * @property {Array} data
+ */
+export {}
```
sarayourfriend, line 14: It looks like we're using `APIResponse` as a generic type here: https://github.com/WordPress/openverse-frontend/pull/173/files#diff-a6daedb006e158dcd94eb06a42f670f76b3c8ffe49b9dfdee53a53f6c492c0a9R51

As well in some other places.

We should expand the definition in that case to include a template parameter:

```suggestion
/**
 * @template T
 * @typedef {Object} APIResponse
 * @property {Array<T>} data
 */
export {}
```

I'm just guessing on the implementation of the generic parameter

src/data/types.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * The search result object
+ * @typedef {Object} MediaResult
+ * @property {number} result_count
+ * @property {number} page_count
+ * @property {number} page_size
+ * @property {MediaResult[]} results
```
sarayourfriend, line 7: It looks like MediaResult is also generic here: https://github.com/WordPress/openverse-frontend/pull/173/files#diff-a6daedb006e158dcd94eb06a42f670f76b3c8ffe49b9dfdee53a53f6c492c0a9R51

Should it be this?

```suggestion
 * The search result object
 * @template {MediaResult} T
 * @typedef {Object} MediaResult
 * @property {number} result_count
 * @property {number} page_count
 * @property {number} page_size
 * @property {T[]} results
```

src/store-modules/types.d.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * The search result object
+ */
+export interface MediaResult {
+  result_count: number
+  page_count: number
+  page_size: number
+  results: AudioDetail[] | ImageDetail[]
+}
```
sarayourfriend, line 9: If we can make this a generic type I think it will be more useful. For example if someone calles `getAudio` it should return a `MediaResult<AudioDetail>` without them having to do a typecheck on the `results` property right?

src/data/AudioService.js
```diff
@@ -20,27 +22,34 @@ const AudioService = {
   /**
    * Retrieve audio details by Id number.
    * SSR-called
+   * @param {{id: string}} params
+   * @return {Promise<{data: Object}>}
```
sarayourfriend, line 26: Fun fact, `Object`, `object` and `{}` all have different meanings in TypeScript. https://stackoverflow.com/questions/49464634/difference-between-object-and-object-in-typescript

It seems like here we actually might want `any` or `unknown` if we don't have a better type for it?

krysal (approved):
This looks good to me! We can continue to add more JSDocs/TypeScript checking in subsequent PRs related to #175.

Thank you so much!

src/data/AudioService.js
obulat: Thank you for this comment, I didn't know! Interesting read :)

PR #188, opened 2021-09-06
UI improvements for audio track

### LINKED ISSUE #147: [Feature] Waveform focus styles
## Problem
The waveform does not have a clear focus style, meaning that sighted keyboard users will have a difficult time knowing that they are able to seek using the keyboard like other "seek bars" that exist.

## Description
We should add a focus style that makes it clear to sighted keyboard users that they can seek using the arrow keys. Screen reader users will get an indiciation from their screen reader that they are in a "slider" which will give the hint that they can use the arrow keys to interact, but for non-screen reader keyboard users this will not apply. Therefore we need some kind of visual indication that the waveform is focused and seekable.

One potential solution is to make the vertical "seek" line a solid color with a circle or bulb at the top. Similar to how iOS represents a video scub in the Photos app, but with a bulb at the top of the line to add an extra indication (especially when the audio is at the `0:00` timestamp something will need to be visible inside of the waveform).

IMG_1064

## Alternatives
I'm sure there are other designs we could do for the focus styles, but I'm not sure of any.

## Implementation

### LINKED ISSUE #190: [Feature] Waveform keyboard seek to start/end
## Problem
Some keyboard navigable audio/video seek bars are able to be jumped to the start or end of the audio by using `cmd/super + left/right` or `home/end`. We don't currently implement this. 

## Description
Add the ability to seek to the start or end of the audio using the keyboard either through `meta + left/right` or `home/end`

## Additional context
I got this idea checking out this other media player's implementation: https://github.com/IET-OU/mediaelement/commit/25eb64a2097f25bffd5c1420807f5605cbe4e0b0#diff-219f920a90e9fec6455e5cb6863cf6e51831182cd586184b53e5ae3a4e94b32aR102-R108

## Implementation

### LINKED ISSUE #191: [Bug] Play/pause button needs rounded corners
## Description
The play/pause button current has square corners, but should be updated to match the design and have rounded corners: https://www.figma.com/file/vEoItpMQyFZb5AZV9aoP7I/Audio-integration?node-id=1107%3A12561

## Resolution

dhruvkb:
## Fixes

Fixes #147 by @sarayourfriend
Fixes #190 by @sarayourfriend
Fixes #191 by @sarayourfriend

## Description

This PR adds the focus indicator to the waveform.

## Known issues
The focus outline of the play/pause button interferences with the waveform seekbar, likely due to complex `z-index` layers inside the `Waveform` component. This might be tricky to solve.

## Screenshots

https://user-images.githubusercontent.com/16580576/132212606-9dd3d87a-1cd3-475e-93e7-da66b9af6a85.mov

Uses <kbd>Tab</kbd> and <kbd>Shift</kbd>+<kbd>Tab</kbd> to focus forward and backward respectively.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/AudioTrack/Waveform.vue
+++ b/src/components/AudioTrack/Waveform.vue
@@ -37,9 +37,7 @@
         class="transform origin-bottom transition-transform duration-500"
         :class="[
           isReady ? 'scale-y-100' : 'scale-y-0',
-          spaceBefore(index) < seekBarWidth
-            ? 'fill-black'
-            : 'fill-dark-charcoal-20',
+          index <= seekIndex ? 'fill-black' : 'fill-dark-charcoal-20',
         ]"
         :x="spaceBefore(index)"
         :y="spaceAbove(index)"
@@ -265,6 +263,8 @@ export default {
       const frac = seekFrac.value ?? currentFrac.value
       return waveformWidth.value * frac
     })
+    const seekIndex = computed(() => getPeaksInWidth(seekBarWidth.value))
+    const seekSpaceBefore = computed(() => spaceBefore(seekIndex.value))
 
     /* Seek timestamp */
 
@@ -360,6 +360,9 @@ export default {
 
       seekFrac,
       seekBarWidth,
+      seekIndex,
+      seekSpaceBefore,
+
       seekTimestamp,
       seekTimestampEl,
       isSeekTimestampCutoff,
--- a/src/components/AudioTrack/Waveform.vue
+++ b/src/components/AudioTrack/Waveform.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     ref="el"
-    class="waveform relative bg-dark-charcoal-04 overflow-x-hidden focus:outline-none"
+    class="waveform relative bg-dark-charcoal-04 overflow-hidden focus:outline-none"
     tabIndex="0"
     role="slider"
     :aria-label="$t('waveform.label')"
--- a/src/components/AudioTrack/Waveform.vue
+++ b/src/components/AudioTrack/Waveform.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     ref="el"
-    class="waveform relative bg-dark-charcoal-04 overflow-hidden focus:outline-none"
+    class="waveform relative bg-dark-charcoal-04 overflow-hidden rounded-sm focus:outline-none focus-visible:ring focus-visible:ring-offset-2 focus-visible:ring-pink"
     tabIndex="0"
     role="slider"
     :aria-label="$t('waveform.label')"
--- a/src/components/AudioTrack/Waveform.vue
+++ b/src/components/AudioTrack/Waveform.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     ref="el"
-    class="waveform relative bg-dark-charcoal-04 overflow-x-hidden"
+    class="waveform relative bg-dark-charcoal-04 overflow-x-hidden focus:outline-none"
     tabIndex="0"
     role="slider"
     :aria-label="$t('waveform.label')"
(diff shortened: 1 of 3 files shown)


zackkrida (approved):
Looks good. Two things:

1. My waveform has a vertical scrollbar, that is easily removed by replacing `overflow-x-hidden` on the waveform with `overflow-hidden`.
2. The play/pause overflow issue is caused becasue the waveform has relative positioning, which gives it a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) and places it in front of the play/pause button. We can add the `relative` class to the play/pause button or any of its parents to fix that.

src/components/AudioTrack/Waveform.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div
     ref="el"
-    class="waveform relative bg-dark-charcoal-04 overflow-x-hidden"
+    class="waveform relative bg-dark-charcoal-04 overflow-x-hidden focus:outline-none"
```
sarayourfriend, line 4: Will this disable the extra outline on high contrast mode on Windows? 🤔 I think we shouldn't necessarily fight the natural focus outline when it exists.

sarayourfriend (commented):
https://user-images.githubusercontent.com/24264157/132225697-bdf78e64-bd64-4a80-98ee-fb44a0ef6c3d.mov

Code looks great! There a couple of issues though. In firefox, when pressing `space` to toggle the play/pause, the default behavior of the space bar activates in the waveform element and it is scrolled down. I guess there is some y overflow. This would also trigger if someone was scrolling over the page and accidentally started their scroll inside the waveform element. I guess we should ensure that overflow-y is hidden?

Secondly, (this probably warrants a separate issue/PR) I finally got to test it using VoiceOver in Safari and the entire interaction is broken. The aria label is not translated (the `$t...` is read out literally) and the `valuetext` is not correct read on interaction. Furthermore, the instructions for interacting with the element don't work correctly, `control+option+left/right` doesn't do anything and `left/right` arrows alone seem to be navigating inside the text of the timestamp rather than the manipulating the slider itself.


https://user-images.githubusercontent.com/24264157/132226792-262b451e-a499-49f4-a584-5196b1944e13.mov

To me the only blocker here though is fixing the y-overflow. The rest can be addressed in a separate PR.

src/components/AudioTrack/Waveform.vue
zackkrida: > The outline-none utility is implemented using a transparent outline under the hood to ensure elements are still visibly focused to Windows high contrast mode users.

No, thankfully! https://tailwindcss.com/docs/outline

sarayourfriend: Right on!

zackkrida:
@dhruvkb final focus mockup is [here](https://www.figma.com/file/vEoItpMQyFZb5AZV9aoP7I/Audio-integration?node-id=1151%3A14566):

CleanShot 2021-09-06 at 10 13 47@2x

dhruvkb:
This looks really good! I'll mark this PR as draft and take it back to the workshop.

sarayourfriend:
> The focus outline of the play/pause button interferences with the waveform seekbar, likely due to complex z-index layers inside the Waveform component. This might be tricky to solve.

I think this would be solved in the newest design that Zack shared above where the play/pause button is separated from the waveform, right?

dhruvkb:
I'll work on the VoiceOver issues separately as there is a bit of reading involved and I wouldn't want to block #195 for the focus styles.

sarayourfriend (approved):
LGTM! Nice work 🚀

Issue #205, opened 2021-09-10
[Feature] Break down megastore into smaller modules


dhruvkb:
## Problem

Currently we have a huge megastore that stores everything. This is not optimal because
- there is a possibility of name collisions in state (actions and mutations are safer due to use of constants)
- all information about the hierarchy of modules is lost when they are flattened and merged into one

Moreover modules are merged with `Object.assign` without a blank target, which alters the first argument to receive all subsequent keys.

## Description

Putting the contents of `store-modules/` inside `store/` can automatically configure Nuxt to use namespaced modules. The need for manually merging all stores with `Object.assign` is also eliminated.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

PR #207, opened 2021-09-13
Active Media Vuex Model + Composable AudioTrack

### LINKED ISSUE #182: [Bug] Managing playback of multiple media files
## Description

The current setup allows for multiple audio files to be played concurrently, which is a bad user experience.

## Reproduction

1. View any page with multiple audio players
2. Press play on multiple audio players
3. Listen to the resulting 'chaos orchestra'

## Expectation

When pressing 'play' on an audio file, if there is _already_ an active audio file it should be paused.

## Resolution

I have proposed a solution in #183.

### LINKED ISSUE #195: [Feature] Abstract AudioTrack to support different UI layouts
## Problem

The `AudioTrack` component currently supports only two layouts (and that too with quite a significant bit of code repetition). Extending the same component to support many more new layouts will be complicated unless the player, waveform and play/pause facilities are abstracted and made composable.

Screenshot on 2021-09-06 at 15-34-34

## Description

There needs to be a higher level of abstraction over the audio player functionality so that these three components are composable.

## Additional context

This is closely related to #182 and #183 as breaking the components will need a Vuex store for handling state.

## Implementation

dhruvkb:
## Fixes

Fixes #182 by @zackkrida 
Fixes #195 by @dhruvkb

## Description

This PR
- creates a store for holding the ID of the active media
- automatically pauses playing media when something new plays
- breaks down `AudioTrack` into `AudioController` (with `Waveform`) and `PlayPause`
- enables composing these components in different layouts
- creates 3 layouts: Full and Row (S + M)
- creates a new component `License`
- creates a new component `AudioThumb` (just a yellow square)

## To-do

The examples show how to lay the different layouts. Work on the contents is still a WIP (especially considering that new components will be needed to display the license icon)

## Screenshots

`AudioTrack` (feat. `AudioThumb`):
image

`License`:
image

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/AudioTrack/PlayPause.vue
```diff
@@ -1,8 +1,8 @@
 <template>
   <button
     type="button"
-    class="flex items-center justify-center bg-dark-charcoal h-20 w-20 rounded-sm transition-shadow duration-100 ease-linear disabled:opacity-70 focus:outline-none focus-visible:ring focus-visible:ring-offset-2 focus-visible:ring-pink"
-    @click="toggle"
+    class="play-pause flex items-center justify-center bg-dark-charcoal transition-shadow duration-100 ease-linear disabled:opacity-70 focus:outline-none focus-visible:ring focus-visible:ring-offset-2 focus-visible:ring-pink"
```
obulat, line 4: Can `@apply` be used here?

src/components/AudioTrack/layouts/Row.vue
```diff
@@ -0,0 +1,89 @@
...
+          <div class="part-a">
+            <i18n
+              tag="span"
+              class="font-semibold leading-snug"
+              path="audio-track.creator"
+            >
+              <template #creator>{{ audio.creator }}</template>
+            </i18n>
+            <span v-if="!isSmall">
+              • {{ audio.duration }} • {{ audio.category }}
```
obulat, line 27: Duration needs to be converted from milliseconds to min:sec.
And category needs to be translated. Or is it passed here already translated?

src/components/AudioTrack/meta/AudioTrack.stories.mdx
```diff
@@ -110,3 +144,35 @@ The component is capable of playing audio in many formats, but this largely
 depends on the browser and engine. Refer to
 ['Can I use...'](https://caniuse.com/?search=audio%20format) for more up-to-date
 information on browser compatibility.
+
+## Single active player
+
+The audio players use a centralised store for holding the details of the active
```
obulat, line 150: Works great, and thank you for Mozart!
It would be closer to the real usage if the `AudioTrack`s here used
`{ layout: 'row', size: 'm' }`, as in Related Audio.

src/components/AudioTrack/layouts/Row.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+  <div class="row-track flex flex-row" :class="`size-${size}`">
+    <AudioThumb
+      class="flex-shrink-0"
+      :class="isSmall ? 'w-20 mr-4' : 'w-30 mr-6'"
+      :audio="audio"
+    />
+    <div class="flex" :class="isSmall ? 'flex-row gap-8' : 'flex-col gap-4'">
+      <div class="flex-shrink-0">
+        <p class="font-heading font-semibold text-2xl">{{ audio.title }}</p>
```
obulat, line 10: The Audio title should be a link. I'm not sure if it should be a link _everywhere_, but on the search result page and on related audios for sure.

src/components/AudioTrack/layouts/Row.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+  <div class="row-track flex flex-row" :class="`size-${size}`">
```
obulat, line 2: From our conversation with @zackkrida , 

> each [search result] should probably also be an article tag, semantically.
?

src/components/AudioTrack/meta/AudioTrack.stories.mdx
dhruvkb: Resolved in b7baf27.

src/components/AudioTrack/layouts/Row.vue
dhruvkb: Should it be wrapped in an `article` or itself be an `article`?

src/components/AudioTrack/layouts/Row.vue
dhruvkb: Won't the search results page uses the (upcoming) 'box' layout?

src/components/AudioTrack/layouts/Row.vue
dhruvkb: Converted and formatted duration in f11dfae. Category translated in f5e63b2.

src/components/AudioTrack/AudioThumb.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <!-- Should have a fixed width -->
```
zackkrida, line 2:
```suggestion
  <!-- Should have a fixed width -->
```

I'm not 100% sure what this comment means. It reads like a `@TODO` but I think it's meant to be more of an suggestion for implementing the audio thumb? I'm not certain.

src/components/AudioTrack/layouts/Row.vue
zackkrida: I _believe_ the box layout is used in the "all" results view, and the row used when _exclusively_ searching for audio. I agree about linking to the single result view in this Row component.

src/components/AudioTrack/layouts/Row.vue
zackkrida: I think it should itself be an article, only for the sake of fewer dom nodes in the tree.

src/components/License/License.vue
```diff
@@ -0,0 +1,68 @@
...
+      validator: (val) =>
+        [
+          'by',
+          'by-sa',
+          'by-nd',
+          'by-nc',
+          'by-nc-sa',
+          'by-nc-nd',
+          'cc0',
+          'pdm',
```
zackkrida, line 44: surprised this isn't a constant yet somewhere, but we can add that to the frontend later if needed

src/store-modules/active-media-store.js
```diff
@@ -0,0 +1,39 @@
...
+  UNSET_ACTIVE_MEDIA_ITEM,
+} from './mutation-types'
+
+/**
+ * Stores information about the active media item.
+ * @type {import('./types').ActiveMediaState}
+ */
+const state = {
+  activeMediaType: null,
+  activeMediaId: null,
```
zackkrida, line 12: Small stylistic question, do we need to be this verbose? `activeMediaType` vs. `type` for example. Is the context of the ActiveMediaStore enough? `</bikeshedding>`

zackkrida (approved):
Amazing work.

src/components/AudioTrack/AudioThumb.vue
dhruvkb: It's an instruction for the users of this component. I should probably add an explanation for this in the Storybook for the component.

zackkrida: thanks!

zackkrida:
One more thing I noticed: the focus outline appearance is less than ideal here:

CleanShot 2021-09-13 at 20 29 01@2x

dhruvkb:
Tagging @panchovm as this will need some input from the design perspective. My (non-professional) opinion is that the parent component should be padding the audio component when using in a full-screen context.

src/store-modules/active-media-store.js
dhruvkb: Since our store modules are not namespaced, I'm concerned that using single-word names like `type` and `id` can be 
- too ambiguous to know what store they represent
- easy to conflict if used by some other store

src/components/License/License.vue
obulat: I had a lot of license-related code in the chooser: https://github.com/creativecommons/chooser/blob/main/src/utils/licenses.js :)

src/components/AudioTrack/layouts/Row.vue
dhruvkb: Added the link in c80de9e.

src/components/AudioTrack/layouts/Row.vue
dhruvkb: Changed to an `article` element in c80de9e.

src/components/License/License.vue
dhruvkb: Extracted constants separately in a67800c.

src/components/AudioTrack/PlayPause.vue
dhruvkb: I'm sorry but I didn't understand what you meant.

src/store-modules/active-media-store.js
zackkrida: fine with me 😄

dhruvkb: We might be able to switch to smaller names after #205.

src/components/AudioTrack/PlayPause.vue
zackkrida: She's referring to using a class here, like ".play-button", and putting all these tailwind classes in an @apply in the style block of the component.

Which in general we need to decide if that's a pattern we want to embrace or not.

dhruvkb: Using `@apply` feels fundamentally wrong **to me** because I feel like Tailwind's full potential can be utilised when you don't have to write any CSS at all. I only use @apply for situations where 
- the same classes are being repeated in many places
- the classes (if combined) make a meaningful "micro-component"

but those are just my thoughts. We should definitely discuss this in the chat.

zackkrida:
@sarayourfriend I think this would be a good fit for you to review this week. You have a strong eye for component composition that would be appreciated here!

sarayourfriend:
I will take a close look at this, but it will take some time to review.

src/components/AudioTrack/AudioThumb.vue
```diff
@@ -0,0 +1,18 @@
...
+  <div>
+    <div class="h-0 w-full pt-full bg-yellow">&nbsp;</div>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'AudioThumb',
```
sarayourfriend, line 10: Where does the name "Thumb" come from? I've only ever seen thumb used to describe the knob on a slider before so this is new to me.

src/components/AudioTrack/AudioThumb.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <!-- Should have a fixed width -->
+  <div>
+    <div class="h-0 w-full pt-full bg-yellow">&nbsp;</div>
+  </div>
```
sarayourfriend, line 5: Does this need a wrapping `div` or could it just be:

```suggestion
  <div class="h-0 w-full pt-full bg-yellow">&nbsp;</div>
```

src/components/AudioTrack/AudioTrack.vue
```diff
@@ -1,106 +1,51 @@
...
-          :duration="duration"
-          :peaks="audio.peaks"
-          :show-duration="isCompact"
-          @seeked="setPosition"
+    <Component
+      :is="layoutComponent"
+      :audio="audio"
+      :size="size"
+      :aspect="aspect"
+    >
```
sarayourfriend, line 12: This is very cool, like component polymoprhism 😁

src/components/AudioTrack/PlayPause.vue
```diff
@@ -1,12 +1,12 @@
 <template>
   <button
     type="button"
-    class="flex items-center justify-center bg-dark-charcoal h-20 w-20 rounded-sm transition-shadow duration-100 ease-linear disabled:opacity-70 focus:outline-none focus-visible:ring focus-visible:ring-offset-2 focus-visible:ring-pink"
-    @click="toggle"
+    class="play-pause flex items-center justify-center bg-dark-charcoal text-white transition-shadow duration-100 ease-linear disabled:opacity-70 focus:outline-none focus-visible:ring focus-visible:ring-offset-2 focus-visible:ring-pink"
```
sarayourfriend, line 4: Where does the `play-pause` class come from?

src/components/AudioTrack/PlayPause.vue
```diff
@@ -43,13 +48,16 @@ export default {
...
+    },
     label() {
       return this.$t(this.isPlaying ? 'play-pause.pause' : 'play-pause.play')
     },
   },
   methods: {
-    toggle() {
-      this.$emit('toggle', !this.isPlaying)
+    handleClick() {
+      this.$emit('toggle', this.isPlaying ? 'paused' : 'playing')
```
sarayourfriend, line 60: Nice, sending an enum is much clearer here.

src/components/AudioTrack/meta/AudioTrack.stories.mdx
```diff
@@ -90,14 +109,25 @@ export const Template = (args, { argTypes }) => ({
   <Story name="Default">{Template.bind({})}</Story>
 </Canvas>
 
-The component can also be rendered in a smaller format, when used in the
-"Related audio" section for example.
+The component can also be rendered in a multiple different layouts, by
+specifying a `layout`. The layout can further by customised by the `size` and
```
sarayourfriend, line 113:
```suggestion
specifying a `layout`. The layout can further be customised by the `size` and
```

src/store-modules/active-media-store.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+const mutations = {
+  /**
+   * Set the active media item.
+   * @param {import('./types').ActiveMediaState} _state
+   * @param {'image' | 'audio' | null} type - the nature of the active media item
+   * @param {string} id - the ID of the active media item
```
sarayourfriend, line 20: For object destructuring to work with TypeScript compiler we'd need to do the following:

```suggestion
   * @param {import('./types').ActiveMediaState} _state
   * @param {object} payload
   * @param {'image' | 'audio' | null} payload.type - the nature of the active media item
   * @param {string} payload.id - the ID of the active media item
```

sarayourfriend (commented):
Just reviewing the code so far, I still haven't tested it locally, but will do that next.

Overall I really like this approach 🙂

src/components/AudioTrack/AudioThumb.vue
obulat: Short for thumbnail?

sarayourfriend (approved):
LGTM. Tests well locally and I really like the approach you took. Nice work!

src/components/AudioTrack/AudioThumb.vue
dhruvkb: The `div` is needed because the aspect ratio hack doesn't work on `flex` children. I'm desperately waiting for general support for [CSS `aspect-ratio` property](https://caniuse.com/mdn-css_properties_aspect-ratio).

src/components/AudioTrack/AudioTrack.vue
dhruvkb: Thanks 😃!

src/components/AudioTrack/PlayPause.vue
dhruvkb: It's not a utility that's coming from any library. The `play-pause` class is declared here to allow the layouts to style the component from their stylesheets by referencing it.

src/components/AudioTrack/PlayPause.vue
dhruvkb: Yeah, boolean was ambiguous. Plus we can expand to more granular statuses like loading, errored etc. in the future.

src/components/AudioTrack/AudioThumb.vue
dhruvkb: Yes, it was short for thumbnail. But I can understand the confusion so I've expanded that in bdb2f1c.

src/components/AudioTrack/AudioThumb.vue
dhruvkb: But the responsibility of wrapping should be on the layout and not the thumbnail, so it's fixed in da4ff50.

src/store-modules/active-media-store.js
dhruvkb: Thank you! I was stupidly blaming WebStorm for not being able to infer the types 😂!

src/components/AudioTrack/meta/AudioTrack.stories.mdx
dhruvkb: Thanks for catching this!

src/components/AudioTrack/PlayPause.vue
sarayourfriend: Ah okay... In generally I prefer to avoid that approach because it creates the problem of "magic strings" that are hard to change or understand in the future. Would it be possible to in the future instead pass a new prop for the specific styles that should be added to the component? For example allow the user to pass a `button-class` prop or something like that would allow them to define specific scoped styles or use CSS Modules for example and would be easier tie the heirarchy of style application (at least in my experience).

Issue #149, opened 2021-08-23
Accessible dropdown button (for audio downloads)


zackkrida:

panchovm:
I was thinking of a dropdown menu similar to the _filters applied_ menu when hovering the filter badge next to expanding filters action. Yet a pending mockup I need to polish.

sarayourfriend:
Is there a figma for the dropdown yet?

Also, when clicking one of the dropdown options, does it immediately trigger the download or does it just change the format that will be downloaded when clicking download?

zackkrida:
> Is there a figma for the dropdown yet?

> Also, when clicking one of the dropdown options, does it immediately trigger the download or does it just change the format that will be downloaded when clicking download?

One more question from me:

> Anything on the button to show what the currently-selected format is? Like "download audio" is essentially the button label, can we add some text for "MP3 V0" or whatever the selected format is?

These are good questions for @panchovm 😄 . I think the answers are both "not yet" and "but soon" for all questions.

panchovm:
Haha, like @zackkrida said, not yet but soon. Although I have to say that I was thinking of **immediately trigger the download**. However, in all audio views, the default option will be the same and the user may have to select the favorite option every time, which might be annoying in the long-term run. Thoughts about that?

Regarding: 

> can we add some text for "MP3 V0" or whatever the selected format is?

Yes. Very considered

sarayourfriend:
> However, in all audio views, the default option will be the same and the user may have to select the favorite option every time, which might be annoying in the long-term run. Thoughts about that?

In implementing the dropdown we could also store their preference in `localStorage` so that their previously selected format is pre-selected the next time they visit an audio result page.

zackkrida:
> In implementing the dropdown we could also store their preference in `localStorage` so that their previously selected format is pre-selected the next time they visit an audio result page.

Which we already do for filter sidebar visibility!


Would the idea be that a click of a drop down choice 1. Downloads the file and 2. Sets the default download format to the clicked format?

panchovm:
That sounds excellent. I have an idea but I need to make it.

sarayourfriend:
> Would the idea be that a click of a drop down choice 1. Downloads the file and 2. Sets the default download format to the clicked format?

What's what I was thinking, if clicking a dropdown choice does indeed trigger a download. If not then just step 2.

sarayourfriend:
[...]

PR #213, opened 2021-09-14
Add DownloadButton component

sarayourfriend:
## Fixes

Related to #149 by @zackkrida 

## Description
Adds a new DownloadButton component based on the DropdownButton component.

## Technical details
Adds the `filesize` dependency which is a function for displaying filesizes with their units appropriatly. It's pretty small: https://bundlephobia.com/package/filesize

## Tests
Most of the behavior of this component is already tested in the DropdownButton component's tests. The behavior around the localStorage would be worth testing though, so I'll add some unit tests for that (assuming JSDOM supports it)

To test this locally run `npm run storybook` and navigate to `/?path=/story/components-downloadbutton--default` to test the component manually. It should even let you download the files 🙂 

## Screenshots
Captura de Tela 2021-09-14 às 09 03 12

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/DownloadButton.vue
+++ b/src/components/DownloadButton.vue
@@ -5,7 +5,7 @@
         v-bind="buttonProps"
         class="whitespace-nowrap"
         :href="selectedFormat.download_url"
-        :download="downloadName"
+        download=""
       >
         <span>{{ $t('download-button.download') }}</span>
         <span class="ml-4 selected-format">
@@ -69,10 +69,6 @@ export default {
         return formats.every((format) => properties.every((p) => p in format))
       },
     },
-    fileName: {
-      type: String,
-      required: true,
-    },
   },
   data() {
     const savedFormatExtension =
@@ -89,11 +85,6 @@ export default {
 
     return { selectedFormat: format }
   },
-  computed: {
-    downloadName() {
-      return `${this.fileName} - ${this.selectedFormat.extension_name}`
-    },
-  },
   methods: {
     getFormatSize(size) {
       return filesize(size, { locale: this.$i18n.locale })
@@ -105,7 +96,6 @@ export default {
       )
       this.selectedFormat = format
     },
-    onClick() {},
   },
 }
 </script>
--- a/src/components/DownloadButton.vue
+++ b/src/components/DownloadButton.vue
@@ -42,7 +42,7 @@
               setSelectedFormat(format)
               toggleOpen()
             "
-            @keydown="onItemKeydown"
+            @keydown="onItemKeydown($event)"
           >
             <span class="font-bold mr-2">{{ format.extension_name }}</span>
             <span>{{ getFormatSize(format.extension_name) }}</span>
@@ -57,17 +57,19 @@
 import filesize from 'filesize'
 import axios from 'axios'
 import local from '~/utils/local'
+import DropdownButton from '~/components/DropdownButton'
 
 const LS_DOWNLOAD_FORMAT_EXTENSION_KEY = 'openverse:download-format-extension'
 
 export default {
   name: 'DownloadButton',
+  components: { DropdownButton },
   props: {
     formats: {
       type: Array,
       required: true,
       validator: (formats) => {
-        const properties = ['extension_name', 'download_url', 'filesize']
+        const properties = ['extension_name', 'download_url']
         return formats.every((format) => properties.every((p) => p in format))
       },
     },
--- a/src/components/DownloadButton.vue
+++ b/src/components/DownloadButton.vue
@@ -56,6 +56,7 @@
 <script>
 import filesize from 'filesize'
 import axios from 'axios'
+import local from '~/utils/local'
 
 const LS_DOWNLOAD_FORMAT_EXTENSION_KEY = 'openverse:download-format-extension'
 
@@ -73,7 +74,7 @@ export default {
   },
   data() {
     const savedFormatExtension =
-      localStorage.getItem(LS_DOWNLOAD_FORMAT_EXTENSION_KEY) ?? null
+      local.get(LS_DOWNLOAD_FORMAT_EXTENSION_KEY) ?? null
     let format = this.formats[0]
     if (savedFormatExtension) {
       const found = this.formats.find(
@@ -114,10 +115,7 @@ export default {
       return filesize(size, { locale: this.$i18n.locale })
     },
     setSelectedFormat(format) {
-      localStorage.setItem(
-        LS_DOWNLOAD_FORMAT_EXTENSION_KEY,
-        format.extension_name
-      )
+      local.set(LS_DOWNLOAD_FORMAT_EXTENSION_KEY, format.extension_name)
       this.selectedFormat = format
     },
   },
--- a/src/components/DownloadButton.vue
+++ b/src/components/DownloadButton.vue
@@ -1,0 +1,111 @@
+<template>
+  <DropdownButton>
+    <template #default="{ buttonProps }">
+      <a
+        v-bind="buttonProps"
+        class="whitespace-nowrap"
+        :href="selectedFormat.download_url"
+        :download="downloadName"
+      >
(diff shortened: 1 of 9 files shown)


zackkrida:
This is looking great 🤩 
One item of interest, there's a small localStorage wrapper in https://github.com/WordPress/openverse-frontend/blob/main/src/utils/local.js you can use, with `local.get(key)` and `local.set(key, value)` methods. This exists to include a simple `window` check in SSR scenarios.

sarayourfriend:
Awesome, thanks for the heads up @zackkrida!

test/unit/specs/components/download-button.spec.js
```diff
@@ -0,0 +1,84 @@
...
+  await DownloadButton.fetch.call(wrapper.vm)
+  return wrapper
+}
+
+describe('DownloadButton', () => {
+  beforeEach(() => {
+    axios.head.mockReset()
+  })
+
+  // TODO(@sarayourfriend) convert this to testing-library once it's possible to get the `vm` from the wrapper
```
sarayourfriend, line 54: This is blocked by this PR https://github.com/testing-library/vue-testing-library/pull/250

src/components/DownloadButton.vue
```diff
@@ -0,0 +1,125 @@
...
+import local from '~/utils/local'
+import DropdownButton from '~/components/DropdownButton'
+
+const LS_DOWNLOAD_FORMAT_EXTENSION_KEY = 'openverse:download-format-extension'
+
+export default {
+  name: 'DownloadButton',
+  components: { DropdownButton },
+  props: {
+    formats: {
```
obulat, line 68: Very thorough validation!

src/components/DownloadButton.vue
```diff
@@ -0,0 +1,125 @@
...
+          <button
+            class="flex justify-between w-full"
+            :class="[
+              itemClass,
+              selectedFormat.extension_name === format.extension_name
+                ? activeItemClass
+                : '',
+            ]"
+            type="button"
+            v-bind="itemA11yProps"
```
obulat, line 40: Can we change the `aria-label` from dropdown to something like 'Select the download file format'?

obulat:
The button works and looks amazing! I have a couple of questions about functionality. 
When you open a dropdown and select the filetype, shouldn't the dropdown close after selection? That was my expectation.
When using keyboard to navigate, do you click on <kbd>Shift</kbd><kbd>Tab</kbd> to go back from the dropdown to download button?

obulat (approved):

src/components/DownloadButton.vue
```diff
@@ -0,0 +1,125 @@
...
+    return { selectedFormat: format, filesizes: null }
+  },
+  async fetch() {
+    const extensionsToFilesizes = await Promise.all(
+      this.formats.map(async (format) => {
+        try {
+          const response = await axios.head(format.download_url)
+          return [format.extension_name, response.headers['content-length']]
+        } catch (e) {
+          return [format.extension_name, undefined]
```
zackkrida, line 99: Minor point but I love seeing undefined in use 😄

sarayourfriend:
> When you open a dropdown and select the filetype, shouldn't the dropdown close after selection?

It should be doing that, if it's not it's a bug 😅 

> When using keyboard to navigate, do you click on ShiftTab to go back from the dropdown to download button?

Yup! I think that matches the behavior on the WARIA examples but I'll double check.

src/components/DownloadButton.vue
```diff
@@ -0,0 +1,125 @@
...
+      local.get(LS_DOWNLOAD_FORMAT_EXTENSION_KEY) ?? null
+    let format = this.formats[0]
+    if (savedFormatExtension) {
+      const found = this.formats.find(
+        (format) => format.extension_name === savedFormatExtension
+      )
+      if (found) {
+        format = found
+      }
+    }
```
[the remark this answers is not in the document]
sarayourfriend: Oh huh, what's the advantage to do it that way?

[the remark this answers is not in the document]
zackkrida: I tried to research a bit and I can't find anything, so you can leave this as is, fwiw the execution order is:

1. `data()`
2. `created()`
3. `mounted()`

sarayourfriend:
> I think there's still an issue with events firing multiple times, perhaps due to some event bubbling. Keyboard controls work flawlessly but clicks seem unpredictable. This video isn't the most helpful but should show some instances where 2,3, or even 5 clicks are necessary to reveal the dropdown.

For some reason I'm not able to reproduce this... or the bug where clicking doesn't cause the dropdown to open 🤔 I'll keep digging and see if I can identify the problem though.

sarayourfriend:
So I've just done some experimenting and discovered something interesting... I wonder if it's at the root cause of what's going on here. When you `@click` bind an event in Vue on a `button`, it will also trigger the click when you hit space or enter... So if you combine `@click` with `@keydown` it'll fire BOTH events at the same time! You can see this happening here in this codesandbox:

https://codesandbox.io/s/reverent-lucy-j7kf3?file=/src/App.vue

I have absolutely no idea how to circumvent this... I suppose we could just not use the `button` element and use a `div` with some appropriate `aria` attributes applied... but is that the right solution? Or is there a different vue-y way to do this?

zackkrida:
CleanShot 2021-09-23 at 13 28 20@2x

Also, small Safari bug where the button needs `appearance-none`

zackkrida:
@sarayourfriend perhaps we just remove the keydown events? Is that all we need?

sarayourfriend:
> perhaps we just remove the keydown events? Is that all we need?

We can do that for the dropdown button I think... but for the format buttons I don't think it's possible because we need to respond to specific keys (tab, arrows, page down/up, end/home, etc) as well as the click event... I wonder if there's a way of telling the click event to only fire if it's a true click, given we're handling the keydown event ourselves 🤔

sarayourfriend:
Apparently you can use `https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail` on the `MouseEvent` to tell whether it was triggered by a `click` or a keypress 🤔

sarayourfriend:
Hmmm, actually we can't do that because we are depending on the `@click` behavior to handle when the dropdown item is actually selected... `onItemKeydown` only handles navigational keys, not selection (which makes sense because the `DropdownButton` shouldn't know anything outside of the a11y interactions for each item...)

sarayourfriend:
Additionally, I'm just noticing that because the link for the download button looks like a button, it doesn't actually behave like a button (<kbd>Space</kbd> doesn't trigger the download, in fact it scrolls the screen) so I wonder if there's something we need to do there to handle that situation?

zackkrida:
> Additionally, I'm just noticing that because the link for the download button looks like a button, it doesn't actually behave like a button (Space doesn't trigger the download, in fact it scrolls the screen) so I wonder if there's something we need to do there to handle that situation?

@sarayourfriend I think it's okay for it to behave like a link and look like a button.

sarayourfriend:
I think I figured out the issue with clicking the dropdown button. The click only triggers if you click the button and not the icon 🤯

zackkrida:
@sarayourfriend `pointer-events: none` on the icon then should do it!

sarayourfriend:
Thanks @zackkrida !

sarayourfriend:
> The dropdown does not close after selecting an item, again I think because of an event double fire

I still can't figure this out... or reproduce it 😞

zackkrida (approved):
This now looks and works perfectly 🥳. The only bit of feedback I have is that I'd like to figure out how to show the focus ring on dropdown items only when the user is using the keyboard. If someone is using the mouse with the dropdown, it's confusing to see two interaction states: the gray background of the selected item and the pink ring defaulted to the first item.

PR #239, opened 2021-09-17
Change AudioTrack component according to filters sidebar

### LINKED ISSUE #238: [Bug] Update component in Audio results page
## Description

Audio results page is broke after the update of the Audio Track component.

## Reproduction

1. Run the project as usually `npm run dev`
2. Search anything, for example `space` and click in the Audio tab
3. See error.
 
CleanShot 2021-09-17 at 10 35 13@2x

## Expectation

This page should show the list of audio results with the new **row medium** component if there are results that match the search query, otherwise, it should show there are no audios found without crashing. It's desirable to add tests that prevent these errors.

## Resolution

krysal:
## Fixes

Fixes #238 by @krysal

## Description

This PR updates the AudioTrack component to render even if the `audio` object doens't specify `duration` or `category`. Also, changes the variant according to if the filter sidebar is visible or not, as in mockups [[ref1][mockup1], [ref2][mockup2]]

[mockup1]: https://www.figma.com/file/vEoItpMQyFZb5AZV9aoP7I/Audio-integration?node-id=1070%3A10966
[mockup2]: https://www.figma.com/file/vEoItpMQyFZb5AZV9aoP7I/Audio-integration?node-id=1147%3A12234

## Technical details

Adds an 'Undefined category' to avoid showing the locale key and shows `--:--` in case the duration of the track is unknown. These fields should always be present once we add them to the API, but I believe they are not strictly required to show the component.

## Tests

I added some minimal tests to verify the component is working, we should add more to test for accessibility aspects of the Audio player but I don't want to block this issue any longer.

### With sidebar open
sidebar_open

### With sidebar closed
sidebar_closed

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "storybook": "nuxt storybook",
     "test": "jest",
     "test:watch": "jest --watch",
-    "test:unit": "jest",
+    "test:unit": "jest --collectCoverage=false",
     "test:e2e": "cypress open",
     "test:e2e:ci": "cypress run",
     "lint": "eslint --ext .js,.vue --ignore-path .gitignore .",
--- a/src/components/AudioTrack/layouts/Full.vue
+++ b/src/components/AudioTrack/layouts/Full.vue
@@ -50,6 +50,7 @@ export default {
         date.setSeconds(ms / 1e3)
         return date.toISOString().substr(11, 8).replace(/^00:/, '')
       }
+      return '--:--'
     }
 
     return {
--- a/src/components/AudioTrack/layouts/Full.vue
+++ b/src/components/AudioTrack/layouts/Full.vue
@@ -45,9 +45,11 @@ export default {
      * @returns {string} the duration in a human-friendly format
      */
     const timeFmt = (ms) => {
-      const date = new Date(0)
-      date.setSeconds(ms / 1e3)
-      return date.toISOString().substr(11, 8).replace(/^00:/, '')
+      if (ms) {
+        const date = new Date(0)
+        date.setSeconds(ms / 1e3)
+        return date.toISOString().substr(11, 8).replace(/^00:/, '')
+      }
     }
 
     return {
--- a/src/components/AudioTrack/layouts/Row.vue
+++ b/src/components/AudioTrack/layouts/Row.vue
@@ -7,7 +7,7 @@
       class="flex"
       :class="isSmall ? 'flex-row gap-8' : 'flex-col justify-between'"
     >
-      <div class="flex-shrink-0">
+      <div class="flex-shrink-0" :class="isSmall ? 'w-96' : ''">
         <NuxtLink
           :to="localePath(`/audio/${audio.id}`)"
           class="font-heading font-semibold text-2xl"
@@ -35,12 +35,12 @@
             </span>
           </div>
 
-          <div class="part-b">
-            <template v-if="isSmall">
+          <div class="part-b inline-flex space-x-1">
+            <div v-if="isSmall">
               {{ timeFmt(audio.duration) }} {{ $t('interpunct') }}
               {{ $t(`audio-categories.${audio.category}`) }}
               {{ $t('interpunct') }}
-            </template>
+            </div>
             <License class="inline" :license="audio.license" />
           </div>
         </div>
--- a/src/components/AudioTrack/layouts/Row.vue
+++ b/src/components/AudioTrack/layouts/Row.vue
@@ -7,7 +7,7 @@
       class="flex"
       :class="isSmall ? 'flex-row gap-8' : 'flex-col justify-between'"
     >
-      <div class="flex-shrink-0" :class="isSmall ? 'w-96' : ''">
+      <div class="flex-shrink-0" :class="isSmall ? 'w-70' : ''">
         <NuxtLink
           :to="localePath(`/audio/${audio.id}`)"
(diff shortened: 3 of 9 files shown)


krysal:
I added width to the 'small row' variant (second screenshot) to fix the misalignment of audio players and fit the info in one line but the long name of [retired licenses](https://creativecommons.org/licenses/nc-sampling+/1.0/) may put that info out of place anyway. CC @panchovm

package.json
```diff
@@ -19,7 +19,7 @@
     "storybook": "nuxt storybook",
     "test": "jest",
     "test:watch": "jest --watch",
-    "test:unit": "jest",
+    "test:unit": "jest --collectCoverage=false",
```
dhruvkb, line 22: Why not collect coverage?

src/components/AudioTrack/layouts/Full.vue
```diff
@@ -45,9 +45,11 @@ export default {
...
      */
     const timeFmt = (ms) => {
-      const date = new Date(0)
-      date.setSeconds(ms / 1e3)
-      return date.toISOString().substr(11, 8).replace(/^00:/, '')
+      if (ms) {
+        const date = new Date(0)
+        date.setSeconds(ms / 1e3)
+        return date.toISOString().substr(11, 8).replace(/^00:/, '')
+      }
```
dhruvkb, line 52: The return statement from [here](https://github.com/WordPress/openverse-frontend/pull/239/files#diff-7f028a3652b640df123de004df0126bc9acecee3dc7d52a724ceaf52506b6897R79) should be added.

src/components/AudioTrack/layouts/Row.vue
```diff
@@ -7,7 +7,7 @@
       class="flex"
       :class="isSmall ? 'flex-row gap-8' : 'flex-col justify-between'"
     >
-      <div class="flex-shrink-0">
+      <div class="flex-shrink-0" :class="isSmall ? 'w-96' : ''">
```
dhruvkb, line 10: Instead of a fixed width, maybe a `max-width` or `min-width` would be preferable? Using hardcoded widths might negatively affect responsiveness.

src/locales/en.json
```diff
@@ -617,7 +617,8 @@
     "title": "Audio Category",
     "music": "Music",
     "sound-effects": "Sound Effects",
-    "podcast": "Podcast"
+    "podcast": "Podcast",
+    "undefined": "Undefined category"
```
dhruvkb, line 621: @panchovm Can we omit the category if we don't know instead of showing "Undefined category"?

dhruvkb (approved):
Some minor non-blocking questions. It looks great to me otherwise!

Ideally we would want the left and right column (info and waveform respectively) to have fixed width percentages/pixels (with the option to either reflow or truncate any long text). But these are design-level questions so deferring to @panchovm might be wiser.

src/locales/en.json
panchovm: Yes. Definitely.

panchovm:
I wonder if we can use the name **NC Sampling Plus 1.0** instead of **NonCommercial Sampling Plus 1.0**, or the icons without the license name. It seems that CC did not work on short versions of [retired licenses](https://creativecommons.org/retiredlicenses) as shown below.

Screen Shot 2021-09-28 at 08 57 50

package.json
krysal: While writing the tests, I don't wanted to see the coverage table every time (it's long!), only the test results. The `test` and `test:unit` command are the same so I though we can modify one to skip the table. I can revert it if this was made for a specific purpose of course.

src/locales/en.json
krysal: Deleting then... 🧹  Thanks!

src/components/AudioTrack/layouts/Row.vue
krysal: I was thinking the same but sadly with `min-w-` or `max-` the [tracks are misaligned](https://cln.sh/mTkrfg).

src/components/AudioTrack/layouts/Full.vue
krysal: Added in [`14e8072`](https://github.com/WordPress/openverse-frontend/pull/239/commits/14e80720abebc9a4479b75affd08b2b9a662a01e).

krysal:
@panchovm We're showing the `sampling+` license like this in the audio details:

CleanShot 2021-09-28 at 13 44 13@2x

We need to standardize the rest too. However, I'm leaning towards displaying the icons without the license name in the "row small" variant.

zackkrida:
> icons without the license name in the "row small" variant.

I think that's a great idea @krysal

krysal:
@zackkrida can you take a look at this? As Sara is AFW this week. We can refine the component styles in successive improvements, as this was mainly to fix the audio results page.

zackkrida (approved):
This is great work! One thing I noticed—it's odd when you show/hide the filters on mobile and it changes the audio layout. Probably need to do something with `window.matchMedia` to make sure it's the same component on mobile regardless of filter visibility.

Also, we need to make the `AudioTrack` look better on mobile regardless of which layout is used. I'll make a separate ticket there

krysal:
Thanks!

Issue #280, opened 2021-10-01
[Bug] Media fetching status and errors are all combined


obulat:
## Description

Currently, all fetch state and errors for _**single image**_, _**image search result**_ and _**related images**_ are saved into two Vuex state variables: 
- `isFetching.images`
- `isFetchingError.images`

This is wrong, and sometimes causes error messages for irrelevant image fetch error.

## Expectation

The fetch data for three different fetch points should be kept separately. Only the search result data needs to be kept in Vuex store, other data can be kept in the components.

## Resolution

- [x] 🙋 I would be interested in resolving this bug.

PR #281, opened 2021-10-01
Add ImageGrid component and use it for RelatedImages

obulat:
## Fixes

Related #280 by @obulat

## Description

This PR extracts the functionality of `SearchGridManualLoad` that has to do with displaying images and an optional `Load more` button into a new `ImageGrid` component.
`RelatedImages` component is refactored to be responsible for fetching related images, and use `ImageGrid` to display them. This makes the `images` part of `related` Vuex store unnecessary, and also simplifies the Vuex `isFetching` and `isFetchingError` states: now they are only changed by Image search and Single image result [which should actually control its state separately].

I think the `SearchGrid` component will also need to be refactored to use the `ImageGrid` component later.
The `ImageCell` is copied as-is from the `SearchGridCell` to ensure that possible changes later do not introduce unwanted bugs.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/test/unit/specs/components/image-grid.spec.js
+++ b/test/unit/specs/components/image-grid.spec.js
@@ -1,0 +1,82 @@
+import ImageGrid from '~/components/ImageGrid/ImageGrid'
+import { render, screen } from '@testing-library/vue'
+
+const propsData = {
+  images: undefined,
+  canLoadMore: undefined,
+  isFetching: undefined,
+  fetchingError: undefined,
+  errorMessageText: undefined,
+  isFinished: undefined,
+}
+
+test('ImageGrid loads with default values', async () => {
+  render(ImageGrid, {
+    props: propsData,
+  })
+  expect(screen.getByRole('button')).toHaveTextContent('browse-page.load')
+})
+
+test('ImageGrid renders images without load more button if canLoadMore is false', async () => {
+  let propsData = {
+    images: [
+      { id: 'image1', url: 'http://localhost:8080/image1.png' },
+      { id: 'image2', url: 'http://localhost:8080/image2.jpg' },
+      { id: 'image3', url: 'http://localhost:8080/image3.svg' },
+      { id: 'image4', url: 'http://localhost:8080/image4.jpeg' },
+    ],
+    canLoadMore: false,
+  }
+  render(ImageGrid, {
+    props: propsData,
+    stubs: ['NuxtLink', 'LicenseIcons'],
+  })
+  expect(screen.queryAllByRole('img').length).toEqual(propsData.images.length)
+  expect(screen.queryAllByRole('figure').length).toEqual(
+    propsData.images.length
+  )
+  expect(screen.queryByRole('button')).not.toBeInTheDocument()
+})
+
+test('ImageGrid renders images and load more button if canLoadMore is true', async () => {
+  let propsData = {
+    images: [
+      { id: 'image1', url: 'http://localhost:8080/image1.png' },
+      { id: 'image2', url: 'http://localhost:8080/image2.jpg' },
+      { id: 'image3', url: 'http://localhost:8080/image3.svg' },
+      { id: 'image4', url: 'http://localhost:8080/image4.jpeg' },
+    ],
+    canLoadMore: true,
+  }
+  render(ImageGrid, {
+    props: propsData,
+    stubs: ['NuxtLink', 'LicenseIcons'],
+  })
+  expect(screen.queryAllByRole('img').length).toEqual(propsData.images.length)
+  expect(screen.queryAllByRole('figure').length).toEqual(
+    propsData.images.length
+  )
+  const loadMoreButton = screen.queryByRole('button')
+  expect(loadMoreButton).toBeVisible()
+  expect(loadMoreButton).toHaveTextContent('browse-page.load')
+})
+
+test('ImageGrid shows LoadingIcon instead of LoadMoreButton when isFetching', async () => {
+  let propsData = {
+    canLoadMore: true,
+    images: [
+      {
+        id: 'image1',
+        url: 'http://localhost:8080/image1.png',
+        title: 'image1',
+      },
+    ],
+    isFetching: true,
+  }
+  render(ImageGrid, {
+    props: propsData,
+    stubs: ['NuxtLink', 'LicenseIcons'],
+  })
+  // getByRole('button') does not find the button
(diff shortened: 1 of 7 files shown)


src/composables/use-related.js
```diff
@@ -0,0 +1,26 @@
...
+  // in components, so there's no need to export them,
+  // see https://composition-api.nuxtjs.org/lifecycle/usefetch/
+  // eslint-disable-next-line no-unused-vars
+  const { fetch, fetchState } = useFetch(async () => {
+    const response = await service.getRelatedMedia({
+      id: mediaId.value,
+    })
+    media.value = response.data.results
+  })
+  fetch()
```
zackkrida, line 24:
```suggestion
  useFetch(async () => {
    const response = await service.getRelatedMedia({
      id: mediaId.value,
    })
    media.value = response.data.results
  }).fetch()
```

Perhaps _this_ style would look less weird, for hooks that don't return anything and are just being called? it does feel extremely weird for `fetch` and `fetchState` not to be returned and not declaring them as vars might help ease that confusion.

cc @sarayourfriend is this an awful idea lol?

zackkrida: The linked documentation shows something weird:

https://composition-api.nuxtjs.org/lifecycle/usefetch/

```vue
import { defineComponent, ref, useFetch } from '@nuxtjs/composition-api'
import axios from 'axios'

  setup() {
    const name = ref('')

    const { fetch, fetchState } = useFetch(async () => {
      name.value = await axios.get('https://myapi.com/name')
    })

    // Manually trigger a refetch
    fetch()

    // Access fetch error, pending and timestamp
    fetchState

    return { name }
  },
})

```

What's with fetchState not being called? Is that just a typo?

zackkrida (approved):
Great. There used to be such an irritating bug where if loading related images failed, the site would show related images from the _previous_ image, due to these being held in a global store.

src/components/ImageDetails/RelatedImages.vue
```diff
@@ -0,0 +1,55 @@
...
+import useRelated from '~/composables/use-related'
+import { IMAGE } from '~/constants/media'
+
+export default {
+  name: 'RelatedImages',
+  props: {
+    imageId: {
+      type: String,
+      default: null,
+    },
```
krysal, line 30: I would think that at least the `imageId` should be mandatory for this component. What happens if is not present?
```suggestion
    imageId: {
      type: String,
      required: true,
    },
```

krysal (commented):
Now these components seem more in line with the "Single-responsibility principle". Awesome!

src/composables/use-related.js
sarayourfriend: `fetchState` is an object I believe... I think the example you shared is just indicating that you _can do something_ with it there.

I think it's fine to destructure `fetch` and call it separately... I don't have a preference there really (it's probably easier to read as seprate lines?) 

We should avoid destructuring unused variables though (like `fetchState`) and just throw them out completely... There's no need to get a reference to it anyway so I think it's better to just do `const { fetch } = ...` instead.

zackkrida: You're right, `fetchState` is an object so that must just be an example. Thanks for the suggestion and I've updated my _code_ suggestion to remove `fetchState` from the destructuring assignment.

src/components/AudioDetails/Related.vue
```diff
@@ -3,22 +3,56 @@
...
+        v-for="audio in audios"
+        :key="audio.id"
+        :audio="audio"
+        layout="row"
+        size="m"
+        class="mb-12"
+      />
+      <LoadingIcon v-show="$fetchState.pending" />
+    </template>
+    <p v-show="!!$fetchState.error">
```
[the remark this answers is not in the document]
obulat: `fetchState.error` can be null, if there is no error, or be an object containing the error message. Double negation is to convert this to Boolean: if it's null, it would be converted to false (`null`, `!null` = `true`, `!!null` = `!true`), and if it's an object, it would be converted to true (`{error: ''}`, `!{error: ''}` = `false`, !!{error: ''}` = `true`).

src/components/ImageDetails/RelatedImages.vue
obulat: Definitely true!

test/unit/specs/components/related-images.spec.js
```diff
@@ -1,14 +1,45 @@
-import RelatedImage from '~/components/RelatedImages'
+import RelatedImages from '~/components/ImageDetails/RelatedImages'
```
sarayourfriend, line 1: Would it be possible to conver this test suite to use `testing-library`? We're re-writing it already 🙂

test/unit/specs/components/related-images.spec.js
```diff
@@ -1,14 +1,45 @@
...
-      },
-    }
-    const wrapper = render(RelatedImage, options)
-    expect(wrapper.find('.photo_related-images').element).toBeDefined()
+  it('should render content when finished loading related images', async () => {
+    const wrapper = await doRender()
+
+    const header = wrapper.find('h3').text()
+    expect(header).toEqual('photo-details.related-images')
+    console.log(wrapper.html())
```
sarayourfriend, line 38: Looks like a stray console.log here 😱

test/unit/specs/components/related-audios.spec.js
```diff
@@ -0,0 +1,46 @@
+import RelatedAudios from '~/components/AudioDetails/Related'
+import render from '../../test-utils/render'
```
sarayourfriend, line 2: Here too, could we use `testing-library`?

test/unit/specs/components/image-grid.spec.js
```diff
@@ -0,0 +1,82 @@
...
+const propsData = {
+  images: undefined,
+  canLoadMore: undefined,
+  isFetching: undefined,
+  fetchingError: undefined,
+  errorMessageText: undefined,
+  isFinished: undefined,
+}
+
+test('ImageGrid loads with default values', async () => {
```
sarayourfriend, line 13: Rather than repeating `ImageGrid` at the start of each test, could we wrap the whole suite in a `describe('ImageGrid')` block?

sarayourfriend (commented):
I love getting rid of global state 🎉 I'm running into some local dev issues so I can't test it though 😢

PR #288, opened 2021-10-04
Add download button to the single audio result page

### LINKED ISSUE #270: [Feature] Implement download buttons on single audio result view
## Problem

The single audio result view is missing the download dropdown for the audio files.

## Description

Now that #213 has landed we can implement the download buttons on the single audio result view.

## Implementation

obulat:
## Fixes

Fixes #270 by @zackkrida

## Description

This PR adds `DownloadButton` component to the Single audio result page: 

download_audio

Currently, the file extension is being set in the Catalog as `mp3`. However, Jamendo provides two types of `mp3` of different quality. Should we also set the filetype as mp31 and mp32 as Jamendo does?

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/AudioTrack/layouts/Full.vue
+++ b/src/components/AudioTrack/layouts/Full.vue
@@ -2,7 +2,7 @@
   <div class="full-track w-full">
     <slot name="controller" />
 
-    <div class="flex flex-row space-between mx-16 my-6">
+    <div class="flex flex-row justify-between mx-16 my-6">
       <div class="left-content flex flex-row items-center gap-6">
         <slot name="play-pause" />
 
@@ -28,8 +28,10 @@
         </div>
       </div>
 
-      <!-- TODO: Download dropdown -->
-      <div class="right-content bg-pink h-12 w-30 rounded-sm ml-auto" />
+      <DownloadButton
+        class="right-content h-14 flex"
+        :formats="getFormats(audio)"
+      />
     </div>
   </div>
 </template>
@@ -52,9 +54,30 @@ export default {
       }
       return '--:--'
     }
-
+    /**
+     * Creates a list of { extension_name, download_url } objects
+     * for DownloadButton
+     * @param {AudioDetail} audio
+     */
+    const getFormats = (audio) => {
+      let formats = [
+        { extension_name: audio.filetype, download_url: audio.url },
+      ]
+      if (audio.alt_files) {
+        console.log('audio has alt_files')
+        formats = formats.concat(
+          audio.alt_files.map((altFile) => ({
+            extension_name: altFile.extension,
+            download_url: altFile.url,
+          }))
+        )
+        console.log('formats: ', formats)
+      }
+      return formats
+    }
     return {
       timeFmt,
+      getFormats,
     }
   },
 }
--- a/src/components/DropdownButton.vue
+++ b/src/components/DropdownButton.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="relative block max-w-min">
+  <div class="relative flex max-w-min">
     <!-- rounded-X-none is required to fight Edge UA styles that apply a 2px border radius to all `button` elements -->
     <div class="flex">
       <slot
@@ -11,7 +11,7 @@
       <button
         ref="dropdownButton"
         type="button"
-        class="dropdown-button ml-1 rounded-r-sm rounded-l-none"
+        class="dropdown-button ml-1 rounded-r-sm rounded-l-none w-14"
         :class="{ 'dropdown-button-active': isOpen }"
         aria-haspopup="menu"
         :aria-label="safeDropdownAriaLabel"
--- a/src/data/sample-audio-responses.json
+++ b/src/data/sample-audio-responses.json
@@ -30,6 +30,7 @@
             "name": "vocal"
           }
         ],
+        "filetype": "mp3",
         "fields_matched": ["tags.name"],
         "genres": ["classical", "gospel"],
         "detail_url": "http://localhost:8000/v1/audio/8071128a-bebc-4c24-823d-6e26a2e6dfe2",
@@ -58,6 +59,7 @@
             "name": "vocal"
           }
         ],
+        "filetype": "mp3",
         "fields_matched": ["tags.name"],
         "genres": ["metal"],
         "detail_url": "http://localhost:8000/v1/audio/73e00493-9189-42ed-955b-4f9d26d0d9f4",
@@ -250,6 +252,14 @@
       "name": "Distance & Temps",
       "url": "https://www.jamendo.com/album/1290/distance-and-temps"
     },
+    "alt_files": [{
+      "extension": "flac",
+      "url": "https://mp3d.jamendo.com/download/track/8561/flac"
+      },
+      {
+      "extension": "mp31",
+      "url": "https://mp3d.jamendo.com/download/track/8561/mp31"
+    }],
     "duration": 248000,
     "detail_url": "http://localhost:8000/v1/audio/242c1a4c-0623-436c-a600-7847412d28b6",
     "related_url": "http://localhost:8000/v1/audio/242c1a4c-0623-436c-a600-7847412d28b6/recommendations"


sarayourfriend:
> Should we also set the filetype as mp31 and mp32 as Jamendo does

This is where I think we still need to talk about how to store the file format name. I separated the format name from the URL because I assumed the true format name couldn't be derived from some download URLs (this is an example). We should have an MP3 96kbs and an MP3 V0 for jamendo for each of their available MP3 download formats.

src/components/AudioTrack/layouts/Full.vue
```diff
@@ -52,9 +54,30 @@ export default {
...
+     * Creates a list of { extension_name, download_url } objects
+     * for DownloadButton
+     * @param {AudioDetail} audio
+     */
+    const getFormats = (audio) => {
+      // TODO: finalize the property names in
+      //  formats objects and alt_files objects
+      let formats = [
+        { extension_name: audio.filetype, download_url: audio.url },
+      ]
```
sarayourfriend, line 67: If the `audio` object already uses `filetype` and `url` for the properties I think we should just update the download button component to accept those properties, so we can just pass a list of audio objects instead.

The only concern is whether `filetype` is a truly human readable filetype (like it needs to be `MP3 96kbs` instead of `mp31` for example)

src/data/sample-audio-responses.json
```diff
@@ -250,6 +252,14 @@
...
       "url": "https://www.jamendo.com/album/1290/distance-and-temps"
     },
+    "alt_files": [{
+      "extension": "flac",
+      "url": "https://mp3d.jamendo.com/download/track/8561/flac"
+      },
+      {
+      "extension": "mp31",
+      "url": "https://mp3d.jamendo.com/download/track/8561/mp31"
+    }],
```
sarayourfriend, line 262: Why use `extension` here instead of `filetype` to be consistent with the main result?

sarayourfriend (commented):
Sorry for the early review but just some preliminary comments and thoughts. Otherwise I'm hopeful that this looks like it shouldn't be too bad to implement using the component 😁

src/components/AudioTrack/layouts/Full.vue
```diff
@@ -28,8 +28,10 @@
         </div>
       </div>
 
-      <!-- TODO: Download dropdown -->
-      <div class="right-content bg-pink h-12 w-30 rounded-sm ml-auto" />
+      <DownloadButton
+        class="right-content h-14 flex"
```
zackkrida, line 32:
```suggestion 
```

These don't seem to have any affect on the styling. Removing them doesn't change the appearance

src/components/AudioTrack/layouts/Full.vue
zackkrida: Yeah we might want to keep an enum of human readable file name mappings here, or more likely push this down to the catalog level and prettify the `format` before saving the rows.

src/components/DropdownButton.vue
```diff
@@ -11,7 +11,7 @@
       <button
         ref="dropdownButton"
         type="button"
-        class="dropdown-button ml-1 rounded-r-sm rounded-l-none"
+        class="dropdown-button ml-1 rounded-r-sm rounded-l-none w-14"
```
zackkrida, line 14: I don't love the idea of a hardcoded width inside of this component, I'd rather see this be `w-full` and do the `w-14` on a parent at the page, implementation-level.

src/data/sample-audio-responses.json
zackkrida: yep, might be best to match the catalog fields, no?

https://github.com/WordPress/openverse-catalog/blob/a63fdf81755632f9c6ba83c822fd0fe8eef853a6/openverse_catalog/dags/provider_api_scripts/freesound.py#L198-L204

src/components/DropdownButton.vue
sarayourfriend: I think a fixed width for the dropdown arrow button is fine... the download button should expand to fill the remaining space (`w-full` I guess?). I tried to implement that with flex but I guess I didn't do it right? I don't think we need to actually set explicit widths on anything? Except maybe a min-width on the download button in case the word "download" is very short in any given language.

zackkrida: Oh i totally lazily missed that this was on the dropdown arrow button, oops. I agree with your thoughts @sarayourfriend

src/data/sample-audio-responses.json
obulat: This is the main reason why this PR is a draft: I wanted to discuss and finalized the naming and values for `extension` ( filetype?, human-readable-filetype?). I used `extension` for now because that is what's used in the `DownloadButton` component.
So, to make the naming consistent, I guess we could use `filetype` and `url` properties for main audio, and change the [`alt_files`](https://github.com/WordPress/openverse-catalog/blob/a63fdf81755632f9c6ba83c822fd0fe8eef853a6/openverse_catalog/dags/provider_api_scripts/freesound.py#L198-L204) to have the same properties (filetype instead of format):
```
alt_files: {
        'url': media_data.get('download'),
        'filetype': media_data.get('type'),
        'filesize': media_data.get('filesize'),
        'bit_rate': media_data.get('bitrate'),
        'sample_rate': media_data.get('samplerate'),
}

Then I think it would be easier to have a technical name in the catalog (ie. `mp31`, `mp32`), and convert it to a human-readable form on the front end. So, the front end would have a mapping to convert `mp31` coming from the API to `MP3 96kbs` for display on the button.

src/components/AudioTrack/layouts/Full.vue
obulat: I agree. I thought that `left-content` and `right-content` were left in their for some styling later on, so I decided to leave it in.

src/components/AudioTrack/layouts/Full.vue
obulat: >  I separated the format name from the URL because I assumed the true format name couldn't be derived from some download URLs (this is an example). We should have an MP3 96kbs and an MP3 V0 for jamendo for each of their available MP3 download formats.

There is no _standard_ way of deriving the audio format from the URL: for example, Jamendo has format information after the last `/`, and Wikimedia often has it as an extension on the download URL. This is why we want to add `filetype` as a column to the Catalog database.

src/data/sample-audio-responses.json
obulat: I've opened a discussion about this: https://github.com/WordPress/openverse/discussions/85

zackkrida:
Hm, it looks like a visual regression where the dropdown "drawer" _covers_ the button, instead of appearing below it.

obulat:
> Hm, it looks like a visual regression where the dropdown "drawer" _covers_ the button, instead of appearing below it.
> 
> CC @sarayourfriend would appreciate 👀 here

The regression was caused by changing the `DropdownButton`'s outer container's `display` property from `block` to `flex`, so that it takes up the 100% of parent's height, and becomes the same height as the play-pause button on the left. 
To fix it, I set the `top` property for the dropdown to the height of the play-pause and arrow buttons. But I used a CSS variable set on the `Full` audio track component, because the actual `DropdownButton` that controls the dropdown height is a generic-ish component, and it shouldn't have a set top offset for it. What is the best way of doing it using Tailwind?

Here is the result:
download_audio

sarayourfriend:
> The regression was caused by changing the DropdownButton's outer container's display property from block to flex, so that it takes up the 100% of parent's height, and becomes the same height as the play-pause button on the left.

This is totally naive question, but is it possible to keep it as `block` but use `h-full` to make it fill the parent container's height? Not sure if that's possible or if it works.

obulat:
Copying the message from a PR to a PR :)
I've added some changes:
- made the Full layout top-aligned to prevent button moving downwards on a narrow screen, and adjusted the h1's line height so that it looks more center-aligned anyways. 
- added flex-shrink-0 to the download button so that it stays the same width.
- adjusted padding and line-height (leading-) on the download button so that it's height more or less matches the height of the download button. 

Does it look close to what you envisioned, @panchovm, @sarayourfriend ?
Screen Shot 2021-10-15 at 2 28 05 PM
Screen Shot 2021-10-15 at 2 27 46 PM

PR #301, opened 2021-10-08
Add replay action and played state to audio track

### LINKED ISSUE #289: [Bug] {Audio} Play button doesn't reset when audio track ends
## Description

When an audio track ends, the play/pause button remains in its active state and the track remains at the end.

## Reproduction
1. Play any audio and allow it to reach its end.
2. Observe that the "pause" icon is still visible, instead of the button resetting to the "play" button. 
3. Observe that the waveform progress bar stays at the end.

## Expectation

1. The button should show the play arrow.
2. The track should reset to the starting, unplayed state

sarayourfriend:
## Fixes

Fixes #289 by @zackkrida 

## Description
Adds a new `played` state to the AudioTrack triggered when the current time of the audio matches or exceeds the duration of the audio. When seeking during this state, it toggles back to `paused`. When hitting the play/pause button during the state, it will restart the audio from the beginning.

https://user-images.githubusercontent.com/24264157/136586058-cf0fe93a-fb29-4ce1-a555-40cd132e09c2.mov

I tried to implement this in the least intrusive way possible... In the course of that I added two new events to the AudioControlled (`finished` and `seeked`). The `seeked` event is probably the most problematic as it doesn't emit the timestamp or anything like that, which you could argue would make sense for a `seeked` event to do. But it wasn't necessary for this particular application of that event I was sort of thinking of a YAGNI kind of thing, to keep the event as minimal as possible for the time being. If in the future we need to add the timestamp it would be trivial to do (but I'm also open to adding it now if folks think it would make more sense that way, even though it's not needed for this particular instance).

I tried to think of a few different approaches to solving this problem but then all eventually hit dead ends... this was the least intrusive version of these changes I could come up with.

## Testing Instructions
Checkout this PR and run `npm run storybook` and navigate to `/?path=/story/components-audio-track--default-story`. Play around with the audio track and the new replay functionality and verify that the existing behavior continues to work (especially seeking) and that the new replay behavior works and makes sense.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/assets/icons/replay.svg
+++ b/src/assets/icons/replay.svg
@@ -1,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <path id="icon" fill-rule="evenodd" clip-rule="evenodd" d="M12 8V11L8 7L12 3V6C15.3137 6 18 8.68629 18 12C18 15.3137 15.3137 18 12 18C8.68629 18 6 15.3137 6 12H8C8 14.2091 9.79086 16 12 16C14.2091 16 16 14.2091 16 12C16 9.79086 14.2091 8 12 8Z" fill="currentcolor"/>
+</svg>
--- a/src/components/AudioTrack/AudioController.vue
+++ b/src/components/AudioTrack/AudioController.vue
@@ -61,7 +61,7 @@ export default {
     status: {
       type: String,
       required: true,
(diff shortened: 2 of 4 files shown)


zackkrida:
@sarayourfriend in that case...LGTM!

PR #319, opened 2021-10-14
Fix filter display wrapping and clean up tests

### LINKED ISSUE #269: [Bug] Fix active filter display on medium screens
## Description

When on medium-ish viewports, the text in the 'active filters row' takes up multiple lines and looks terrible! 

## Expectation

We should fix this in one of two ways (cc @panchovm):

1. Make the bar scrollable with a width > 100%, and force a single line of text
2. Truncate the text in each active filter with '...'.

## Resolution

CleanShot 2021-09-28 at 09 13 32@2x

sarayourfriend:
## Fixes

Fixes #269 by @zackkrida 

## Description
Fixes the active filter display by making them flex wrap. Also converts it to tailwind for the most part except for the `button` and `tiny` classnames which I assume we'll get rid of when we do the full redesign anyway.

Also removes all the existing tests for the `FilterDisplay` component in favor of a single snapshot tests.

I don't _love_ snapshot tests, but in this case the tests that existed there previously either (a) didn't actually do anything (in particular the "checked" test was completely useless) or (b) were manual snapshot tests, directly testing the markup. So I just replaced it with a single snapshot test that at least verifies that the component renders as expected and requires updating if anything changes.

It's probably, ultimately, a fragile snapshot test, but I'd personally like to see this component refactored in a number of ways to make it more testable and simpler. For one, we could split it into a container and display component. The container would connect to the store and the display would taking things as normal props. It would massively simplify the tests against the display component because we could just use normal props instead of having to mock out a whole store and getters and stuff.

In any case, that's a project for another day. I think the current set up makes the most sense from a perspective of (a) preserving what testing was already being done while (b) removing tests that weren't actually doing anything.

## Testing Instructions
Checkout the branch and run `npm run dev`. Then perform a search and activate enough filters to cause the active filters list to wrap and verify that it looks okay.

Captura de Tela 2021-10-14 às 10 13 27

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/Filters/FilterTag.vue
+++ b/src/components/Filters/FilterTag.vue
@@ -1,18 +1,12 @@
 <template>
-  <div
+  <button
     class="button tiny tag"
-    :aria-label="$t('filters.filter-tag-aria', { filterLabel: filterLabel })"
+    :aria-label="$t('filters.aria.remove-filter', { label: filterLabel })"
+    @click="onClick"
   >
     <span>{{ filterLabel }}</span>
-    <span
-      role="button"
-      class="close ml-2 p-2"
-      tabindex="0"
-      @click="onClick"
-      @keyup.enter="onClick"
-      ><i class="icon cross"
-    /></span>
-  </div>
+    <span class="close ml-2 p-2"><i class="icon cross" /></span>
+  </button>
 </template>
 <script>
 export default {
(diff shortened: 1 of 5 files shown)


sarayourfriend:
@obulat Thanks for the review. I made the changes you suggested... and then went down a rabbit hole to fix the `FilterTag` tests as well and fix some accessibility issues in that component, so this could do with a re-review from you if you have time 🙂

PR #367, opened 2021-10-26
Avoid error using only local translation files

### LINKED ISSUE #298: [Bug] Frontend fails to run if translations aren't fetched first
## Description
Sometimes new translations will be available for the frontend but if you don't run `i18n:get-translations` before running `npm run dev` the API will crash after building.

## Reproduction

1. Checkout the repository and delete a translation `.json` file
2. Run `npm run dev`
3. See error about missing translation files

## Expectation
Missing translations shouldn't crash the server

## Additional context
Expectation according to @zackkrida 

## Resolution

krysal:
## Fixes

Fixes #298 by @sarayourfriend

## Description

This addresses the problem of the project failing to run if not all translation files are present, avoiding doing `i18n:get-translations` previously.

It adds a step to the `dev` command to generate the locales list. The downside of this approach is that it modifies the `valid-locales.json` file and we will need to be alert in PR changes, but do we really need to track it? Looks like it is actually not used.

## Testing Instructions

1. Delete any translation file other than the `en.json`
2. Run the project in this branch: `npm run dev`
3. See no errors.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (approved):

package.json
```diff
@@ -12,7 +12,7 @@
   "author": "sebworks",
   "private": true,
   "scripts": {
-    "dev": "cross-env ENABLE_AUDIO=true nuxt",
+    "dev": "node src/locales/scripts/get-validated-locales && cross-env ENABLE_AUDIO=true nuxt",
```
dhruvkb, line 15: We could use the `get-translations` script; or write a different one to just get the validated locales and call it?
```suggestion
    "dev": "npm run i18n:get-translations && cross-env ENABLE_AUDIO=true nuxt",
```

dhruvkb (approved):
LGTM! Not sure why CI is failing though.

sarayourfriend (approved):
LGTM. +1 to what Dhruv said. Super great to not track generated files 🙂

krysal:
The `/src/locales/scripts/valid-locales.json` file is used in `nuxt.config.js` so it's critical that the file is present to run the tests. Adding the command to get translations solves it. Do you want to take a second look at this @zackkrida?

package.json
```diff
@@ -12,14 +12,14 @@
...
-    "dev": "node src/locales/scripts/get-validated-locales && cross-env ENABLE_AUDIO=true nuxt",
+    "dev": "npm run i18n:get-translations && cross-env ENABLE_AUDIO=true nuxt",
     "build": "npm run i18n:get-translations && nuxt build",
     "generate": "nuxt generate",
     "start": "nuxt start",
     "storybook": "nuxt storybook",
     "test": "jest",
     "test:watch": "jest --collectCoverage=false --watch",
-    "test:unit": "jest --collectCoverage=false",
+    "test:unit": "npm run i18n:get-translations && jest --collectCoverage=false",
```
sarayourfriend, line 22: Should this be added to all the `test*` scripts?

krysal: Not sure it is necessary as this was sufficient to pass tests in the GH action and adding it to `dev` makes it more likely to be local. I didn't want to overcrowd this file but keeping uniformity sounds like a good reason to add it to the other commands as well :)

sarayourfriend:
If that file is a pre-requisite for running the tests, even locally, then we should probably add it to the test scripts.

I wonder if there's a way to just add it to the jest setup though instead of having to duplicate it across all the NPM scripts.

Which brings up the other option, of using `just` instead of npm scripts as `just` has proper dependency support for each recipe and cross-environment ENV variables built in (we wouldn't need `cross-env`) and much easier support for overriding variables (like right now to run the app locally without audio you need to modify the `package.json`, with `just` we could just make it an argument).

But that's controversial... idk many frontend projects (if any) that don't use `npm` scripts.

krysal:
Thinking again, it makes more sense to add the get-translation step in the Github action! And keep the `package.json` lean, in my view. Do you think the same?

krysal:
I remembered why I had chosen to use just the `get-validated-locales` node script instead of the npm `i18n:get-translations`... Network requests are slow! and doing it whitin every `dev` script run is a pain in slow connections. I'm changing that back.

sarayourfriend:
> I remembered why I had chosen to use just the get-validated-locales node script instead of the npm i18n:get-translations... Network requests are slow! and doing it whitin every dev script run is a pain in slow connections. I'm changing that back.

💯

zackkrida:
@krysal I would add a `"i18n:get-valid-locales": "node src/locales/scripts/get-validated-locales"` script, and add _that_ to the `dev` script, then get this merged 👍

krysal:
Done! Thanks for your suggestions pals 😄

PR #372, opened 2021-10-27
Add base Button component

### LINKED ISSUE #365: Create base Button component
Create a base button component based on the figma styles here:

https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=354%3A2764

## Implementation ideas

A rough sketch of what this could look like, some feedback would be welcome to confirm this approach makes sense to everyone 🙂 

```vue
<template>
  <!-- All events pass through to the root component, but we'll explicitly `v-bind` $props so they pass through as well -->
  <component :is="as" v-bind="$props" :class="classes"><slot /></component>
<template>

<script>
  name: 'VButton',
  props: {
    // would allow passing `<VButton as="a" href="https://wp.org">This is a link</VButton>` so a link with button styles
    as: {
      type: [String, Object],
      default: 'button',
    },
    type: {
      type: String,
      // override the default type of an HTML button from `submit` to the more common `button`
      default: 'button',
    },
    variant: {
      type: String,
      default: 'primary',
      validate: (v) => ['primary', 'secondary', ...etc].includes(v)
    },
  },
  setup(props) {
    // do some basic a11y checks (if `props.as === 'a'` for example, ensure that `props.href` is defined)
    const style = useCssModule()

    return {
      classes: [style.button, style[props.variant]]
    }
  },
})
</script>

<style module>
.button {
  @apply base button styles
}

.primary {
  @apply tailwind-classes-for-primary-button
}

.secondary {
  @apply etc etc etc
}
</style>
```

## Usage

Usage would basically look like the regular `button` element, would just have default styles applied to it and the ability to easily switch to an `anchor` element with the same styles.

sarayourfriend:
## Fixes

Fixes #365 by @sarayourfriend

## Description

Adds a base `VButton` component that basically just wraps the default `button` element with styles as well as some sensible defaults like `type = 'button'`. There is also explicit `active` state handling to allow for situations where a button might need the "active" appearance without being actually `:active` (which only applies to anchor elements anyway).

## Testing Instructions

Checkout this branch and run `npm install` then `npm run storybook` and verify that the `VButton` component in Storybook looks the same as the figma designs linked in the issue attached to this PR.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VButton.vue
+++ b/src/components/VButton.vue
@@ -42,7 +42,7 @@ const VButton = defineComponent({
       validate: (v) => ['primary', 'secondary', 'tertiary'].includes(v),
     },
     /**
-     * Allows for programatically setting the active state of a button,
+     * Allows for programmatically setting the active state of a button,
      * i.e., in the case of a button opening a menu.
      */
     active: {
--- a/src/components/VButton.vue
+++ b/src/components/VButton.vue
@@ -16,7 +16,6 @@
     :tabindex="tabIndex"
     v-on="$listeners"
   >
-    {{ type }}
     <slot />
   </Component>
 </template>
@@ -108,17 +107,12 @@ const VButton = defineComponent({
     // We check all this stuff against the `node` directly because `as` could be a Vue component!
     // In that case we need to know what element it ultimately renders to decide what a11y props to configure
     watch(nodeRef, (element) => {
-      if (element.tagName === 'BUTTON') {
-        typeRef.value = propsRef.type.value
-      } else {
-        typeRef.value = undefined
-      }
-
       // https://github.com/reakit/reakit/blob/41bd33be66ebcd0b4dcb63155e3ff03e1ae9dfc8/packages/reakit/src/Button/Button.ts#L34
       if (!isButton(element)) {
         if (element.tagName !== 'A') {
           roleRef.value = 'button'
         }
+        typeRef.value = undefined
       }
 
       if (!isNativeTabbable(element)) {
--- a/src/components/VButton.vue
+++ b/src/components/VButton.vue
@@ -1,0 +1,102 @@
+<template>
+  <Component
+    :is="as"
+    :type="as === 'button' ? $props.type || 'button' : undefined"
+    v-bind="filteredProps"
+    :class="[
+      $style.button,
+      $style[variant],
+      active && $style[`${variant}-active`],
+      $style[`size-${size}`],
+    ]"
+  >
+    <slot />
+  </Component>
+</template>
+
+<script>
+import { defineComponent } from '@nuxtjs/composition-api'
+
+/**
+ * A button component that behaves just like a regular HTML `button` element
+ * aside from pre-applied styles based on the passed in variant.
+ *
+ * All props available for the basic `button` component are available here as
+ * well, including an `as` prop which allows for component polymorphism. The
+ * most common use case for this prop is to turn the `VButton` component into
+ * an `anchor` element, so that you can render a link instead of a `button`.
+ *
+ * This also implements basic accessibility checks.
+ */
+const VButton = defineComponent({
+  name: 'VButton',
+  props: {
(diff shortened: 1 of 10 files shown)


src/components/VButton.vue
```diff
@@ -0,0 +1,231 @@
...
+ * most common use case for this prop is to turn the `VButton` component into
+ * an `anchor` element, so that you can render a link instead of a `button`.
+ *
+ * The accessibility helpers on this component are critical and are completely
+ * adapted from Reakit's Button, Clickable, and Tabbable component implementations.
+ */
+const VButton = defineComponent({
+  name: 'VButton',
+  props: {
+    as: {
```
dhruvkb, line 45: Would be nice to add docstrings for the props as they show up in the Storybook args table.

dhruvkb (approved):

sarayourfriend:
@obulat or @krysal do either of y'all have time sometime to review this? Thanks 😁

test/unit/specs/components/v-button.spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+    const element = await screen.findByText(/code is poetry/i)
+
+    screen.debug(element)
+
+    expect(element.tagName).toBe('A')
+    expect(element).not.toHaveAttribute('type')
+  })
+
+  it('should render as a div with role and tabindex', async () => {
```
obulat, line 79: Do we really want to allow creating a 'div' button? What would be the use case?

test/unit/specs/components/v-button.spec.js
```diff
@@ -0,0 +1,125 @@
...
+    const element = await screen.findByRole('button')
+
+    expect(element.tagName).toBe('DIV')
+    expect(element).toHaveAttribute('role', 'button')
+    expect(element).toHaveAttribute('tabindex', '0')
+  })
+
+  it('should render the disabled attribute on a button when the element is explicitly unfocusable and is disabled', async () => {
+    render(VButton, {
+      props: { disabled: true, focusable: false },
```
obulat, line 94: I don't understand why setting a button 'disabled' doesn't also make it unfocusable? It seems to do so when I add `disabled="disabled"` to the Openverse Search button: I can't tab there...

obulat (approved):
Such a fundamental setup!

I've added a couple of questions about div as a button, and focus/disabled inline, but other than that, it's great!

test/unit/specs/components/v-button.spec.js
sarayourfriend: There are cases where you would want to render a button as disabled but still focusable, for example, the "submit" button of a form. You want there to still be a submit form that the screen reader is capable of navigating to but you want to communicate that it is disable due to conditions of the form (for example, required fields not being filled out or fields being filled out with invalid input).

`disabled="disabled"` makes the button completely ignored by the screen reader, meaning you would not be able to convey that state that users who can see the button will get to users who cannot see the button.

test/unit/specs/components/v-button.spec.js
sarayourfriend: Hmm, I'm not sure. It was necessary to cover because of how lax the `as` type is, it's just a pass through to the `is` prop on `Component` and so could be anything.

We could narrow it though and require the `as` prop to only be either `button` or `a` instead of handling every possible thing. That would limit certain scenarios like if you needed to pass another Vue component to `as`. In those cases we can't even properly do the validation because the `ref` wouldn't resolve to the HTML node anyway (I'm just realizing this is a problem in implementing this for Vue as there is no ref forwarding in Vue).

Probably we should just restrict this to either a basic anchor or button element instead of allowing any Vue component or HTML element...

I'll push an update for that.

PR #380, opened 2021-10-28
Create the `InputField` and `SearchBar` components

### LINKED ISSUE #374: Component: InputField
## Description
The `InputField` control is used for entering single line short text. The planned `SearchBar` component is a special case of a generic `InputField` component.

## API

The component will be a shallow wrapper over the native `<input>` element, allowing it to be combined with neighbouring elements like a joint button or an indicator icon. It should be functional and stateless.

### Props

| Name | Type | Description |
|-|-|-|
| value | string | Used with `v-model` |
| inputId | string | For assigning an `id` attr to the field that can then be used with `<label>` |

### Slots

| Name | Slot props | Description |
|-|-|-|
| default | - | Extra information, that which is seen on the end-side, goes here |

### Events

|-|-|-|
| input | string; the new value of the field | Used with `v-model` |

## References

- **Figma:** Refer to the SearchBar designs in #375

## Implementation

### LINKED ISSUE #375: Component: SearchBar
## Description

This consists of an `InputField` connected to a `SearchButton` on the end-side. Each of the two are individually focusable.

### Props

| Name | Type | Description |
|-|-|-|
| `value` | string | Used with `v-model` |

### Slots

| Name | Slot props | Description |
|-|-|-|
| default | - | Extra information, that which is seen on the end-side of the input field, goes here |

### Events

|-|-|-|
| `input` | string; the new value of the search query | Used with `v-model` |
| `submit` | string; the new value of the field | Should be used to detect form submission via <kbd>Enter</kbd> or button click |

## Dependencies

Depends on
- #374

## References

- **Figma:** [Search bar](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=558%3A2278)

## Implementation

dhruvkb:
## Fixes

Fixes #374 by @dhruvkb 
Fixes #375 by @dhruvkb

## Description

This PR creates an input field component and search field component.

## Testing Instructions

0. Run Storybook with `npm run storybook`.
1. Go to `http://localhost:6006/`.
2. Using the left pane, open the docs for
   - Components > Input field
   - Components > Header > Search bar
3. Check if the component looks and behaves as expected. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (changes requested):
These look beautiful. Spot on to the figma designs. A couple of functionality notes:

- really like that the `extraText` is "dumb"—the parent can handle what this displays 👍 
- I don't seem to be able to add a submit event listener to the <SearchBar />
- Just noting that the search input will become a `combobox` at some point to show the user their recent searches

dhruvkb:
@zackkrida 
- !
- I've made some changes to the way form submissions work. Let me know if that's what you were looking for?
- Thanks for the heads-up! Are there any changes needed at the moment to prepare for this change?

dhruvkb:
@zackkrida thanks! The use of `box-shadow` to draw the white line inside the button was new for me!

- The goal was to keep all components as dumb as possible, leaving the functionality to higher-level components.
- I've changed the way the submit event works. Now it should be possible to directly attach to the `@submit` event on the `SearchBar` component.
- Thanks for the heads-up. Do we need to do something at the moment to prepare for this change?

dhruvkb:
I'm also looking for ways to unify the `AudioTrack.PlayPause` button and this new `SearchButton` with Sara's `VButton`, but I'm having a hard time doing that. Maybe that can be a different future PR.

.storybook/decorators/with-rtl.js
```diff
@@ -16,15 +16,19 @@ export const WithRTL = (story, context) => {
     setup() {
       const element = ref()
       const { i18n } = useContext()
+      const setLanguage = (direction) => {
+        element.value.ownerDocument.documentElement.setAttribute(
+          'dir',
+          direction?.value ?? 'ltr'
+        )
+      }
```
sarayourfriend, line 24: Instead of combining `watch` and `onMounted` you can just add `{ immedate: true }` as the third parameter to `watch` I think.

src/components/InputField/InputField.vue
```diff
@@ -0,0 +1,86 @@
...
+      type: String,
+      default: '',
+    },
+    /**
+     * the ID to associate with the internal `<input>` element; This should be
+     * used to associate a label and is recommended for a11y.
+     */
+    inputId: {
+      type: String,
+    },
```
sarayourfriend, line 58: Given that a label is required for accessibility, most input field components will combine the label and input together to ensure that the label is always rendered. Gutenberg's Input component does this and has a `visuallyHideLabel` prop that makes it only visible for screen readers. If we do that on the component level then there's no way for us to create an inaccessible input.

src/components/InputField/InputField.vue
```diff
@@ -0,0 +1,86 @@
...
+  },
+  setup(props, { emit, attrs }) {
+    const text = computed({
+      get() {
+        return props.value
+      },
+      set(value) {
+        emit('input', value)
+      },
+    })
```
sarayourfriend, line 76: This is cool!

src/components/Header/SearchBar/SearchBar.vue
```diff
@@ -0,0 +1,81 @@
...
+</script>
+
+<style>
+/* Removes the cross icon to clear the field */
+input[type='search']::-webkit-search-decoration,
+input[type='search']::-webkit-search-cancel-button,
+input[type='search']::-webkit-search-results-button,
+input[type='search']::-webkit-search-results-decoration {
+  -webkit-appearance: none;
+}
```
sarayourfriend, line 80: Why remove this? It's especially useful on mobile.

cc @panchovm 

Also the way these styles are applied means they will be applied globally. We should use either `scoped` or `module` and associate these modifications to a specific class at the very least.

sarayourfriend (commented):
LGTM so far. Just a couple of things I think we should address, specifically removing the "clear input" button and the lack of accessible label in the input field component.

.storybook/decorators/with-rtl.js
dhruvkb: I've been doing immediate watch wrong my whole life. Fixed in f06e44e.

src/components/Header/SearchBar/SearchBar.vue
dhruvkb: This can't be scoped because the input-field being styled is in a different component. But made the selector more "selective" in a7d1e98.

panchovm:
It looks great. Excellent work. A couple of comments:

* I just noticed there is no style for the placeholder text in the Design Library. I have been using the shade `dark charcoal 70%` color since it meets the AA level.
* In Header, the **search bar** (input + button) has a slightly different style. On `hover`, both elements change to their hover style, as shown in the [Views page in the Design Library](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=751%3A2800).

src/components/Header/SearchBar/SearchBar.vue
sarayourfriend: That's fine, I still think we should make sure removing it is the right choice. Why are we removing it? The button is especially useful on mobile devices where selecting the contents of an input isn't necessarily easy across devices.

krysal:
The components are looking neat 👌 I second Sara regarding the button to remove input. I would understand if it's too disruptive for the design but just want a confirmation from @panchovm.

src/components/Header/SearchBar/SearchBar.vue
dhruvkb: I agree with the utility of the cross on mobiles. It's good from an a11y perspective. My main complaint against it was that it also appears on desktops, where its utility is doubtful. I'll let @panchovm take the call on this one.

src/components/InputField/InputField.vue
dhruvkb: I made this change in b34a843. Boiling down these changes:
- `inputId` is gone (you can directly attach `id` via `$attrs`)
- new prop `labelText` allows setting an SR-only label inside the `<label>` that now wraps `<input>`
- skipping this prop means the `<span>` is not rendered (allowing you to set an `id` and BYO label)

I went with the SR-only label instead of a visual one because we don't have any labelled fields in the design library.

dhruvkb:
@panchovm thanks for the careful inspection, absolutely love your increased presence on GitHub. I incorporated both your suggestions in c7f8032 and ef73c50 respectively.

src/components/Header/SearchBar/SearchBar.vue
```diff
@@ -0,0 +1,81 @@
...
+      text,
+
+      handleSearch,
+    }
+  },
+}
+</script>
+
+<style>
+/* Removes the cross icon to clear the field */
```
obulat, line 74: Why do we remove the standard cross icon from a search bar? Isn't it something that a user expects from it?

obulat (approved):
These look and work amazing!

src/components/Header/SearchBar/SearchBar.vue
obulat: I should definitely read previous comments before reviewing 😆

obulat (commented):
Some tiny things I've noticed after looking more closely:
- The SearchBar's search button is rounded on the right in Figma, but not rounded in Storybook.
- In Figma, when the Search Input is focused, only the Search Input's border becomes pink, the border of the Search Button stays gray. When the Search Button is focused, on the other hand, the Search Button's border becomes pink, and the Search Input's border is gray. And in both states the input background is still white.
- If I understand correctly, only when we have a combobox, and when it is open and shows selections, is when the input background becomes gray, and the Search Button becomes pink. Am I right, @panchovm ?

src/components/Header/SearchBar/SearchBar.vue
panchovm: Sorry I am confused, what are we removing?

panchovm:
You are right @obulat when it is a combo box, both elements react differently as separated.

src/components/Header/SearchBar/SearchBar.vue
panchovm: Now I understand, sorry for the confusion. The clear action never was in the redesign, but you are right; it adds so much value to the search experience. I will work on adding it to the Design Library's components.

zackkrida (approved):
LGTM after the small remaining fixes from @obulat. Great work!

dhruvkb:
The button is actually rounded on the 'end' side, it's a bug with the Tailwind RTL plugin ([fix merged](https://github.com/20lives/tailwindcss-rtl/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Amerged+author%3Adhruvkb)) that prevents it from showing. Changing the direction to RTL and then back to LTR whould reveal the rounded corners in Storybook, but be2b67f provides a workaround for this.

dhruvkb:
@obulat what you say about focused states in point#2 is already true. They behave in the same manner as you mentioned:

Input focus:
image
Button focus:
image

When you say 'focused', do you mean 'hovered'? Because if so, @panchovm mentioned in point#2 of [his comment](https://github.com/WordPress/openverse-frontend/pull/380#issuecomment-964409801) that the during `hover` both elements change into their hover style simultaneously.

obulat:
Thank you for detailed explanations with all the links, @dhruvkb ! 
I'm actually a little confused. I had an impression that the 'hovered' style is the same as the 'focused' styles. What I mean by that is: both the input and the button have a pink border. And only one element (input or button) is pink when it's focused.
And the styles from the [Header Interaction section](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=411%3A2177) are maybe when you show the search suggestions, or leave the search bar (when it's not as prominent, which is shown using the gray background)?
@panchovm , could you please clarify this behavior? Sorry for tagging you here so much :)

panchovm:
No worries at all @obulat. I would try to explain it in two examples

## Search navigated by keyboard
When the keyboard navigates, the first focused element is the search input, and then the search button. The style for both is `focus` since it's **a selected element**. 





## Search activated
When the user clicks on the search input to search for something, search input and search button change to `focus & active` style and the popover of recent searches appears below the input. This is because the search input **was activated by the user**.

image

obulat:
One more quick question, @panchovm : What happens if the user simply hovers over the input, without focusing it or activating it?

dhruvkb:
I'm also confused by the `focus & active` style (but in a different way). For users of the keyboard, the focus state is the active state. In that case, there will not be any hovering so will the input bar never be grey for them? In a sense, I'm confused what `active` means in terms of `hover` and `focus`?

https://user-images.githubusercontent.com/16580576/141303782-cb1559f2-ff75-4027-b76c-71f407bb0080.mov

panchovm:
@obulat @dhruvkb 
You are right about the input element as focus and active at the same time. In that case, the style is `active focus`. I missed the active focus for input on the prototype. Here are two examples for hover and active using cursor.

https://user-images.githubusercontent.com/895819/141306299-1f4f6002-9376-4400-aeb4-85ca1edfd8b3.mp4

My understanding about `active` states is that it depends on the component. For search input, the style is `active focus` since users can start writing immediately. In the case of buttons, the style is `focus` and changes to `active focus` when is activated.

Please correct me if I misunderstand it.

dhruvkb:
I'm still having a bit of a hard time to understand so help me please. Let's not use the context dependent word 'active' for some time. Let's use the words 'hover', 'focus' and 'opened'/'closed' for the suggestion box below. With this vocabulary, so far I have this:

State `hover`: When I hover over the search bar (either input or button), input becomes gray & button becomes pink.

State `focus`: When my mouse is outside the search bar and I <kbd>Tab</kbd> to the input, it gets a pink border.
State `focus`: When my mouse is outside the search bar and I <kbd>Tab</kbd> to the search button, it gets a pink border.

State `hover & focus`: When my mouse is inside the search bar (input becomes gray and button becomes pink) and I <kbd>Tab</kbd> to the input, it gets a pink border also.
State `hover & focus`: When my mouse is inside the search bar (input becomes gray and button becomes pink) and I <kbd>Tab</kbd> to the search button, it gets a pink border.

panchovm:
Replying per point:

(1): Yes.
(2): I'm not sure. Since you said that for input "focus state is the active state", I think search input should be gray with pink border. Use the `active focus` style.
(3): Yes.

Let me stop here. For points 4 and 5, I think we're understanding `hover & focus` as a state, but I defined that in the Design Library to indicate that `hover` state and `focus` state share the same style. It is neither a single state nor a single style.

(4): Yes
(5): I'm not sure. In terms of styles that make sense. But the pink with border style is for an element that is active. In this case, the user has not submitted a new search. I don't have a proper answer for this.

And your sentence about recent searches box is correct.

dhruvkb:
@panchovm thanks! I think I understood your meaning a little more clearly now. I implemented
- point (2) of your comment in 616400a. This makes the input focus state same as active. 
- point (5) of your comment in cc1ab67. This makes search button active state based on the HTML's `:active`. 

Let me know if you think these changes match your vision of the component's behaviour.

src/components/InputField/InputField.vue
```diff
@@ -0,0 +1,94 @@
...
+     */
+    connectionSides: {
+      type: Array,
+      default: () => [],
+      validator: (val) => val.every((item) => ['start', 'end'].includes(item)),
+    },
+  },
+  setup(props, { emit, attrs }) {
+    const type = attrs['type'] ?? 'text'
+    const fieldId = attrs['id'] ?? btoa(Date.now().toString())
```
sarayourfriend, line 69: This, unfortunately, isn't guaranteed to be unique.

We can use a `getUniqueId` function to guarantee a unique ID across all inputs.

```js
let currentId = 0

/**
 * @param {string} prefix
 */
export const getUniqueId(prefix) {
  return `${prefix}-${currentId++}`
}
```

sarayourfriend (commented):
Looking good so far! I noticed a couple of things. Additionally, the input does not span the visual width of its container:

Captura de Tela 2021-11-12 às 07 56 34

This can be fixed by adding `flex` to the label container and `flex-grow` to the input itself.

src/components/InputField/InputField.vue
```diff
@@ -0,0 +1,94 @@
...
+    :class="[
+      {
+        // Padding is set to 1.5px to accommodate the border that will appear later.
+        'border-s-0 ps-1.5px rounded-s-none': connectionSides.includes('start'),
+        'border-e-0 pe-1.5px rounded-e-none': connectionSides.includes('end'),
+      },
+    ]"
+  >
+    <label class="flex-grow" :for="fieldId">
+      <span v-if="labelText" class="sr-only">{{ labelText }}</span>
```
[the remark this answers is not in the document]
zackkrida: Thanks for addressing this @sarayourfriend I forgot to mention the label in my review. I think your proposed solution makes sense.

src/components/InputField/InputField.vue
```diff
@@ -0,0 +1,94 @@
...
+      <input
```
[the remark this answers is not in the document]
obulat: `$listeners` is [deprecated in Vue3](https://v3.vuejs.org/guide/migration/listeners-removed.html#listeners-removed). I wonder if it's possible to integrate this change before moving to Vue3?... Or will we have change all usages of $listeners once we move to Vue3?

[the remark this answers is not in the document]
sarayourfriend: Oh wow, good to know.

It doesn't look like there's a way to use just `$attrs` in Vue 2 but [there is a backwards compat flag for Vue 3](https://v3.vuejs.org/guide/migration/migration-build.html#global-config) (as long as I'm understanding that page correctly)

src/components/InputField/InputField.vue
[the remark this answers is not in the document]
dhruvkb: Making `labelText` required does make sense to prevent inaccessible usage. Thanks!

My main argument against the `hideLabelFromVision` prop is that in the Design Library there are no inputs with labels at all so
- that prop would always be `true`
- there are no examples of how the labels would be styled

So I went ahead and hardcoded the label as `.sr-only`.

src/components/InputField/InputField.vue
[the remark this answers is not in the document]
dhruvkb: Yes, as long as we are on Vue 2 we can use `$listeners` and we'll have to use a compat-mode till we update all usages after upgrading to Vue 3.

src/components/InputField/InputField.vue
dhruvkb: Made this a required prop in 61687b6.

src/components/InputField/InputField.vue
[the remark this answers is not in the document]
dhruvkb: Attached listeners to `<input>` in 76a97e1. We can update when migrating to Vue 3.

src/components/InputField/InputField.vue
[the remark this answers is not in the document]
dhruvkb: Made `labelText` required in 61687b6. Also separated the `<label>` from the `<input>` by updating ESLint rules in 447a973. As there are no custom labels in the designs, we are good for now. We can implement an option to hide labels if and when the need arises.

PR #382, opened 2021-10-28
Refactor search results components and media fetching

### LINKED ISSUE #381: Extract common functionality from Audio and Image search result pages
## Problem

Currently, we have two separate components, `AudioResultsList` and `SearchGridManualLoad` which both have similar functionality. Media fetching for these components is done in those two components, but also in the `search.vue` page. This means the code is duplicated and confusing.

## Description

We can extract a presentational component that has the following functions:
- showing results metadata (results count, `SearchRating`, `SaferBrowsing`)
- showing a list of results
- showing Error/Load more button/ Loading spinner.
- emitting an event to load additional media when Load more button is called.
It should also have a slot for image or audio results.

The media loading and anything that interacts with Vuex store can be done on the `search.vue` page.

## Implementation

### LINKED ISSUE #383: App shows "no results for {query}" while results are loading
## Description

While the current search is loading, the application says there are no results available. This could confuse users on slow network speeds. 

This will be helped when we have placeholder loading ui that pushes the meta search down, but we should still fix this. I consider it a state management issue—the application shouldn't think there are 'no results' until a search has been completed.

CleanShot 2021-10-28 at 12 39 04@2x

obulat:
## Fixes

Fixes #381 by @obulat
Fixes #383 by @zackkrida 

## Description

This PR extracts a `SearchGrid` component that is responsible for displaying search results metadata for images and audio: 
- results count, `SearchRating` and `SaferBrowsing` components
- `LoadMoreButton`, `MetaSearchForm` and error messages.
It also moves data fetching on `LoadMore` button click from the `SearchGridManualLoad` and `AudioResultsList` components to the `search.vue` page, which is already responsible for data fetching on query change.
This way, there are fewer components that need to interact with Vuex store.

## Testing Instructions

Note from @zackkrida: If you see a 'Your focus-trap must have at least one container with at least one tabbable node in it at all times' error, just `npm install`.

Run the app and try searching for 'all media', images and audios, changing the applied filters and tabs. There should be no errors in the console. Also, test SSR by going to the search pages directly, for example:
http://localhost:8443/search/?q=cat&license=cc0,by
http://localhost:8443/search/image?q=cat&license=cc0,by
http://localhost:8443/search/audio?q=cat&license=cc0,by

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (changes requested):
Everything is working, but I see two visual regressions:

- The sidebar width varies wildly based on the length of items in the filters (you can view this by switching content types)
- The audio rows no longer have vertical spacing between them

zackkrida (approved):

src/components/ImageDetails/RelatedImages.vue
```diff
@@ -9,9 +9,10 @@
     <ImageGrid
       :images="images"
       :can-load-more="false"
-      :is-fetching="$fetchState.pending"
-      :fetching-error="$fetchState.error"
-      :error-message-text="null"
+      :fetch-state="{
+        isFetching: $fetchState.pending,
+        fetchingError: $fetchState.error,
+      }"
```
zackkrida, line 15: I really like this!

src/components/ImageGrid/ImageGrid.vue
```diff
@@ -4,15 +4,15 @@
...
     <LoadMoreButton
       v-if="canLoadMore"
       :is-error="isError"
-      :is-fetching="isFetching"
-      :is-finished="isFinished"
+      :is-fetching="fetchState.isFetching"
+      :is-finished="fetchState.isFinished"
       data-testid="load-more"
-      @onLoadMore="onLoadMoreImages"
+      @onLoadMore="onLoadMore"
```
zackkrida, line 15: Great to make this media type agnostic 👍

src/components/ImageGrid/LoadMoreButton.vue
```diff
@@ -31,10 +31,14 @@ export default {
...
     },
+    mediaType: {
+      type: String,
+      default: 'image',
+    },
   },
   computed: {
     finishedLabel() {
-      const type = this.$t('browse-page.search-form.image')
+      const type = this.$t(`browse-page.search-form.${this.mediaType}`)
```
zackkrida, line 41: Interpolation like this will break our source url detection code for the `.pot` file, correct? I think it's okay if a few translations are missing the reference to the source code. Alternatively this would have to be written as a switch or if/else block with each media type:  `this.$t(`browse-page.search-form.image`)`, `this.$t(`browse-page.search-form.audio`)`, and so on.

zackkrida (changes requested):
Unfortunately, clicking the 'load more' button results in an error:

CleanShot 2021-12-01 at 15 10 00@2x

I'm really loving the work here and can't wait for it to be done!

src/components/ImageGrid/LoadMoreButton.vue
obulat: I haven't thought about it :)

zackkrida (changes requested):
The video route: http://localhost:8443/search/video?q=dogs does not set the media type correctly, it fails on client-side load and says 'no images' when loaded via ssr.

> client side
CleanShot 2021-12-03 at 09 24 53@2x

> via ssr 
CleanShot 2021-12-03 at 09 24 55@2x

zackkrida (approved):
Finally! 

One thing I noticed is that the "Load more" button currently jumps to the top of the page after loading more results. Seems odd, probably unrelated to this PR.

src/components/ImageDetails/RelatedImages.vue
sarayourfriend: How does this work with reactivity? Is it working as a side-effect of the object being redeclared on each render?

src/pages/search/audio.vue
```diff
@@ -1,24 +1,75 @@
...
   name: 'AudioSearch',
-  data() {
-    return {
-      // Only show audio results if non-image results are supported
-      supported: process.env.enableAudio,
-    }
+  props: {
+    mediaResults: {},
+    fetchState: {},
+    isFilterVisible: {},
```
sarayourfriend, line 36: Should we add prop documentation for these?

sarayourfriend (approved):
The E2E tests for audio are failing for me locally, but I can access the audio results page in the browser, so I'm not sure it's a blocker. I'm also assuming they're passing for everyone else?

Anyway this LGTM! Nice work pull all this apart and cleaning it up 🎉

src/pages/search/audio.vue
obulat: I tried adding the prop types, @sarayourfriend , and ended up converting the search pages to use composition API, and creating a new composable.  So, now I'm scared to merge this without having someone look at the PR again...

obulat:
> Finally!
> 
> One thing I noticed is that the "Load more" button currently jumps to the top of the page after loading more results. Seems odd, probably unrelated to this PR.

I think the problem was caused by `:v-show` instead of `v-show` property on ScrollButton. At least, it is no longer happening for me.

src/pages/search/audio.vue
obulat: I rebased on the main to check with `npm types`, and the composable passes the type check!!! 
By the way, maybe you have the audio set as unsupported? That would probably cause the e2e tests to fail as there wouldn't be an Audio tab.

sarayourfriend: > the composable passes the type check

Nice! That's awesome 🤩

zackkrida (changes requested):
@obulat Now I'm seeing "no results" when searching from the home page. Client-side searches from the '/search' endpoints seem to work fine!

obulat:
Thank you for testing! New bugs mean more e2e tests to write :)

sarayourfriend (approved):
LGTM! Tested locally and the critical path appears to work perfectly as far as I can tell 😁 I even found an unrelated, pre-existing bug #511 during my testing 🎉

src/components/VSearchRating.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <div class="caption font-semibold">
+  <div>
```
krysal, line 2: The exported name of this component needs the `V` too.

krysal (approved):
Work all well for me. Great component organization! 👍

zackkrida (approved):

src/components/VSearchRating.vue
obulat: Great catch!

Issue #387, opened 2021-10-28
e2e test suite


zackkrida:
## Problem

We currently make lots of small regressions in search functionality that aren't covered by our unit tests.

## Description

Add a robust e2e test suite either using the existing Cypress code or by implementing https://github.com/microsoft/playwright, which probably has a smaller learning curve and size footprint.

## Alternatives

There is no alternative 😆

sarayourfriend:
I think @obulat already completed this!

PR #394, opened 2021-11-02
Add playwright e2e tests

obulat:
## Fixes

Related to #387 by @zackkrida

## Description

This PR sets up e2e testing using [Playwright](https://playwright.dev/docs/intro), replacing the previously added e2e stubs that used Cypress.

### Tests added:
1. the URL parameters are correctly converted into the search state [server rendered]:
- media type from search path `openverse/search/<image>?q=cat` is used to choose the 'Image' tab on the search page.
- query `q` (search term) parameter `openverse/search/?q=<cat>` is used as the value for search input.
- query parameters are converted into filters.  Opening `openverse/search/image?q=cat&license=cc0,by&aspect_ratio=tall` sets the two filters of type `license` checked (CC0 and BY), as well as one filter of type `aspect_ratio`.

2. When searching from the homepage, the client-rendered search page displays the following search state items correctly: query, media type, filter parameters.

To complete this PR, the following pieces need to be added:

- [x]  filters: selection from the sidebar/toggling using the tags

- [x] loading additional media using the "Load more" button.
- [x] search from the content pages (ie `About us`) works.
- [ ] clicking on a single image/audio displays a page with correct info filled in.

Based on [the comment below](https://github.com/WordPress/openverse-frontend/pull/394#issuecomment-957362406), we should create new tickets for the tests of the last two items.

## Testing Instructions

Run the app using `npm run dev`, and then run the tests using `npm run test:e2e`. To see the browser, you can add `--headed` flag to the `test:e2e` script.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
@obulat If all of the code here works, I think it would be best to get this reviewed and merged, and create an issue for each new test you think should be in the suite. Great work!

README.md
```diff
@@ -56,6 +56,14 @@ You can run the unit tests by executing:
 npm run test
 ```
 
+To run the e2e tests, run:
+
+```bash
+npm run test:e2e
+```
```
sarayourfriend, line 63: Should we add a note that you might have to run `npx playwright install` to get the browsers installed if `test:e2e` fails?

package.json
```diff
@@ -20,8 +20,8 @@
     "test": "jest",
     "test:watch": "jest --collectCoverage=false --watch",
     "test:unit": "jest --collectCoverage=false",
-    "test:e2e": "cypress open",
-    "test:e2e:ci": "cypress run",
+    "test:e2e": "npx playwright test --headed",
+    "generate-e2e-tests": "npx playwright codegen localhost:8443/",
```
sarayourfriend, line 24: Fun fact: In `npm` scripts you don't need to use `npx` because `npm` scripts checks `node_modules/.bin` for executables before checking the current $PATH.

```suggestion
    "test:e2e": "playwright test --headed",
    "generate-e2e-tests": "playwright codegen localhost:8443/",
```

Also, what was the reason to run the tests `--headed` instead of headless? They run too quickly be default to be able to visually audit them... and we'd have to have a different script than the base one to run in CI (as it has to run headless there).

sarayourfriend (approved):
This is an amazing start! Thank you for taking this on and adding so many tests right off the bat 🤩 

When I run the tests locally the following fail for me:

```
  4 failed
    filters.spec.js:47:1 › filters are updated when media type changes =============================
    search-query-client.spec.js:41:1 › selecting `audio` on homepage, you can search for audio =====
    search-query-server.spec.js:42:1 › url path /search/ is used to select `all` search tab ========
    search-query-server.spec.js:53:1 › url path /search/audio is used to select `audio` search tab =
```

It seems to just be the audio specific tests so maybe there's a configuration issue with my local environment that is causing these errors. We should figure out what is causing that discrepancy before merging this PR.

I think there's also a devex issue I ran into where running `npm run test:e2e` assumes you've also run `npm run dev` in the background. I'm not sure if this is standard practice for how e2e infrastructure is set up but we could add a small script to check that the API is running before executing the tests and emit a helpful warning for the person running them that they should run the server first in the background.

Additionally, it'd be really nice if we could continue to use jest for the test running and `expect` function, for consistency with the rest of our test suite. Playwright has official (recommended) support for this via the [`jest-playwright`](https://playwright.tech/blog/using-jest-with-playwright) package.

We can do that in this PR or a future PR, I'll leave that up to you 🙂

zackkrida (approved):
I agree with all of @sarayourfriend's excellent suggestions. Here's the results of my test run:

```cli

Running 17 tests using 4 workers

  ✓  filters.spec.js:20:1 › can unset filters using filter tags (7s)
  ✓  homepage.spec.js:3:1 › can change type and search for audio/modification from homepage (6s)
  ✓  search-query-client.spec.js:30:1 › q query parameter is set as the search term (6s)
  ✓  search-query-server.spec.js:32:1 › q query parameter is set as the search term (6s)
  ✓  search-query-client.spec.js:41:1 › selecting `audio` on homepage, you can search for audio (2s)
  ✓  search.spec.js:24:1 › does not show an error message before search (2s)
  ✓  search-query-server.spec.js:42:1 › url path /search/ is used to select `all` search tab (2s)
  ✓  filters.spec.js:47:1 › filters are updated when media type changes (2s)
  ✓  search-query-client.spec.js:58:1 › selecting license type filter on homepage applies filters (2s)
  ✓  search.spec.js:39:1 › shows search result metadata (2s)
  ✓  search-query-server.spec.js:53:1 › url path /search/audio is used to select `audio` search tab (2s)
  -  search-query-client.spec.js:75:6 › url filter parameters not used by current mediaType are discarded
  -  search-query-client.spec.js:89:6 › url filter types not used by current mediaType are discarded
  ✓  filters.spec.js:69:1 › new media request is sent when a filter is selected (2s)
  ✓  search-query-server.spec.js:65:1 › url query to filter, all tab, one parameter per filter type (1s)
  ✓  search-query-server.spec.js:87:1 › url query to filter, image tab, several filters for one filter type selected (2s)
  ✓  search-query-server.spec.js:107:1 › url mature query is set, and can be unchecked using the Safer Browsing popup (2s)

```

package.json
obulat: The reason for adding `--headed` was local testing, should have deleted it :)

package.json
```diff
@@ -18,10 +18,13 @@
...
     "storybook": "nuxt storybook",
     "test": "jest",
+    "test:all": "npm run e2e:ci & jest",
     "test:watch": "jest --collectCoverage=false --watch",
     "test:unit": "jest --collectCoverage=false",
-    "test:e2e": "npx playwright test --headed",
-    "generate-e2e-tests": "npx playwright codegen localhost:8443/",
+    "test:e2e": "playwright test",
+    "run-server": "cross-env ENABLE_AUDIO=true nuxt build && npm run start",
+    "e2e:ci": "start-server-and-test run-server http://localhost:8443 test:e2e",
```
sarayourfriend, line 26: Cool!

sarayourfriend (approved):

obulat:
> Additionally, it'd be really nice if we could continue to use jest for the test running and `expect` function, for consistency with the rest of our test suite. Playwright has official (recommended) support for this via the [`jest-playwright`](https://playwright.tech/blog/using-jest-with-playwright) package.

I tried to add the `jest-playwright`, but then I read that [`jest-playwright`](https://github.com/playwright-community/jest-playwright#jest-playwright) recommends using the playwright's official test runner in their GitHub readme. I also don't quite understand what you mean by "continue to use the `expect` function", @sarayourfriend . Don't the tests here already use `expect`? I guess there is something else I'm not familiar with...

sarayourfriend:
Oh interesting, I hadn't read that notice on the `jest-playwright` github page... somehow I glanced over it, despite it being so large!

I just meant for consistency between the unit and e2e tests it would be nice to be able to use the same globals for `describe`, `test`, and `expect` across both. I guess right now we just have to import them in each test file (which isn't the end of the world) but there could be API differences...

However, I just learned that it seems that playwright uses Jest's `expect` function anyway, so there isn't really a difference there, just some minor differences in the test-runner API as far as describe, beforeAll etc blocks are defined, which seem perfectly tolerable to me.

Anyway, all of that to say, sounds good, let's stick to the official test runner 😀

zackkrida:
sounds like we're ready for...LGTM!

obulat:
> Oh interesting, I hadn't read that notice on the `jest-playwright` github page... somehow I glanced over it, despite it being so large!

I did the same several times before finding it :)

> Anyway, all of that to say, sounds good, let's stick to the official test runner 😀
🚀

PR #397, opened 2021-11-05
Add VPopover component

### LINKED ISSUE #366: Popover component
Figma design here: https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=411%3A2176

We'll use [Popper.js](https://popper.js.org) to simplify the popover positioning logic. A Vue example for implementation is here: https://vuecomponent.com/integrations/popperjs.html

Probably also a good idea to reference the WAI-ARIA spec and other implementations like Reakit:

https://github.com/reakit/reakit/tree/master/packages/reakit/src/Popover
https://github.com/reakit/reakit/tree/master/packages/reakit/src/Dialog#non-modal-dialogs

sarayourfriend:
## Fixes

Fixes #366 by @sarayourfriend 

## Description

Adds a new `VPopover` component which composes a "trigger" and popover content to create a single, acessible, popover unit.

`VPopoverContent` can be reused generically if, for any reason (there shouldn't really be any for the most part) a popover is needed with a different trigger mechanism. It implements 95% of the accessibility functionality (only `aria-haspopup` and `aria-expanded` are handled by the parent `VPopover` component).

Most of the logic for this component is siloed into individual responsibility composables. Most of this logic is copied from Reakit's Dialog and Popover implementations.

`VPopoverContent` emits all normal `div` events.

`VPopover` emits `open` and `close` events.

## Testing Instructions
Run `npm run storybook` and visit `http://localhost:6006/?path=/docs/components-vpopover--default`. Play around with the popover and the various props to make sure it behaves as expected according to the prop descriptions. Unit tests cover almost everything except one edge case that isn't possible to test in JSDom specific to text selection.

To test this particular behavior, open the popover and make sure `hideOnClickOutside` is either unset or set to `true`. Then click and drag to select text inside the popover but be sure to finish the drag _outside_ the popover and then release the mouse click. This should not close the popover.

This is important because the `click` event is actually happening outside the popover, but in the `useHideOnClickOutside` composable we conditionally hide the popover only when the `mousedown` event target matches the click target or both are outside the popover. This ensures that the popover doesn't accidentally close if someone is selecting the text inside of it.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/composables/use-focus-on-blur.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+function useBlurTracker() {
+  const blurredRef = ref(0)
+
+  /**
+   * @param {number} n
+   */
+  const scheduleFocus = () => (blurredRef.value += 1)
+
+  return [blurredRef, scheduleFocus]
```
sarayourfriend, line 39: I think this could be turned into a `customRef`

src/components/.eslintrc.js
```diff
@@ -3,7 +3,11 @@ module.exports = {
   rules: {
     'unicorn/filename-case': [
       'error',
-      { case: 'pascalCase', ignore: ['.eslintrc.js', '.*\\.stories\\.js'] },
+      // Allow things like `Component.stories.js` and `Component.types.js`
+      {
+        case: 'pascalCase',
+        ignore: ['.eslintrc.js', '.*\\..*\\.js'],
+      },
```
zackkrida, line 10: thanks!

src/utils/warn.js
```diff
@@ -0,0 +1,2 @@
+export const warn =
```
zackkrida, line 1: Great addition to the codebase

src/components/VPopover/VPopoverContent.vue
```diff
@@ -0,0 +1,111 @@
...
+import { propTypes } from './VPopoverContent.types'
+
+export default defineComponent({
+  name: 'VPopover',
+  props: propTypes,
+  /**
+   * This is the only documented emitted event but in reality we pass through `$listeners`
+   * to the underlying element so anything and everything is emitted. `@keydown` is the
+   * only one this component overrides and controls (but ultimately still emits).
+   */
```
zackkrida, line 36: ✅ nice!

src/components/VPopover/VPopover.vue
```diff
@@ -0,0 +1,123 @@
...
+      :hide-on-click-outside="hideOnClickOutside"
+      :auto-focus-on-show="autoFocusOnShow"
+      :auto-focus-on-hide="autoFocusOnHide"
+      :gutter="gutter"
+      :hide="close"
+      :aria-label="label"
+      :aria-labelledby="labelledBy"
+    >
+      <!-- @slot The content of the popover -->
+      <slot name="content" />
```
zackkrida, line 32: Since this doesn't appear to have a ref or any props passed to the slot, should we make this the _default_ slot to make the component api a little simpler? I think there's also something compelling about having the two explicit slots in the current implementation, so up to you.

zackkrida (approved):
This is _such_ comprehensive work, and the new composables created here will be a great resource for other future components. I wish I had more feedback to offer!

src/components/VPopover/VPopover.vue
sarayourfriend: Yeah! We can do that.

zackkrida:
The improvements are really nice, especially the test with two popovers.

sarayourfriend:
@obulat Do you think you will have time to review this PR this week?

src/composables/use-hide-on-click-outside.js
```diff
@@ -0,0 +1,87 @@
...
+  })
+
+  useEventListenerOutside({
+    containerRef: popoverRef,
+    triggerRef: popoverPropsRefs.triggerElement,
+    eventType: 'focusin',
+    listener: (event) => {
+      const document = getDocument(popoverRef.value)
+      // Fix for https://github.com/reakit/reakit/issues/619
+      // On IE11, calling element.blur() triggers the focus event on
```
obulat, line 79: I thought we no longer support IE11?

obulat (approved):
I love the thoroughness of the implementation! It all works great, and is accessible.

obulat:
```
 WARN  in ./.nuxt/components/index.js                                                                                                            friendly-errors 06:43:56

"export 'default' (reexported as 'VPopoverContentTypes') was not found in '../../src/components/VPopover/VPopoverContent.types.js' 
```
Do you know why I get this warning when I run `npm run dev` or `npm run storybook`?

src/composables/use-focus-on-blur.js
```diff
@@ -0,0 +1,68 @@
...
+  getDocument,
+  getActiveElement,
+  getNextActiveElementOnBlur,
+} from 'reakit-utils'
+import { ref, watch } from '@nuxtjs/composition-api'
+
+/**
+ * @typedef Props
+ * @property {import('./types').Ref<HTMLElement>} popoverRef
+ * @property {import('./types').ToRefs<import('../components/VPopover/VPopover.types').Props>} popoverPropsRefs
```
obulat, line 11: WebStorm shows a warning that `VPopover.types` is not found.

src/composables/use-focus-on-blur.js
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/**
+ * @returns {[import('./types').Ref<number>, (n: number) => void]}
+ */
+function useBlurTracker() {
+  const blurredRef = ref(0)
+
+  /**
+   * @param {number} n
```
obulat, line 35: This parameter doesn't seem to be used here.

src/composables/use-focus-on-hide.js
```diff
@@ -0,0 +1,59 @@
...
+  if (!popover) return false
+
+  const activeElement = getActiveElement(popover)
+
+  if (!activeElement) return false
+  if (contains(popover, activeElement)) return false
+  if (isTabbable(activeElement)) return true
+  if (activeElement.getAttribute('data-popover') === 'true') return true
+
+  return false
```
obulat, line 28:
```suggestion
  return activeElement.getAttribute('data-popover') === 'true'
```

src/composables/use-focus-on-hide.js
```diff
@@ -0,0 +1,59 @@
...
+  watch(
+    [
+      popoverRef,
+      popoverPropsRefs.triggerElement,
+      popoverPropsRefs.visible,
+      popoverPropsRefs.autoFocusOnHide,
+    ],
+    /**
+     * @param {[HTMLElement, HTMLElement, boolean, boolean]} deps
+     */
```
obulat, line 44:
```suggestion
     * @param {boolean} previousVisible
     */
```

src/components/VPopover/VPopoverContent.vue
```diff
@@ -0,0 +1,101 @@
...
+      popoverPropsRefs: propsRefs,
+    })
+
+    usePopper({
+      popoverRef,
+      popoverPropsRefs: propsRefs,
+    })
+
+    /**
+     * @param {Event} event
```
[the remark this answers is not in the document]
sarayourfriend: Hmm, it shouldn't reference either, it should reference the global `Event` type from TypeScript's built in `dom.d.ts`: https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L5034

When I hover over it it shows the description from the global dom types but if I "go to definition" on it in VSCode it does show React and the global dom types. I think the React types get detected because of interface merging and because React declares global types. Luckily they don't modify the Event type in this way. I have no idea what WebStorm pulls types from sentry without an explicit reference to it.

src/composables/use-hide-on-click-outside.js
sarayourfriend: Yup! I think we can remove part of this check.

src/composables/use-focus-on-blur.js
sarayourfriend: I don't think WebStorm is using TypeScript to interpret the JSDoc comments.

sarayourfriend:
> Do you know why I get this warning when I run npm run dev or npm run storybook?

Yeah, it's because of Nuxt's global component resolution trying to find a component in every single file in the VPopover folder 😞 We could remove it from the global component resolution configuration in `nuxt.config` and I think it would remove that warning.

src/components/VPopover/VPopoverContent.vue
```diff
@@ -0,0 +1,101 @@
...
+   * @param {import('@nuxtjs/composition-api').SetupContext} context
+   */
+  setup(props, { emit, attrs }) {
+    if (!attrs['aria-label'] && !attrs['aria-labelledby']) {
+      warn('You should provide either `aria-label` or `aria-labelledby` props.')
+    }
+
+    const propsRefs = toRefs(props)
+    const popoverRef = ref()
+    const focusOnBlur = useFocusOnBlur({
```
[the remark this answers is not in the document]
sarayourfriend: Sure!

PR #398, opened 2021-11-05
Search and media store refactoring

### LINKED ISSUE #383: App shows "no results for {query}" while results are loading

obulat:
## Fixes

Fixes #383 by @zackkrida

## Description

This is the first part of another big store refactoring. Instead of a `search` and `filter` store, there are a `media` and a `search` stores.
### Media store
This store is responsible for keeping information about the search results: the AudioResult or ImageResult arrays, the current Audio or Image, as well as the data about the fetching state (whether it is fetching or not, whether there is a fetching error).
### Search store
This is where all the information about applied filters and search query is kept.

I've also added docstrings to actions, and found that they didn't have the necessary `await`s in them, which caused problems with loading (I still need to add several docstring descriptions).

### Next steps
The next step should be simplifying the query and filter handling:
The filters should be the source of truth for all query parameters except for `q` and `mediaType`. This would also simplify media fetching for different media types.

## Testing Instructions

You can run the unit tests. Another thing I did was to create a new branch, merge the e2e tests from #394 into it, and test with `npm run test:e2e`. This has helped me see bugs that I didn't realize exist when testing manually. They all currently pass.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal:
Shouldn't ticket #382 precede this one? I mean, should we review that first or the order doesn't matter?

obulat:
It would be better to review this one first,  @krysal . I should probably rewrite #382 as I've worked on it before writing e2e tests, and there probably aren't bugs there that do nothing exist here

test/unit/specs/store/search-store.spec.js
```diff
@@ -1,573 +1,576 @@
...
+      ${'licenses'}        | ${'cc0'}          | ${0}
+      ${'licenseTypes'}    | ${'modification'} | ${1}
+      ${'imageExtensions'} | ${'svg'}          | ${3}
+      ${'imageCategories'} | ${'photograph'}   | ${0}
+      ${'searchBy'}        | ${'creator'}      | ${0}
+      ${'mature'}          | ${undefined}      | ${-1}
+      ${'aspectRatios'}    | ${'tall'}         | ${0}
+      ${'sizes'}           | ${'medium'}       | ${1}
+    `(
+      "TOGGLE_FILTER should set filter '$code' of type '$filterType",
```
krysal, line 531: A m a z i n g ✨

krysal (approved):
This looks really good to me, I tested around manually and everything seems to work well as expected. The stores get simplified step by step, and the same for the tests. Great job ⭐️

test/unit/specs/store/search-store.spec.js
obulat: Thank you, @krysal! I tried using the parametrized before, but it didn't work for some reason. This time, I got really tired of changing the same tests over and over :)

obulat:
Several changes added:

- API search query parameters are set in `search.query` object. It can be changed by dispatching `UPDATE_QUERY` action with either `q` or `searchType` parameters. `searchType` is the content tab that is selected in the UI, and the `query.mediaType` is the media type that is queried from the API (currently, only `image` and `audio` are supported (`audio` is supported only under flag). Other search query parameters are updated through `filters`. Some API parameters have the same names but different value types for different media types (eg. `extensions` for `audio` can be `mp3` or `ogg`, and for `image` - `jpg`, `svg`, etc.), so special filter-to-query transformations are applied.

 - If the search page is opened directly by going to a URL like `<openverse>/search/image?q=cat&license=cc0`, it is server-rendered, and the initial search state is set from the `$route` parameter in `created` function in the `search.vue` page. Instead of parsing the query string to extract information about applied filters, I used the more convenient `$route.query` object.
 
 - `media` store state is simplified: the information about fetching status is in `fetchingState` and the information about the search API request results is in the `results` object, and the component use `getters` to directly access the data for selected media type.

sarayourfriend (changes requested):
LGTM so far. E2E tests pass locally. However when I manually tested it locally I'm unable to enter a new search term in the search results page. Well, I'm able to enter a new one, but when I submit it I keep getting the results from the original search.

Steps to reproduce:

1. Go to `/` and enter a search term
2. Enter a new completely unrelated search term
3. Observe that the results have not changed

The query string never updates either.

obulat:
Oh no, I've checked and un-checked the filters and changed the search tabs so many times, but never tried to change the query, @sarayourfriend  :) I've fixed the search term not updating, and also the "No image results for 'cat'" showing up in the MetaSearchForm while the page is still fetching. And removed a stray log. 
Will need to add the search term test to e2e, too.

sarayourfriend (approved):
Yay! Everything works great and thank you so much for adding another e2e test for the failing case, I love to see that! Awesome work on this really great PR 🙂

PR #399, opened 2021-11-05
Create a global audio player (amongst many smaller improvements)

### LINKED ISSUE #395: Global Audio Player
Create a layout of AudioPlayer to work as the unified global player across different pages.

## References

- [Figma](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=405%3A2998)

dhruvkb:
## Fixes

Fixes #395 by @dhruvkb

## Description

This PR creates a `Global` layout component for the `AudioTrack` that represents the L&F of the player when presented globally.

It also makes a number of tiny improvements in the surrounding components:
- A `VIcon` component for rendering SVG icons (also updated some components to use it)
- Fixes to the mutations made to the active media store and addition of paused/stop differentiation
- Documentation and i18n improvements for the 'replay' state of `PlayPause`
- Code cleanup in `Waveform` allowing absolute height adjustment (later used in `Global` layout)
- Updates to the `Row` layout to match new designs in Figma (discussion with @panchovm about consistency in audio meta info)

## Testing Instructions

Run Storybook and navigate to Audio Track > Global.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


panchovm:
It looks awesome. Excellent work ✨ 

Some comments that might apply to this round:
* Close button is a `53px` square, as the action button to play/pause/replay.
* Close button has no focus style. I can add the styles to the Design Library file but would like your opinion since you asked if it was a one-off button.
* When navigating with the keyboard on the wave area, I was expecting to play/pause with `spacebar`. That interaction will help a lot.

src/store/active.js
```diff
@@ -10,6 +11,7 @@ import {
 const state = () => ({
   type: null,
   id: null,
+  status: 'stopped', // can be 'playing' or 'paused' as well
```
sarayourfriend, line 14: Just thinking out loud, but the distinction between `stop` and `pause` is maybe too subtle for this situation (they're synonyms outside the world of physical audio/video players)... Would `ejected` be clearer than `stopped`?

src/components/License/License.vue
```diff
@@ -85,3 +75,10 @@ export default {
   },
 }
 </script>
+
+<style scoped>
+.v-icon {
+  height: 1.1667em;
+  width: 1.1667em;
+}
+</style>
```
sarayourfriend, line 84: Instead of rely-ing on the `VIcon` component implementation details (the `v-icon` class) can we just create a new local class to pass to the component in this template? Vue will forward the classes and merge them together.

src/components/AudioTrack/PlayPause.vue
```diff
@@ -65,11 +40,27 @@ export default {
...
+     * Get the button icon based on the current status of the player.
+     */
     icon() {
-      return STATUS_TO_ICON[this.status]
+      const statusIconMap = {
+        playing: pauseIcon,
+        paused: playIcon,
+        played: replayIcon,
+      }
+      return statusIconMap[this.status]
```
sarayourfriend, line 63: Just curious, it's not a big perf cost or anything, but why put this (and the `statusVerbMap`) constants in the function body instead of on the module level? Inside the function body it'll be redefined every time the function runs and also redefined for each instance of the component rendering to the page.

It's not a big deal ultimately, they're small objects memory wise, but just curious if you prefer this for some reason over what existed before?

src/components/AudioTrack/layouts/Global.vue
```diff
@@ -0,0 +1,69 @@
...
+  @apply h-14 w-14;
+}
+
+.global-track .audio-controller {
+  @apply absolute inset-0;
+}
+
+.global-track .waveform {
+  @apply h-full;
+}
```
sarayourfriend, line 68: Same deal here with the classes transcending component borders. It's a bit of implementation detail leak that we can avoid by passing the class prop directly to the component.

sarayourfriend (commented):
LGTM so far. A couple of accessibility problems related to focus rings:

The focus ring of the play/pause button is covered by the waveform and the audio artwork. Additionally, if the focus ring is going to overlap varying background colors we will probably need to include some kind of outline to the ring, similar to how subtitles without a background render as white text with a black border so that the subtitles are still readable in both bright and light colored scenes.

The focus ring will also overlap the audio time-stamp when at 00:00. I'm not sure how to address this.

Captura de Tela 2021-11-05 às 13 12 26

The close button needs focus styles. It is currently using the browser defaults which are insufficient contrast:

Captura de Tela 2021-11-05 às 13 12 32

The tab order does not make much sense visually either. Tabs should follow a logical pattern according to the functionality of the page, but to tab from the play/pause button to the waveform requires tabbing _through_ the close button (the tab order is `play/pause, close, waveform`). Ideally this would be `play/pause, waveform, close`.

I also noticed that pressing `space` on the waveform does not toggle the audio to play or pause. I filed an issue here for us to address that: https://github.com/WordPress/openverse-frontend/issues/402

Update: I should have read Francisco's comment first, I just repeated two of the things he said 🤦

src/store/active.js
zackkrida: I think finished/completed/ended could be good terms here. `ejected` is fun if I picture cassette tapes

zackkrida:
Is there going to be separation between a bottom-fixed drawer (i.e, the positioning) and this audio layout (the actual player)?

If so, I think it might make sense for the close button to be in that component (it could still be automatically positioned over this player with css). That seems to be the right separation of concerns and could fix some of the tab index issues that were shared.

src/store/active.js
dhruvkb: Most music players have a very explicit "stop" functionality (indicated by the square icon)

src/components/AudioTrack/PlayPause.vue
dhruvkb: I preferred this change because it meant keeping the code of the component inside the component itself. As you said it's not a performance hit so it made sense to me to keep it all inside the top level component `export`.

src/components/License/License.vue
dhruvkb: Great suggestion, implemented the change in edafcbd.

src/store/active.js
dhruvkb: I think finished/completed/ended would not be correct, because as per my usage here, we reach 'stopped' when the close button is pressed and the player disappears. In that sense, ejected makes more sense to me (and is fun to picture so win-win)

src/components/AudioTrack/layouts/Global.vue
dhruvkb: In case of the layouts, these classes have to leak because they're coming from slots. It's not possible to pass the `class` attribute to slotted content.

However, in our case, the different layouts and the `AudioTrack` component are anyway tightly coupled together, decoupling just the classes wouldn't help very much.

dhruvkb: Technically, moving them to `data` might make more sense, right?

src/components/AudioTrack/PlayPause.vue
dhruvkb: Technically, moving them to `data` might make more sense. Updated in 3714540.

sarayourfriend: Oh interesting, why does moving them to `data` make more sense? Wouldn't that turn them into reactive objects? I always thought it was weird that the only way to get data into templates in the options API, even static data, was to make it reactive 🤔

src/store/active.js
dhruvkb: Updated the name in 00243ec.

dhruvkb:
I think we should track the focus changes in a separate issue and implement them in a separate PR. This one already touches 21 files, making it pretty complicated to review as is.

zackkrida:
@dhruvkb I'll review more fully today, just wanted to make sure you saw the above:

> Is there going to be separation between a bottom-fixed drawer component (i.e, the positioning) and this audio layout (the actual player)?
> 
> If so, I think it might make sense for the close button to be in _that_ component (it could still be automatically positioned over this player with css). That seems to be the right separation of concerns and could fix some of the tab index issues that were shared.

dhruvkb:
@zackkrida sorry I missed this comment. Yes, the cross should not be a part of the audio player but rather be on a layout component enclosing it.  I'll implement that change, please hold off the review for some time.

sarayourfriend (approved):
LGTM! Fixing the focus styles in a separate PR sounds good to me.

krysal (commented):
It looking great @dhruvkb! I only have a _not so small_ note to comment:

Since this component is interactive and involves updates to the global state, it should have some tests to facilitate refactoring in the future, don't you all think?

Though you may want to merge this as is and wait for Olga's PR #382 and/or #394 to merge to add them.

PR #388, opened 2021-10-29
Improve `Dockerfile` and enable pipeline to automate docker image generation

rbadillap:
## Description
In this PR I'm implementing the following features:
- Stabilize the `Dockerfile` to follow common standards and allow us to create a docker image of the application that can be uploaded in a registry
- Enable two GitHub Actions workflows:
  - `pre-build` that runs `npm run lint` to make sure the code follows our existing code standards
  - `build` that generates a docker image and pushes in a registry when a pre-released has been created

## Testing Instructions
- This is under development, a documentation about how to test this workflow will be provided soon

PR #401, opened 2021-11-05
Fix `dependencies` vs `devDependencies`

### LINKED ISSUE #400: Dependencies are miscategorized as `devDependencies`
## Description

Some dependencies are mis-categoized as `devDependencies`. We should identify them and move them into the `dependencies` block.

## Reproduction

1. `rm -rf node_modules` to remove existing dependencies.
2. `npm ci --only="prod" to install dev-only deps
3. `npm run build` to attempt to build the application
4. Observe errors about missing dependencies

## Expectation

The application builds successfully.

## Resolution

zackkrida:
## Fixes

Fixes #400 by @zackkrida 

## Description

Moves all dependencies required to run `npm run build` into the `dependencies` list, so we can use `npm ci --only=prod` to install dependencies in our Dockerfile (see #388).

## Testing Instructions

1. `npm ci --only=prod && npm run build`
1. Verify that the app built successfully
1. (optional) Click around the application after `npm run start`
1. (optional) `npm run test` to verify tests pass

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


package.json
```diff
@@ -40,49 +40,55 @@
...
+    "postcss": "^8.3.6",
     "reakit-utils": "^0.15.2",
-    "sass": "^1.34.0",
     "sass-loader": "^10.1.1",
+    "sass": "^1.34.0",
+    "tailwindcss-rtl": "^0.7.3",
+    "tailwindcss": "^2.2.7",
     "uuid": "^8.3.2",
+    "vue-i18n-extract": "^2.0.0",
     "vue-i18n": "^8.26.5"
```
zackkrida, line 79:
```suggestion
    "vue-i18n": "^8.26.5",
    "vue-i18n-extract": "^2.0.0"
```

krysal (commented):
Removing the `node_modules` folder and running `npm ci --only=prod` gave me an error:
`sh: husky: command not found`. Easy to solve doing first `npm install --only=prod` (I also ran `npm audit fix` to solve some moderate vulnerabilities but... not much difference really 🤷🏻‍♀️)

zackkrida:
@krysal 

> `running npm ci --only=prod` gave me an error: sh: husky: command not found

hm, that's surprising to me because I moved husky to `dependencies`! It _does_ need to be in prod because it's used to do a `postinstall` script that patches a package used in production.

rbadillap (changes requested):
You might not need `husky` in `dependencies`.
Based on the feedback received in #388 we can run `npm ci --only=production --ignore-scripts` so we avoid `prepare` and `postscript` executions.

Now the question: is it ok to ignore them? I believe I'm ok with `prepare` but `postscript` seems to be still needed in production.

So I have two workarounds:
1. Change the Dockerfile in #388 and set a blank prepare script manually by running

```docker
# remove prepare script
RUN npm set-script prepare ""

# install without ignoring scripts
RUN npm ci --only=production
```
2. Use a .js file to have more control of what we want to be executed.
[Taking this example](https://docs.npmjs.com/cli/v7/using-npm/scripts#examples) we can do something like this:

```json
{
  "scripts" : {
    "install" : "scripts/install.js",
    "postinstall" : "scripts/install.js"
  }
}
```
and `scripts/postinstall.js`

```js
if (process.env.NODE_ENV !== 'production') {
  // .. husky can be called here
}
```

What are your thoughts about this?

package.json
```diff
@@ -40,49 +40,55 @@
     "@nuxtjs/i18n": "^7.0.3",
     "@nuxtjs/sentry": "^5.0.3",
     "@nuxtjs/sitemap": "^2.4.0",
+    "@nuxtjs/style-resources": "^1.0.0",
+    "@nuxtjs/eslint-module": "^3.0.2",
     "@nuxtjs/svg": "^0.1.12",
-    "axios": "^0.21.1",
+    "@nuxtjs/tailwindcss": "^4.2.1",
```
dhruvkb, line 46: Tailwind also should be a dev-dependency, the build step will generate the actual CSS used in prod.

dhruvkb: Same for PostCSS and Sass.

zackkrida: There's an unclear distinction with 'devDependencies' on whether they should be 'deps not included in the production bundle' _or_ 'deps needed for local development'. to put the question differently:

If we run `npm install --production && npm run build`, should the build work?

cc: @WordPress/openverse-frontend

sarayourfriend: Well, when you add in an SSR server that needs to be able to build things inside the container, the distinction gets muddied. Production dependencies should include everything required to "run the app." If we built a standalone package with styles that docker was able to consume then yeah, tailwind would be a dev dependency because the only production dependency would be its output. But because our "production" release includes building the app from scratch inside the container then we need to include some of these dependencies that seem to cross the boundaries.

zackkrida: I agree 👍

rbadillap: > If we run npm install --production && npm run build, should the build work?

From a DevOps perspective, it shouldn't.
Build your application is one step before getting it up and running. Being very simplistic, production doesn't care how you came to create a distribution package.
They are only interested in having the necessary files and commands to be able to run freely and independently.

So the question we should make is: **what packages should be placed in dependencies to make sure to run the app as soon as I have the dist folder ready**, that's how I consider `nuxt` should be one of the dependencies (because we need to run `nuxt start`) but `husky` shouldn't.

In addition, that's why I'm using two stages in `Dockerfile`. The first stage requires `devDependencies` to build the application, but, the Production stage doesn't them, that's why I'm transferring the `dist` folder from one stage to the other, because that's the only assets I'm care of.

More info: https://www.docker.com/blog/speed-up-your-development-flow-with-these-dockerfile-best-practices/

sarayourfriend: Definitely agree that `husky` shouldn't be a regular dependency. Nothing it does is necessary for building the app, it's just meant to run developer level verification tools.

If you're doing a two-stage docker build like that then yeah, you can and should move anything that isn't necessary for _after_ `nuxt build` is run I guess.

krysal: I like how the approach of @rbadillap sounds with two stages build, keeping the running production environment lean. Moving dependencies around I got this warning message in one opportunity:

>  WARN  Module @nuxtjs/tailwindcss not found. Please ensure @nuxtjs/tailwindcss is in devDependencies and installed. HINT: During build step, for npm/yarn, NODE_ENV=production or --production should NOT be used.

zackkrida:
I'm going to close this, since we will be following a multi-stage build approach I think it will be simpler to reduce size there rather than fuss too much over devDependencies vs. dependencies.

PR #406, opened 2021-11-09
Add Checkbox component

### LINKED ISSUE #405: Component: Checkbox input
### Props

| Name | Type | Description |
|-|-|-|
|id|String| Checkbox `id` is used for the input id property, connecting the label to the checkbox. id is also used in the `change` event payload as the `name` and `value` parameters if they are not set.
|label|String|The checkbox can be labeled either using the `label` prop, or the default slot (for labels that can contain more than simple strings).
|checked|Boolean| True/false (no indeterminate), false by default.
|name|String| This is usually the category the checkbox input belongs to (eg. `LicenseTypes` for a `Commercial` filter. `name` is passed in the emitted event payload. If no value is provided, `id` is used.
|value|String| This is usually the name of the checkbox input belongs to (eg. `cc0` for a `CC0 Mark` filter. `value` is passed in the emitted event payload.
|disabled|String/Boolean| Set to `true` to disable the input.

### Slots

| Name | Slot props | Description |
|-|-|-|
|default| - | For the inputs that need more than a simple String as a label. For example, the license checkbox inputs can use the `License` component with icons and text as the label.

### Events

|-|-|-|
|change| { name, value, checked} | Is emitted on the checkbox native `change` event.

## Code samples

```
<Checkbox id='tall' name='aspect' :checked="true" label="Tall" />
<Checkbox id='by-sa' name='license' :checked="true">
  <License license="by-sa" />
</Checkbox>
```

## References

Currently, the styles for the checkbox are written in `styles/form/checkbox-radio.sass` file. They will need to be transferred to the `Checkbox` component.

- **Figma:**
https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=25%3A3615
## Implementation

obulat:
## Fixes

Fixes #405 by @obulat

## Description

This PR creates a standalone `Checkbox` component with the new styles, and uses it throughout the app:
Screen Shot 2021-11-09 at 5 53 12 PM
The last checkbox in the screenshot is disabled. I couldn't find the actual disabled variant in Figma, @panchovm .

## Testing Instructions

Checkout the branch and run `npm run storybook` to see the Checkbox story there, and verify that it works both with a string label, and a `License` component label.
Then, run `npm run dev` and test the Checkboxes in use on the home page (license types inputs) and the filter sidebar. Check/uncheck them to see that the filters are applied, and new media is fetched.
`npm run test` should also run without problems :)
## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/Checkbox.vue
```diff
@@ -0,0 +1,122 @@
...
+.checkbox {
+  @apply appearance-none w-5 h-5 border-dark-charcoal border rounded-sm me-2 flex-shrink-0 relative;
+  @apply focus:outline-none focus-visible:ring focus-visible:ring-offset-2 focus-visible:ring-pink;
+  @apply transition-colors disabled:opacity-50;
+  @apply checked:bg-dark-charcoal;
+}
+
+.checkbox:checked::after {
+  content: '';
+  @apply w-2 h-4 border-2 border-white absolute top-0 left-1/3;
```
obulat, line 116:
```suggestion
  @apply w-2 h-4 border border-white absolute top-0 left-1/3;
```

src/components/Checkbox.vue
```diff
@@ -0,0 +1,122 @@
...
+  border-width: 1px;
+  border-top: 0;
```
obulat, line 118:
```suggestion
  border-top: 0;
```

src/components/Checkbox.vue
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * The checkbox can be labeled either using the `label` prop, or the default
+     * slot (for labels that can contain more than simple strings).
+     *
+     * @default ''
+     */
+    label: {
+      type: String,
+      default: '',
+    },
```
sarayourfriend, line 42: Having two ways to set the label seems confusing to me and also potentially prone to someone forgetting to add a label. What if we go with just the slot and somehow verify that it is passed? Is there a scenario where we wouldn't be able to pass the slot value and would have to use a prop instead?

dhruvkb:
The padding on the bottom and right sides seems smaller than the other two.

test/unit/specs/components/checkbox.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  it('should render icons if provided in the default slot', async () => {
+    // @obulat: I cannot pass a required prop to `License` object to use it in the slot,
+    // nor can I check that svg is rendered using its accessible role. But if you use
+    // `screen.debug()`, you can see it is rendered :)
+    render(Checkbox, {
+      propsData: { id: 'simple', label: 'Is Code Poetry?' },
+      slots: { default: ByIcon },
+    })
+    const checkbox = screen.getByRole('checkbox')
+    await checkbox.click()
```
sarayourfriend, line 28: Firing events like this work most of the time, but it's recommended to use `userEvent.click` instead of calling `click` on the element directly. Calling `click` on the element only fires the `click` event, whereas an actual click using a mouse device will cause a `mousedown, click, mouseup` series of events. `userEvent` simulates this kind of interaction.
```suggestion
    await userEvent.click(checkbox)
```

src/components/Checkbox.vue
```diff
@@ -0,0 +1,121 @@
...
+      this.$emit('change', {
+        name: this.inputName,
+        value: this.inputValue,
+        checked: !this.checked,
+      })
+    },
+  },
+}
+</script>
+<style>
```
sarayourfriend, line 103: We should use `scoped` or `module` to avoid introducing global classes and/or clashing with existing global classes.

src/components/Checkbox.vue
```diff
@@ -0,0 +1,121 @@
...
+      <slot>
+        {{ label }}
+      </slot>
+    </span>
+  </label>
+</template>
+
+<script>
+export default {
+  name: 'Checkbox',
```
sarayourfriend, line 22: Single word component names are not recommended in Vue because you could inadvertently override some future HTML native element by the same name (all native elements have a single word name, even `textarea` which is two words in the prototype name `HTMLTextAreaElement` 😱)

For that reason I was appending `V` to `VButton` for example. This helps future proof the Vue components (even though it makes the names less pleasant 😞)

src/components/Checkbox.vue
```diff
@@ -0,0 +1,121 @@
...
+      :checked="checked"
+      :name="inputName"
+      :value="inputValue"
+      :disabled="disabled"
+      @change="onChange"
+    /><span>
+      <slot>
+        {{ label }}
+      </slot>
+    </span>
```
sarayourfriend, line 16: I get why prettier auto-formats things this way because of Vue's whitespace handling, but in this case the whitespace means nothing, so for readability's sake I think it would be better to put each tag on a separate line.

```suggestion
    />
    <span>
      <slot>
        {{ label }}
      </slot>
    </span>
```

src/components/Filters/FilterChecklist.vue
```diff
@@ -34,35 +34,33 @@
...
+          :checked="item.checked"
+          :disabled="isDisabled(item)"
+          @change="onValueChange"
+        >
           <LicenseIcons v-if="filterType === 'licenses'" :license="item.code" />
-          {{ itemLabel(item) }}
-        </label>
-        <button class="appearance-none" type="button">
+          {{ itemLabel(item) }}</Checkbox
+        >
```
sarayourfriend, line 47: Same thing here with the whitespace, it doesn't affect the output so we can put the Checkbox closing tag all on one line to make it easier to find compared to the opening tag.

```suggestion
          {{ itemLabel(item) }}
        </Checkbox>
```

test/unit/specs/components/checkbox.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    render(Checkbox, { propsData: { id: 'simple', label: 'Is Code Poetry?' } })
+    const checkbox = screen.getByText(/code/i)
+    await checkbox.click()
+    screen.getByText(/code/i, { checked: true })
+  })
+
+  it('should render icons if provided in the default slot', async () => {
+    // @obulat: I cannot pass a required prop to `License` object to use it in the slot,
+    // nor can I check that svg is rendered using its accessible role. But if you use
+    // `screen.debug()`, you can see it is rendered :)
```
sarayourfriend, line 22: Can we just use `querySelector` for `svg` on the container that `render` returns?

```js
const { container } = render(Checkbox, ...)

const icon = container.querySelector('svg')
```

Also for being able to render a full output with slots with props, I think it's appropriate to create a TestWrapper style component:

```js
const TestWrapper = Vue.component('TestWrapper', {
  template: `<Checkbox id="simple" label="Is Code Poetry?"><Licence :whatever="'foo'" /></Checkbox>`,
  components: { Checkbox, License },
})
```

That way you can render anything and everything that a regular Vue application would allow you to without having to juggle around the slots options that the `render` function gives you and its limitations.

test/unit/specs/components/checkbox.spec.js
```diff
@@ -0,0 +1,31 @@
+import { render, screen } from '@testing-library/vue'
+import Checkbox from '~/components/Checkbox'
+import ByIcon from '~/assets/licenses/by.svg?inline'
+
+describe('Checkbox', () => {
+  it('should render a checkbox with a string label', () => {
+    render(Checkbox, { propsData: { id: 'simple', label: 'Is Code Poetry?' } })
+    // Get throws an error if element is not found, so this serves as an assert
+    screen.getByLabelText(/code/i, { checked: false })
```
sarayourfriend, line 9: While these do throw errors if the element isn't found, I think for consistency and clarity it would still be nice to add an explicit assertion.

```suggestion
    const checkbox = screen.getByLabelText(/code/i)
    expect(checkbox).toHaveAttribute('checked', 'false')
```

For example

src/components/HeroSection.vue
```diff
@@ -92,15 +94,24 @@ export default {
...
       return this.form.searchType
     },
+    toggleFilter({ code, checked }) {
+      this.filters[code] = checked
+    },
     onSubmit() {
       this.setSearchTerm({ q: this.form.searchTerm })
+      Object.keys(this.filters)
+        .filter((filterName) => !!this.filters[filterName])
+        .map((f) => this.checkFilter({ filterType: 'licenseTypes', code: f }))
```
sarayourfriend, line 114: This can just be a `forEach` instead of a `map` I think. Might save some tiny amount of CPU cycles by not throwing away the `map` result 😅 Also it's more semantically obvious what is meant... usually with a `map` or `reduce` operation I expect the result to be used, whereas `forEach` is explicitly just a loop over the array and nothing more.

src/components/Checkbox.vue
```diff
@@ -0,0 +1,121 @@
...
+    <input
+      :id="id"
+      type="checkbox"
+      class="checkbox"
+      :checked="checked"
+      :name="inputName"
+      :value="inputValue"
+      :disabled="disabled"
+      @change="onChange"
+    /><span>
```
sarayourfriend, line 12: Is the `span` wrapper necessary for the label text? I removed it locally and it didn't seem to effect the visual output at all.

sarayourfriend (commented):
LGTM so far! I love the implementation, sweet and simple 🙂 Just a few thoughts of things we could try differently and some small things to fix here and there. But overall this is awesome 😀

test/unit/specs/components/checkbox.spec.js
obulat: Thank you so much for this, @sarayourfriend ! I haven't seen `container` yet, and the `TestWrapper` seems just the solution I was looking for!

src/components/Checkbox.vue
obulat: It did change the visual output for me in the beginning, but that might be because I hadn't deleted all the sass styles before, I will check tomorrow.

panchovm:
The color for `active` state looks good, but the check icon within the box looks bigger and misaligned. The component is based on the WordPress Design Library version. [Here is the storybook](https://wordpress.github.io/gutenberg/?path=/story/components-checkboxcontrol--default).

I noticed there is no style for the `disabled` state in Figma either. But my guess is there is one in the code version of the component.

zackkrida:
Related to @panchovm's reference to Gutenberg, I think we'll want to take their approach and use `appearance: none;` on the actual checkbox while positioning an svg over the real element. Just inspecting their implementation in the browser should be really useful for reviewing. Thanks for sharing that @panchovm

src/components/Checkbox.vue
obulat: I've removed the label prop. I guess it was my first thought to add the label using prop, and then I just went with it without much thought. AND got confused with whether I need to pass a prop or not myself, so thank you for this comment!

src/components/HeroSection.vue
obulat: Thank you for your explanation :)

zackkrida:
There's a regression where in the checked state I don't even see a checkmark anymore:

CleanShot 2021-11-10 at 15 33 25@2x

obulat:
> There's a regression where in the checked state I don't even see a checkmark anymore:
> 
The first implementation kept the Vue `checked` state from the parent, and did not have any local state. This is why on click the checkbox input was checked in HTML, and used the `.checkbox:checked` styles (i.e. black background). But the JavaScript `checked` value wasn't update by the parent, and therefore stayed `false`, which prevented the checkmark from showing.

I've rewritten the component, adding local checked state, so that it can be in sync with the HTML `checked` attribute, but can also be updated by the parent. I also used composition API, and removed the `value` prop from the component as it doesn't really add value, and collides with the `ref.value` values.

krysal (commented):
The component looks identical to the Figma mockup, great! I just wonder if we can add a `value` prop for the _HTML value attribute_. The licenses filter is a perfect example for when the `name` is the name of a field (`licenses`) and the value is its possible values, worth the redundancy, like licenses codes in this case, and they might not be unique so using the `id` always might not be the desire in those cases.

src/components/meta/Checkbox.stories.js
```diff
@@ -0,0 +1,75 @@
+import VCheckbox from '~/components/VCheckbox'
+import License from '~/components/License/License'
```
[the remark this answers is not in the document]
zackkrida: Nice detail testing the Storybook _build_ @krysal, thank you!

src/components/VCheckbox.vue
```diff
@@ -0,0 +1,133 @@
...
+    <!--  @slot The checkbox label  --><slot />
+  </label>
+</template>
+
+<script>
+import Checkmark from '~/assets/icons/checkmark.svg?inline'
+import { computed, defineComponent, ref, watch } from '@nuxtjs/composition-api'
+
+const VCheckbox = defineComponent({
+  name: 'VCheckbox',
```
dhruvkb, line 27: Might be nice to set [the `model` option](https://vuejs.org/v2/api/#model) to the component to use with `v-model`.

src/components/VCheckbox.vue
```diff
@@ -0,0 +1,133 @@
...
+  components: { Checkmark },
+  props: {
+    /**
+     * Checkbox `id` is used for the input id property, connecting the label to
+     * the checkbox. id is also used in the `change` event payload as the `name`
+     * and `value` parameters if they are not set.
+     */
+    id: {
```
dhruvkb, line 35: I'm not 100% sure but since ID is an HTML attribute, this might conflict. For disambiguation, I used `fieldId` in the `InputField` component.

src/components/VCheckbox.vue
```diff
@@ -0,0 +1,133 @@
...
+      width="20"
+      height="20"
+      role="img"
+    />
+    <!--  @slot The checkbox label  --><slot />
+  </label>
+</template>
+
+<script>
+import Checkmark from '~/assets/icons/checkmark.svg?inline'
```
dhruvkb, line 23: Nice, I didn't know this was possible! (Might be good to standardise usage between inline SVG imports and the `VIcon` component.)

dhruvkb (approved):

src/components/VCheckbox.vue
obulat: I wanted to use the `v-model` at first, too. The problem with model for checkbox is it's binding `value` and `input` event. And in checkbox, `value` is the name of the checkbox value (ie., 'CC BY' for licenses), and the thing that actually needs to change is the `checked` property.

src/components/VCheckbox.vue
obulat: It's the [`vue-svg-loader` that is used by the Nuxt `svg-module`](https://github.com/nuxt-community/svg-module#vue-svg-loader). The sad thing about it is I don't think they support Nuxt 3...

src/components/VCheckbox.vue
obulat: This `id` prop is used both for the HTML attribute (mainly to get the `<label for="id">` working well, since I've read that simply wrapping an input in a label might not be sufficient for some screen readers), and to use as a fall-back for `name` and/or `value` properties. I can see that a non-unique id might cause a problem, but don't know if it can be significant. Can it?

sarayourfriend (changes requested):
Looks really good so far! Just a couple of suggestions, none of which are actually blockers.

The only blocker is a regression in the tooltip for each checkbox. On `main` you can open the tooltip for each license type, but on this branch it doesn't work.

src/components/Filters/FilterChecklist.vue
```diff
@@ -1,147 +1,101 @@
...
       <button
-        v-if="!filtersExpandedByDefault"
-        :aria-label="$t('filter-list.category-aria', { categoryName: title })"
-        class="filter-visibility-toggle is-white py-2"
+        v-if="filterType === 'licenses'"
+        :ref="`${item.code}licenseIcon`"
+        :aria-label="$t('browse-page.aria.license-explanation')"
+        class="appearance-none"
+        type="button"
+        @click.stop="toggleLicenseExplanationVisibility(item.code)"
```
obulat, line 29: @sarayourfriend , There were two event handlers that got triggered when you clicked the License explanation button: one opened it, and the other closed it right away. Adding `.stop` to the `click` event fixed this.
There is still a problem of the explanation opening in weird places, that I think the `VPopover` could solve.

sarayourfriend: Nice! Yeah using the popover component would be great here! We'd have to add support for the arrow though, I left that out cause there weren't any examples I found in the designs.

I think it would also fix another issue I saw where if you have a license explanation popover open and click on another license explanation button it only closes the existing one but doesn't open the one you clicked on. I think VPopover would fix this too.

sarayourfriend (approved):
🚀 🌵 🤠 🐄 🌵 🚀

src/components/VCheckbox.vue
```diff
@@ -0,0 +1,122 @@
...
+    watch(props, (props) => {
+      if (props.checked !== localCheckedState.value) {
+        localCheckedState.value = props.checked
+      }
+    })
+
+    const onChange = () => {
+      localCheckedState.value = !localCheckedState.value
+      emit('change', {
+        name: inputAttrs.value.name,
```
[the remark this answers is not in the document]
obulat: I like having the properties similar to the native HTML checkbox ones...

Issue #329, opened 2021-10-18
Rework styles of the homapage audio/image media type switcher


zackkrida:
The homepage media switcher implemented in #271 should be updated to the following design from the _transitional styles_ figma file:

![137492436-1d86ee22-d7ee-45ec-9088-0d8010b61334](https://user-images.githubusercontent.com/6351754/137764505-22089c1c-12b2-4cd2-a690-7ed49caf0c91.gif)

zackkrida:
Closing, this design will no longer be used and a new one has been proposed in the redesigned homepage.

PR #410, opened 2021-11-11
Add VItemGroup component

### LINKED ISSUE #370: Content switcher on mobile
## Description
Since we will start implementing the audio integration based on the redesign (#364), there is room for improvement of the content switcher on mobile.

The current one has a horizontal distribution of content items and a vertical distribution of internal pages. That adds unnecessary complexity to identifying the elements. Besides, it seems reasonable to follow the general style on desktop view.

  

## Alternatives
I was working on different ideas and here is the version I like most.

  

The changes are:
* All elements are vertically distributed.
* The active item has the same style as on desktop and a `check` icon (from [WordPress Design Library](https://www.figma.com/file/e4tLacmlPuZV47l7901FEs/WordPress-Design-Library?node-id=8111%3A19295)) for a11y purposes.
* There is more space between sections.
* **Content type** title has `dark-charcoal` color.

sarayourfriend:
## Fixes

Fixes #370 by @panchovm

## Description

Adds new `VItemGroup` and `VItem` components for rendering a grouped set of buttons.

This is currently blocked on focus styles for the items as the current focus styles won't work:

Captura de Tela 2021-11-11 às 12 59 57

It's also currently missing the "direction" behavior to render a horizontal list, useful for https://github.com/WordPress/openverse-frontend/issues/329

## Testing Instructions

Run `npm run storybook` and visit `http://localhost:6006/?path=/story/components-vitemgroup--default`.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
@panchovm whenever you get a chance, the current focus styles are show in the screenshot in the PR description. Would love some advice on how to fix this. One option I had was to raise the z-index of the focused item so that the focus styles will show over adjacent items. I think this looks quite nice but I wanted to verify it with you before committing to it:

Captura de Tela 2021-11-11 às 13 16 33

sarayourfriend:
This is fully ready for review aside from the focus styles issue which should hopefully be simple to implement anyway.

sarayourfriend:
Adding a video to help show @panchovm the current behavior due to local environment issues:

https://user-images.githubusercontent.com/24264157/141479723-741748aa-8ce6-4440-9e84-b0bf79a87704.mov

I'm noticing a couple issues:

1. Duplicated top border on items other than the first one
2. Each item "jiggles" when it is focused

panchovm:
Could this be related to #370 and [this answer](https://github.com/WordPress/openverse-frontend/issues/370#issuecomment-967140504)? I'm seeing the mockups and here are the two uses I found.

| Audio format | Mobile menu |
| ----------- | ----------- |
| Screen Shot 2021-11-12 at 11 04 50 | Screen Shot 2021-11-12 at 11 05 47 |

sarayourfriend:
I think we'll need to create a slightly separate implementation for the popover contents rather than standalone grouped items like this.

Here're the updated focus styles:

https://user-images.githubusercontent.com/24264157/141483847-b2ffc741-3a95-42db-8e09-7736f5fa6d6c.mov

sarayourfriend:
Okay, I fixed some issues and added a borderless design that will allow this component to be used inside and outside of a popover 🙂 

https://user-images.githubusercontent.com/24264157/141490752-e1b40392-387d-404e-bd6f-efbb63a28a5a.mov

panchovm:
It looks wonderful. Great job 👏

src/components/VIcon/VIcon.vue
```diff
@@ -29,7 +29,10 @@ export default {
      * importing an SVG should give us the path to the file.
      */
     iconPath: {
-      type: String,
+      /**
+       * In `just` our icons get transformed to Vue components
+       */
+      type: process.env.NODE_ENV === 'test' ? Object : String,
```
obulat, line 35: Thank you for explaining this!

src/components/VItemGroup/VItemGroup.vue
```diff
@@ -0,0 +1,76 @@
...
+export default defineComponent({
+  name: 'VItemGroup',
+  props: {
+    /**
+     * The direction to render the items in.
+     *
+     * @default 'vertical'
+     */
+    direction: {
+      type: /** @type {import('@nuxtjs/composition-api').PropType<'vertical' | 'horizontal'>} */ (String),
```
obulat, line 38: I didn't know you can type the props this way :)

src/assets/icons/check.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none">
```
obulat, line 1: I think the same icon is used in the VCheckbox component.

obulat (approved):
I love the way the components work and look, and they sound great in Voice Over. There are so many code patterns to learn from :)

src/components/VIcon/VIcon.vue
zackkrida: This comment should read `jest`, not `just`, right?

zackkrida:
This is so awesome, some thoughts/questions.

-  I think `z-index: 1;` on the focused element would be ideal; I see a small 1/1.5px imperfection on the focus ring without it

    > before

    CleanShot 2021-11-19 at 08 01 34@2x

    > after

    CleanShot 2021-11-19 at 08 01 54@2x

- Some `radiogroup` [implementations](https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/radio/radio.html) I've tested use `tab` to focus in and out of the entire group, but use keydown/up/left/right to cycle through the radio options. Have you considered implementing that here? Alternatively, I think it could just be an aria `group` instead of a radiogroup?
- There's a visual bug in the horizontal popover example, where each item is missing the correct 'end' padding:
    CleanShot 2021-11-19 at 08 12 09@2x

zackkrida (approved):

src/components/VIcon/VIcon.vue
sarayourfriend: Yup!

sarayourfriend:
> Some radiogroup implementations I've tested use tab to focus in and out of the entire group, but use keydown/up/left/right to cycle through the radio options. Have you considered implementing that here? Alternatively, I think it could just be an aria group instead of a radiogroup?

That's a good point, I completely forgot to implement that interaction! Our DropdownMenu component already does that anyway 🙂

sarayourfriend:
I don't think the group role works for this case, BTW. I'll work on fixing the radiogroup implementation 👍 

https://www.w3.org/TR/wai-aria-1.1/#group

src/components/VItemGroup/VItemGroup.vue
```diff
@@ -70,7 +94,69 @@ export default defineComponent({
...
+          break
+        case 'ArrowDown':
+        case 'ArrowRight':
+          if (targetIndex === items.length - 1) {
+            // If navigation down/right on the last item, ignore it
+            return
+          }
+          ensureFocus(items[targetIndex + 1])
+          break
+      }
```
sarayourfriend, line 136: I think this might need some tweaking for RTL support with the Left/Right handlers.

zackkrida: s/could down/right on the last item actually cycle back to the first item?

sarayourfriend: Ah you're right, it should. I missed that in the example.

sarayourfriend: Fixed!

PR #413, opened 2021-11-12
Node 16 and NPM 8

zackkrida:
Previously, only production issues were keeping us on Node 14 and npm 6. This PR upgrades both and updates the local configuration for `volta` in the `engines` key of the package.json file.

This also builds a new `package-lock.json` with the updated deps.


package.json
```diff
@@ -109,16 +109,15 @@
     "vue-jest": "^3.0.7"
   },
   "engines": {
-    "node": ">= 12.0.0",
-    "npm": ">= 6.0.0"
+    "node": "16.13.0",
+    "npm": "8.1.0"
```
dhruvkb, line 113: Should we be pinning `engines` this aggressively? I liked the old approach that only pinned the major version.

package.json
```diff
@@ -109,16 +109,15 @@
...
   },
   "browserslist": [
     "> 1%",
     "last 2 versions",
     "not ie <= 8"
   ],
   "volta": {
-    "node": "14.17.5",
-    "npm": "7.21.0"
```
dhruvkb, line 122: Why aren't we pinning `npm`?

dhruvkb (commented):
Nice to upgrade to the new LTS. I have a few non-blocking questions.

sarayourfriend (approved):
🚀

dhruvkb (approved):
LGTM!

PR #420, opened 2021-11-17
Update video demos for Meta Search

### LINKED ISSUE #418: Recreate Meta Search marketing page documentation with Openverse screenshots
The gifs on https://wordpress.org/openverse/meta-search currently use old 'footage' from CC Search. This should be replaced with Openverse.

dhruvkb:
## Fixes

Fixes #418 by @zackkrida 

## Description

This PR
- records and updates <s>video</s> image demos for Openverse replacing the old CC videos
	- uses <s>WebM for smoother playback and</s> PNG for smaller file sizes
- updates the language to be future-proof for when audio becomes a first-class media type
	- also replaces hard-coded text for buttons and sections with references to the actual buttons and sections

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/locales/po-files/openverse.pot
```diff
@@ -1518,103 +1515,105 @@ msgctxt "meta-search-page.content-types"
...
 msgctxt "meta-search-page.audio-video.title"
-msgid "How does the Meta Search feature work for Audio and Video?"
+msgid "How does the Meta Search feature work for non-indexed media types?"
 msgstr ""
 
-#: src/pages/meta-search.vue:50
+#: src/pages/meta-search.vue:65
 msgctxt "meta-search-page.audio-video.content"
-msgid "Openverse does not yet support built-in audio and video search. For Audio and Video, simply click on the appropriate tab, and you’ll see what sources you can find CC-licensed content on."
+msgid "Openverse does not yet support built-in search for some media types, such as videos. For these media types, simply click on the appropriate tab, and you’ll see what sources you can find CC-licensed content on."
```
sarayourfriend, line 1582:
```suggestion
msgid "Openverse does not yet support built-in search for some media types, such as videos. For these media types, simply click on the appropriate tab and you’ll see sources where you can find CC-licensed content."
```

sarayourfriend (changes requested):
Thanks for updating this and reducing the filesize, that's a really big deal!

The only thing is that these videos are way too tall for many screen sizes. The originals were not very tall and easily fit on most screen sizes.

On mobile it is too difficult to read anything because the video includes so much detail that all the text is blurry and small. The original gifs were not _much_ better but they were a little better.

We should also add the `loop="true"` attribute to these videos so that they replay like the gifs do.

dhruvkb:
@sarayourfriend from a usability and accessibility perspective, would images be more preferable than a video or a GIF? My understanding here is that in a short video without commentary, it's very hard to grasp what actually happened whereas an image can have both alt text and drawn-on markers, making the content quite clear.

src/locales/po-files/openverse.pot
dhruvkb: Fixed in 10f9fff.

sarayourfriend:
Hmm, probably an image would be better. When I see a video or GIF like this playing on a page I tend to skip right over it, but I don't know if that's how most people behave. Maybe @panchovm would know.

panchovm:
I would use images instead of GIFs or videos. The sequences recreate the whole flow unnecessarily. Having screenshots of the pages where Meta Search areas are centered is a solution that solves the need. Also agree with @sarayourfriend, the crop can be a laptop-screen ratio.

For the last part, where we explain the why of this feature, I will get rid of the image. It doesn't add more value than the current description.

obulat (changes requested):
The screenshot alt text does not make it clear what it contains. Something like "Screenshot showing the meta search form with buttons to search for openly-licensed media of unsupported type on other sites" would be better for a11y.

sarayourfriend:
> The screenshot alt text does not make it clear what it contains. Something like "Screenshot showing the meta search form with buttons to search for openly-licensed media of unsupported type on other sites" would be better for a11y.

Just a nit pick, but I think "screenshot showing" is redundant in alt text. At least I know it's recommended to not say "picture of ..." because the screen reader will already announce that it is an image. Maybe screenshot is extra useful information in this case though? 🤔

dhruvkb:
To be fair, we could entirely make the image invisible to screen-readers for these reasons: 
- the information is already covered in the text, and the image is not needed to understand it
- the image adds nothing more than visual help, so it's zero value add for visually impaired users

sarayourfriend:
@dhruvkb Nice thinking! Makes sense to me!

obulat (approved):

PR #422, opened 2021-11-17
Give licenses the full name and retire `LicenseIcons`

### LINKED ISSUE #358: License icons/titles accessibility
## Problem

We display licenses using icons and their names in several places, but they are not accessible. 

For example, if you open the search filter sidebar with VoiceOver in Firefox, you will hear "Circle C C Circle Human Figure Circle Dollar sign with overlay backslash Circle equal sign  B Y N C N D unchecked checklist" for  CC BY-NC-ND, with each letter pronounced separately. When reading the license label, VoiceOver pronounces CC BY-SA license name as "B Y S A text element", and it is very difficult to understand that it refers to a license.

## Description

We are using two similar components to display licenses (name and icons): `src/components/License/License.vue` and `src/components/LicenseIcons.vue`. `LicenseIcons` is using VocabularyIcons icon font which we have decided to deprecate, so the solution should be to:
- deprecate `LicenseIcons` component, and use License component everywhere instead.
- make the `License` component more flexible to allow it to be used both as a span and as a checkbox label
- add a screen-reader-friendly label to the licenses: instead of the short abbreviation, it should use the full license name (eg. "CC BY-SA" -> "CC Attribution Share-Alike license")
- make the actual svgs hidden from screen readers.

## Implementation

krysal:
## Fixes

Fixes #358 by @obulat

## Description

This PR helps describe better the license in the `VLicense` component adding an `aria-label` with the readable name of the license. This way using VoiceOver in the filter sidebar you should hear the full name of the license instead of the spelling of the letters one by one.

It also replaces the old `LicenseIcons` component in favor of this one. For this, two properties had to be added: `hideName` and `filledBg`, more details in the Storybook story.

## Testing Instructions

Run storybook and see the new VLicense story showing only icons.
```
npm run storybook
```
Also, run the project normally and verify the replaced component in the Search image results page and in the individual result page.
```
npm run dev
```

CleanShot 2021-11-17 at 18 34 50@2x

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/Filters/FilterChecklist.vue
```diff
@@ -57,7 +57,7 @@
 
 <script>
 import LicenseExplanationTooltip from '~/components/Filters/LicenseExplanationTooltip'
-import VLicense from '~/components/License/VLicense'
+import VLicense from '~/components/License/VLicense.vue'
```
obulat, line 60: You should not need to include the `.vue` extension in your imports.

src/locales/en.json
```diff
@@ -640,6 +640,16 @@
     "by-nc-sa": "BY-NC-SA",
     "by-nc-nd": "BY-NC-ND"
   },
+  "license-readable-names": {
+    "cc0": "Zero",
+    "pdm": "Public Domain Mark",
+    "by": "Attribution",
+    "by-sa": "Attribution ShareAlike",
```
obulat, line 647: I wonder what the best solution is for the "readable" names. I know these are the official names, but since they are used for the screenreader, maybe it's better to split the 'ShareAlike'? Because at the moment, it is read as 'sharelike', and does not give screenreader users the most accurate representation...

obulat (changes requested):
@krysal , could you please rebase this PR onto master? There have been quite a few changes in the filter sidebar, including related to the a11y (eg removal of from 'role="list"' and addition of the fieldset), that might affect the a11y of the license icons as well.

zackkrida:
The way it's read by Voiceover on cc.org sounds decent to me: 

"Attribution-ShareAlike"

https://creativecommons.org/licenses/by-sa/4.0/

https://user-images.githubusercontent.com/6351754/142625388-f46e15f2-d069-415f-ade8-30d02f3321eb.mp4

src/components/Filters/FilterChecklist.vue
krysal: The "Go to Definition" doesn't work in VS Code without the extension, sadly, which is the main reason to add the import.

src/locales/en.json
krysal: That is an important question here. I was wondering the same thing and whether to add "CC" at the beginning and/or "license" at the end to be more descriptive. Does it change the meaning significantly if we split "ShareAlike"? I'm not the best person to answer this so I'll apply what the rest suggest. CC @sarayourfriend

zackkrida: https://github.com/WordPress/openverse-frontend/pull/422#issuecomment-974043628

VoiceOver reads 'Attribution-ShareAlike' correctly for me, but I think it would also read 'Share-Alike' correctly if we wanted to add a hyphen there. Since these strings are being _heard_, not seen, I think how it sounds is more important and it's okay to change the spacing/hyphenation as-needed.

krysal:
@obulat This branch is already rebased onto main, but I see the license's names are not announced by VoiceOver when you're on the checkbox, only when it's checked/unchecked, I'm not sure why this changed, but I will review it.

zackkrida (approved):
The hyphenated names sound good to me. Nice!

src/components/Filters/FilterChecklist.vue
dhruvkb: I think I read Evan You's comment somewhere that it's recommended to use the extension.

sarayourfriend (changes requested):
LGTM! Just one edge case with VoiceOver we should fix but otherwise this is great 😁

src/locales/en.json
```diff
@@ -640,6 +640,16 @@
     "by-nc-sa": "BY-NC-SA",
     "by-nc-nd": "BY-NC-ND"
   },
+  "license-readable-names": {
+    "cc0": "Zero",
```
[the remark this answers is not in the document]
zackkrida: Yeah I think something like "CC Zero" or even "CC Zero, No Rights Reserved" would be preferable.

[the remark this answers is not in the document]
krysal: Luckily as this is the first item, after the license name, the name of the group is mentioned, "License or Public Domain", so it should be clear that it's a CC license in this context (I hope so!)

src/components/License/VLicense.vue
```diff
@@ -1,16 +1,27 @@
...
       />
     </div>
-    <span class="name">
+    <span
+      v-if="!hideName"
+      class="name"
+      :aria-label="$t(`license-readable-names.${license}`)"
+    >
       {{ $t(`license-names.${license}`) }}
     </span>
```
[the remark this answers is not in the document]
krysal: The problem was due to the duplicated `SearchGridFilter`; now that I rebased, the names should be announced correctly.

sarayourfriend (approved):
🚀 works great, thanks!

PR #439, opened 2021-11-19
Toggle `SearchGridFilter` completely according to view width

### LINKED ISSUE #438: Ids of filters input are not unique
## Description

The ids of the filters, like licenses, use, image type, etc, are not unique. This is due to the `SearchGridFilter` component being repeated (and visually hidden conditionally) for the desktop (as the filter sidebar) and mobile view (as a modal), you can see this on the search page.

https://github.com/WordPress/openverse-frontend/blob/bea2783fc589638d2c37a119f96f46cad9abc3a2/src/pages/search.vue#L4-L14

Losing the uniqueness of the id makes them ineffective, currently, this is preventing us to make the filters checkboxes fully accessible to screen readers despite improvements in #406 and #422.

## Reproduction

1. Run the project and search for something
2. On the search results page, using the browser **Inspector** you can see the structure of the `SearchGridFilter` repeated even though one is visually hidden.

CleanShot 2021-11-19 at 16 52 01@2x

Checkbox does not recognize its label text.
Screen Shot 2021-11-19 at 4 55 04 PM (2)

## Expectation

Ever id should be unique in order to work with them effectively.

## Resolution

krysal:
## Fixes

Fixes #438 by @krysal
Related to #422 by @krysal 

## Description

This PR solves the problem of having a duplicated `SearchGridFilter` component and avoid repeating the ids of its subcomponents. See also #445.

## Testing Instructions

1. Run the project and search for something to go to the search result page.
2. Use the inspector of Developer Tools and observe a unique block of code for `SearchGridFilter`.
3. Extra: On the same search page, right-click on any checkbox of the filter sidebar and on "Inspect Accessibility Properties", each `checkbutton` should have its text set. This is the manual test I used to confirm it.

CleanShot 2021-11-19 at 17 23 54@2x

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (approved):
Nice, simple solution 👍

obulat (changes requested):
This is really nice and simple, but unfortunately, because the `isDesktop` function is not reactive, if you open the filters and change the screen width, the filters stay the same until it's rerendered (after closing and reopening). So, it doesn't change to modal, if you narrow your screen down from the desktop. And then there can be a state where there is no button to open the filter, and no filter visible.

obulat:
filterChange

krysal:
This was greatly improved by Olga's PR mostly and I tried to add some tests but is hard given the structure of the components. I think is ready to review and merge @dhruvkb @zackkrida

obulat (approved):

src/pages/search.vue
```diff
@@ -95,6 +101,14 @@ const BrowsePage = {
         this.isAnyFilterApplied
       )
     },
+    searchFilter() {
+      return {
+        classes: this.isLargeScreen
+          ? 'column is-narrow grid-sidebar max-w-full bg-white'
+          : '',
```
dhruvkb, line 108: Can use the cleaner `{ 'list of classes': boolean }` notation here instead of ternary with blank string.

src/pages/search.vue
```diff
@@ -35,14 +28,31 @@ import {
...
+  },
+  setup() {
+    const defaultWindow = typeof window !== 'undefined' ? window : undefined
+    const isLargeScreen = useMediaQuery(
+      `(min-width: ${MIN_SCREEN_WIDTH_FILTER_VISIBLE_BY_DEFAULT}px)`,
+      defaultWindow
+    )
+    return {
+      isLargeScreen,
+    }
```
dhruvkb, line 54: Mixing `setup()` with the older options API is not very nice. Moving `computed` inside `setup` would be better.

src/pages/search.vue
```diff
@@ -69,12 +79,8 @@ const BrowsePage = {
       local.get(process.env.filterStorageKey)
         ? local.get(process.env.filterStorageKey) === 'true'
         : true
-
-    const MIN_SCREEN_WIDTH_FILTER_VISIBLE_BY_DEFAULT = 768
-    const isDesktop = () =>
-      screenWidth() > MIN_SCREEN_WIDTH_FILTER_VISIBLE_BY_DEFAULT
     this.setFilterVisibility({
-      isFilterVisible: isDesktop() ? localFilterState() : false,
+      isFilterVisible: this.isLargeScreen ? localFilterState() : false,
```
dhruvkb, line 83: Can be simplified as
```suggestion
      isFilterVisible: this.isLargeScreen && localFilterState(),
```

dhruvkb (changes requested):

src/components/ScrollButton.vue
```diff
@@ -9,7 +9,6 @@
     <svg
       viewBox="0 0 24 24"
       xmlns="http://www.w3.org/2000/svg"
-      role="img"
```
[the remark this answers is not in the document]
krysal: Since the icon is hidden, it is not necessary to give it the `img` role[^1]. If we are going to treat the icon as an image, we must give it a title and (optionally) a description.

[^1]: https://css-tricks.com/accessible-svg-icons/

src/pages/search.vue
krysal: I agree and also would prefer not to mix the two Vue API, but to change it completely will require to move methods and watchers too, and this page component depends heavily on the global state using Vuex, whose current functioning as we know it will probably change. I'm honestly wondering if it's worth changing the rest to the composition API if there is no a clear conflict to resolve and given the [`useStore` situation in Nuxt3](https://github.com/nuxt/framework/issues/899).

zackkrida (approved):
lgtm ✨

PR #448, opened 2021-11-25
VLogoLoader

### LINKED ISSUE #371: Component: Header Logo (`VLogoLoader`)
## Description

The `VLogoLoader` is a component that displays the Openverse logo but also acts as a global loader to indicate that results aren't ready. In the header this component will be wrapped in a `<NuxtLink />` so that it also serves as a link to the homepage.

Initially this component can be very simple, but I am exploring the usability of making it use `aria-busy`[^1] to announce the loading state. We'd have to make the results area an live region so we need to explore that more.

A good writeup of live regions can be found here: https://www.smashingmagazine.com/2015/04/its-alive-apps-that-feed-back-accessibly/

### Props

| Name | Type | Description |
|-|-|-|
state | string enum `IDLE\|LOADING` | The current state of the component, used to show the loading state

It might make sense to make the hover state a prop at some point but I don't think it's necessary yet.

## Code samples

Usage:

```vue
<NuxtLink to="/">
  <VLogoLoader :state="isFetchingMedia ? LOADING : IDLE" />
</NuxtLink>
```

## References

- **Figma:** https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=558%3A2279

## Implementation

[^1]:https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-busy

zackkrida:
## Fixes

Fixes #371 by @zackkrida 

## Description

Adds a new `VLogoLoader` component and some supporting composables:

- One for using event listeners (no longer being used, so I could remove)
- One for using media queries (added by @obulat modified by me)
- One for using `prefers-reduced-motion`

The loader itself is an `aria-hidden`, visual enhancement to the site. It does **not** have any inherent accessibility qualities, nor is it focusable. Originally I thought this component should be an ARIA Live Region[^1] but after some research that will be a role better served by our actual search grid, which can announce it's own loading status, the # of newly added results after the user hits the 'load more' button, and so on. I'll make a new ticket soon to spec out those improvements. 

## Testing Instructions

This component can be tested with `npm run storybook` and viewing the `VLogoLoader` story. There are a few things you can do:

- Toggle the `status` prop from "loading" to "idle" with the handles
- Set `prefers-reduced-motion` in your browser while in the "loading" status, observe the change in animation to a more subtle, non-moving effect. This is currently a gentle pulse from full to partial opacity. @panchovm might have a better idea for a subtle animation that doesn't add movement, but still visually conveys to the user that loading is occurring. Prefers reduced motion !== wants _no_ motion. Here's a quick video on how to test for prefers reduced motion:

https://user-images.githubusercontent.com/6351754/144527413-bc1c3bd7-6d94-408e-8088-f283d9089e8d.mp4

## Follow-up questions

1. Please suggest any test cases I should add. I admittedly don't love unit testing of front-end components and am always looking for improvements there.
2. The component is extremely small in terms of pixel values in the mockups (cc @panchovm), with the logo being 20px tall with 14px of padding surrounding. This has been increased to 32px tall with 16px of padding in my component. Otherwise the Openverse logo is barely legible, even on my 5k monitor.
  
    > My sizing
    CleanShot 2021-12-02 at 19 55 00@2x
  
    > Figma sizing
    CleanShot 2021-12-02 at 19 54 04@2x
    CleanShot 2021-12-02 at 19 54 29@2x

3. The mobile menu includes the logotext "Openverse", I think that can simply exist _next to_ this loader but perhaps be wrapped in the same `<NuxtLink />` component.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

[^1]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions


src/components/VLogoLoader/VLogoLoader.vue
```diff
@@ -0,0 +1,93 @@
+<template>
+  <div
+    :data-prefers-reduced-motion="prefersReducedMotion"
+    :data-fake-aria-label="loadingLabel"
```
zackkrida, line 4: These are testing values I will remove 😄

src/composables/use-event-listener.js
```diff
@@ -0,0 +1,33 @@
+import {
+  isRef,
+  watch,
+  onMounted,
+  onBeforeUnmount,
+  unref,
+} from '@nuxtjs/composition-api'
+
+/**
+ * Use an event listener. Shamelessly stolen from https://logaretm.com/blog/my-favorite-5-vuejs-composables/#useeventlistener
```
obulat, line 10: Vueuse library steals from logaretm, too 😆

src/composables/use-media-query.js
```diff
@@ -0,0 +1,35 @@
+import { useEventListener } from '~/composables/use-event-listener'
+import { computed, ref } from '@nuxtjs/composition-api'
+
+/**
+ * Use a media query.
+ * @param {Ref<string> | string} query
```
obulat, line 6: I used the same code here 😆 : https://github.com/WordPress/openverse-frontend/pull/445/files#diff-8408da4942f82bca81affaff29a72e22ee03a7d3c7dc3966684abec20418d8fdR1-R38
Window needs to be handled correctly in SSR.

zackkrida: Awesome, I'll pull in your implementation :)

zackkrida:
Note, the current aria live region implementation is incorrect 😄 but will be revised soon!

src/components/VLogoLoader/VLogoLoader.vue
```diff
@@ -0,0 +1,197 @@
...
+
+export default defineComponent({
+  name: 'VLogoLoader',
+  props: propTypes,
+  /**
+   * @param {import('./VLogoLoader.types').Props} props
+   * @param {import('@nuxtjs/composition-api').SetupContext} context
+   */
+  setup() {
+    const defaultWindow = typeof window !== 'undefined' ? window : undefined
```
obulat, line 55: The default window is being used in several places now to ensure that composables work correctly with SSR. Is there a way to extract this code somehow?

src/components/VLogoLoader/VLogoLoader.vue
```diff
@@ -0,0 +1,197 @@
...
+  86% {
+    visibility: visible;
+    transform: translateX(0);
+  }
+}
+
+/**
+  The stops at 20% and 40% here are used to simulate a 'pause' whenever the spheres
+  reach the start or end of the logo. CSS's built in animation-delay property only
+  works on the first cycle of an animation, not  between subsequent runs.
```
obulat, line 171:
```suggestion
  works on the first cycle of an animation, not between subsequent runs.
```

obulat: Silly nitpicking :)

obulat (approved):
The Logo loader component is awesome!

src/components/VLogoLoader/VLogoLoader.vue
zackkrida: I was thinking it could be the default value in the hook, at least. Not sure if it'll work, I'll test

src/components/VLogoLoader/meta/VLogoLoader.stories.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+  },
+}
+
+const SimpleLoaderStory = (_, { argTypes }) => ({
+  props: Object.keys(argTypes),
+  template: `
+    <div>
+      <VLogoLoader v-bind="$props" />
+      <p>Remember to test with the <code class="inline p-0"><pre class="inline bg-light-gray p-0 m-0 leading-normal">prefers-reduced-motion</pre></code> media query. You can find instructions for doing so in Firefox <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion#user_preferences">here.</a>
```
krysal, line 20: Minor detail: the ending tag of the paragraph is missing.
```suggestion
      <p>Remember to test with the <code class="inline p-0"><pre class="inline bg-light-gray p-0 m-0 leading-normal">prefers-reduced-motion</pre></code> media query. You can find instructions for doing so in Firefox <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion#user_preferences">here.</a></p>
```

krysal (approved):
Looks amazing! You can see the attention to details, really nice ✨

src/components/VLogoLoader/VLogoLoader.types.js
```diff
@@ -0,0 +1,9 @@
+export const propTypes = {
+  status: {
+    type: /** @type {'loading'|'idle'} */ String,
```
sarayourfriend, line 3: To get a type cast to work correctly in JSDoc TypeScript you have to wrap the thing you're casing in parenthesis:

```suggestion
    type: /** @type {'loading'|'idle'} */ (String),
```

src/components/VLogoLoader/VLogoLoader.vue
sarayourfriend: I don't think you can use a default value as it'll try to evaluate `window` even when it's undefined and throw an error.

src/composables/use-event-listener.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param {Ref<EventTarget | null> | EventTarget} target The target can be a reactive ref which adds flexibility
```
sarayourfriend, line 12:
```suggestion
 * @param {import('@nuxtjs/composition-api').Ref<EventTarget | null> | EventTarget} target The target can be a reactive ref which adds flexibility
```

src/composables/use-event-listener.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {string} event
+ * @param {(e: Event) => void} handler
+ */
+export function useEventListener(target, event, handler) {
+  // if its a reactive ref, use a watcher
+  if (isRef(target)) {
+    watch(target, (value, oldValue) => {
+      oldValue?.removeEventListener(event, handler)
+      value?.addEventListener(event, handler)
+    })
```
sarayourfriend, line 22: Would it be better to use `onInvalidate` to remove the event listener?

```suggestion
    watch(target, (value, _, onInvalidate) => {
      value?.addEventListener(event, handler)
      onInvalidate(() => value?.removeEventListener(event, handler))
    })
```

I'm not 100% sure there is a difference, I'll do some fiddling to test this out, but i think the watcher may never run on component dismount to remove the old listener but invalidate will run.

src/composables/use-event-listener.js
```diff
@@ -0,0 +1,33 @@
...
+  } else {
+    // otherwise use the mounted hook
+    onMounted(() => {
+      target.addEventListener(event, handler)
+    })
+  }
+  // clean it up
+  onBeforeUnmount(() => {
+    unref(target)?.removeEventListener(event, handler)
+  })
```
sarayourfriend, line 32: Oh wait, I see, I guess `onInvalidate` isn't necessary because the clean up happens in one place here?

test/unit/specs/components/v-logo-loader.spec.js
```diff
@@ -0,0 +1,28 @@
...
+
+  describe('accessibility', () => {
+    it('should render differently when the user prefers reduced motion', () => {
+      render(VLogoLoader, {
+        props: { status: 'loading' },
+      })
+      const element = screen.getByTestId('logo-loader')
+      expect(element).toHaveAttribute('data-prefers-reduced-motion', 'true')
+    })
+  })
```
sarayourfriend, line 27: Should we also add a complimentary test for when `useReducedMotion` returns `false`? Currently all the test cases have it returning the default mock implementation you gave of `true`.

src/composables/use-media-query.js
```diff
@@ -13,6 +13,7 @@ export function useMediaQuery(query, options = {}) {
   if (!window) return ref(false)
 
   const mediaQuery = window.matchMedia(query)
+  /** @type {import('@nuxtjs/composition-api').Ref<Boolean>} */
```
sarayourfriend, line 16: Minor type nit pick: `Boolean` (with a capital B) refers to the primitive constructor, `boolean` with a lowercase b refers to the general type. Same goes for the rest of the primitive types.
```suggestion
  /** @type {import('@nuxtjs/composition-api').Ref<boolean>} */
```

sarayourfriend (commented):
Nice! The CSS animations are magic to me 🤩 Just a couple documentation things but otherwise looks great!

With respect to the "pulsing" reduced motion version... I have decent vision and I can barely tell anything is happening. In fact, it might as well not even pulse if it's that subtle.

src/composables/use-media-query.js
zackkrida: D'oh! I always do this with `Boolean` in particular, thanks!

panchovm:
It looks awesome. Great work. 

I agree with @sarayourfriend as the fading effect is not very noticeable. And I would add that it might be unnecessary. The idea of motion reduction is great, but in this case, the animation is very subtle and lives in a context where the skeleton interface fading is more proper for showing the loading instance.

zackkrida:
@panchovm that's a great point about the skeleton loaders. I think I will go with no animation then!

test/unit/specs/components/v-logo-loader.spec.js
zackkrida: Added, great suggestion.

src/composables/use-event-listener.js
zackkrida: I do like your take on it! 😄

PR #450, opened 2021-11-25
Update Sentry version and turn it on only for production

### LINKED ISSUE #446: Upgrade to the latest version of the Sentry SDK
The Sentry SDK version we're using has fallen behind—the latest is `6.15.0` at the time of writing.

obulat:
## Fixes

Fixes #446 by @zackkrida

## Description

This PR updates the `@nuxtjs/sentry` package version to the most recent one, `5.1.5`, which uses Sentry SDK version `6.14.1`, not the most recent`6.15.0`. It also loads Sentry only on production builds to make the development environment faster.

## Testing Instructions

I only ran:
1. `npm ci`
2. `npm run dev` 
3. `npm run build`, `npm run start`. Locally, this raises a CORS error.
After deploy, we can see if there are any problems (which there shouldn't be).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -205,6 +205,7 @@ export default {
       process.env.SENTRY_DSN ||
       'https://3f3e05dbe6994c318d1bf1c8bfcf71a1@o288582.ingest.sentry.io/1525413',
     disabled: process.env.NODE_ENV === 'development',
+    environment: process.env.NODE_ENV
     lazy: true,
   },
   tailwindcss: {
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -203,7 +203,7 @@ export default {
   sentry: {
     dsn:
       process.env.SENTRY_DSN ||
-      'https://3f3e05dbe6994c318d1bf1c8bfcf71a1@o288582.ingest.sentry.io/1525413',
+      'https://53da8fbcebeb48a6bf614a212629df6b@o787041.ingest.sentry.io/5799642',
     disabled: process.env.NODE_ENV === 'development',
     environment: process.env.NODE_ENV
     lazy: true,
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -205,7 +205,7 @@ export default {
       process.env.SENTRY_DSN ||
       'https://53da8fbcebeb48a6bf614a212629df6b@o787041.ingest.sentry.io/5799642',
     disabled: process.env.NODE_ENV === 'development',
-    environment: process.env.NODE_ENV
+    environment: process.env.NODE_ENV,
     lazy: true,
   },
   tailwindcss: {
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -204,6 +204,7 @@ export default {
     dsn:
       process.env.SENTRY_DSN ||
       'https://3f3e05dbe6994c318d1bf1c8bfcf71a1@o288582.ingest.sentry.io/1525413',
+    disabled: process.env.NODE_ENV !== 'production',
     lazy: true,
   },
   tailwindcss: {
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -204,7 +204,7 @@ export default {
     dsn:
       process.env.SENTRY_DSN ||
       'https://3f3e05dbe6994c318d1bf1c8bfcf71a1@o288582.ingest.sentry.io/1525413',
-    disabled: process.env.NODE_ENV !== 'production',
+    disabled: process.env.NODE_ENV === 'development',
     lazy: true,
   },
   tailwindcss: {
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@nuxt/components": "^2.1.6",
         "@nuxtjs/composition-api": "^0.27.0",
         "@nuxtjs/i18n": "^7.0.3",
-        "@nuxtjs/sentry": "^5.0.3",
+        "@nuxtjs/sentry": "^5.1.5",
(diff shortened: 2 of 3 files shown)


nuxt.config.js
```diff
@@ -204,6 +204,7 @@ export default {
     dsn:
       process.env.SENTRY_DSN ||
       'https://3f3e05dbe6994c318d1bf1c8bfcf71a1@o288582.ingest.sentry.io/1525413',
+    disabled: process.env.NODE_ENV !== 'production',
```
krysal, line 207: Nice!

krysal (approved):

sarayourfriend (approved):
🚀

nuxt.config.js
zackkrida: I think we'll actually want to make this _explicitly_ disabled locally somehow, because soon we'll have cloud staging and production environments and we'll want to have Sentry running on both.

zackkrida (changes requested):
Looks great, let's just make sure Sentry will still run on our cloud staging environment too.

sarayourfriend:
> let's just make sure Sentry will still run on our cloud staging environment too

Why do we want Sentry in staging?

zackkrida:
> > let's just make sure Sentry will still run on our cloud staging environment too
> 
> Why do we want Sentry in staging?

My thinking was if there's errors getting logged while we're testing staging that we don't notice, that Sentry could help make those more visible. Just more assurance against bugs. 

I'd frankly be happy to run it _locally_, too, but for the sake of devex and nuxt initialization speed (it can make a difference of a few seconds) I think it's fine to omit.

sarayourfriend:
If we're going to run it in staging should we make the sentry URL configurable and have a separate environment for staging? Or is there a way to partition the sentry dashboard by the source's host or something like that?

nuxt.config.js
dhruvkb: Can we have a separate var like `ENABLE_SENTRY`? It'll allow Sentry to be enabled or disabled.

zackkrida:
Sentry has a concept of environments[^2] that by default uses `process.env.NODE_ENV`, but we should explicitly add configuration for it in our Sentry config, as described here: https://sentry.nuxtjs.org/sentry/options#config


[^2]: https://docs.sentry.io/product/sentry-basics/environments/

obulat:
> Sentry has a concept of environments[1](#user-content-fn-2-4df0377d957d04b1a1b4b5ae601573ed) that by default uses `process.env.NODE_ENV`, but we should explicitly add configuration for it in our Sentry config, as described here: https://sentry.nuxtjs.org/sentry/options#config
> ## Footnotes
> 
>     1. https://docs.sentry.io/product/sentry-basics/environments/ [↩](#user-content-fnref-2-4df0377d957d04b1a1b4b5ae601573ed)

@zackkrida , I have currently set the Sentry disabled when NODE_ENV is `development`, would it be `development` in staging environment, too?
I would definitely like to remove Sentry from dev, because it prevents the app from loading in Firefox. Or does it actually mean that we need to look into Sentry logs to find what the problem is?
Oh, and I don't understand how I can set the Sentry config environments ...

nuxt.config.js
```diff
@@ -204,6 +204,7 @@ export default {
     dsn:
       process.env.SENTRY_DSN ||
       'https://3f3e05dbe6994c318d1bf1c8bfcf71a1@o288582.ingest.sentry.io/1525413',
+    disabled: process.env.NODE_ENV === 'development',
```
zackkrida, line 207:
```suggestion
    disabled: process.env.NODE_ENV === 'development',
    environment: process.env.NODE_ENV
```

zackkrida: @obulat this is how we'd add the environment support, which would be either `'staging'|'production'`.

zackkrida: And a note, this is happening by default but it's probably better for us to be explicit about it.

nuxt.config.js
```diff
@@ -204,6 +204,7 @@ export default {
     dsn:
       process.env.SENTRY_DSN ||
       'https://3f3e05dbe6994c318d1bf1c8bfcf71a1@o288582.ingest.sentry.io/1525413',
```
zackkrida, line 206:
```suggestion
      'https://53da8fbcebeb48a6bf614a212629df6b@o787041.ingest.sentry.io/5799642',
```

The DSN was for Creative Common's Sentry account 😭. Noting that the Production DSN in our infra is still correct.

zackkrida (changes requested):

nuxt.config.js
obulat: Was that the reason that it wasn't loading locally? Or is it the reason that we won't see the dev errors in our sentry logs?

nuxt.config.js
```diff
@@ -203,7 +203,9 @@ export default {
   sentry: {
     dsn:
       process.env.SENTRY_DSN ||
-      'https://3f3e05dbe6994c318d1bf1c8bfcf71a1@o288582.ingest.sentry.io/1525413',
+      'https://53da8fbcebeb48a6bf614a212629df6b@o787041.ingest.sentry.io/5799642',
+    disabled: process.env.NODE_ENV === 'development',
+    environment: process.env.NODE_ENV
```
obulat, line 208:
```suggestion
    environment: process.env.NODE_ENV,
```

sarayourfriend (approved):
🎉 🤠

zackkrida (approved):

PR #454, opened 2021-11-25
Converge `NavSection` components

### LINKED ISSUE #433: Make `EmbeddedNavSection` the standard `NavSection`
## Description
Maintaining two copies of a component, one of which is not used now (and won't be in the foreseeable future) is a waste of effort. To simplify we must deprecate the current `NavSection` and change `EmbeddedNavSection` to `NavSection`.

## Expectation

There is one `NavSection` and it is the embedded one.

## Additional context

The removal of one navsection would lead to the removal of a lot of scaffolding code the enables one of the two sections to be used for example the `?embedded=<boolean>` query param.

## Resolution

dhruvkb:
## Fixes

Fixes #433 by @dhruvkb

## Description

This PR makes `EmbeddedNavSection` the main `NavSection`. It also updates usages and layouts to reflect this merge-and-replace.

## Testing Instructions

Open any URL with the `?embedded=false` query params. While the old code would show a different, unmaintained, taller header, the new one remains the same. Other embedded features such as message emits to parent frames work as intended, unchanged.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/NavSection.vue
```diff
@@ -1,42 +1,50 @@
...
+  >
+    <div class="navbar-brand flex-grow md:flex-grow-0">
+      <NuxtLink
+        to="/"
+        class="navbar-item"
+        style="align-self: center; line-height: 0"
+      >
+        <!-- width and height chosen w.r.t. viewBox "0 0 280 42" -->
+        <OpenverseLogo
+          :style="{ width: '160px', height: '24px' }"
```
obulat, line 16: The reason for setting the width and height _HTML properties_ (not inline style) here was to minimize CLS as described here: https://web.dev/optimize-cls/#images-without-dimensions

dhruvkb: Nice. I wasn't aware of their utility. Restored in 072672c.

sarayourfriend: Should we remove the inline styles now or are both necessary?

src/components/NavSection.vue
```diff
@@ -1,42 +1,52 @@
...
           width="160"
           height="24"
+          alt="Openverse logo mark"
```
sarayourfriend: "mark" seems redundant with `logo` to me.

```suggestion
          alt="Openverse logo"
```

src/components/NavSection.vue
```diff
@@ -51,18 +61,21 @@
           </div>
         </form>
       </div>
+
       <div class="navbar-end">
         <Dropdown v-slot="{ onFocus }" :text="$t('header.about-tab')">
           <NuxtLink
             class="navbar-item"
             :to="localePath('/about')"
+            role="menuitem"
```
sarayourfriend, line 65: Nice catch, but instead of duplicating this (and requiring that the `Dropdown` consumer know to repeat it everywhere) could we just add an `a11yProps` to the slot that includes the role and then `v-bind` that to all the dropdown items?

src/components/NavSection.vue
dhruvkb: You're right that the `style` attr is no longer necessary. Removed in ad5a6ed.

sarayourfriend (approved):
LGTM! Just one suggestion but not a blocker.

src/components/NavSection.vue
dhruvkb: Makes sense, implemented in c7231c1.

src/components/NavSection.vue
dhruvkb: Fixed in 53058d1.

sarayourfriend (approved):
🚀

zackkrida (approved):
Nice work! It makes me ask a larger question of 'do we need embedded mode?' but I think it's still nice that the functionality that _requires_ a parent to communicate with are behind a flag.

PR #456, opened 2021-11-26
Make single result page for images RTL-compatible

### LINKED ISSUE #435: Make single result page for images RTL compatible
Issues identified on this page:
- [x] Go to image's website" button needs to have its icon padding adjusted
- [x] The attribution box and tabs need to have their borders adjusted to be on the correct side for RTL languages
- [ ] The plain text attribution also needs to have its quotation marks corrected as currently they're being moved around by the browser into strange places
- [x] The license icon needs to have its padding adjusted against the associated text
- [x] The information table needs to have its padding adjusted as the "dimensions" text overlaps the listed dimensions
- [x] The leading edge of the information/reuse table needs to be fixed for RTL languages to align with the rest of the text in the result information

krysal:
## Fixes

Fixes #435 by @krysal

## Description

 This PR:
- Adjusts the padding of the "Go to Image's website"
- Adjusts the padding of license icons against the associated text on the "Reuse" tab
- Adjusts the padding of texts in the "Information" tab
- Deletes unused files, like `src/utils/get-provider-logo.js`, and old styles
- Removes the icon of the copy button
- Fixes the style of the image's title according to transitional mockups
- Adjusts overall padding and margin on the page
- ~The plain text attribution also needs to have its quotation marks corrected as currently they're being moved around by the browser into strange places~

## Testing Instructions

Set a right to left language (e.g. Arabic) using the locale chooser in the footer of a non-embedded page, like `http://localhost:8443?embedded=false` and check the fixed styles.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/CopyButton.vue
```diff
@@ -6,11 +6,11 @@
     :data-clipboard-target="el"
   >
     <span v-if="!success">
-      <i class="icon cc-share me-2" />
+      <i class="icon cc-share mx-1" />
```
obulat, line 9: The [transitional styles](https://www.figma.com/file/BlUTIXRHLOq5mRITpDpi3E/Transitional-styles?node-id=530%3A2051) don't have an icon on the copy button, so I think we can remove the icons altogether.

zackkrida: Agreed, we should remove the icon. CC @panchovm

zackkrida:
@krysal could you update this with some detail on what isn't complete? It's looking great!

src/components/CopyButton.vue
krysal: Good call, thanks. I should have checked the mockups before, It seemed strange to me that the title was so small, and of course, it's fixed in the transitional mockups :) I'll make that change too.

krysal:
@zackkrida I added the info to the first post. The main thing that is left is to fix the quotation marks of the attribution text, I'll complete that today with some other minor styles fixes.

krysal:
About the last point, I'm not sure if we can do anything here, as it seems correct for RTL languages that the opening quotation mark is at the right (beginning) of the text and the closing mark at the end of the word, but it's confusing with the text in English. It's the same case with periods on the index page.

However, I added the `dir="ltr"` to the textarea of the HTML, as I believe that would be always read from left to right.

obulat (changes requested):
Looks great, and it's nice to see the code cleaned up more :) Can you also add the `dir="ltr"` to the HTML attribution span, as the quotes there are also placed incorrectly. This behavior is really weird, to be honest.

krysal:
@obulat Could you clarify more specifically which `span` needs the `dir` attribute? I'm not sure which one you are referring to.

obulat (approved):
Since the rich text IS supposed to be translated, it doesn't need to be set to ltr. Great improvements overall 👌

zackkrida (approved):
The page looks beautiful in Firefox, Safari, and Chrome at all breakpoints with `lang=rtl`. Awesome work!

PR #457, opened 2021-11-29
Add translation notification banner

### LINKED ISSUE #419: Banner for incomplete translations
## Problem

Currently, if a wp.org Rosetta site is only partially translated, especially if the % translated is _low_, like 1, or 2%, or the translation threshold is _high_  but not for key UI elements like the navigation, users might think there's something wrong with the site, their browser, or other issues when they attempt to navigate to a localized site.

## Solution 

We should implement [the banner Learn uses](https://learn.wordpress.org/?locale=dv) to make incomplete translations clear to the user. Learn uses data from the rest api to accomplish this: https://github.com/WordPress/learn/pull/202/files#diff-ddff3385c614ff999459619af1b876da8a6de11f0a5f0a774453db8d9bec1e5bR39-R52

obulat:
## Fixes

Fixes #419 by @zackkrida

## Description

This PR adds a banner inviting contributors to add translations for locales where fewer than 90% of strings are translated.
The translated percentages for locales are scraped from the HTML page instead of the API due to lack of CORS header on the wp.org site.
It uses the previously created `NotificationBanner` component, but refactors it to be more flexible through using slots for any actions. The `Notification` Vuex store is also removed.

This PR replaces a 18n-sync mixin with a similar composable, and adds two more composables: use-storage and use-event-listener, mainly based on the Vueuse library code.

To add the correct link to the translation banner, I had to use a full locales list instead of the `valid-locales.json` list of locales that have at least one translated string. To make the use of both lists together easier, I refactored the locales list to use camelCase property names instead of the snake_case that was previously used.

## Screenshots
I've used the banner color that was used previously on CC Search. @panchovm , what is the best color and position for this banner? Should it be full-width as now, or have the same maximum width as the header?

Screen Shot 2021-12-02 at 12 37 58 PM

Screen Shot 2021-12-02 at 3 15 54 PM

## Testing Instructions

> from Zack: You can also test this in a hacky way by editing the `onMounted` hook of use-i18n-sync with the following code: 
>   localeMsgHandler({
>     data: { type: 'localeSet', value: { locale: 'ar' } },
>   })

1. Update Nuxt locale files using `npm run i18n:create-locales-list` and then `npm run i18n:get-translations`.
2. Run the Nuxt app using `npm run dev`.
3.  Run the Openverse WordPress theme (instructions in the theme Readme).
4. Open the theme Customizer and set the embed URL to http://localhost:8443
5. Change the WP theme locale on `http://localhost:8888/wp-admin/options-general.php` options page. Note the icon next to the language select, it will help to change the language back from an unfamiliar one.
6. Reload the WP Openverse site, or open it anew by going to http://localhost:8888. You should see the Translation Banner below the header, with a correct link to the translation project on translate.wordpress.org. Note: for some reason, if I try to refresh 'http://localhost:8888/openverse', it doesn't work. Only going to the top page at 'http://localhost:8888' works.
7.  If you dismiss the banner, you should see the language code in the dismissed locales list in the localStorage. The localStorage saves the dismissed banner locale code so that if the user changes the locale, they are encourage to help with the new locale, too :)

This PR still doesn't update the translated status, which should be done regularly.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


package.json
```diff
@@ -35,7 +35,8 @@
     "i18n:get-valid-locales": "node src/locales/scripts/get-validated-locales",
     "i18n:get-translations": "node src/locales/scripts/get-translations && npm run i18n:get-valid-locales",
     "i18n:generate-pot": "node src/locales/scripts/json-to-pot",
-    "i18n:update-locales": "node src/locales/scripts/update-locale-list"
+    "i18n:update-locales": "node src/locales/scripts/update-locale-list",
+    "i18n:get-translated-status": "node src/locales/scripts/get-translations-status.js"
```
dhruvkb, line 39: Maybe drop the `.js` for consistency with other scripts.

dhruvkb (dismissed):
From a surface level it looks very promising. I'll take a deeper look soon.

zackkrida:
Looks really good so far, thanks @obulat! Please add testing instructions when you can. My only other question—I think we could _calculate_ the translation percentage by comparing the # of string keys in each `{locale}.json` file to the main `en.json` file. It would be nice to not rely on scraping for the percentages. I am open to the scraping if there's a benefit though.

panchovm:
For the transitional style and the redesign, I think we should follow the current notification style and place it above the Openverse header. I was looking for uses on WP.org and just found this in Learn section.

image

image

I will task myself to create a mockup for the redesign version.

obulat:
My brain dump of the process:
### Changes to locale json files

There are two lists of locales:
1. `locales-list.json` contains all of the GlotPress locales, even several that are not present on translate.wp.org.
2. `valid-locales.json` contains only the locales that have at least one translated string.

`locales-list.json` keys are `slug`, because those are the codes used for translation URL. (is this a good enough reason? Why not wpLocale?
This list is created in `update-locale-list.js` by scraping the GlotPress GitHub file with a list of locales. I want to update this list to 1. delete the locales that are not available on translate.wp.org, 2. add translation percentage. 
The scraping/update of the list of locales available on translate.wp.org should not happen often. Once a year?
However, this list needs to be regularly updated with the `translated` status.

So the process can be:

#### Setup or yearly update
1. `update-locale-list` can be called with a script that gets a list of locales by scraping GlotPress, and then filtering it, removing locales that are not available in translate.wp.org.

#### Weekly? update of translations and translation statuses
1. Call `update-translation-status` to update `translated` property in the `locales-list.json`. 2. `i18n:get-translations` updates the `valid-locales` list that also include the `translated` property with percent translated, as well as the actual `.po` - `.json` language files.
3. `valid-locales` are used when setting up the Nuxt i18n.locales list. This way, the `translated` property is available on i18n.locale or i18n.localeProperties.

obulat:
I've refactored some i18n scripts to improve fetching the locale data and updating it.
1. When setting up the app, you need to run `npm run i18n:create-locales-list`. This will fetch the list of locales and data associated with each locale from GlotPress. There are more locales in this list than what is available on `translate.wordpress.org`, so then the script parses the list from `translate.wordpress.org`, removes locales that are not found there, and also adds the status of translations to the list.

2. Next, you will need to run `npm run i18n:get-translations`. This will load the translated `.po` files from `translate.wordpress.org`, convert them to `.json` files that are usable in Vue i18n. It will also create two files: `valid-locales.json` for locales that have at least one translated string, and that will be loaded in Vue, and `untranslated-locales.json`, which is used for retrieving language name, translation status and the slug to create a link to this language's page on `translate.wordpress.org` for the translation banner.

The second script will need to be run regularly.

src/composables/use-i18n-sync.js
```diff
@@ -0,0 +1,147 @@
...
+   * @param {string} value.lang - the iso code for language
+   * (and sometimes country), e.g. 'en-US'.
+   * Default lang value on wp.org is 'en', and on wp.org/openverse - 'en-US'.
+   * @param {'rtl'|'ltr'} [value.dir] - the locale text direction.
+   */
+  const localeMsgHandler = async ({ data: { type, value } }) => {
+    if (type !== 'localeSet') return
+    // If the locale set by wp.org is 'en_US', not 'en', this is not necessary.
+    const wpLocaleValue = value.locale === 'en' ? 'en_US' : value.locale
+    let locale = i18n.locales.find((item) => item.wpLocale === wpLocaleValue)
```
zackkrida, line 104: When I test this PR, I'm not able to see the banner anywhere. It seems that `i18n.locales` does not have the `translated` key for each locale. If I manually add `locale.translated = 2` inside of `localeMsgHandler` however, I'm able to see the banner and everything looks great!

CleanShot 2021-12-09 at 13 53 34@2x

zackkrida (changes requested):
See inline comments. It's so close to working!

zackkrida:
By the way, I tested this quite simply by adding code to the `onMounted` call of use-i18n-sync to manually set the locale to Arabic:

```js
  localeMsgHandler({
    data: { type: 'localeSet', value: { locale: 'ar' } },
  })
```

src/composables/use-i18n-sync.js
obulat: Did you run `npm run i18n:create-locales-list` and `npm run i18n:get-translations`? I should've added that to testing instructions. This way, your `locale` lists should get updated with `translated` prop added (both the ones that are used in Vue in i18n.locales, and the ones that have no translations at all).

I wonder if I should:
**A.** Move the dark yellow border to the right for RTL languages (because it looks ugly with the scrollbar that I get when testing with the wp theme) or 
**B.** Remove it completely.

zackkrida (approved):
Thank you for updating the instructions! It works as-expected. Great work. There's so much more logic here necessary than one would expect; nice work building something so surprisingly complex.

src/composables/use-i18n-sync.js
panchovm: Jumping in to say that moving it to the right for RTL languages. I envision that in the long-term future we will work on a different solution for the whole WordPress site.

panchovm:
I tested the notification position below and above the Openverse header and it looks odd when divides the header and the filter area. Here is a screenshot of my point.

CleanShot 2021-12-13 at 12 18 36@2x

If we move it above, the broken layout gets solved.

CleanShot 2021-12-13 at 12 20 54@2x

zackkrida:
I agree with @panchovm about moving it above

obulat:
@panchovm , 
I have moved the banner above the header, and also tested what it looks like with the redirect banner on the homepage:
Screen Shot 2021-12-14 at 7 58 55 AM

This is the current view with the filters open:
Screen Shot 2021-12-14 at 8 02 19 AM

zackkrida:
Looks great! I don't love the banner with the CC banner, but it's not _terrible_. Great work

dhruvkb (approved):
I merged `main` into the PR to incorporate the `pnpm` changes. LGTM!

zackkrida:
Fixed a formatting issue, if CI passes I'll merge!

Issue #414, opened 2021-11-13
Move away from `nuxt-ssr-cache`


zackkrida:
## Problem

For microcaching of our server-rendered pages, we currently use a poorly-maintained package `nuxt-ssr-cache`. We also use `patch-package` to modify this broken package.

## Solution

There are a few alternative solutions:

1. Switch to an NGINX based approach to microcaching: https://www.nginx.com/blog/benefits-of-microcaching-nginx/
2. Switch to nuxt-perfect-cache with Redis: https://www.npmjs.com/package/nuxt-perfect-cache

I prefer approach 1, as it removes caching from an app-level consideration and into our infra layer.

With either approach we can:

1. Remove the `"postinstall": "patch-package",` script
2. Remove the `patches/` directory

zackkrida:
Fixed via #415

PR #467, opened 2021-12-01
Add /healthcheck endpoint

### LINKED ISSUE #466: Create a health check endpoint
## Problem
Currently the Load Balancer is using `/` as a health check endpoint. 

## Description
Taking into consideration the ELB checks the same endpoint 5 times per minute, I'd suggest to consider to return an empty response instead of loading all the assets. This also will give the benefit to avoid shallowed metrics of requests that belongs to the Load Balancer.

## Additional context
[There is a module, I've found out there](https://github.com/fukuiretu/nuxt-healthcheck) can be helpful. Take a look and let me know your thoughts about it.

## Implementation

obulat:
## Fixes

Fixes #466 by @rbadillap

## Description

This PR adds a `/healthcheck` endpoint that returns 'OK' with the following headers:
```
HTTP/1.1 200 OK
Date: Wed, 01 Dec 2021 06:15:17 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Content-Length: 2
```

## Testing Instructions

Run `npm run dev`, and go to `http://localhost:8443/healthcheck`. You should get a text "OK" response.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/server-middleware/healthcheck.js
+++ b/src/server-middleware/healthcheck.js
@@ -1,0 +1,4 @@
+export default function healthcheck(req, res) {
+  res.setHeader('Content-Type', 'text/plain')
+  res.end('OK')
+}
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -167,6 +167,9 @@ export default {
     scss: ['./styles/utilities/all.scss'],
   },
   modules: ['@nuxtjs/sentry', '@nuxtjs/i18n', '@nuxtjs/sitemap'],
+  serverMiddleware: [
+    { path: '/healthcheck', handler: '~/server-middleware/healthcheck.js' },
+  ],
   i18n: {
     locales: [
       {


src/server-middleware/healthcheck.js
```diff
@@ -0,0 +1,4 @@
+export default function healthcheck(req, res) {
+  res.setHeader('Content-Type', 'text/plain')
+  res.end('OK')
```
dhruvkb, line 3: It's recommended to use `.end` when [not returning any data](https://expressjs.com/en/api.html#res.end). To send data, it's better to use `.send` instead. Returning an object like the following would make the response identical to API and ingestion-server.

Also `.send` sets the `Content-Type` header automatically!

```suggestion
  res.send({ status: '200 OK' })
```

dhruvkb (changes requested):
Although the use of middleware to define an endpoint is not ideal, I can definitely get behind this approach given how succinct the resulting code is!

obulat:
> Although the use of middleware to define an endpoint is not ideal

@dhruvkb, I think that you are referring to `middleware`, not the `serverMiddleware`[^1] used here.
Or is there something I should research more?

[^1]: https://nuxtjs.org/docs/configuration-glossary/configuration-servermiddleware/#servermiddleware-vs-middleware

dhruvkb:
I was referring to the [Custom API endpoint](https://nuxtjs.org/docs/configuration-glossary/configuration-servermiddleware/#custom-api-endpoint) section from the Nuxt docs where they import `express` to create an app and then register routing on the app instead of directly returning a response from the middleware.

I actually like what you have done, which is much simpler and readable so 👍. I requested changes for the response, not the approach.

src/server-middleware/healthcheck.js
obulat: After some research: The `res` here is the Node.js http response object: https://github.com/senchalabs/connect#appuseroute-fn (not express), so using `res.send` causes a `res.send is not a function` error.

dhruvkb: Ooooh, that's interesting! I was under the impression that Nuxt was running Express under the hood. While we can still do something like
```suggestion
  res.setHeader('Content-Type', 'application/json')
  res.end(JSON.stringify({ status: '200 OK' }))
```
it might be too much effort for a response that doesn't even matter to a health-check service.

dhruvkb (approved):
LGTM!

src/server-middleware/healthcheck.js
obulat: I hope @rbadillap can clarify if plaintext 'OK' response is enough, or JSON is better?

rbadillap: The sole purpose of this report is to reduce the size of the response, so from a DevOps perspective there is no difference between the two solutions as we are achieving the main goal.
Thank you for tackling this so quick.

krysal (approved):
LGTM 🚀

zackkrida (approved):
In the current frontend deployment strategy we handle the healthcheck endpoint via the NGINX configuration:

https://github.com/Automattic/openverse-infrastructure/blob/d89f76572b1412ead83b75fb0925454ed529a69d/modules/services/frontend/init.tpl#L149-L152

```nginx
location /healthcheck {
  default_type "text/plain";
  return 200 'healthy';
}
```

@rbadillap I don't know if you've developed a replacement for the microcaching we removed in #414, but if we were to implement that with NGINX we _could_ continue to have this endpoint live there. For now though I think this is a small, lightweight solution that works well. It also ensures that our Nuxt site is actually working—with the current healthcheck we can run into issues where the healthcheck passes but the Nuxt app failed to build.

PR #469, opened 2021-12-02
Add Modal component

### LINKED ISSUE #458: Modal component
We need to add an accessible modal component that can be used for the single result view in the redesign.

There's a lot of reading to keep in mind while working on this:

https://www.w3.org/TR/wai-aria-practices/#dialog_modal
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

There are some good React implementations to keep in mind:

https://assortment.io/posts/accessible-modal-component-react-portals-part-2
https://reakit.io/docs/dialog/

Avoid referencing the [Vue 2 modal example in the docs](https://vuejs.org/v2/examples/modal.html) as it is not an accessible modal implementation.

A lot of the behavior that is needed for an accessible modal can be extracted from the `usePopoverContent` composable. Most of the composables that composable references should be able to be abstracted to handle a modal as well. Because a lot of it is inspired by Reakit, it'd be a good idea to see what Reakit is doing for the base Dialog component when `modal=true`.

One thing we can ignore from the Reakit implementation is multiple dialog handling _I think_... we should make sure to test having a popover inside the modal and that it behaves as expected (escape when the popover is open should only close the popver, not the modal; but clicking outside the modal should close everything). Seems like a real can of worms!

## Usage example

```vue
<Modal @close="handleModalClose" @open="handleModalOpen">
  <template #trigger>
    <!-- Could be a button but I suppose it could also be a result element (like a grid image) with accessible keyboard handling for causing a `click` event... might need to revisit the trigger handling to include activation via keyboard in case a `click` event isn't bubbled... not sure how to prevent duplicate events in that case though -->
    <VButton>Open the modal!</VButton>
  </template>
  <!-- Whatever modal content we want :) -->
</Modal>
```

sarayourfriend:
## Fixes

Fixes #458 by @sarayourfriend 

## Description

Adds a new Modal component based on the same code as the popover.

Refactors the popover composables to be more generic "dialog" composables and applies them to the modal.

https://user-images.githubusercontent.com/24264157/144447616-8bc4a7c8-16d3-4ad1-ad25-c490ac059088.mov

## Testing Instructions

Run `npm run storybook` and visit `http://localhost:6006/?path=/story/components-vmodal--default` and test the modal with different props.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
Outstanding question: should we scroll-lock the background (main page) when the modal is open?

src/components/VButton.vue
```diff
@@ -87,14 +87,15 @@ const VButton = defineComponent({
       default: false,
     },
     /**
-     * The size of the button.
+     * The size of the button. `disabled` removes all internal padding allowing
+     * the consumer of the component to determine the padding.
```
zackkrida, line 91: This is a nice naming convention. Great idea

sarayourfriend:
Also wanted to point out that yes the Modal and Popover implementations are almost identical, but I'm not sure how to abstract them properly to combine them 🤔 I think a base `VDialog` component would make sense that VModal and VPopover would extend but with how the Vue templates work I think it would end up being more or less the same amount of code and prop documentation duplication that we have now.

src/components/VModal/VModal.vue
```diff
@@ -0,0 +1,159 @@
...
+} from '@nuxtjs/composition-api'
+import VModalContent from '~/components/VModal/VModalContent.vue'
+
+export default defineComponent({
+  name: 'VModal',
+  components: { VModalContent },
+  /**
+   * NB: Most of these technically default to `undefined` so that the underlying `VPopoverContent`
+   * default for each of them can take over.
+   */
```
zackkrida, line 47: I was going to ask this while looking at the props, glad to see it answered right here 😆

src/components/VModal/VModalContent.vue
```diff
@@ -0,0 +1,107 @@
...
+      role="dialog"
+      v-on="$listeners"
+      @keydown="onKeyDown"
+      @blur="onBlur"
+    >
+      <div class="w-4/5 flex justify-end">
+        <VButton
+          size="disabled"
+          variant="floating"
+          class="text-white py-2"
```
zackkrida, line 16:
```suggestion
          class="text-white my-2"
```

The focus style is less than ideal with this as padding:

CleanShot 2021-12-02 at 10 36 32@2x

After switching to margin:

CleanShot 2021-12-02 at 10 36 21@2x

Alternatively, `py-1 my-1` produces an in=between result:

CleanShot 2021-12-02 at 10 38 22@2x

zackkrida (changes requested):
This is awesome, and I feel like you wrote this so fast! 😅 I have a question and a suggestion:

1. When testing, I'm seeing the focus ring on these pink buttons is _also_ pink, which seems like really insufficient contrast. Should I make a separate issue for that? I'm not sure if that's been introduced in this PR or not. 
  
https://user-images.githubusercontent.com/6351754/144453697-a8c33fef-22f6-408d-a22f-f5aab79474e9.mp4

2. I think unlike the popover, the modal should _trap_ focus rather than open/close when tabbing away from the first or last focusable element within the modal. In the current form I'm not able to focus on the 'Close' button because when I tab to it the modal seems to close immediately. It's great that focus is returned to the trigger correctly after closing though! The existing https://github.com/WordPress/openverse-frontend/blob/16ae200e79a029d6fd87ea9f436096073eaaf4ba/src/components/AppModal.vue component (used for the current mobile filters) traps focus using a 3rd party dependency, and I'm not sure if it would conflict with your awesome focus composables or not so we might need to once again roll-our-own. 

Edit, to address some of your comments:

> Outstanding question: should we scroll-lock the background (main page) when the modal is open?

Yes, I think so, particularly for the use case when the modal has an internal scrollbar, that would be intuitive for users to have two scrolling contexts to contend with.

> but I'm not sure how to abstract them properly to combine them

I think the focus trapping behavior could make them distinctive enough where this won't be necessary.

sarayourfriend:
I totally forgot about focus trapping, thanks for catching that! Shouldn't be too bad to fix.

The focus styles for the close button in general are bad, I put the regular focus ring in as a placeholder and meant to ping @panchovm about it. I don't think the focus ring by itself works well because the close button is right aligned with the edge of the modal so the focus ring is too "tight" horizontally with the button contents.

I'm confident Francisco can come up with a good solution for it 😁

sarayourfriend:
> When testing, I'm seeing the focus ring on these pink buttons is also pink, which seems like really insufficient contrast. Should I make a separate issue for that? I'm not sure if that's been introduced in this PR or not.

Not sure when that was introduced but this PR doesn't change the button other than adding a new variant so I think it's a separate issue.

src/components/VButton.vue
```diff
@@ -270,4 +271,8 @@ a.button {
 .grouped-pressed {
   @apply bg-dark-charcoal-10 ring-offset-dark-charcoal-10;
 }
+
+.floating {
+  @apply bg-tx focus-visible:ring-pink;
+}
```
sarayourfriend, line 277: We could also name this `plain` instead of `floating`? That might be more generally applicable 🤷

zackkrida: Plain might be better. We also called this type of button `.text` at Creative Commons but I think that's a bit confusing.

zackkrida (approved):
Awesome. Could you make a new issue for button focus styles? We can probably tackle the floating + default styles in one new PR with some guidance from @panchovm

sarayourfriend:
Right now opening the modal via keyboard interaction will always focus the close button first, as it is the first tabbable element inside the `dialogRef` node. I think we could potentially add a prop for this like `skip-close-button-on-focus` (ugh, names!) that would skip the close button and focus the second tabbable element.

Reakit's implementation of this is nice because you can just pass an `initialFocusElement` prop but when I tried to get that to work with the popover I couldn't because at the time when you want to do the focus management, the `ref` for your internal node could still be empty. I wonder if we could somehow create a custom ref for this that would be able to signal like "reference pending re-render" or something so that we could just use `initialFocusElement` instead of trying to do something jankily named prop like `skipCloseButtonOnFocus` or whatever.

zackkrida:
I once implemented a modal in react with something to the effect of an 'initial-focus-element-index' prop where if you wanted, say, the 5th element in a modal to be focused first, you passed the index`5` to the prop. Terrible api design 😆

sarayourfriend:
> you passed the index5 to the prop
😱

panchovm:
I am not sure about the interface/visual consequences when @sarayourfriend says that Modal and Popover are equal. I guess it is just an implementation aspect. In terms of behavior, I like the idea of pressing `escape` any time on the modal to close it. The modal layout in general suggests a sub-navigation where it seems like a glance of a long searching for relevant content journey.

Regarding Modal size and breakpoints, I will work on a proposal. But for now, the desktop version is in the audio and image single result mockups. Whereas the mobile proposal I am handling looks like opening a new page but with a back to results action top-left aligned. However, I am seriously considering replacing it with a top-right aligned close action.

I will share a proposal for the focus style of close button soon. My initial guess is having a similar top-bottom and left-right padding that goes off the modal's width with a `1.5px` white border. After all, the focus style is for highlighting the element itself.

> Outstanding question: should we scroll-lock the background (main page) when the modal is open?

Yes, I agree with @zackkrida on this.

krysal (commented):
It was really super fast, considering all the details that the modal involves, and above the interaction with the Popover. Works great ⭐️  and looks pretty complete to me. I just think the new components need some testing.

sarayourfriend:
> I just think the new components need some testing.

I agree! I've been trying to decide how to do this because most of the things that are actually testable in this component are already tested in the popover. The things the modal adds are:

1. Focus trapping - No need to test this as it's a third party library and we don't need to test it's implementation
2. Body scroll locking - I don't know how to test this without testing the implementation details and, in fact, I don't even know that we could test it in JSDom as I don't believe it emulates scroll position.

Body scroll locking could be tested in the e2e tests but this modal isn't implemented anywhere yet and we don't run the e2e tests against storybook I don't think.

I suppose I could add a test for the open/close behavior and the close button though.

src/components/VModal/VModal.vue
```diff
@@ -0,0 +1,166 @@
+<template>
+  <div ref="nodeRef">
+    <div ref="triggerContainerRef" @click="onTriggerClick">
+      <!--
+        @slot The trigger. Should be a button 99% of the time. If you need custom event handling on the trigger button, ensure bubbling is not prevented or else the dialog will not open.
+          @binding {object} a11yProps Props to v-bind to the trigger element to ensure accessibility
+          @binding {boolean} visible Whether the dialog is currently visible (open)
```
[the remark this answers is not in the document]
sarayourfriend: Sure, though I'm not sure how well wrapping works with this bespoke Vuestyleguidist documentation scheme. I'll try and see if Storybook is able to pick it up though!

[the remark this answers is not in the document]
krysal: Actually, I'm noticing now that Storybook is not showing all the stories for me (e.g. `AudioTrack`s and `VLicense` are missing), not even the index page (!?), and not only on this branch, main has the same problem for me locally. I'll have to look at what happens here before so I can try something.

sarayourfriend:
@krysal can you re-review? I added tests for the close button and the scroll locking. Everything else is already tested by the VPopover tests. Those tests need to be abstracted out of the VPopover test suite. I tried doing this in this PR but I ran into some blockers and weird behavior... and it would increase the size of this PR significantly to re-write those tests.

I opened this issue https://github.com/WordPress/openverse-frontend/issues/501 to address that in the future.

krysal (commented):
Thanks for adding the tests @sarayourfriend, good to have those dependencies updated 💯 A _small_ issue appeared on the page search results page showing this:
```
An error occurred

Your focus-trap must have at least one container with at least one tabbable node in it at all times
```
It's another component using focus-trap: `SaferBrowsing`, binding its `active` attribute instead of using true always seems to solve it for me.
```vue
<!-- File: src/components/SaferBrowsing.vue -->
<FocusTrap :active="showForm">
```
Other than that it looks good to me!

sarayourfriend:
Thanks for catching that @krysal! It's fixed now 🙏

PR #473, opened 2021-12-02
Add e2e test for image detail page

### LINKED ISSUE #472: Required e2e tests for single result page
## Problem

Testing whole pages which involves several components with jest is complicated, requires many congifurations just for the setup, so adding e2e tests test should allow us to still test these pages.

## Description

Add tests to ensure that single result page load and display the media information correctly

## Alternatives

Find a good general configuration for jest with nuxt and the many plugings (especially i18n). 

## Additional context

Related to #394.

## Implementation

krysal:
## Fixes

Fixes #472 by @krysal

## Description

This PR:
- adds a playwright test to ensure the image detail page load correctly from links in the search results page
- renames the `/photos/` route to `/image/` as Openverse serves more than photos
- moves the test of the image's author display to a playwright test 

I wanted to add these tests to the GitHub action workflow but they're still kind of flaky, depending on internet to make requests as the `route.fulfill` method doesn't seem to work.

## Testing Instructions

```
npm run e2e:ci
```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (approved):
LGTM! I ran the tests locally and they passed 🎉 Nice work! 🚀

test/e2e/search.spec.js
```diff
@@ -55,3 +55,11 @@ test('shows search result metadata', async ({ page }) => {
   // All search results have been shown, cannot load more
   await expect(loadMoreButton).toHaveCount(0)
 })
+
+test('navigates to the image details page correctly', async ({ page }) => {
+  await page.goto('/search/image?q=honey')
+  const figure = page.locator('figure').first()
+  const imgTitle = await figure.locator('img').getAttribute('alt')
+  await figure.click()
+  await expect(page.locator('h1')).toHaveText(imgTitle)
```
obulat, line 64: Love the logic of expecting the image alt text to be the h1!

src/components/ImageGrid/ImageCell.vue
```diff
@@ -8,7 +8,7 @@
     <figure class="search-grid_item">
       <i :style="`padding-bottom:${iPadding}%`" />
       <NuxtLink
-        :to="localePath('/photos/' + image.id)"
+        :to="localePath('/image/' + image.id)"
```
obulat, line 11: Thank you for changing /photos/ to /image/! It makes the terminology consistent.

obulat (changes requested):
Did you plan to add any more tests, @krysal ?

I think it is important for e2e tests to test both Server-side rendering and client-side rendering. For single image result I think it might get flaky to try to navigate from search result to single image result, as we cannot guarantee the order of the results. 

And still, I think we should add a test that would check not for the specific author name or title, but that the items like author or title **are there** when navigating from search result to single image result.

zackkrida:
Just a note—if we're going to rename routes we'll need to add a redirect somewhere so old urls aren't broken

krysal:
@obulat I think in this case both modes are tested at least very basically. The client-side navigation in the case I added to `test/e2e/search.spec.js`, and the server-side with the new `test/e2e/image-detail.spec.js` but thinking about this, adding the check for the breadcrumb is perfect here! The unit test does this by modifying a property directly, so these tests double-check it doing what a real user would do and see.

I updated the test to be more general, although before I wanted to make it really specific, controlling the data passed to pages and asserting the specific names, title, and text info in general, as is the only way to make sure the correct data is shown, but this is complicated in the server-side mode.

At first, my intention was to replace the unit test that I had commented and add the test for navigation from the search results. Do you have in mind another thing specifically that should be tested here?
<hr>
@zackkrida Good call, I'll add the redirect.

obulat:
The tests currently mock the client-side requests, but not the server-side ones. So, the first server-side render is happening without mocking.
But I've found a tutorial about mocking the server-side requests in Next.js (the Nuxt of the React world :) ): https://frontend-digest.com/using-playwright-to-test-next-js-applications-80a767540091. I guess I should open a new issue for that.

test/e2e/search.spec.js
```diff
@@ -21,37 +21,47 @@ test.beforeEach(async ({ context }) => {
 
 // TODO(obulat): Fix the search metadata before search
 //  This test currently passes, although everything should be the opposite
-test('does not show an error message before search', async ({ page }) => {
-  await page.goto('/search')
+// test('does not show an error message before search', async ({ page }) => {
```
obulat, line 24: You can add `.skip` to a test that doesn't pass instead of commenting it out:
`test.skip('shows search result metadata',` ...

obulat (approved):
LGTM!

PR #489, opened 2021-12-06
Add VFilterButton

### LINKED ISSUE #476: VFilterButton
## Description

The Filters button is a button with an icon in the header. It always looks the same on Desktop, but on Mobile it shows the filters icon when no filters are applied, or the number of filters checked if any filters are applied.  When clicked, it opens the filter checklist.
### State

| Name | Type | Description |
|-|-|-|
|appliedFilterTags|Array| An array of filters that are applied. Changes the button content for mobile: If empty, the button will only have an icon, and if not, the button will show the number of applied filters.
| appliedFilterCount | Number | The number of applied filters, will be used on mobile.
| isTablet | Boolean | From useMediaQuery composable, will determine the button content.
| isScrolled | Boolean | From useScroll composable, needed for mobile header (when on top of the page, the header is in two lines: logo/content switcher / filters button and SearchBar. When scrolled, the two lines are combined into one.
### Events

|-|-|-|
| toggle-filters | - | Will be emitted when the button is clicked, and will open/close the filters checklist.

## Code samples

```
<VButton>
  <FilterIcon v-show="isTablet || !isAnyFilterApplied" />
  <span v-show="isTablet">Filters</span>
  <span v-show="!isTablet && isAnyFilterApplied>{{ appliedFilterCount }} Filters </span>
</VButton>
```
## Dependencies

- FilterIcon

## **Figma:**
Desktop: https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=352%3A446
Mobile: https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=344%3A1057
## Implementation

obulat:
From Zack:

Building on Olga's work and with confirmation from @panchovm I was able to simplify the state management here considerable and I also believe all the varied states are correct. There's like 5-6 permutations _per_ viewport and getting them right was intense!

The biggest way this was simplified, was previously it was attempting to turn the filter button into the close button for the mobile menu. 

CleanShot 2021-12-15 at 17 41 56@2x

This seems natural given the positioning, but ultimately really added an extreme level of complexity to this component. Instead, I see this Openverse logo and close button in the above screenshot as _contents of the mobile menu_, which appear  _overlapping_ the mobile header. This allows us to keep the close button inside the mobile menu, and keep this component in scope. 

Other changes here that _are_ outside the scope, that I apologize for:

- Header padding optimized
- Logo size fixed to match mockups
- Header switched to sticky instead of fixed
- Hide filter button on the homepage

A note: test coverage only tests for label and icon usage, _not_ the button variation and appearance. I think it's sufficient coverage; you may disagree!

> Original message:
> ## Fixes
> 
> 
> Fixes #476 by @obulat
> 
> ## Description
> 
> 
> This PR adds the `VFilterButton` component in `VHeader`.  This is a button that opens and closes the `FiltersSidebar`, and also shows the number of applied filters on mobile screens.
> 
> ## Testing Instructions
> 
> Run `npm run dev`, and open http://localhost:8443/search?q=cat. You should see a Header with only the `VFilterButton` component. You can open and close the `FiltersSidebar` clicking on it, and if you narrow your screen, it also shows you the number of filters applied.
> 
> ## Outstanding Todos:
> - [x] Improve the styles (sizes, paddings, check for active/hover states).
> - [x] Improve the docstrings.
> - [ ] Improve the Storybook story.
> - [x] Add tests.
> 
> ## Checklist
> 
> - [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
> - [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
> - [x] My commit messages follow [best practices][best_practices].
> - [x] My code follows the established code style of the repository.
> - [x] I added or updated tests for the changes I made (if applicable).
> - [ ] I added or updated documentation (if applicable).
> - [x] I tried running the project locally and verified that there are no visible errors.
> 

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+  <VButton
+    variant="action-menu"
+    class="self-center"
+    :pressed="sidebarVisible"
+    @click="toggleFilters"
+  >
+    <FilterIcon v-show="showIcon" class="w-6 h-6" width="24" height="24" />
```
sarayourfriend, line 8: It's nice to be able to inline icons like this, but should we be using the `VIcon` component instead? It seemed nice to be able to remove the size classes and attribute duplication using it.

src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * On screens wider than tablet: 'Filters'.
+     * On mobile screens:
+     * 1. If no filters are applied - '' (icon only).
+     * 2. If the page is scrolled down - '' (icon only).
+     * 3. If the page wasn't scrolled, and some filters are applied:
+     * the number of applied filters, e.g. '7'.
+     *
+     * @type {ComputedRef<string>}
```
sarayourfriend, line 60: For TS you'll need to explicitly import these types:

```suggestion
     * @type {import('@nuxtjs/composition-api').ComputedRef<string>}
```

src/composables/use-media-query.js
```diff
@@ -13,7 +13,7 @@ export function useMediaQuery(query, options = {}) {
   if (!window) return ref(false)
 
   const mediaQuery = window.matchMedia(query)
-  /** @type {import('@nuxtjs/composition-api').Ref<boolean>} */
+  /** @type {import('@nuxtjs/composition-api').Ref<Boolean>} */
```
sarayourfriend, line 16: `Boolean` with an uppercase b actually refers to the primitive constructor literally rather than the abstract type for `boolean`. Same with the other JS primitives. Object is even more complicated as there are _three_ ways to reference an object type: https://stackoverflow.com/questions/49464634/difference-between-object-and-object-in-typescript

```suggestion
  /** @type {import('@nuxtjs/composition-api').Ref<boolean>} */
```

src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,124 @@
...
+
+      toggleFilters,
+    }
+  },
+})
+
+export default VFilterButton
+</script>
+
+<style>
```
sarayourfriend, line 114: Can we use `scoped` or `module` here?

src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,124 @@
...
+     * the number of applied filters, e.g. '7'.
+     *
+     * @type {ComputedRef<string>}
+     */
+    const buttonLabel = computed(() => {
+      if (isTablet.value) {
+        return i18n.t('header.filter-button.simple')
+      }
+      return isScrolled.value || !isAnyFilterApplied.value
+        ? ''
```
sarayourfriend, line 67: Can we use `header.filter-button.simple` here? Otherwise when you load the page there is pop-in for this where the filter button has no text when the page initial loads and then suddenly has text after the page settles.

sarayourfriend (changes requested):
Looks good so far! I love small components like this 🙂 

Just one thing blocking this (the empty label case) but otherwise I think it's good to go!

src/composables/use-ui-state.js
```diff
@@ -0,0 +1,18 @@
+import { readonly, ref } from '@nuxtjs/composition-api'
+
+const isFilterSidebarVisible = ref(false)
```
sarayourfriend, line 3: Are global refs safe to use this way?

src/pages/search.vue
```diff
@@ -26,34 +27,50 @@ import {
...
     )
+    const { isFilterSidebarVisible, setFilterSidebarVisibility } = useUiState()
 
+    onMounted(() => {
+      const localFilterState = () =>
+        local.get(process.env.filterStorageKey)
+          ? local.get(process.env.filterStorageKey) === 'true'
+          : true
+      setFilterSidebarVisibility(isLargeScreen && localFilterState())
+    })
```
sarayourfriend: I wonder if this could all be built into the `useUiState` composable? In fact, to avoid `useUiState` from growing to encompass the entire UI state, could we rename it to `useFilterSidebarVisibility` or something like that? And then all the logic for determining whether it is open or closed etc can go in there.

test/unit/specs/components/VHeader/v-filter-button.spec.js
```diff
@@ -47,10 +48,13 @@ describe('VFilterButton', () => {
...
-          i18n,
+      propsData: props,
+      mocks: {
+        $nuxt: {
+          context: {
+            store: storeMock,
+            i18n,
+          },
         },
       },
```
sarayourfriend: Did this work for you btw? I tried this to mock the `i18n` object and I could _not_ get it to work at all! That's why i ended up creating the `useI18n` composable so that it would be easier to mock.

obulat: I just confirm, @sarayourfriend , and yes, it does work (I added some console.logs and ran the tests to be sure ) 
Did you also create the i18n object and load the messages as I did at the top of this file?
https://github.com/WordPress/openverse-frontend/pull/489/files/87c23926df95a8bfe4012069674e7485c12c5c18..e7f624fef9d52d8f279490cfd92ed6c96b18f6cb#diff-323a77061c65693ad59cc4cb90840d5fbcdc403428f023ffb3b4a448fc234a00R6-R31

sarayourfriend: I did! I remember everything you've done here but I must have left _something_ out becuase I just could not get it to work! Glad you figured it out though 🎉

src/composables/use-ui-state.js
obulat: This is the solution for using composable for global state: if the ref was inside the function, then it would be recreated in every component that uses it. This way, all the components using `isFilterSidebarVisible` are getting (and updating) the same value.
Here's a helpful article from VueSchool: https://vueschool.io/articles/vuejs-tutorials/state-management-with-composition-api/

src/components/VHeader/VFilterButton.vue
obulat: The style was blank, so I removed it completely.

src/composables/use-media-query.js
obulat: Done

src/components/VHeader/VFilterButton.vue
obulat: I changed the way label is set, and added `sr-only` label for icon button.

src/pages/search.vue
obulat: I had an idea to use `UI state` for state that is mainly about the UI (something shown or not), but you are right, it can grow uncontrollably, so I renamed it.

test/unit/specs/components/VHeader/v-filter-button.spec.js
obulat: It's weird, I really wonder why id didn't work now ... 🤔

src/composables/use-ui-state.js
sarayourfriend: Gotcha, yeah I understand the difference here and the why, I just didn't realize it was _possible_. In React you can't have global refs because you can't call hooks outside of components. Point for Vue! 🙂

src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,140 @@
...
+     * @default false
+     */
+    isScrolled: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  setup(props) {
+    const { i18n, store } = useContext()
+    const { isScrolled = false } = props
```
sarayourfriend, line 50: Destructuring the `props` object will remove reactivity from the properties and their values will get stuck on the first value. Learned that one the hard way! 😭

src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,140 @@
...
+     * Whether the page has been scrolled. If true on mobile, the button
+     * label is blank.
+     * This is currently not implemented, and is always false.
+     * TODO(@obulat): add use-scroll composable to VHeader
+     * @default false
+     */
+    isScrolled: {
+      type: Boolean,
+      default: false,
+    },
```
sarayourfriend, line 46: Why does this component need this as a prop, rather than being able to use the `useScroll` composable itself to tell whether the page is scrolled?

obulat: I guess my thinking at the time was that scroll position will be used by several header components, and could be set once in the header and passed down to children as props. I have to look if it's possible to set isScrolled as a global ref that could be used by all components...

sarayourfriend: Ah I see, to avoid processing it multiple times? A global ref seems like it would be a good use case for that 😁 What a handy tool!

src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,140 @@
...
+    :variant="buttonVariant"
+    class="self-center"
+    :pressed="pressed"
+    aria-controls="filter-sidebar"
+    @click="toggleFilters"
+  >
+    <VIcon
+      v-show="showIcon"
+      :icon-path="filterIcon"
+      :class="{ 'me-2': iconNeedsPadding }"
```
zackkrida, line 12: This padding is displaying on middle breakpoints, where the text label is no longer visible:

CleanShot 2021-12-09 at 13 07 22@2x

zackkrida (changes requested):
Looks great! In addition to the inline comments:

- The hover style seems wrong, there's an odd border over the black
- The font weight should be 600

And some non blocking concerns:

- I think after we build the header we'll need to review sizing and spacing with Francisco (the logo actually looks huge to me now) but I don't think those are blocking concerns
- The button currently displays on the homepage, which seems odd, but perhaps we can also deal with that later.

src/components/VButton.vue
```diff
@@ -72,6 +73,7 @@ const VButton = defineComponent({
           'secondary',
           'tertiary',
           'action-menu',
+          'action-menu-secondary',
```
sarayourfriend, line 76: Hmmm, I'm beginning to worry we have too many variants! I wonder if we can ditch any of these? Or maybe nudge @panchovm to unify the button variants so that we don't have so many 🤔

src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,166 @@
...
+    }
+    // TODO(obulat): Remove buttonVariants. The aim of this object was to lower
+    // the cognitive load of remembering the variant names when debugging the
+    // button.
+    const buttonVariants = {
+      pressed: 'action-menu-pressed',
+      gray: 'action-menu-muted',
+      bordered: 'action-menu',
+      textOnly: 'action-menu-secondary',
+    }
```
sarayourfriend, line 81: Looks like a dangling todo?

src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,166 @@
...
+    const isAnyFilterApplied = computed(() => appliedFilterTags.value > 0)
+
+    const labelWithCount = (count) => {
+      return i18n.tc('header.filter-button.with-count', count)
+    }
+    // TODO(obulat): Remove buttonVariants. The aim of this object was to lower
+    // the cognitive load of remembering the variant names when debugging the
+    // button.
+    const buttonVariants = {
+      pressed: 'action-menu-pressed',
```
sarayourfriend, line 77: I guess this is _technically_ a valid variant that you can pass to the button component in that it will render correctly... but really it should be a combination of the `action-menu` variant and `:pressed="true"` prop. Ideally `pressed` would just be passed through to the button component and all of the pressed styles for the variants used by this component would be the same?

I'm still kind of sad that we need three _slightly_ different variants for this component 😞

src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,166 @@
...
+     * Since the page is rendered as mobile first, i.e. with a visually-hidden
+     * text label initially, running updateButton adds a visual label if the
+     * screen size is larger than mobile when it is mounted.
+     */
+    onMounted(() => updateButton())
+
+    watchEffect(() => {
+      updateButton()
+    })
+
```
sarayourfriend, line 138: This can be condensed into just a `watchEffect` with the `immediate` option:

```suggestion
    /**
     * Since the page is rendered as mobile first, i.e. with a visually-hidden
     * text label initially, running updateButton adds a visual label if the
     * screen size is larger than mobile when it is mounted.
     */
    watchEffect(
      () => {
        updateButton()
      },
      { immediate: true }
    )

```

sarayourfriend (approved):
I'm hopeful we can improve the button variant situation... it seems like it's getting out of hand with the number of `action-menu` based variants we have. I wonder if there's a way to simplify that OR alternatively extract the complexities of those various similar variants into a parent component.

My main concern over having so many variants is that it feels like the UI could end up with lots of similar but slightly different buttons. I'm sure Francisco has an eye on that though!

Otherwise this tests fine locally for me!

I _might_ prefer to not have Vuex inside this component... like instead have it just take in props and emit events when it is clicked and have the Vuex wiring done in a parent component, but it's not really necessary and is just a philosophical discussion we can have some time 🙂

src/components/VHeader/VFilterButton.vue
obulat: I was planning to keep it for the work on the assembling the header, and removing after it all works together.

src/components/VHeader/VFilterButton.vue
obulat: Thank you, I didn't know about that :)

src/components/VHeader/VFilterButton.vue
obulat: I didn't read the code well-enough to notice that there is no 'action-menu-pressed' variant, sorry about that. I removed the 'pressed' variant, and only left the two different 'action-menu' variants: one with a border, and one without. I agree that it should be possible to join these two variants in favor of additional prop? or some other config?

zackkrida (changes requested):
I'll make a video in a minute but I'm seeing a bunch of state management and visual quirks:

- The word "Filters" doesn't appear on desktop
- The video tab is broken with a fetching error
- Extra padding-right "pops in" on desktop page load
- On mobile, after closing the modal the Vilter button still says "Close X"

obulat:
- The word "Filters" doesn't appear on desktop
- Extra padding-right "pops in" on desktop page load

These two problems are related, and I'm really frustrated because I spent almost an hour trying to figure out why re-rendering isn't happening, and thought that I had figured it out this morning.
I can't figure out why Vue/Nuxt isn't re-rendering the component. I set the Label span text to 'Filters', and its class to 'sr-only' by default to make the component mobile-first. Then, onMounted, the label text and class are recalculated, and I can see that they DO update. But the component still has 'sr-only' class !

src/components/VHeader/VFilterButton.vue
sarayourfriend: Maybe we need a `pressed` and `active` prop? 🤔 For most variants this is the same state I think and I'd have a hard time understanding the difference between those two props at face value personally... hmmm. Gutenberg has an `is-busy` prop but that adds an animation, and I don't think it quite fits with our situation here.

We could just remove the `muted` variant and add that as a custom state to the parent VFilterButton instead.

src/components/VButton.vue
panchovm: You are right. I also noticed I forgot to add the content switcher and filter versions for mobile. I would like to know more about your concern @sarayourfriend, since the unskippable variant I detect is the responsive one. Called Big for desktop elements and **Small** for mobile ones.

zackkrida:
I pushed a small addition to make the header sticky instead of fixed, and to show the border when the filter sidebar is opened in _addition_ to on scroll. 

This fix currently means the search bar on the search routes is overlapped by the header, but that searchbar will be removed soon anyway in favor of #491

test/unit/specs/components/VHeader/v-filter-button.spec.js
```diff
@@ -83,12 +83,9 @@ describe('VFilterButton', () => {
     mediaQueryListMock.matches = false
     appliedFilters = []
     const { container } = render(VFilterButton, options)
-
-    // Text is only visible for screen readers.
-    const buttonTextLabel = screen.getByText('Filters')
+    // Screen debug shows that there is aria-label, but findByLabelText doesn't
+    // find the button.
```
sarayourfriend, line 87: When I tried this locally it appeared to work, but I used `getByLabelText` instead of `find*`. The `find*` methods return a Promise... maybe that was the issue?

Specifically it worked when I did `screen.getByLabelText(/filters/i)`

sarayourfriend (commented):
I'm still seeing the pop-in on initial load for the visible "filters" label on the button 🙁 Otherwise this is looking good!

I tried out one of the comments you left in a test and it seemed to work, but maybe I was doing something different than you intended.

panchovm:
It looks great and beautiful. Just one comment that I mentioned to @zackkrida and copying it here for visibility.

On desktop, the width size changes and it feels odd, especially when pushing the content switcher on the left. It might be a good idea to have it fixed and the number center-aligned inside.

filter  width change

The Figma component is a good reference for my point.

Figma component screenshot

obulat:
A fixed width for the whole button would cause a problem with other languages, because the length of the label may be really different for different languages.

krysal (commented):
Noticing two things:
- I'm seeing a gray background for the filter icon on mobile, wherein mockups it has no background
- The number of applied filters is shown on the desktop view if any is applied. The issue mentions that the button does not change on desktop.

Are these expected behavior?

zackkrida:
@krysal you may need to git pull to resolve the first issue, the second is the correct behavior. I'll update the issue!

zackkrida:
@panchovm I agree about the number positioning, but it actually isn't possible yet and adds a great deal of complexity here, because doing so requires us to wrap the number in its own html element which breaks our pluralization rules.

It would still be really complicated, but there's a feature in the latest version of vue-i18n, which Nuxt's u18n plugin does not support yet, which would allow us to do this:

﻿https://vue-i18n.intlify.dev/guide/advanced/component.html#pluralization-usage

panchovm (approved):
It looks wonderful ✨

src/composables/use-filter-sidebar-visibility.js
```diff
@@ -0,0 +1,28 @@
...
+  const setFilterSidebarVisibility = (val) => {
+    isFilterSidebarVisible.value = val
+    local.set(process.env.filterStorageKey, val)
+  }
+
+  onMounted(() => {
+    const localFilterState = () =>
+      local.get(process.env.filterStorageKey)
+        ? local.get(process.env.filterStorageKey) === 'true'
+        : false
```
sarayourfriend, line 20: This can be simplified:

```suggestion
      local.get(process.env.filterStorageKey) === 'true'
```

src/composables/use-media-query.js
sarayourfriend: This seems to have gone back to the original state.

src/components/VHeader/VFilterButton.vue
```diff
@@ -0,0 +1,129 @@
...
+    },
+    pressed: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  setup(props, { emit }) {
+    const { i18n, store } = useContext()
+    const { isHeaderScrolled, pressed } = toRefs(props)
+    const isMinScreenMD = isScreen('md')
```
sarayourfriend, line 49: It would be nice to rename this composable in the future to `isMinScreen`.

test/unit/specs/components/VHeader/v-filter-button.spec.js
```diff
@@ -0,0 +1,140 @@
...
+      const icon = container.querySelector('svg')
+      const button = screen.getByText(`${filterCount} Filters`)
+
+      expect(icon).not.toBeInTheDocument()
+      expect(button).toBeVisible()
+    })
+    it('only shows the count when filters are applied and the user scrolls', () => {
+      isScreen.mockImplementation(() => ({ value: false }))
+      // +2 to guarantee it's plural
+      const filterCount = Math.floor(Math.random() * 10) + 2
```
sarayourfriend, line 128: Curious what you think of this, but in my experience introducing randomness into tests that only run once per suite (as opposed to something like property based testing which would more rigorously test the input/output based on random and vast examples) run can introduce difficult to debug flakiness.

By virtue of that we tend to expect the randomness to cover all the cases we'd have to manually test otherwise, but using `it.each` we can easily repeat tests for different cases and test specific edge cases.

For example, I'd probably recommend testing `[0, 1, 2, 3, 4]` to cover pluralization rules in most languages (though admitedly not all). Right now we're not testing 0 and 1 in the scrolled state at all, which are distinct branches.

sarayourfriend (approved):
LGTM. I won't block this PR because I know you intend to continue iterating on it once it is merged. There's a couple of things I found that concerned me. Primarily the pop-in on the SSR search page.

There's also a regression with the focus outline for the active state of one of the variants that doesn't appear to be an issue in storybook on main:

Captura de Tela 2021-12-16 às 07 48 55

Note the blue outline instead of the pink outline.

src/components/VHeader/README.md
```diff
@@ -1,11 +1,17 @@
...
+
+- TBD
+
 ### VFilterButton
+
+This button toggles the filter drawer on desktop and opens the mobile menu set to the filters pane on mobile.
+
+- When the header is scrolled, remove the button's border
+- When there's applied filters, replace the filter icon with the # of active filters
+- On scroll, hide the word label (the word 'Filters') to save space
```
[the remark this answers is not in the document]
krysal: The word is hidden on scroll for me, and showed again on top, but I agree is a bit odd to have only the number, maybe it is worth leaving the icon also?

test/unit/specs/components/VHeader/v-filter-button.spec.js
zackkrida: That makes sense. I mainly wanted to protect against some specific test where only the hardcoded number of filters worked, but the `it.each` approach sounds lovely. I'm going to add those tests for pluralization before merging 👍

src/components/VHeader/README.md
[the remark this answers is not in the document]
zackkrida: I agree this could be a concern, it's only informed by limited horizontal space to fit everything into a single header on mobile. The search bar will be in the middle.

The word 'Filters' is _only_ removed after the user scrolls, so they'll at least see it initially if they're paying attention. Adjusting things so the icon fits in mobile after scrolling could be a really good compromise, but I'll leave figuring this out to @panchovm

src/components/VButton.vue
```diff
@@ -249,6 +251,9 @@ a.button {
 }
 
 .action-menu {
+  @apply bg-white text-black border border-dark-charcoal-20 hover:border-dark-charcoal-20 focus-visible:ring-pink;
+}
+.action-menu-secondary {
```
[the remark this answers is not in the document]
zackkrida: Fixed!

src/components/VHeader/VHeader.vue
```diff
@@ -1,20 +1,21 @@
...
-      variant="action-menu"
-      class="self-center"
-      @click="closeOverlay"
-    >
-      <span class="text-sr">{{ $t('modal.close') }}</span>
-      <VIcon :icon-path="closeIcon" />
-    </VButton>
+
+    <VFilterButton
+      v-if="isSearch"
```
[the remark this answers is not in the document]
zackkrida: @sarayourfriend Are you certain this is the case? When I SSR load the page I see the filter icon, which indicates the button is rendering right away, but then the label pops in later, because the `isMinScreen` media query is false server-side. Thankfully, I added an option to the media query in #491 that allows to choose the default SSR behavior, so basically we can choose if pop-in occurs on mobile or desktop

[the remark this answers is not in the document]
sarayourfriend: Yup, just double checked and took a video. It's subtle and short but there's definitely pop in. I also noticed another thing in this video is that the filter icon appears instead of the number on the initial render even when a filter is applied

https://user-images.githubusercontent.com/24264157/146492122-de1cb690-19a8-4cc9-ae48-b71b2ef7229b.mov

src/components/VHeader/README.md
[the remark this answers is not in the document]
panchovm: Agree there is a risk in removing words and relying solely on a numeric symbol, but for this release, I think it's a good solution for the navigation on mobile. Yet, I would follow the behavioral statistic to assess its success after a few months.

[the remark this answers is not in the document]
sarayourfriend: Another thing to keep in mind about the mobile version is that there will also be the WordPress header hamburger menu present on the page at the same time. I actually already made the error of confusing the two 😭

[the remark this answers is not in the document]
panchovm: Yes 💔 . We are aware of the double header navigation risk and, hopefully, we can work on a new WordPress header this year. Although, the scroll just hides the WP header whereas Openverse header remains fixed.

Issue #462, opened 2021-11-30
Header


zackkrida:
## Description

The header...it's pretty important! The first thing at the top of the page 😅 .

### Features

- Sticky (persists at the top of the page after the user scrolls)
- Has a bottom border when the filter sidebar is active _and_ after the user scrolls down the page

### Props

| Name | Type | Description |
|-|-|-|

### Slots

| Name | Slot props | Description |
|-|-|-|

### State

| Name | Type | Description |
|-|-|-|

### Events

| Name | Associated data | Description |
|-|-|-|

## Children

- [x] #448 
- [x] #375 
- [x] #453
- [x] #476
- [x] #485 

## References

Desktop: https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=751%3A2802
Mobile: https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=751%3A2808

## Implementation

- [x] 🙋 I would be interested in implementing this component.

zackkrida:
Completed via #606

PR #491, opened 2021-12-06
Add searchbar to header

obulat:
From @zackkrida 😄 

## Todo

- [x] Persist search term from homepage to search pages

## Fixes

Related to #462 by @zackkrida 

## Description

This PR makes the SearchBar in Header functional. Also does the following:

- Adjusts logo alignment in header
- Implements proper header mobile layout
- Adds an option to the use media query hook to set if the media query should pass or fail in SSR
- Prevents 1px header layout shift by adding a white bottom border by default
- Switches the header to a `<header>` tag
- Removes started mobile menu triggering code, to be handled later

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VHeader/VSearchBar/VSearchBar.vue
```diff
@@ -53,7 +53,11 @@ export default {
         return props.value
       },
       set(value) {
-        emit('input', value)
+        // For some reason, the `input` event is fired twice,
+        // with InputEvent as value the second time.
+        if (typeof value === 'string') {
+          emit('input', value)
+        }
```
zackkrida, line 60: Seems like a bug we should fix in the component, rather than apply this check.

src/components/VHeader/VHeader.vue
```diff
@@ -25,33 +63,129 @@ import {
...
   },
   setup() {
-    const { store } = useContext()
+    const { app, i18n, store } = useContext()
     const { isSearch } = useSearchRoute()
     const { isHeaderScrolled } = useWindowScroll()
-    const isMdScreen = isScreen('md')
+    const router = useRouter()
+    // Screen is at least the MD breakpoint wide
+    const isMinScreenMD = isScreen('md', { shouldPassInSSR: true })
```
zackkrida, line 112: `shouldPassInSSR` basically makes this media query desktop-first

src/components/VHeader/VHeader.vue
```diff
@@ -25,33 +63,129 @@ import {
...
+    /** @type {import('@nuxtjs/composition-api').ComputedRef<number>} */
+    const resultsCount = computed(() => store.getters['media/results'].count)
+
+    /**  @type {import('@nuxtjs/composition-api').ComputedRef<boolean>} */
+    const isFetching = computed(() => {
+      return store.getters['media/fetchState'].isFetching
+    })
+
+    /**
+     * Status is blank on mobile screen.
```
zackkrida, line 138:
```suggestion
     * Status is hidden below the medium breakpoint.
```

zackkrida:
@WordPress/openverse-frontend if could figure out how to make the homepage search term carry over to the search endpoints and populate the searchbar, it would be a huge help!

sarayourfriend (commented):
Looks real good! I'm really liking this header design, kudos to @panchovm!

Noticing a couple of oddities here and there that would be good to address:

The tailing edge of the search button is missing in RTL languages

Captura de Tela 2021-12-17 às 05 47 33

In LTR languages, the leading edge of the focus outline is missing (might be solvable with a z-index):

Captura de Tela 2021-12-17 às 04 42 15

On the home page, the header spans the entire width of the screen and has no bottom border, which makes it so that if you have to scroll down the page, the header will cover UI elements. A bottom border would help this. Alternatively, on the homepage we could just "float" the Openverse logo and button in the corner and not have the header span the full width of the screen. cc @panchovm on that one.

Captura de Tela 2021-12-17 às 05 49 22

This one is a nit pick, but on mobile, the search bar isn't quite centered between the logo and the filters button:

Captura de Tela 2021-12-17 às 05 50 19

Potentially related is that the filters button is not square:

Captura de Tela 2021-12-16 às 07 49 54

Scroll to top could use some love, if not now then in the future, as there is a very perceptible delay between when the page has visually finished scrolling to the top and when the header decides to "expand" back into the two-row version:

https://user-images.githubusercontent.com/24264157/146555015-ca164116-499c-4674-abd1-427f6551409f.mov

On search result pages the search bar does not span the full available width of the header and in fact will grow when scrolling (but doesn't ever grow to match the full available width in the header):

https://user-images.githubusercontent.com/24264157/146555125-6e786114-0f2a-489a-9c55-327d3c64fb4a.mov

I did some initial keyboard usability testing and all seems good to me so far!

panchovm:
> On the homepage we could just "float" the Openverse logo and button in the corner and not have the header span the full width of the screen

That's a great idea @sarayourfriend I will keep it in mind when designing the home page. And thanks for the kudos ✨

obulat:
> @WordPress/openverse-frontend if could figure out how to make the homepage search term carry over to the search endpoints and populate the searchbar, it would be a huge help!

I've added a watcher for the search term in the search store, so now the search term does carry over from the homepage.

obulat:
> The tailing edge of the search button is missing in RTL languages. In LTR languages, the leading edge of the focus outline is missing (might be solvable with a z-index).

I couldn't fix this problem using z-index, but [changing the border styling](https://github.com/WordPress/openverse-frontend/pull/491/commits/3bd9875ef8cd231ceac9654653fbd9f222ccac0e) using CSS worked. 

> Scroll to top could use some love, if not now then in the future, as there is a very perceptible delay between when the page has visually finished scrolling to the top and when the header decides to "expand" back into the two-row version:

I agree, but it's probably better to work on scroll to top when the content switcher is ready, as it will affect the placement of components in the header.

zackkrida (approved):
The fixes look great! Glad to see the search term persisted too.

krysal (commented):
It's looking good to me! Both for LTR and RTL modes and the shift in layout on scroll works well on the search results page, but on the image detail page it is not changing, then the full logo persists and causes the page to have a horizontal scroll on mobile.

CleanShot 2022-01-04 at 12 57 44@2x

Also, I'm getting some warnings on the console about _the client-side rendered virtual DOM tree not matching server-rendered content..._ when a page is rendered server-side, it might need to be addressed before these errors pill up.

zackkrida:
@krysal I'm not able to replicate the rtl/ltr issue, could you test that again?

> the client-side rendered virtual DOM tree not matching server-rendered content...

Olga has found a fix for this, it's related to conditionally rendering things via media queries and the media query failing in SSR but passing client-side during the hydration process. We can and will fix this later.

src/components/VHeader/VHeader.vue
```diff
@@ -1,22 +1,64 @@
...
+      class="rounded-sm ring-offset-1 focus:outline-none focus-visible:ring focus-visible:ring-pink -ms-2 inline-flex items-center hover:bg-yellow"
+      :class="{
+        'pe-3': !isHeaderScrolled || !isSearchRoute,
+        'md:px-0': isSearchRoute,
+      }"
+    >
       <VLogoLoader :status="isFetching ? 'loading' : 'idle'" />
+      <svg
+        v-if="!isHeaderScrolled || !isSearchRoute"
+        class="-ml-1 mt-1"
```
krysal, line 22: This is what is causing the text still appear, I think the second part can be ommited.
```suggestion
      <svg
        v-if="!isHeaderScrolled"
        class="ml-1 mt-1"
```

src/components/VHeader/VHeader.vue
```diff
@@ -1,22 +1,64 @@
...
       'border-b border-dark-charcoal-20':
-        isHeaderScrolled || isFilterSidebarVisible,
+        isSearchRoute && (isHeaderScrolled || isFilterSidebarVisible),
+      'flex-wrap gap-y-4': !isMinScreenMD && !isHeaderScrolled,
     }"
   >
-    <NuxtLink to="/">
+    <NuxtLink
+      to="/"
+      class="rounded-sm ring-offset-1 focus:outline-none focus-visible:ring focus-visible:ring-pink -ms-2 inline-flex items-center hover:bg-yellow"
```
krysal, line 13:
```suggestion
      class="rounded-sm ring-offset-1 focus:outline-none focus-visible:ring focus-visible:ring-pink ms-2 inline-flex items-center hover:bg-yellow"
```

src/components/VHeader/VHeader.vue
```diff
@@ -1,22 +1,64 @@
...
+        />
```
krysal, line 33: Perhaps we can extract this path to an asset like the `VCheckbox`'s [checkmark](https://github.com/WordPress/openverse-frontend/blob/main/src/components/VCheckbox.vue#L23)? That looked neat!

src/components/VHeader/VHeader.vue
```diff
@@ -1,22 +1,64 @@
...
     </NuxtLink>
 
+    <VSearchBar
+      v-show="!isHomeRoute"
+      v-model.trim="searchTerm"
+      class="mx-auto md:ms-0 md:me-auto lg:w-1/2 2xl:w-1/3"
+      :class="{
+        'order-4 w-full md:order-none md:w-auto': !isHeaderScrolled,
+        'w-2/3': isHeaderScrolled,
+      }"
```
krysal, line 44: nitpick: The array syntax fit better here.
```suggestion
      :class="[
        isHeaderScrolled ? 'w-2/3' : 'order-4 w-full md:order-none md:w-auto',
      ]"
```

krysal (changes requested):
Sorry I didn't explain myself well. What I wanted to say is that the **logo _with text_** persists when scrolling on mobile in non-search views, when the expected behavior is that the part of the text should be hidden and only show the icon in those cases. Here is a short video showing it.

https://user-images.githubusercontent.com/9145885/148299247-bae50557-12cd-4ff6-a4ba-b9026c57bad9.mp4

The good thing is I think I identified the cause and seems easy to fix, that is the only blocking thing here. The rest of my comments are just about typos/minor suggestions. After I made the adjustments the horizontal scrollbar persist so the cause it's something else that does not concern this PR.

src/components/VHeader/VHeader.vue
```diff
@@ -1,22 +1,64 @@
 <template>
-  <div
-    class="sticky top-0 flex py-4 px-6 md:px-7 align-center justify-between z-40 w-full bg-white"
+  <header
+    class="sticky top-0 flex py-4 px-4 md:px-7 items -center justify-between z-40 w-full bg-white gap-x-2"
```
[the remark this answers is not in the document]
zackkrida: Nice catch!

src/components/VHeader/VHeader.vue
zackkrida: The `-ms-2` is actually intentional, it is a negative margin-start value.

src/components/VHeader/VHeader.vue
zackkrida: Done! Great idea.

zackkrida:
Thanks for all the suggestions @krysal and fixing that mobile bug! Should be resolved.

krysal (approved):
Nice, confirmed the logo text is hidden on scroll in non-search pages.

PR #497, opened 2021-12-07
Add a transform step after receiving data from the API

dhruvkb:
## Description

This PR makes some improvements to the media stores and services to improve the general code quality.

- changes the results from an array to an object so that referencing results by ID becomes easier
- changes the decoding to be an additional transformation step in the media service
- cleans up the code for merging new results with the old ones

## Testing Instructions

Searching and fetching more results should work as previously.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/AudioResultsList.vue
+++ b/src/components/AudioResultsList.vue
@@ -7,8 +7,8 @@
     </div>
     <div class="px-6 pt-6">
       <AudioTrack
-        v-for="audio in results.items"
-        :key="audio.id"
+        v-for="(audio, index) in results.items"
+        :key="index"
         :audio="audio"
         :size="audioTrackSize"
         layout="row"
@@ -56,7 +56,7 @@ import { MEDIA, SEARCH } from '~/constants/store-modules'
 export default {
   name: 'AudioResultsList',
   async fetch() {
-    if (!this.results.items.length) {
+    if (!Object.keys(this.results.items).length) {
       await this.fetchMedia({
         ...this.query,
         mediaType: AUDIO,
--- a/src/data/audio-service.js
+++ b/src/data/audio-service.js
@@ -13,9 +13,11 @@ const AudioService = {
    * @return {Promise<{data: any}>}
    */
   search(params) {
-    return config.dev
-      ? Promise.resolve({ data: sampleAudioResponses.search })
-      : ApiService.query('audios', params)
+    if (config.dev) {
+      const data = JSON.parse(JSON.stringify(sampleAudioResponses.search))
+      return Promise.resolve({ data })
+    }
+    return ApiService.query('audios', params)
   },
 
   transformResults(data) {
--- a/src/data/audio-service.js
+++ b/src/data/audio-service.js
@@ -1,12 +1,15 @@
 import config from '../../nuxt.config.js'
 import sampleAudioResponses from './sample-audio-responses.json'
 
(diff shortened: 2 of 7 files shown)


src/data/base-media-service.js
```diff
@@ -0,0 +1,15 @@
+import decodeMediaData from '~/utils/decode-media-data.js'
+
+const BaseMediaService = (mediaType) => ({
+  transformResults(data) {
+    data.results = Object.fromEntries(
+      data.results.map((item) => {
+        item = decodeMediaData(item, mediaType)
+        return [item.id, item]
+      })
+    )
```
sarayourfriend, line 10: Couldn't this be done in a single pass using `reduce`? I'm not sure how big `data.results` can get but it could be significant looping `n`, allocating `n` arrays then throwing them all out to do another `n` loop.

```suggestion
    data.results = data.results.reduce((acc, item) => {
      acc[item.id] = decodeMediaData(item, mediaType)
      return acc
    }, {})
```

src/data/audio-service.js
```diff
@@ -1,18 +1,26 @@
...
    * Search for audios by keyword.
    * @param {Object} params
    * @return {Promise<{data: any}>}
    */
   search(params) {
-    return config.dev
-      ? Promise.resolve({ data: sampleAudioResponses.search })
-      : ApiService.query('audios', params)
+    if (config.dev) {
+      const data = JSON.parse(JSON.stringify(sampleAudioResponses.search))
```
sarayourfriend, line 20: Just to make sure, we won't need the `stringify -> parse` stuff once we're using real results right?

sarayourfriend (changes requested):
Something strange is happening with this PR, I haven't pinpointed the cause yet, but it seems that the _first_ search always results in a "no results" page, but subsequent searches (even for the exact same search term) will return the expected results.

I verified this doesn't happen on `main` so I believe it is a regression specific to this branch.

src/data/audio-service.js
dhruvkb: Yes, it's just as long as we're using a dummy JSON file as the result response.

src/data/base-media-service.js
dhruvkb: That's a great idea!

dhruvkb:
Checking the "no-results" bug from Sara's comment...

dhruvkb:
@sarayourfriend thanks for finding the regression! I believe 54a8efd should fix it.

sarayourfriend (approved):
Works great now! 🚀 🤠 🎉

zackkrida (approved):
Great idea and approach!

src/components/AudioResultsList.vue
```diff
@@ -7,8 +7,8 @@
     </div>
     <div class="px-6 pt-6">
       <AudioTrack
-        v-for="audio in results.items"
-        :key="audio.id"
+        v-for="(audio, index) in results.items"
+        :key="index"
```
zackkrida, line 11: What is this change for?

dhruvkb: This is just to make the code for audio same as that for images. The output is the same because the `index` is the ID itself.

PR #503, opened 2021-12-08
Add progressive TypeScript compilation

### LINKED ISSUE #175: Explore adding progressive typechecking where possible
## Problem
We currently use JSDoc comments to type hint our files, but we're not actually running the TypeScript compiler on the files that are being annotated. We could run `tsc` in some places to help us verify the code we're writing, even while still just using JSDoc for the type annotations.

## Description
Let's explore adding a `tsconfig.json` file to replace the `jsconfig.json`. This should use a similar configuration to potentially some of the existing WordPress projects (like Gutenberg) except that we don't need a composite style project here as we only have a single project to type check.

We can copy the existing configuration from Gutenberg for example and apply it to the API service files and see what it would take to get the types up to scratch for the compiler to pass on those files. It might take a lot of work to actually be able to enable type checking as a PR check for example.

## Alternatives
We could just not do this, but then we don't reap the benefits of our type hints fully.

## Implementation

sarayourfriend:
## Fixes

Fixes #175 by @sarayourfriend

## Description

Replaces `jsconfig.json` with a `tsconfig.json` that mirrors Gutenberg's with some changes to accomodate not being a composite project.

Currently this only checks a single file, basically in an effort to minimize the size of this PR while also still proving that it does work as intended. I had to make a small update to the type declarations in this file to get the compiler to pass. Hooray for type feedback! 🎉 

Future work will include slowly adding each file in `src/composables` to the `include` array. We can also start running the type checker on other parts of the codebase like the services 🙂 I would propose doing this in small isolated bits and pieces to avoid having to make too many changes in a single PR to address the inevitable error here and there.

## Testing Instructions

Checkout the PR and run `npm run types`. Everything should come back without an error. Next, update `use-event-listener-outside.js` to include a type error. For example, add the following somewhere in the file:

```js
/** @type {string} */
const foo = null
```

This is a type error becuase you can't assign `null` to a `string`. Your IDE should pick this up and give you some squiggly lines or a warning (however your IDE normally picks up compilation errors). You should now also be able to run `npm run tsc` and have it report back the error to you.

Go ahead and try to commit that (you'll have to disable the eslint `no-unused-vars` rule temporarily to be able to run the pre-commit) and pre-commit should fail on the `tsc` step.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (approved):
Nice! Adding a type error produced the expected result.

zackkrida:
`npm run format:test` is warning on the tsconfig.json file, however...

obulat (approved):
Great start! I tried adding `use-related` to the `include` array - and Oh wow, it's so much work :)

dhruvkb (approved):
This is a major boost for DX!

krysal (approved):
This extra help with types is excellent! ✨

PR #514, opened 2021-12-13
Fix audio RTL visual bugs

### LINKED ISSUE #421: RTL touchups
## Description
While the site still functions under RTL language direction, it looks extremely unpolished and will give a bad impression (in addition to the frustrating partial translations). What follows is a list of the individual places that need to be touched up to get the site "presentable" for RTL languages.

The following descriptions are all what it should be for RTL languages. For each fix, ensure that it does not cause a regression to the LTR languages.

- [x] #426
    - The navigation items should be aligned with the left side of the screen
    - The dropdown items should not have a large leading gap
    - Should the openverse logo be on the opposite side of the openverse brand text? cc @panchovm
- [x] #424
    - The search input and button should have the correct corners rounded
    - The "I want something I can..." checkboxes should have their spacing fixed to match the LTR language spacing
- [x] #431
    - The table has a double border on the left side
    - The table is missing the border between the Source and Domain column
    - The spacing of the "Suggest a new Source" button with the text and the icon is a little off compared to the LTR version (the left side of the button has too much padding)
    - Trailing slashes for all domains listed should be removed as they get automatically put at the front of the domain in RTL languages (`https://asia.si.edu/` gets visually transformed by the browser into `/https://asia.si.edu`)
- [x] #428
    - The search bar needs to have its button and input corners correctly rounded as well as its padding adjusted for RTL
    - The "safer browsing" flag icon needs to have its padding adjusted to match the LTR language padding
    - The scroll to top button should be moved to the opposite corner
    - License icons and their corresponding text need to have their separation padding adjusted for RTL
- [x] #435
    - "Go to image's website" button needs to have its icon padding adjusted
    - The attribution box and tabs need to have their borders adjusted to be on the correct side for RTL languages
    - The plain text attribution also needs to have its quotation marks corrected as currently they're being moved around by the browser into strange places
    - The license icon needs to have its padding adjusted against the associated text
    - The information table needs to have its padding adjusted as the "dimensions" text overlaps the listed dimensions
    - The leading edge of the information/reuse table needs to be fixed for RTL languages to align with the rest of the text in the result information
- [x] #512
    - The audio tags need to be adjusted for RTL languages, their borders are all over the place
- [x] #513
  -   The audio result license icon spacing compared to the track length needs to be adjusted, see [screenshots and description in the comments](https://github.com/WordPress/openverse-frontend/issues/428#issuecomment-974044511).

## Reproduction
The easiest way to reproduce is to set your language to an RTL language (like Arabic). I'm not sure what the best way to do this is, but the way I did it was to edit to `i18n_redirected` cookie value to be `ar`. I don't know if this cookie is always present, it might exist only because I tested the WordPress.org Openverse theme translations at one point. If that is the case, then follow [the instructions here](https://github.com/WordPress/wordpress.org/blob/master/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md) to set up the theme on a local WordPress install and set your language via wp-admin. Alternatively, manually edit the `dir` attribute on each page to be `rtl`. The attribute will be reset to `ltr` on each page load, so this is a very tedious option for testing as you'll have to edit the attribute every time you reload the page or navigate to a new one.

Because we will inevitably have partial translations down the line as well (for example when new strings are introduced or existing strings are updated), the site should look passable as RTL even with English strings, so for now the partial translations work in our favor as we fix these issues.

## Expectation

After fixing the issues above, the site should look polished in both LTR and RTL language directions.

For most of these the best solution might be to convert the underlying markup to use Tailwind CSS with the RTL ready start/end based styles.

## Environment

 - Device: 
 - OS: 
 - Browser: 
 - Version: 
 - Other info: 

## Resolution

### LINKED ISSUE #512: Single result page for Audio RTL touchups
The audio tags need to be adjusted for RTL languages, their borders are all over the place:

Captura de Tela 2021-12-13 às 09 51 03

### LINKED ISSUE #513: Audio search result RTL issues
The audio result license icon spacing compared to the track length needs to be adjusted:

Captura de Tela 2021-12-13 às 09 51 52

sarayourfriend:
## Fixes

Fixes #513 by @sarayourfriend
Fixes #512 by @sarayourfriend

Closes #421 as this will address the last two problems in that issue.

## Description

Fixes the RTL visual bugs outlined in the issues.

## Testing Instructions

Checkout this branch and run it with `npm run dev`. Then go to http://localhost:8443/audio/8071128a-bebc-4c24-823d-6e26a2e6dfe2 and verify that the two elements outlined in the issue look good in both RTL and LTR mode. Follow the instructions in [this issue](https://github.com/WordPress/openverse-frontend/issues/421) under the Reproduction headline for how to quickly test RTL and LTR locally.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/AudioTrack/layouts/Row.vue
+++ b/src/components/AudioTrack/layouts/Row.vue
@@ -50,7 +50,7 @@
             </span>
           </div>
 
-          <div class="part-b inline-flex space-x-1">
+          <div class="part-b inline-flex gap-1">
             <div v-if="isMedium">
               {{ timeFmt(audio.duration) }} {{ $t('interpunct') }}
               <template v-if="audio.category">
--- a/src/components/AudioDetails/Tags.vue
+++ b/src/components/AudioDetails/Tags.vue
@@ -7,7 +7,7 @@
       <MediaTag
         v-for="(tag, index) in getValidTags()"
         :key="index"
-        class="me-4"
+        class="me-4 inline"
         @click="searchByTagName"
       >
         {{ tag.name }}
--- a/src/components/MediaTag/MediaTag.vue
+++ b/src/components/MediaTag/MediaTag.vue
@@ -2,7 +2,7 @@
   <Component
     :is="tag"
     v-bind="$attrs"
-    class="media-tag text-dark-charcoal-60 text-sm font-semibold py-1 px-2 border border-dark-charcoal-60 rounded-full"
+    class="text-dark-charcoal-60 text-sm font-semibold py-1 px-2 border border-dark-charcoal-60 rounded-full"
   >
     <!-- @slot Content goes here -->
     <slot />
@@ -23,7 +23,7 @@ export default {
      */
     tag: {
       type: String,
-      default: 'span',
+      default: 'div',
     },
   },
 }
--- a/test/unit/specs/components/MediaTag/media-tag.spec.js
+++ b/test/unit/specs/components/MediaTag/media-tag.spec.js
@@ -11,10 +11,10 @@ describe('MediaTag', () => {
     options = { propsData: props }
   })
 
-  it('should render an span tag by default', () => {
+  it('should render an div tag by default', () => {
     const wrapper = render(MediaTag, options)
-    expect(wrapper.vm.$props.tag).toEqual('span')
-    expect(wrapper.vm.$el).toBeInstanceOf(HTMLSpanElement)
+    expect(wrapper.vm.$props.tag).toEqual('div')
+    expect(wrapper.vm.$el).toBeInstanceOf(HTMLDivElement)
   })
 
   it('should render the supplied tag', () => {


src/components/AudioTrack/layouts/Row.vue
```diff
@@ -50,8 +50,8 @@
             </span>
           </div>
 
-          <div class="part-b inline-flex space-x-1">
-            <div v-if="isMedium">
+          <div class="part-b inline-flex">
+            <div v-if="isMedium" class="pe-1">
```
dhruvkb, line 54: `space-x-` classes apply to all children as in
```css
.space-x-1 > * + * {
  margin-left: 0.25rem;
}
```
Replacing it with `pe-` classes will mean adding it to all children except the first. Alternatively we could use `gap-` classes on the flex itself?

sarayourfriend: Leaving padding on the last element doesn't seem to introduce any visual problems that I could see. While not being a 1:1 replacement for `space-x-*` I think it's appropriate in this case? I guess the tailwindcss-rtl plugin doesn't cover the `space-<dir>-<unit>`.

`gap` seems like a better solution though. I'll update to that.

sarayourfriend: Fixed!

zackkrida (approved):
LGTM!

dhruvkb (approved):

Issue #483, opened 2021-12-03
Storybook issues


krysal:
## Description

The Storybook is not showing the stories written on `.mdx` files and the ability to change props with _knobs_ is not visible. I'm not sure with what changes this started to happen.

## Reproduction

1. Just run `npm run storybook`
2. Go to the open browser windows obverse there is not showing the index page
3. Search for the `VLicense` story for example and see it's not found neither

## Expectation

We want to show these stories too, using this format is especially important because it enables us to add documentation about the components.

## Screenshots

![CleanShot 2021-12-03 at 16 50 49@2x](https://user-images.githubusercontent.com/9145885/144671116-0b8af816-70e7-4d4b-9279-bb7c695f6fa1.png)

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

obulat:
I wasn't able to fix this yet, unfortunately. All I could find that at the moment before the docker-improvements branch was merged) the mdx files were built fine: https://github.com/WordPress/openverse-frontend/tree/7042449170caec9dda9716cc3236c4261229721a . And the issue is that `@storybook/addon-essentials` cannot be resolved.
A very crazy workaround I've been using is to checkout the code before the `docker_improvements` branch, remove node_modules, run `npm install`, then checkout my branch (header) and run storybook :)

dhruvkb:
An alternative workaround (discovered by comparing the `node_modules/` directory structure before and after commit 7042449) is by moving Storybook dependencies from the Nuxt Storybook’s internal node-modules to the upper level.

```bash
$ mv node_modules/@nuxtjs/storybook/node_modules/@storybook/addon-essentials node_modules/@storybook
$ mv node_modules/@nuxtjs/storybook/node_modules/@storybook/addon-docs node_modules/@storybook
```

I still have no idea why the folder structure changed but it seems to be some dependency conflict which causes these two packages to into a subfolder. The rest don’t conflict and they are placed normally. Since moving these folders to their expected location fixes the issue, whatever the conflict is must be largely superficial.

zackkrida:
I haven't looked but I wonder if it's possible there was a nuxt storybook minor upgrade that actually had a breaking change. I know the package lock was rebuilt when the docker branch was merged

dhruvkb:
What I'm going to do (in my next debugging-friendly time slot) is remove our packages one by one till it starts working and then add them back one by one.

sarayourfriend:
[...]

PR #520, opened 2021-12-13
Add Storybook smoketest to GitHub CI

### LINKED ISSUE #518: Add Storybook compilation to CI to ensure that no PR breaks Storybook
## Problem

Currently it is possible to accidentally merge a change that breaks Storybook in unexpected ways. For example:

https://github.com/WordPress/openverse-frontend/issues/516
https://github.com/WordPress/openverse-frontend/issues/483

We can add Storybook compilation to our CI pipelines to make sure that this doesn't happen.

## Description

We can pass `--ci --smoke-test` to the storybook command to verify that it is working.

Additionally we could do some kind of snapshot testing of the story list to ensure they're all present? That seems like a good _next_ step but first we should just do the `ci smoke-test` route to ensure baseline integrity.

## Implementation

sarayourfriend:
## Fixes

Fixes #518 by @sarayourfriend

## Description

Adds a step to our CI to run `storybook --ci --smoke-test` to ensure PRs don't break Storybook.

Note: this PR will be broken until https://github.com/WordPress/openverse-frontend/pull/517 is merged and a fix for https://github.com/WordPress/openverse-frontend/issues/483 is found. Obviously both of those bugs are rendering our current storybook broken, so adding it to CI while it's broken will well... be broken 😂 

## Testing Instructions

Verify that the CI is failing on Storybook compilation issues currently. In the future of this PR you'll have to look at past runs before this PR inevitably gets rebased when the referenced bugs are fixed. Of course after those bugs are fixed the CI should pass once this PR is rebased.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (approved):

dhruvkb (approved):
LGTM but I agree on keeping this blocked till we can first get it to pass 😄.

obulat (approved):
Nice to be able not to break things :)

zackkrida:
@sarayourfriend while you're out I've rebased this with main to get the license filename issue fixed, I'll merge as well if CI passes.

zackkrida:
Boo, that got us closer, but gave a new error: 

https://github.com/WordPress/openverse-frontend/runs/4537426501?check_suite_focus=true#step:9:195

zackkrida:
@sarayourfriend I _think_ this works locally if I do `pnpm exec nuxt-storybook --smoke-test --ci`, but locally `pnpm storybook -- --ci --smoke-test` gives the same error as here:

https://github.com/WordPress/openverse-frontend/runs/4537426501?check_suite_focus=true#step:9:195

I'm not confident what the fundamental difference is there.

sarayourfriend:
@zackkrida interesting... I still get a non-zero exit code when I use that alternative command. My understanding is that it shouldn't be different from `pnpm storybook -- --ci --smoke-test` but the output _is_ different for some reason 🤔

sarayourfriend:
Figured out how to get Nuxt component discovery to ignore the `.types.js` files, but still not sure how to get around the eslint rule for filename casing.

It appears that the rule doesn't actually allow for matching against the full path of the file, just the filename, which means that we can't, for example, exclude the `meta` folders completely from the rule.

We could get it to ignore it by adding a suffix to the file, like `targets.utils.js` or something like that, but I really loathe having to use arbitrary filenames like that just to get around a linter rule. I commented on [this issue](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/686#issuecomment-999633679) about the rule configuration. If the maintainers agree that it would be helpful I'll work on implementing an update to that rule.

In the meantime, I think it might just be best to do something like `targets.util.js` or whatever until it's possible to plop it into a folder named in a way that it can be broadly ignored or something. This will also come into play if we do something like create component-specific composables that don't make sense to put into the general `composables` folder.

krysal (approved):
The smoke test and the removal of warnings for `.types.js` are already nice improvements here ✨

PR #530, opened 2021-12-16
Column based image grid

### LINKED ISSUE #482: image result grid

zackkrida:
https://user-images.githubusercontent.com/6351754/146311902-0f6ff3d0-6a92-4177-a7ac-9a9097d87869.mp4

## Fixes

Fixes #482 by @sarayourfriend 

## Description

- Code is written to support image width and height to prevent layout shift, once the API supports it

### Todo

- [ ] Write tests
- [ ] Make related images single-col on mobile
- [ ] Cleanup related image padding
- [ ] Maybe min+max aspect ratio + object-fit to tighten up edges
- [ ] Implement skeleton loaders

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/ImageDetails/RelatedImages.vue
```diff
@@ -6,13 +6,14 @@
...
     </h3>
-    <ImageGrid
+    <VImageGrid
       :images="images"
       :can-load-more="false"
       :fetch-state="{
         isFetching: $fetchState.pending,
         fetchingError: $fetchState.error,
       }"
+      :columns="3"
```
zackkrida, line 16: Need to change this!

src/components/VImageGrid/VImageGrid.vue
```diff
@@ -0,0 +1,151 @@
...
+    const lastImageRefs = ref([])
+
+    // make sure to reset the refs before each update
+    onBeforeUpdate(() => {
+      lastImageRefs.value = []
+    })
+
+    onUpdated(() => {
+      console.log('last image refs!')
+      // console.log(lastImageRefs)
```
zackkrida, line 113: So these dynamic refs are meant to grab the last image of each column. It seems to work, I can see the ref value in Vue Devtools, but I'm not able to log them in `onUpdated` or `onMounted`. In this case, not even my plain string console log is firing at any point. I have no idea why! 😿

src/components/VImageGrid/VImageGrid.vue
```diff
@@ -0,0 +1,151 @@
...
+  grid-template-columns: minmax(0, 1fr);
+  row-gap: var(--gutter);
+}
+
+.image-grid .image-column img {
+  width: 100%;
+  height: auto;
+  background-color: lightgray;
+}
+</style>
```
zackkrida, line 151: @todo: Scope/modularize these styles

src/components/VLoadMore.vue
```diff
@@ -0,0 +1,59 @@
+<template>
```
zackkrida, line 1: I want to refactor the state here a bit, it's needlessly complicated.

panchovm:
It looks great. So smooth. Excellent work.

sarayourfriend (commented):
Couple of initial impressions:

1. It looks great!
2. The tab order is wrong, it's vertically instead of horizontally across the screen
3. Focusing an image result no longer shows the license information
4. VoiceOver now longer announces the image is a figure and doesn't try to read the figure caption

zackkrida:
@sarayourfriend I added some horizontal tab/focus management, it needs to be fixed for the _last_ element but you can still play with it. It feels rather jumpy with the jagged image rows; perhaps I should also do some aspect ratio work, it seems like it'd increasingly be beneficial.

zackkrida:
@WordPress/openverse-frontend this PR has fallen so out of sync with the codebase that I think it's unwise to try and cleanup the merge commits and would be better suited to a _new_ PR that uses some of this code. I also don't think it's best for me to take on this PR moving forward.

@panchovm do you think it's best to have someone else complete this work, or stick to to the row-based grid with some slight improvements? I know we talked a bit this week about both ideas, and I'm curious to hear your thoughts in light of the information in this post.

zackkrida:
I'm _astonished_ that this is 67 days old! I'm officially going to close it. There's definitely work here that would be _useful_ if we ever decide to pick this work back up, but this PR is likely too much trouble to merge.

sarayourfriend:
@panchovm For sighted users that use keyboard navigation, I think the column layout but horizontal tabbing will be confusing.

There's some interesting discussion about masonry layouts of the exact sort we're evaluating here: https://github.com/w3c/csswg-drafts/issues/5675

krysal:
@sarayourfriend for which users more exactly? If the only issue is the order of the elements it is really a problem in this case? I don't think that whether a result comes in second or eighth is really important when we're talking in magnitudes of thousands.

sarayourfriend:
Well we're talking about accessibility, which means that the numbers are going to inherently be on the low end of "statistical significance."

This comment in the issue describes some of the users for whom a masonry layout with weird, unexpected or jumpy tab order could be a bad experience for:

https://github.com/w3c/csswg-drafts/issues/5675#issuecomment-718097362

Copied here for posterity:


- Neuro-divergent (cognitive) may get lost on where to focus their attention next
- Low vision, no AT may lose where their focus indicator has moved to, or may need to read content from a block that isn't proximate to where they are looking
- Low vision, screen reader will be hearing one thing while trying to locate it on the page because it doesn't follow the visual flow of the page
- Keyboard only - expect a linear flow of tabstops. If the tabstops follow code order, not visual order, they may get lost trying to track things

I tend to agree with this.

Again, we're not talking comparatively huge numbers of people, but you rarely are when you're discussing accessibility concerns; and frankly that shouldn't be the focus.

krysal:
Just to clarify, I wasn't talking about a potentially _low_ number of users affected, I was referring to the order of the images on the search results page doesn't matter much, in my view. If images move one, two, or ten places, results are already quite random (aside the importance to the search term/popularity).

The same author of the mentioned issue said:

> If the individual items are standalone, like articles in a newspaper, then this isn't an issue ...

Displayed images are not related to each other on the search page other than by the search term (and filters if applied) so each is displayed independently.

But now I get what was the concern, that the problem more specifically is the disparity in the expected order vs what is announced by a screen reader/asistant software. Given the Mansory layout is not so common and more of a novelty I'd expect that to be the case for users not visually impaired too but if the experience is excessively frustrating for the rest, I will understand that we discard this path. Although the mentioned thread suggests that there is a way to do it right and the concerns are more for default values of the CSS standard.

sarayourfriend:
Oh sorry, I misunderstood your comment as well, my apologies.

Yes, the issue isn't with the result relevancy. I agree that we don't have to worry about that right now given that our results are not optimized very well for most queries.

> Although the mentioned thread suggests that there is a way to do it right and the concerns are more for default values of the CSS standard.

I don't really understand what that "good" implementation actually is though :disappointed:

I'm actually skeptical that there even is a solution that matches this requirement: 

> We need a method though that will ensure that users will be able to read things in their proper comprehension order and that tab order matches visual order. A thought here could be "item priority" which defaults to code order (or no priority), while other options allow you to specify any number of blocks to accept a value that then makes sure the tab order changes to match the visual.

Considering the problem when you have a column that is packed very tightly and navigation sideways could go up more than one image, it wouldn't fall under that description.

> But now I get what was the concern, that the problem more specifically is the disparity in the expected order vs what is announced by a screen reader/asistant software

And additionally as is pointed out in the issue, there are people who will not be using assistive technologies at all but for whom a UI with jumpy visual ordering could be confusing or inaccessible.

It's actually got me wondering if there's a hybrid solution to this problem.

What if we do masonry layout by default but then add some kind of switcher for the results that moves it to a plain grid pattern (to match all results view)?

We could even potentially enable this view automatically if we detect someone using keyboard navigation for example, though I'm not sure that's something we want to store in the UI state cookie and it would cause the layout to shift significantly after the first interaction where we determine that keyboard navigation is being used. Then again, the UI wouldn't be _moving_ so much as the list of images would be getting replaced wholesale. There would probably be less movement than with the banners or sidebar popping in and out.

What do y'all think of that solution?

zackkrida:
That's an interesting idea @sarayourfriend . I wonder if these users would correlate to users who would use `prefers-reduced-motion` and if that would be a reliable way to detect them.

Otherwise, I planned on this PR to constrain images with a min and max aspect ratio, to prevent at best and limit at worse the order disparity issue.

sarayourfriend:
I'm not sure whether that would correlate but preferring reduced motion could be a good indication for someone who does want a simpler UI layout than a masonry style grid would give.

panchovm:
I agree with @krysal that each image is a standalone item, but my main concern lies in how images arrangement mirrors the navigational logic that users expect. If I understand correctly, the row-based layout does solve this problem. Right?

I like @sarayourfriend's idea, but I would not go with an automatic solution. Instead, having a view setting option in the top-right area of the results grid sounds more friendly. Shifting the grid automatically is unexpected and might be confusing at first. 

I can explore a solution for our next release, but for now, I am drawn using a row-based grid.

sarayourfriend:
> I like @sarayourfriend's idea, but I would not go with an automatic solution. Instead, having a view setting option in the top-right area of the results grid sounds more friendly. Shifting the grid automatically is unexpected and might be confusing at first.

Agreed. We can store the option in the UI state cookie as well so that subsequent renders use the simplified view.

PR #547, opened 2021-12-21
Stop importing from nuxt.config.js

### LINKED ISSUE #524: Stop importing `nuxt.config.js` into the app runtime
## Description

Currently we import `nuxt.config.js` into the app's runtime, mainly to reference the `dev` property of the config object.

This is fine in Nuxt 2 but will not be supported in Nuxt 3 because Nuxt 3's configuration file requires you to import from Nuxt 3's build-only toolkit. Trying to import this toolkit during runtime will cause webpack to crash.

Likewise, it also bundles our entire configuration file into the frontend bundle, which isn't necessary and could be unexpected, especially if secrets get interpolated into that file by Webpack's environment variable resolution.

## Reproduction
1. Visit https://wordpress.org/openverse and open the JavaScript debugger. Inspect the files coming from search.openverse.engineering and in each file search for a pattern that would match the config file (like `dev:`) and you'll be able to find the configuration file in total bundled into the frontend code.

Captura de Tela 2021-12-14 às 12 03 48

This is not currently a vulnerability but it could unexpectedly lead to one via an unintentionally leaked secret.

## Expectation
We should move the `dev` variable into it's own module (`dev.js`) and import _that_ into the config file and in the runtime. This would allow us to avoid duplicating the logic around the app while still preventing the import of the config file.

## Additional Context
https://github.com/nuxt/framework/issues/1320

## Resolution

sarayourfriend:
## Fixes

Fixes #524 by @sarayourfriend 

## Description

Stops importing from `nuxt.config.js` during the runtime. See related issue for the rationale behind this.

## Testing Instructions

Run `pnpm dev` and `pnpm storybook` and verify that they work as expected. There should be no changes to the behavior of the application.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/composables/use-i18n-sync.js
+++ b/src/composables/use-i18n-sync.js
@@ -1,5 +1,4 @@
 import { sendWindowMessage } from '~/utils/send-message.js'
-import config from '../../nuxt.config.js'
 import untranslatedLocales from '../locales/scripts/untranslated-locales.json'
 import {
   computed,
@@ -130,7 +129,6 @@ export default function useI18nSync() {
   onMounted(() => {
     window.addEventListener('message', localeMsgHandler)
     sendWindowMessage({
-      debug: config.dev,
       type: 'localeGet',
       value: {},
     })
--- a/src/utils/env.js
+++ b/src/utils/env.js
@@ -1,0 +1,19 @@
+// Use relative import to resolve during nuxt runtime
+import stringToBoolean from './string-to-boolean'
+
+/**
+ * Default environment variables are set on this key. Defaults are fallbacks to existing env vars.
+ * All boolean values should be designed to be false by default.
+ */
+export const env = {
+  apiUrl: process.env.API_URL ?? 'https://api.openverse.engineering/v1/',
+  enableGoogleAnalytics: stringToBoolean(process.env.ENABLE_GOOGLE_ANALYTICS),
+  googleAnalyticsUA: process.env.GOOGLE_ANALYTICS_UA ?? 'UA-2010376-36',
+  filterStorageKey: 'openverse-filter-visibility',
+  notificationStorageKey: 'openverse-show-notification',
+  enableInternalAnalytics: stringToBoolean(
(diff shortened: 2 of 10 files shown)


src/locales/scripts/wp-locales.json
```diff
@@ -467,7 +467,7 @@
     "facebookLocale": "el_GR",
     "wpLocale": "el",
     "code": "el",
-    "translated": 9
+    "translated": 28
```
AetherUnbound, line 470: This looks like the change from #545, should this PR be pointed at that branch?

sarayourfriend: Whoops, I didn't mean to include that file! I'll remove it from the commit.

src/composables/use-i18n-sync.js
```diff
@@ -130,7 +129,6 @@ export default function useI18nSync() {
   onMounted(() => {
     window.addEventListener('message', localeMsgHandler)
     sendWindowMessage({
```
obulat, line 131: Can we import `dev` from the `utils/dev` here, too? It's usually helpful to see the message exchange when debugging the iframe.

obulat (approved):
LGTM

src/utils/env.js
```diff
@@ -0,0 +1,19 @@
+// Use relative import to resolve during nuxt runtime
+import stringToBoolean from './string-to-boolean'
+
+/**
+ * Default environment variables are set on this key. Defaults are fallbacks to existing env vars.
+ * All boolean values should be designed to be false by default.
+ */
+export const env = {
```
dhruvkb, line 8: I wonder if there is a performance boost to using `Object.freeze` on this object literal?

dhruvkb (approved):
LGTM, this is great!

src/utils/env.js
sarayourfriend: Maybe? I doubt it's a measurable difference though.

Issue #554, opened 2021-12-23
Add E2E tests to CI


sarayourfriend:
## Problem

It is currently possible to catastrophically break the frontend in a PR that seems unrelated without knowing: https://github.com/WordPress/openverse-frontend/pull/551

## Description

We need to run the E2E test suite in CI to help prevent this.

## Additional context

@dhruvkb recommended we look into this package to facilitate this: https://github.com/bahmutov/start-server-and-test

However, it's just occurred to me as I write this that we should run the E2E tests with the wp.org theme wrapper, to most closely replicate the production environment.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

obulat:
Closing this issue as the #555 and adding more tests after it have added decent E2E check to our CI. We can, however, open a new issue for the new E2E tests that we should write.

PR #555, opened 2021-12-23
Add stop-gap solution for e2es

sarayourfriend:
## Fixes

Part of #554 by @sarayourfriend 

## Description

Adds a step to run the e2e suite in the CI workflow. This isn't the ideal situation as it doesn't actually replicate the production environment because it doesn't have the wp.org wrapper. We can continue to explore a solution that accomodates that but for now this is better than nothing.

## Testing Instructions

Let the CI run and verify that it works.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,3 +48,6 @@ jobs:
 
       - name: Run Storybook smoke-test
         run: pnpm storybook -- --ci --smoke-test
+
+      - name: Run E2E
+        run: pnpm e2e:ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,5 +49,8 @@ jobs:
       - name: Run Storybook smoke-test
         run: pnpm storybook -- --ci --smoke-test
 
+      - name: Install E2E browsers
+        run: pnpx playwright install
+
       - name: Run E2E
         run: pnpm e2e:ci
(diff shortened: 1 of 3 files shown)


obulat:
The tests are failing because I added a really hacky test for 'Meta search' sources. It checks for the number of buttons, and it wasn't updated when a new image source was added recently. So, it returns 6 instead of the previously-correct value of 5.

Also, I wonder if it's a good idea to merge this before adding the SSR mocking. Currently, we only mock the network requests on Client-side rendering, and anything rendered on the Server-side goes through the actual server, which takes a long time, and is flaky.

sarayourfriend:
The tests currently fail also fail for me locally.

sarayourfriend:
> Also, I wonder if it's a good idea to merge this before adding the SSR mocking. Currently, we only mock the network requests on Client-side rendering, and anything rendered on the Server-side goes through the actual server, which takes a long time, and is flaky.

Ah, good point. We should do the mocking for the server side as well 😕

sarayourfriend:
@obulat what do you think about us merging this as is for now and then opening another issue to investigate how to get Audio passing consistently (while also relying only on text on the screen that users will be able to interact with?)

obulat:
I agree, @sarayourfriend

obulat (approved):

dhruvkb (approved):

PR #557, opened 2021-12-23
Update API urls to avoid redirects

### LINKED ISSUE #556: Use new API URLs in requests to avoid redirects
## Problem

The API URL structure changed a little bit when the regular URL conf was replaced with DRF router. Now, there are trailing slashes, and the requests from the frontend are redirected using 301.

## Description

It would be cleaner to add the trailing slashes to the requests by the frontend to avoid redirects.

## Implementation

obulat:
## Fixes

Fixes #556 by @obulat

## Description

This PR adds trailing slashes to the old request URLs to avoid 301 redirects.

```
https://api.openverse.engineering/v1/images?q=cat[HTTP/2 301 Moved Permanently 579ms]
https://api.openverse.engineering/v1/images/?q=cat [HTTP/2 200 OK 288ms]
```
These changes are added to the `search`, 'single result pages' for image and audio, and the `thumbs` endpoints.

## Testing Instructions

Run the app using `pnpm run dev`, and try searching or looking at the single result pages. You should not see any 301 redirects in the console.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/data/audio-service.js
+++ b/src/data/audio-service.js
@@ -39,7 +39,7 @@ const AudioService = {
 
     return dev
       ? Promise.resolve({ data: sampleAudioResponses.detail })
-      : ApiService.get('audio', params.id)
+      : ApiService.get('audio/', params.id)
   },
 
   /**
--- a/src/data/audio-service.js
+++ b/src/data/audio-service.js
@@ -39,7 +39,7 @@ const AudioService = {
 
     return dev
       ? Promise.resolve({ data: sampleAudioResponses.detail })
-      : ApiService.get('audio/', params.id)
+      : ApiService.get('audio', params.id)
   },
 
   /**
--- a/src/data/api-service.js
+++ b/src/data/api-service.js
@@ -1,11 +1,34 @@
 import axios from 'axios'
+import { warn } from '~/utils/warn'
 
 const DEFAULT_REQUEST_TIMEOUT = 30000
+
+const validateRequest = (errorCondition, message, config) => {
+  if (errorCondition) {
+    warn(
+      `There is a problem with the request url: ${message}.
+Please check the url: ${config.baseURL}${config.url}`
+    )
+  }
+}
 
 export const createApiService = (baseUrl = process.env.apiUrl) => {
   const client = axios.create({
     baseURL: baseUrl,
     timeout: DEFAULT_REQUEST_TIMEOUT,
+  })
+  client.interceptors.request.use(function (config) {
+    validateRequest(
+      !config.url.endsWith('/'),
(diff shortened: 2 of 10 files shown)


krysal (commented):
Great! This is critical for `report/` requests. Could we include them here? Ty!

obulat:
Thank you for spotting this, @krysal ! I haven't thought about other requests at all :(

sarayourfriend:
@obulat I wonder if the ApiService shouldn't enforce this somehow? We could do this using an [axios interceptor](https://axios-http.com/docs/interceptors) for example.

obulat:
> @obulat I wonder if the ApiService shouldn't enforce this somehow? We could do this using an [axios interceptor](https://axios-http.com/docs/interceptors) for example.

It's a nice idea to use one place to enforce the correct URL usage, but I'm not sure it can be used here. The search URLs (https://api.openverse.engineering/v1/images/?q=cat) doesn't need a trailing slash, and all the others do.

sarayourfriend:
Yeah I guess we would have to split the URL parts, append the trailing slash if it's missing, then put it back together? 🤔

dhruvkb (approved):
I like that we have the trailing slash appended to all the URLs in use (and those that don't will not fail, just redirect). Enforcing this in the ApiServices seems excessive to me, especially because it shows clear warnings in the console (that can be easily fixed) and does not break anything.

krysal (approved):

src/data/audio-service.js
```diff
@@ -39,7 +39,7 @@ const AudioService = {
 
     return dev
       ? Promise.resolve({ data: sampleAudioResponses.detail })
-      : ApiService.get('audio', params.id)
+      : ApiService.get('audio/', params.id)
```
zackkrida, line 42: Won't this result in double slashes? The `get` function in the api service already adds a slash after the resource

sarayourfriend: @dhruvkb automating it via some axios adapter would prevent confusions like this ^. That's the main benefit to me, then we never have to think about it again. I see it the same as when we adopt an ESLint rule, just in this case we can fix the problem easily at runtime.

obulat:
> Yeah I guess we would have to split the URL parts, append the trailing slash if it's missing, then put it back together? 🤔

It turned out to be simpler than that, as the axios request already splits the url into parts, and it is the`config.url` param that needs to have a trailing slash. Thank you for suggesting validation, @sarayourfriend !

PR #560, opened 2021-12-23
Add `VContentLink` component

### LINKED ISSUE #453: Component: Content link
## Description
We need to add the mobile version of the **Content Link component**. The Design Library has a new `small` version with all three states, as the other instances of the same Figma component.

### Props

| Name | Type | Description |
|-|-|-|
| `media` | string | One of the current media types we support (likely `IMAGE` or `AUDIO`) |
| `results-count` | number | The number of results for the indicated media type |

### Events

|-|-|-|
| click | `media` | Indicates the media selected to narrow the search results |

## References
**[Design Library: Content link component](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=558%3A2284)**

- [Use on mobile](https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=133%3A11555)
- [Use on desktop](https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=133%3A34915)

krysal:
## Fixes

Fixes #453 by @panchovm 

## Description

This PR adds the mobile version of the Content Link component.

## Testing Instructions

Run `pnpm storybook` and try its interactions.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VContentLink/VContentLink.vue
```diff
@@ -36,6 +40,10 @@ export default defineComponent({
       type: Number,
       required: true,
     },
+    isSelected: {
+      type: Boolean,
+      default: false,
+    },
   },
   setup(props) {
     const { i18n } = useContext()
```
obulat, line 49: We should probably extract this `resultsCountLabel` function in the future, because it's used in the search bar as well.

zackkrida:
Looks great, but should the desktop version be included here too? Maybe instead of looking at is as desktop vs. mobile it should be prop-based, like 'stacked' vs. 'horizontal' with a layout prop?

CleanShot 2022-01-05 at 15 38 32@2x

panchovm:
It looks great 🎉 

Just one comment. I noticed it says `Image` instead of `Images`. It should have the **S** at the end of **Image**, not in Audio. 

To @zackkrida's comment. I would add the version when filter sidebar is expanded and the element slightly reduces its width.

krysal:
@zackkrida I like that approach! In my mind, I was overcomplicating it with breakpoints. I'd add the desktop version too.

@panchovm Good catch, thanks, I'll fix it. Also, are you referring to [this component](https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=133%3A11596)?

CleanShot 2022-01-06 at 11 36 39@2x

I think that @obulat is developing that in #568 (not sure). If not, it will probably be another component anyway, called `VContentTypeMenu` or something like that. I see how they are closely related and I can work on it after this.

zackkrida:
@krysal The desktop version is actually used on the "All" results page, above the results:

Figma: https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=133%3A22596

CleanShot 2022-01-06 at 10 48 22@2x

panchovm:
Sorry @krysal 

I was referring to these versions.

## Large
Desktop with a hidden filter sidebar

CleanShot 2022-01-06 at 17 55 06@2x

## Medium
Desktop with a visible filter sidebar

CleanShot 2022-01-06 at 17 53 43@2x

## Small
Mobile 

CleanShot 2022-01-06 at 17 55 31@2x

src/components/VContentLink/VContentLink.vue
krysal: I extracted it into a new `use-i18n-utilities.js` composable and tried to replace the code on the VHeader but seems not to work well within `watchEffect`, I couldn't make it work there.

zackkrida (approved):
This looks really nice! I can implement this in the actual user interface in my image and all grid pull requests.

test/unit/specs/components/v-content-link.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    render(VContentLink, options)
+    const btn = screen.getByRole('radio')
+    expect(btn).not.toHaveAttribute('aria-checked')
+  })
+
+  it('is marked as selected when indicated with the isSelected prop', () => {
+    options.props.isSelected = true
+    render(VContentLink, options)
+    const btn = screen.getByRole('radio')
+    expect(btn).toHaveAttribute('aria-checked')
```
sarayourfriend, line 39: For some extra assurance it might be good to also check the value of `aria-checked` is `"true"`. I suppose it could also be `"false"` (or any other arbitrary value) and this test would pass.

test/unit/specs/components/v-content-link.spec.js
```diff
@@ -0,0 +1,48 @@
...
+  })
+
+  it('should emit selected event on click', () => {
+    const { emitted } = render(VContentLink, options)
+    const btn = screen.getByText('search-tab.image')
+    userEvent.click(btn)
+    expect(emitted().selected).toBeTruthy()
+  })
```
sarayourfriend, line 47: Non-blocker, kind of philosophical question 😅 I kind of wonder if instead of emitting our own custom event when the button is clicked, if it wouldn't be easier to just `v-on="$attrs"` on the `button` in the component so that consumers can just attach whatever events they want to it instead of just this `selected` event (which users will have to remember what the signature of it is every time they want to use this component).

src/components/VContentLink/VContentLink.vue
```diff
@@ -0,0 +1,95 @@
...
+      default: false,
+    },
+    /**
+     * `stacked` intended for mobile and `horizontal` for desktop.
+     */
+    layout: {
+      type: String,
+      default: 'stacked',
+      validator: (val) => ['stacked', 'horizontal'].includes(val),
+    },
```
sarayourfriend, line 73: It's a shame we have changes other than just organization that get tied to this (like font size I think?) because it seems like something that `flex-wrap` should be able to take care of.

Do we envision the usage of this prop to be a situation where the component is rendered in distinct places where the user will know whether it's a mobile or desktop context? Or are we potentially just pushing the media-query up the stack so that the code for this prop will be something like like `:layout="isMinScreen('md') ? 'horizontal' : 'stacked'"`?

If it's the last case then my inclination would be to just move that logic into the component and hide the fact that it has different layouts depending on screen size from the user (so the user doesn't have to worry about it).

This is not a blocker or even a solid suggestion, just wondering what the intention is.

sarayourfriend (approved):
LGTM! Just a couple questions, non of which are blocking or even necessarily include specific suggestions for changes. Feel free to ignore them or we can push them off to a later day to solve later 🙂 

Good work! The component looks good and the implementation is really nice.

src/components/VContentLink/VContentLink.vue
zackkrida: I also started thinking we might be able to handle this with just css-based media queries. I can look at that when I implement this component in #530 and my 'all' results PR.

test/unit/specs/components/v-content-link.spec.js
krysal: I forgot it, thanks! Added the check in [`dbe772a`](https://github.com/WordPress/openverse-frontend/pull/560/commits/dbe772ae272177968abcce54e1d5138d7383eb1a).

obulat:
Krystle, I can repeat what Francisco told me about the borders: in hover and/or hover state, there shouldn't be a gray border, only the pink outline:
Screen Shot 2022-01-14 at 9 22 06 PM
Can you see the tiny gray border inside the pink outline here?

obulat (approved):
The content link looks amazing, @krysal ! I only saw one tiny issue with the border of a hovered button in the comment.

panchovm:
I saw what @obulat found before and forgot to comment the change. Thanks Olga for bringing it.

test/unit/specs/components/v-content-link.spec.js
krysal: I was trying before and the native event wasn't working for me, then I realized it was using the wrong instance property and I had to use `v-on="$listeners"`, thanks for making me look again at this! It should be a transparent component for events now 😄

src/components/VContentLink/VContentLink.vue
krysal: Using the `isMinScreen()` is giving me mismatched HTML problems when the component is server-side rendered. I'll have to take a closer look at how those errors are being handled but for now I'd like to tie this together. I agree that it would be ideal to hide this in the component logic and save us a property here as well.

krysal:
Thanks everyone for reviewing! Looks all good now :)

Issue #543, opened 2021-12-20
Add a report content action in single result views


panchovm:
## Problem
The current mockups do not have an action for reporting content, and we need to add it as users can find content that needs to be reported.

## Alternatives
The current flow is a two-step process where the radio option shows different info in step two based on the alternative selected. This flow has interaction risks on mobile since it requires an internal navigation to close it and go back and forth.

The proposal is a one-step popover with the original description and reporting reasons, but it displays different content once on the same popover once clicking on an alternative. The interaction goes as it follows:
1. The content information title ("Audio information", "Image information") has a right-aligned link.
2. On hover, the link turns `dark-charcoal`.
3. On click, it opens the popover with radio alternatives unselected and the report button as `disabled`.
4. As clicking on each option, new information is displayed and the report button turns `enabled` as the info is filled correctly. See the image below for the flow.
5. In option **Other**, the text area has a character counter to assure a maximum length. See the image below for the flow.

<img width="1254" alt="CleanShot 2021-12-20 at 15 48 46@2x" src="https://user-images.githubusercontent.com/895819/146817581-16930016-5014-40c1-a26d-efb0ee306dcb.png">

I am curious to see your thoughts, especially on the following:
* **Mobile view.** The single result view is a modal that looks like a page, and it might be a good idea to show this popover as a modal since it focuses the content on a single vertical layout interaction.
* **Clicking outside**. In other popovers, clicking outside closes the box without affecting the viewing info significantly. However, this box contains flows that might frustrate users if it gets closed by unexpected interactions. Like resetting the reason info in the third option.
* **External icon size.** I talked to @dhruvkb days ago about icon sizes, and this one is based in the WordPress Design Library but scaled down to be `16px` square.

https://user-images.githubusercontent.com/895819/146820069-78532fab-b9e1-45fe-b842-ef6717bc2a33.mp4

## Additional context
The popover design has a **shadow popover** style from the [WordPress Design Library](https://www.figma.com/file/e4tLacmlPuZV47l7901FEs/WordPress-Design-Library).

sarayourfriend:
> Mobile view. The single result view is a modal that looks like a page, and it might be a good idea to show this popover as a modal since it focuses the content on a single vertical layout interaction.

Agreed! A blocking modal makes the most sense on small devices for this case.

[...]

PR #562, opened 2021-12-26
Refactor Content report

### LINKED ISSUE #561: Keep content report form state local
## Problem

 Currently, the state for the content report form is kept in Vuex. 

## Description

It is recommended to use the Vuex state for data that is used by several components. The content report state is used by only one component, `ContentReportForm` (other smaller components such as `DoneMessage` or `DmcaNotice` can use props). This means that moving the state from a separate namespaced Vuex store to local state will simplify the app Vuex store and the component itself.

## Alternatives

We can leave the state in Vuex, however, this makes testing more difficult (it requires more mocking for the store in all the components that are using it). With the issues with Nuxt3 `useStore`, it is probably better to reduce the Vuex usage to the data that really needs it.

## Implementation

obulat:
## Fixes

Related to #543 by @panchovm 
Fixes #561 by @obulat

## Description

This PR removes the `report-content` Vuex store module, moving the functionality to the `VContentReportForm` component.

Other changes in this PR include:
- some styling adjustments to the report form (more work on this will be needed)
- renaming of the touched components to add the `V` prefix.
- refactoring `PhotoDetails` and all the components in `ContentReport` directory to use composition API.
- replacing Icon font icons with svg icons (`back`, `flag`).
- adding component imports to remove warnings in test output.
- rewriting the `photo-details.spec.js` tests to use `testing-library`.

## Testing Instructions

Run the app using `pnpm run dev`, open a single image result page, and click on 'Report'. The popup form should open with 3 options, and it should work correctly.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/locales/en.json
```diff
@@ -535,21 +535,20 @@
...
-        "link": "DMCA form",
-        "aria-form": "{provider} form"
-      }
+      "thanks": "Thank you for reporting an issue with the results of Openverse!",
+      "provider": "We recommend doing the same at the source, {link}.",
+      "link": "DMCA form",
+      "aria-form": "{provider} form",
+      "submission-error": "There was an error with the report submission. Please try again.",
+      "dmca-content": "You must fill out this {link} to report copyright infringement. No action will be taken until the form is filled out and submitted.",
+      "placeholder": "Issue description required (with at least 20 characters)"
```
zackkrida, line 551: What is the reason to flatten these and remove them from a 'dmca' namespace? This is a nitpick, just curious

test/unit/specs/components/ContentReport/content-report-form.spec.js
```diff
@@ -1,146 +1,173 @@
...
+
+    expect(getByRole('button', { name: /submit/i })).toBeDisabled()
+
+    // Even though the button is disabled, if we `fireEvent.click`, the report
+    // in this test is sent
+    // expect(serviceMock.sendReport).toHaveBeenCalledWith({
+    //   identifier: props.image.id,
+    //   reason: 'other',
+    //   description,
+    // })
```
zackkrida, line 171: Should this test be removed?

src/components/VContentReport/VContentReportForm.vue
```diff
@@ -0,0 +1,182 @@
...
+  name: 'VContentReportForm',
+  components: {
+    VDoneMessage,
+    VDmcaNotice,
+    VReportError,
+    VOtherIssueForm,
+    VButton,
+    VIcon,
+  },
+  props: ['image', 'providerName', 'reportServiceProp'],
```
zackkrida, line 108:
```suggestion
  props: ['image', 'providerName', 'reportService'],
```

Why does this prop need the word 'prop' in its name?

zackkrida (approved):
This looks great, very nice refactor. One existing bug is still present, the provider name doesn't display correctly. I think we could remove the providerName prop and just access the provider from the `image` prop instead.

> providerName is undefined
CleanShot 2022-01-05 at 14 46 38@2x

src/locales/en.json
obulat: It was originally because the 'DoneMessage' also used strings from 'dmca'. But I should have only flattened the strings that are used by both, and kept the dmca-only strings in the nested object.

obulat:
> One existing bug is still present, the provider name doesn't display correctly. I think we could remove the providerName prop and just access the provider from the image prop instead.

`providerName` can be different from the `image.provider` prop, and I think it's useful to have both provider codes and display names (i.e. we can shorten some of the very long display names when using them in the code). The problem with not displaying `providerName` was embarassingly simply: I forgot to add the `return` in the getter function.

zackkrida:
@obulat ah, thank you for clarifying, I forgot there was a difference that one was formatted and the other was the code.

src/components/ImageDetails/VPhotoDetails.vue
```diff
@@ -0,0 +1,323 @@
...
+      tabClass,
+      setActiveTab,
+
+      providerName,
+    }
+  },
+}
+</script>
+
+<style lang="scss" scoped>
```
dhruvkb, line 301: This block could be replaced entirely with Tailwind utilities (including the `44rem` max-height, now that JIT is supported).

dhruvkb (approved):
I have a suggestion but it's non-blocking considering it runs tangential to the goal of the PR. Great refactor, LGTM!

Issue #368, opened 2021-10-27
Filters sidebar


obulat:
## Description

Filters sidebar display all filters that are available for current media type(except for `mature` filter, which is not displayed), and their selected/not selected state. The visibility of the filters should be set by the parent. The filter sidebar should have a [max-width/ set width] ?? of [316px/ 1/5th of the screen width ] ?? on screens that are larger than [mobile - 375px / tablet ???] ?? On mobile screens, it should be full-width.

### Props

`filters`: an object with `filterType` keys and an array of filter objects:

| Name | Type | Description |
|-|-|-|
| filters | an object with `filterType` keys and an array of filter objects | `{ licenses: [ { code: 'cc0', name: 'filters.licenses.cc0', checked: false } ...  ], ... } `

### State

| Name | Type | Description |
|-|-|-|

### Events

| Name | Associated data | Description |
|-|-|-|
| hide | none | The parent should hide the filters on receiving this event
| toggle-filter | `filterType`, `code` | This event should be called on child checkbox 'change' event, and handled by Vuex TOGGLE_FILTER action.
| clear-filters | none | This event should be called on 'Clear filters' button click, and should call Vuex CLEAR_FILTERS action.

## Dependencies

- [x] #405
- [x] License explanation popup depends on #366

## References

- **Figma:**
Desktop:
https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=133%3A48756
Mobile:
https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=133%3A12225

## Implementation

- [ ] 🙋 I would be interested in implementing this component.

obulat:
@panchovm , 
Could you please clarify a couple of questions about the filters list:
### Width:
Should it be max width  or set width? 
316px (or something else), or 1/5th of the screen width?
### Tablet width:
I guess the mobile design should be used?

Also, are we going to have the 'License explanation' buttons with a popup (what would the popup look like on mobile?) and the 'Clear filters' button?

panchovm:
## Width

The width should be fixed as `316px` and right-aligned. The area that changes its size is the content area where results appear. Here is an example of what I was thinking of for the expanding behavior.

https://user-images.githubusercontent.com/895819/140318644-4740441d-9593-49dc-ac37-dc3c01f2926a.mp4

## Breakpoint
[...]

PR #569, opened 2022-01-03
Filter modal

obulat:
## Fixes

Related to #368 by @obulat

## Description

This PR adds a dynamic component for filters: one that is a full-page modal for mobile widths, but a sidebar for tablet and up. This helps us to avoid having two elements with the same ids for inputs, one hidden on mobile screens and one hidden on desktops, as we previously had. 

The VHeaderFilter component is a wrapper (with implementation mostly based on VModal component) over a VFilterButton, and the VTeleport to the sidebar on desktop, or a VModalContainer on mobile screens.

To overcome the Nuxt SSR hydration problem described in the `Additional details` section, the component is initially set to the mobile version, and is changed to the desktop version in `onMounted` hook, if necessary.

Other changes:
- `VModalContent` is updated to allow for creating the full-page mobile modals.
- The `VButton` is using `aria-pressed` and `aria-expanded` states to add the `pressed` styles.
- `tertiary` `VButton` variant is now a bordered button, both in resting and in hovered states. It is turned into `action-menu` variant, which is a bordereless button, when the header is scrolled.
- `VSearchGridFilter` is refactored to use composition API.

## Additional details
Using a `computed` property returning a dynamic component based on the screen width causes problems. By default, when server-rendered, the `isMdScreen` value is false, so the full-screen modal is rendered on the server. Then if you open the app on a desktop, the `computed` returns the sidebar teleport. This causes a SSR - CSR mismatch, and a full re-render. Sometimes, this full re-render also causes Nuxt to mount some components twice.

```
Parent:  
<button class="transition-colors self-c…Ks-oJ size-medium_faRVV" type="button" aria-controls="filter-sidebar" aria-label="0 Filters" aria-haspopup="dialog">
vue.common.dev.js:6459
Mismatching childNodes vs. VNodes:  
NodeList(3) [ svg.v-icon.w-6.h-6, #text, span ]
Array(3) [ {…}, {…}, {…} ]
vue.common.dev.js:6460
[Vue warn]: The client-side rendered virtual DOM tree is not matching server-rendered content. This is likely caused by incorrect HTML markup, for example nesting block-level elements inside <p>, or missing <tbody>. Bailing hydration and performing full client-side render. vue.common.dev.js:630
```

The same error is thrown when we render something using `v-if` based on screen width (as in when we decide whether to show an icon on a button or not).
It seems that this is a know issue in Nuxt without a solution: https://github.com/nuxt/ nuxt.js/issues/9441.
## Testing Instructions

Run `pnpm run dev`.
On any page besides the search pages, there should be no filter button in the header, and no filter sidebar, even if the it is set to open in the local storage (you can open the sidebar on the search page to set the local storage value to true, and then reload the homepage to test this).
On a search page, the header should have a filters button. You can test that it opens a sidebar on desktop, but a modal on mobile. The desktop version should change into mobile if you change screen width.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
@obulat I don't feel like I can give this a full review, because I'm not sure what the expected behavior is. I see the sidebar on desktop, which looks nice, but on mobile the filter button disappears and no modal is visible. 

Here's a simplified example of how we should handle the styling for the sidebar scrolling independently: https://codepen.io/zackkrida/pen/BawVrmr?editors=1100

The fix for SSR seems to make sense to me; I also wanted to make you aware of something in the searchbar PR. Media queries were given a 'shouldPassInSSR' option which allows for setting the default value of the media query. This allows us to choose to optimize for mobile or desktop, and reduce layout shift for desktop users (which the majority of our users are) https://github.com/WordPress/openverse-frontend/blob/ea04790327f182c21d775aafc876d250fb8639c7/src/composables/use-media-query.js

zackkrida:
@obulat this is really difficult to merge the up to date VHeader branch into, I tried exploring this in a separate PR based on this one: https://github.com/WordPress/openverse-frontend/pull/590

Maybe we could talk about it Monday and even click through together, I removed the `VHeaderFilter` abstraction but kept most of the work in this PR. Maybe we can work together to figure out what the best solution is! Mine has a really weird bug when the filters are open by default on an SSR page, but I bet you could help me fix it 😆

obulat:
> @obulat this is really difficult to merge the up to date VHeader branch into, I tried exploring this in a separate PR based on this one: #590
> 
> Maybe we could talk about it Monday and even click through together, I removed the `VHeaderFilter` abstraction but kept most of the work in this PR. Maybe we can work together to figure out what the best solution is! Mine has a really weird bug when the filters are open by default on an SSR page, but I bet you could help me fix it 😆

Working together on Monday sounds like the best way forward! I've also worked on the ContentSwitcher, making it work both on Mobile and Desktop, using dynamic components based on the media query: https://github.com/WordPress/openverse-frontend/tree/refactor-cs . The problem in this branch is the layout and the focus management of the combined pages and content types popover in the mobile modal.

PR #606, opened 2022-01-14
Open filters in sidebar or modal from the header

obulat:
## Fixes

Related to #462 by @zackkrida
Reworked #569 to remove merge conflicts.

## Description

This PR adds a dynamic component for filters: one that is a full-page modal for mobile widths, but a sidebar for tablet and up. This helps us to avoid having two elements with the same ids for inputs, one hidden on mobile screens and one hidden on desktops, as we previously had.

The `VHeaderFilter` component is a wrapper (with implementation mostly based on `VModal` component) over a `VFilterButton`, and the `VTeleport` to the sidebar on desktop, or a `VModalContainer` on mobile screens.

## Testing Instructions

Run `pnpm run dev`.
On any page besides the search pages, there should be no filter button in the header, and no filter sidebar, even if the it is set to open in the local storage (you can open the sidebar on the search page to set the local storage value to true, and then reload the homepage to test this).
On a search page, the header should have a filters button. You can test that it opens a sidebar on desktop, but a modal on mobile. The desktop version should change into mobile if you change screen width.
## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VHeader/VFilterButton.vue
```diff
@@ -97,18 +95,15 @@ const VFilterButton = defineComponent({
...
       return true
     })
 
     // Hide the label entirely when no filters are applied on mobile.
     const showLabel = computed(() => {
-      if (!isMinScreenMD.value && !filtersAreApplied.value) {
-        return false
-      }
-      return true
+      return !(!isMinScreenMd.value && !filtersAreApplied.value)
```
zackkrida, line 106: thanks for the nice refactor 😄

src/components/VHeader/VHeader.vue
```diff
@@ -1,37 +1,44 @@
 <template>
   <header
-    class="sticky top-0 flex py-4 px-4 md:px-7 items-center justify-between z-40 w-full bg-white gap-x-2"
+    class="fixed top-0 flex py-4 px-4 md:px-7 items-stretch z-40 w-full bg-white gap-x-2"
```
zackkrida, line 3: If I switch this back to `sticky` and remove the hardcoded margins from the default layout, everything appears to work fine. That seems preferable. Do you remember the reason to switch it to fixed?

src/components/VHeader/VFilterButton.vue
obulat: It wasn't me, it was Webstorm's nagging 😆

src/components/VHeader/VHeader.vue
obulat: When one of the parents has an overflow value set to anything but `visible`, the `sticky` stops working: https://css-tricks.com/dealing-with-overflow-and-position-sticky/ 
And we set overflow to hidden on `app` div in the default layout

obulat:
> on about the sticky header, which we could fix on redesign branch, this looks good to me. Is this meant to be merged before the content switcher bra

Yes, this is meant to be merged first, and we could fix bugs after merging it into redesign.

src/components/VHeader/VHeader.vue
zackkrida: I realized with some modification it _actually_ doesn't need sticky _or_ fixed, because our grid layout is keeping at the top! See #609

obulat: I wish I read this comment before trying to figure out why it didn't have sticky 😆

zackkrida:
@WordPress/openverse-frontend I think this PR is the most essential to review, if anyone has review time today or Monday!

src/components/VTeleport/VTeleportTarget.vue
```diff
@@ -1,16 +1,23 @@
...
   data: () => ({ children: [] }),
   created() {
-    if (this.name in targets)
-      throw new Error(`VTeleportTarget: duplicate name ${this.name}`)
+    console.log('created')
+    if (this.name in targets) {
+      warn(
+        `VTeleportTarget: duplicate name ${this.name}, deleting previous teleport`
+      )
+      delete targets[this.name]
```
sarayourfriend, line 19: I wonder if this actually makes sense as a warning or if it should throw... duplicate teleport targets seems like a programming error that we wouldn't want to accidentally let into production.

src/composables/use-event-listener.js
```diff
@@ -9,7 +9,7 @@ import {
 /**
  * Use an event listener. Shamelessly stolen from https://logaretm.com/blog/my-favorite-5-vuejs-composables/#useeventlistener
  *
- * @param {import('@nuxtjs/composition-api').Ref<EventTarget | null> | EventTarget} target The target can be a reactive ref which adds flexibility
+ * @param {import('./types').MaybeRef<EventTarget | null> | EventTarget} target The target can be a reactive ref which adds flexibility
```
sarayourfriend, line 12: I'm not sure where `MaybeRef` was introduced but I think the type for it is wrong. I'll open a bug ticket to fix that. In the mean time, just so that I understand the intention for it, is it meant to make it unnecessary to add `null` or `undefined` to the inner type?

Also before I open the issue, just to verify the type. Should it be this instead:

```js
/**
 * @template T
 * @typedef {import('@nuxtjs/composition-api').Ref<T | null | undefined>}
 */
```

Cause right now the type of it will result in a variable that is either a Ref<T> or T itself (which doesn't seem to match the name of `MaybeRef` and I can't figure out a use case for it).

src/composables/use-media-query.js
```diff
@@ -41,8 +46,11 @@ export function useMediaQuery(query, options = { shouldPassInSSR: false }) {
 }
 
 /**
- * Check whether the curent screen meets
+ * Check whether the current screen meets
  * or exceeds the provided breakpoint size.
+ * @param {'sm'|'md'|'lg'|'xl'|'2xl'} breakpointName
+ * @param {Parameters<typeof useMediaQuery>} options
```
sarayourfriend, line 52: Nice! Just one little fix. `Parameters` returns a tuple of the types for the function's arguments, and `options` is the second one argument so we need to pluck that one specifically.

```suggestion
 * @param {Parameters<typeof useMediaQuery>[1]} options
```

Now that you've added types to this file, if you feel up for it, it might be worth enabling TypeScript checking on this file by adding it to the `includes` list in `tsconfig.json` 😱

sarayourfriend (approved):
Okay I tested it locally and it seems to work well for me. I tried to review as much of the code as possible but it's a big PR (which I can tell is necessary given the scope of the changes).

Only a couple things came up as I was reviewing but absolutely no blockers that I could see. Nice work wrangling all this together!

src/composables/use-event-listener.js
obulat: I copied this type from Vueuse: https://github.com/vueuse/vueuse/blob/8782e1761203cdc0bf41e95e38504fed98351c77/packages/shared/utils/types.ts#L15, 
And yes, it is a type that's either a ref or just a simple value.

PR #607, opened 2022-01-14
Add Content Switcher component

### LINKED ISSUE #485: Internal Pages Menu
Figma link: https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=25%3A2962

image

## Questions

- Does this need to be a named, standalone component, or just a popover + button in the header component?
- This three dot icon sticks out to me as odd. It's a lighter weight than any of the other icons in the system
- What are the focus + hover styles of these inner links? Related to a similar question in #479

obulat:
## Fixes

Fixes #485 by @zackkrida
This PR adds the Content Switcher to the Header. This is a new version of #591 that does not have merge conflicts. It is also based off of the `filter_sidebar` branch, so it shouldn't be merged before that.

## Description

The `VContentSwitcher` component combines the Internal Page menu and the Content type switcher. 
There is also a `VPageMenu` component used on internal pages (eg. `About`), where only the page menu, and not content switcher, are used.

This component turned out to be very complex.
On mobile screens and on initial SSR, the page menu button does not show, and clicking the content switcher button opens a modal with the Content types menu and the Pages menu.
On larger screens, clicking the Pages menu button (ellipsis icon) opens up a Popover with links to internal pages, and clicking the Content type button opens up a Popover with different content types available. Clicking on the items changes the search content type, and closes the popover.

### Issues
There is a problem with tabbing through the items in the mobile modal. I am not sure why they can't be tabbed through, it worked previously.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal (commented):
This is looking really good! It's noticeable that these components are broad in functionality and styles variants. I want to check some files in more detail to learn more from your approach :) but at a first glance seems pretty complete.

I'm just seeing that the `VPageMenu` doesn't close after clicking on a link, it effectively navigates but the popover/full page modal keeps open, one has to click outside/close the modal to see the target page.

src/composables/use-scroll.js
```diff
@@ -0,0 +1,43 @@
...
+      ? throttle(scrollHandler, throttleMs)
+      : scrollHandler
+
+    useEventListener(element, 'scroll', handler, {
+      capture: false,
+      passive: true,
+    })
+  }
+
+  return { y, isScrolled }
```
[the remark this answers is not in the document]
obulat: Their purposes are similar. At first, we were scrolling the whole window, but then we updated the layout so that it consists of a grid, in which the main content and the sidebar can be scrolled separately. This means that we can no longer detect if the header has scrolled or not by using `window`'s scroll offset, and need to get the main content `scrollTop` instead. I did not remove the `use-window-scroll.js` just yet. But if it turns out after all the redesign work that we don't, we should delete `use-window-scroll.js`.

src/components/VContentSwitcher/VContentTypes.vue
```diff
@@ -0,0 +1,60 @@
...
+import VItemGroup from '~/components/VItemGroup/VItemGroup.vue'
+
+export default {
+  name: 'VContentTypes',
+  components: { VIcon, VItem, VItemGroup },
+  props: {
+    bordered: {
+      type: Boolean,
+      default: true,
+      // !isMinScreenMd
```
sarayourfriend, line 39:
```suggestion
```

src/components/VContentSwitcher/VMobileMenuModal.vue
```diff
@@ -0,0 +1,133 @@
...
+        close()
+      } else {
+        open()
+      }
+    }
+    const selectItem = (item) => {
+      emit('select', item)
+    }
+
+    const isLinkExternal = (item) => !item.link.startsWith('/')
```
sarayourfriend, line 106: Nice! This would be something we could add to the `VButton` component as well 😮

src/composables/use-content-type.js
```diff
@@ -0,0 +1,34 @@
...
+import allIcon from '~/assets/icons/all-content.svg'
+import audioIcon from '~/assets/icons/audio-content.svg'
+import imageIcon from '~/assets/icons/image-content.svg'
+
+const icons = {
+  all: allIcon,
+  audio: audioIcon,
+  image: imageIcon,
+}
+export default function useContentType() {
```
sarayourfriend, line 14: I really like this way of encapsulating parts of Vuex into a composable. It feels really clean and creates a much nicer API for the components to work with. As a bonus I suspect this would make it easier to switch to pinia in the future.

src/store/search.js
```diff
@@ -346,7 +346,9 @@ const actions = {
       }
       commit(CLEAR_OTHER_MEDIA_TYPE_FILTERS, { searchType })
     }
-    await dispatch(UPDATE_QUERY_FROM_FILTERS, queryParams)
+    if (queryParams !== {}) {
```
sarayourfriend, line 349: This python-ism won't work in JavaScript 😢  If we need to test for an empty object there are two ways we could do it. Either we could test `Object.keys(...).length` and check it for `0` (which I'm assuming has some caveats that I can't think of, it seems like it should work fine) or we could use the `lodash.isEmpty` function.

src/utils/search-query-transform.js
```diff
@@ -216,3 +216,12 @@ export const queryStringToQueryData = (queryString) => {
...
 }
+
+export const areQueriesEqual = (oldQuery, newQuery) => {
+  for (let key of Object.keys(oldQuery)) {
+    if (oldQuery[key] !== newQuery[key]) {
+      return false
+    }
+  }
+  return true
+}
```
sarayourfriend, line 227: Maybe this doesn't matter for queries but this wouldn't work if `newQuery` has keys that aren't present in `oldQuery`. For example, this will return `true` for:

```js
areQueriesEqual({}, { foo: 'bar' })
```

We could use `lodash.isEqual` or else do something like is described in [this SO answer](https://stackoverflow.com/a/35951373) after converting both arguments to Map objects or adapting that solution to just work on regular objects and avoid the conversion.

sarayourfriend (changes requested):
Just one blocker on the `areQueriesEqual` function and an invalid object literal comparison, but otherwise tests well and LGTM! Nice work!

src/utils/search-query-transform.js
obulat: @sarayourfriend, I added `lodash.isEqual`. I was using this comparison because the queries were supposed to have the same structure, and the same number of properties. In fact, we probably won't need this comparison after the All content page is actually different from the Image page.

sarayourfriend (approved):
LGTM!

src/utils/search-query-transform.js
sarayourfriend: Gotcha! I figured that maybe this specific consideration didn't matter for queries for that reason. Either way works for me!

krysal (approved):
Works perfectly on a desktop layout! This is an incredible work 🌟 

I noticed some issues on mobile, like the label of the content switcher button not changing, or the Openverse logo text not hiding on non-search pages, but I think those can be safely addressed in separate PR.


https://user-images.githubusercontent.com/9145885/149981543-44829c9a-2cfa-49fe-8b68-5d6fa2949793.mp4

PR #609, opened 2022-01-14
Remove hardcoded scrollbar styles, fixed header, and hide sidebar on non search views

zackkrida:
Makes some small adjustments to css grid and the header, along with removing the hardcoded scrollbar styles. 

The main improvement here is that when the sidebar isn't visible, the first column (the main content) will span both grid cells.

The 'main' element is always full height, and it either has 1 full-width child that can scroll, or two (the sidebar added) that scroll independently.


obulat (approved):
Much simpler! :)
I've added a quick fix for scroll button.

PR #647, opened 2022-01-20
Replace plain `<img>` tag with `VAudioThumbnail` component

### LINKED ISSUE #636: Audio single result page does not show default artwork if no thumbnail is available
## Description

The search results show default artwork for audio results that do not have a thumbnail. The single result page shows a broken image.

## Reproduction

1. Go to single audio result.
2. See error: the broken thumbnail in the audio details section. Note that related audio correctly shows the default artwork.

## Screenshots

Screen Shot 2022-01-19 at 7 49 08 AM

## Expectation

For audio without thumbnails, single result page should show default artwork.

## Resolution

dhruvkb:
## Fixes

Fixes #636 by @obulat 

## Description

This PR
- replaces the plain `` with the `VAudioThumbnail` component capable of automatically falling back to generated artwork
- makes the artwork generation deterministic so that it remains the same in different appearances of the file
- removes the extra space above the media tags when audio does not have any associated description

## Testing Instructions

0. Check out the branch and run the dev server.
1. Visit the page for an audio track without artwork.
2. See the generated art in the information section.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VAudioThumbnail/VAudioThumbnail.vue
+++ b/src/components/VAudioThumbnail/VAudioThumbnail.vue
@@ -49,12 +49,43 @@ export default {
       required: true,
     },
   },
-  setup() {
+  setup(props) {
     /* Switching */
 
     const ok = ref(true)
     const handleError = () => {
       ok.value = false
+    }
+
+    /* PRNG */
+
+    /**
+     * Small Fast Counter is a seedable pseudo-random number generator.
+     * @see {@link https://github.com/bryc/code/blob/master/jshash/PRNGs.md#sfc32}
+     */
+    const sfc32 = (a, b, c, d) => () => {
+      a |= 0
+      b |= 0
+      c |= 0
+      d |= 0
+      let t = (((a + b) | 0) + d) | 0
+      d = (d + 1) | 0
+      a = b ^ (b >>> 9)
+      b = (c + (c << 3)) | 0
+      c = (c << 21) | (c >>> 11)
+      c = (c + t) | 0
+      return (t >>> 0) / 4294967296
+    }
+    const rand = (seed) => sfc32(0x9e3779b9, 0x243f6a88, 0xb7e15162, seed)
+    const hash = (str) => {
+      let hash = 0
+      if (str.length === 0) return hash
+      for (let i = 0; i < str.length; i++) {
+        let chr = str.charCodeAt(i)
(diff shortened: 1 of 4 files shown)


src/components/VAudioThumbnail/VAudioThumbnail.vue
```diff
@@ -49,14 +49,45 @@ export default {
...
+    const hash = (str) => {
+      let hash = 0
+      if (str.length === 0) return hash
+      for (let i = 0; i < str.length; i++) {
+        let chr = str.charCodeAt(i)
+        hash = (hash << 5) - hash + chr
+        hash |= 0 // Convert to 32bit integer
+      }
+      return hash
+    }
```
sarayourfriend, line 89: Would it make sense to move these into library functions instead of redefining them for every instance of the audio thumbnail?

It also occurred to me (can be a separate issue) but we can avoid calculating or spending any time on the generated artwork when the audio has a thumbnail using an early return.

sarayourfriend (approved):
LGTM! I think it'd be a nice refactor to move the static functions into their own module to avoid redefining them for every instance of the component.

sarayourfriend:
Oh I noticed another thing when testing this out locally. For some reason the `alt` text of the artwork flashes briefly when a page loads. I would have expected the generated artwork to at least always be present.

Could this be because we're waiting for an error to happen when resolving `audio.thumbnail`? Given we're always generating the artwork anyway, could we actually flip the conditional and only switch to showing the `img` if loading is successful, and display the generated artwork in the meantime? The pop-in with the alt text temporarily is just a little strange.

src/components/VAudioThumbnail/VAudioThumbnail.vue
zackkrida: +1 to Sara's suggestion; I am using a prng in #618 and would be content to use this one instead!

zackkrida (approved):
Nice!

src/components/VAudioThumbnail/VAudioThumbnail.vue
dhruvkb: Makes sense, exporting functions into utils in 418dbfc.

PR #654, opened 2022-01-20
Fix global audio syncing and seeking issues

### LINKED ISSUE #632: Display a loading animation on the play/pause button when an audio is being loaded
## Problem

Switching between several audio tracks results in small (or sometimes larger) periods of seeming unresponsiveness while the audio track's URL is set to the audio element's current `src` and the audio loads.

Try playing several different audio tracks in a related audio section one after the other to see this feeling of unresponsiveness.

## Description

We could add some kind of loading animation to the play/pause button to indicate that the audio track is being loaded.

cc @panchovm for designs. Can we just use the loader? Or a different spinner? Maybe something else.

## Alternatives

Somehow pre-fetch audio tracks when someone hovers over the play/pause button for a determined period of time, or when someone seeks the audio.

## Implementation

### LINKED ISSUE #653: Update global audio to swap between many audio objects
## Problem

Currently global audio uses a single audio element that all audio tracks watch and update themselves according to.

This creates some significant complications with respect to syncing (and especially re-syncing) the state of any given audio track with the internal state of the audio element.

Doing this should fix all the weird edge case bugs we're seeing with browser differences when seeking paused or unplayed tracks and switching between tracks.

## Description

Rather than manage a single audio element between every track on the page, we can use an audio object per AudioTrack, and then swap the currently played one into a global ref for the active audio element.

Then the global audio will just watch the active audio and update state accordingly.

Additionally, rather than trying to apply status to the audio element, let's treat the audio element as the source of truth for a track, and sync it the other way around. That is, listen to the `play` `pause` `ended` and `timeupdate` events on the audio element and update our component's status and play time accordingly.

## Implementation

sarayourfriend:
## Fixes

Fixes #653 by @sarayourfriend
Closes #632 by @sarayourfriend — All audio tracks on the page buffer immediately so there's no longer a delay between playing them... do we need to prevent audio from loading until the "play" button is hit on it or it is seeked? Probably so.

## Description

Switches to using a global ref for the currently active audio track and swapping between multiple audio elements (one per track on the page, aside from the global track).

Some notes/key concepts:

1. We derive all state from the state of the audio object, whether the global "active audio" or the local audio. This means our state is updated when the local audio object emits the corresponding events. The only exception to this is the current time. I originally implemented this to always update the `currentTime` ref (which is used for displaying the play progress on the waveform) using the `timeupdate` event exclusively. But to achieve a smooth animation, when a track is playing, we'll update the `currentTime` ref in a `requestAnimationFrame` loop. `timeupdate` is still used to update the current time when the audio is not playing (for example if it is seeked while it is paused or ended). Otherwise the `timeupdate` event is ignored when the track is playing.
2. The `GlobalAudioTrack` and `AudioTrack` components share much of the same logic. At a glance it does appear that some of this stuff could be abstracted and I did spend a short amount of time thinking about how to do this. However, most of the "simple" ideas for how to do this, were thwarted by various issues with syncing global and local state selectively. The current implementation works, and while there is some duplicated code, most of it is subtly different. The approaching deadline for the audio release means spending time on refactoring this PR right now is pretty low value. If we decide to spend more time on this in the future, it could alleviate the maintainability burden of these two components, as long as we're sure their implementations won't diverge in the future. Given the already existing subtle differences between the two, I'd make a pretty confident bet that trying to abstract these behaviors into a shared utility would present difficulties in the future.
3. The local audio track has to be careful not to use the `Audio` object on the server. I tried to minimize the amount of code that was shifted into the `onMounted` hook, mostly because putting everything in there seems like an organizational misstep. If we did move everything in there, however, we could probably get rid of a lot of the optional chaining on the `localAudio` variable, but it seems like a small victory that could cause delays in mounting the component that are easily avoided (thanks to ergonomic features like optional chaining).
4. I tried to add explanatory comments to any logic that was non-obvious or could be implemented in different ways but needed to be done the way I did. If there's anything else that warrants explanation, let me know and I am happy to add more comments.

## Testing Instructions

Checkout this branch and run `pnpm dev`.

Visit an audio result page like http://localhost:8443/search/audio?q=birds and play around with seeking, playing, pausing, replaying tracks, both with and without the global audio player. Also make sure to test navigation between the search results and single result pages.

## Checklist

- [Not possible yet] I added or updated tests for the changes I made (if applicable).

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -220,6 +220,12 @@ export default defineComponent({
      * into our local audio, then we need to initialize
      * the local state syncing from the audio object
      * to our local refs.
+     *
+     * This lives here instead of closer to where `localAudio`
+     * is defined because `initLocalAudio` and several of
+     * the functions it depends on also all depend on the
+     * `localAudio` variable. This is the earliest in
+     * `setup` that this can be called.
      */
     if (localAudio) initLocalAudio()
 
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -35,7 +35,7 @@ import {
   useStore,
   watch,
   onUnmounted,
-  onMounted,
+  useRoute,
 } from '@nuxtjs/composition-api'
 
 import { useActiveAudio } from '~/composables/use-active-audio'
@@ -48,7 +48,7 @@ import VRowLayout from '~/components/VAudioTrack/layouts/VRowLayout.vue'
 import VBoxLayout from '~/components/VAudioTrack/layouts/VBoxLayout.vue'
 import VGlobalLayout from '~/components/VAudioTrack/layouts/VGlobalLayout.vue'
 
-import { ACTIVE } from '~/constants/store-modules'
+import { ACTIVE, MEDIA } from '~/constants/store-modules'
 import {
   PAUSE_ACTIVE_MEDIA_ITEM,
   SET_ACTIVE_MEDIA_ITEM,
@@ -59,7 +59,7 @@ const propTypes = {
    * the information about the track, typically from a track's detail endpoint
    */
   audio: {
-    type: /** @type {import('@nuxtjs/composition-api').PropType<{ url: string, duration: number, id: string }>} */ (
+    type: /** @type {import('@nuxtjs/composition-api').PropType<import('~/store/types').AudioDetail>} */ (
       Object
     ),
     required: /** @type {true} */ (true),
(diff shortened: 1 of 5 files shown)


zackkrida (approved):
This is amazing, and eliminates all browser inconsistencies and quirks I've encountered in the past.

> As I was writing up this PR description I found one bug where if you play an audio track on the search results page but then visit a different single result page, the previous audio track will continue playing. 

Regarding this, I noticed the track that is left playing is completely decoupled from any UI in that it can't be paused from it's original audio player or the global player. I think we can handle it in a separate PR but just wanted to note the behavior I'm seeing.

sarayourfriend:
@zackkrida the issue was pretty easy to fix, but did require some complicated changes to the initialization logic in `VAudioTrack`. However, as a side effect of this, I was able to delay loading audio until playback or seeking happens 👍 

All page transitions should now be working as smoothly as reasonably possible. Hopefully I haven't made any incorrect assumptions about how the search results are stored and what's presented on the page!

PR #656, opened 2022-01-20
Fix homepage a11y

### LINKED ISSUE #652: Animating the placeholder text on homepage is bad for accessibility
## Description

Animating the placeholder text of the search bar on the homepage every 100ms is a big problem for accessibility. We should find alternative approaches for this animation.

## Reproduction

1. Visit homepage using AT.
4. Experience the horror.

## Expectation

The animation should be a purely ornamental aspect of the homepage, and should completely cease to exist for assistive technologies.

## Additional context

This issue was branched from the homepage PR #593 with a possible solution in https://github.com/WordPress/openverse-frontend/pull/593#issuecomment-1017668491.

## Resolution

zackkrida:
## Fixes

Fixes #652 by @dhruvkb 

## Description

Moves the animated text from the input placeholder to an aria-hidden span positioned absolutely above the search input, which is hidden on focus and blur of the input.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VHeader/VSearchBar/VSearchBar.vue
+++ b/src/components/VHeader/VSearchBar/VSearchBar.vue
@@ -12,6 +12,7 @@
       type="search"
       name="q"
       :model-value="searchText"
+      v-on="{ blur: $listeners.blur, focus: $listeners.focus }"
       @update:modelValue="updateSearchText"
     >
       <!-- @slot Extra information such as loading message or result count goes here. -->
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -22,10 +22,21 @@
       </h2>
       <VSearchBar
         v-model.trim="searchTerm"
-        class="max-w-[40rem] mt-4 lg:mt-8"
-        :placeholder="featuredSearchText"
+        class="max-w-[40rem] mt-4 lg:mt-8 relative"
+        :placeholder="featuredSearch.term"
         @submit="handleSearch"
+        @blur="toggleAnimatedText"
+        @focus="toggleAnimatedText"
       >
+        <span
+          v-show="showAnimatedText && searchTerm === ''"
+          type="text"
+          aria-hidden
+          class="bg-white ml-4 w-[30ch] rounded-sm absolute block left-0 pointer-events-none text-dark-charcoal-60 group-hover:bg-dark-charcoal-06 group-focus:bg-dark-charcoal-06 group-focus-within:bg-dark-charcoal-06"
+          @click="toggleAnimatedText"
+        >
+          {{ featuredSearchText }}
+        </span>
         <VContentSwitcherPopover
           v-if="isMounted && isMinScreenMd"
           ref="contentSwitcher"
@@ -107,6 +118,7 @@ import {
   useRouter,
   useStore,
   useContext,
+  computed,
 } from '@nuxtjs/composition-api'
 
 import VContentSwitcherPopover from '~/components/VContentSwitcher/VContentSwitcherPopover.vue'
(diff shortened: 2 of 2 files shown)


src/components/VHeader/VSearchBar/VSearchBar.vue
```diff
@@ -12,6 +12,7 @@
       type="search"
       name="q"
       :model-value="searchText"
+      v-on="{ blur: $listeners.blur, focus: $listeners.focus }"
```
zackkrida, line 15: We don't want to pass _all_ listeners here because it breaks the input behavior, so I'm only passing the ones I need.

src/pages/index.vue
```diff
@@ -155,12 +167,18 @@ const HomePage = {
...
 
     const featuredSearchIdx = ref(-1) // immediately updates to 0 on mount
+    const featuredSearch = computed(
+      () =>
+        featuredSearches[
+          featuredSearchIdx.value >= 0 ? featuredSearchIdx.value : 0
+        ]
+    )
+
     const nextIdx = () => {
```
zackkrida, line 177: @dhruvkb does this break the animate-in effect I think you were going for with the `-1`?

src/pages/index.vue
```diff
@@ -22,10 +22,21 @@
...
         @submit="handleSearch"
+        @blur="toggleAnimatedText"
+        @focus="toggleAnimatedText"
       >
+        <span
+          v-show="showAnimatedText && searchTerm === ''"
+          type="text"
+          aria-hidden
+          class="bg-white ml-4 w-[30ch] rounded-sm absolute block left-0 pointer-events-none text-dark-charcoal-60 group-hover:bg-dark-charcoal-06 group-focus:bg-dark-charcoal-06 group-focus-within:bg-dark-charcoal-06"
+          @click="toggleAnimatedText"
```
dhruvkb, line 36: Since it has pointer-events set to `none`, will this `@click` be fired at all?

src/pages/index.vue
```diff
@@ -238,7 +256,14 @@ const HomePage = {
       })
     }
 
+    const showAnimatedText = ref(true)
+    const toggleAnimatedText = () => {
+      console.log('Just toggled to', toggleAnimatedText)
```
dhruvkb, line 261: Should be removed before merging.

src/pages/index.vue
zackkrida: No, and it's not needed. Thanks for noticing

zackkrida, line 36:
```suggestion
```

src/pages/index.vue
zackkrida, line 261:
```suggestion
```

src/pages/index.vue
```diff
@@ -22,10 +22,21 @@
       </h2>
       <VSearchBar
         v-model.trim="searchTerm"
-        class="max-w-[40rem] mt-4 lg:mt-8"
-        :placeholder="featuredSearchText"
+        class="max-w-[40rem] mt-4 lg:mt-8 relative"
+        :placeholder="featuredSearch.term"
```
dhruvkb, line 26: The placeholder shows up for a brief moment at the start of the animation, and it's a different font than the animated text so it has a flash there.

dhruvkb:
The animation seems to be jittering as if the browser is clamping it. Text doesn't update for several seconds and then the entire word appears.

src/pages/index.vue
sarayourfriend: We should remove the placeholder altogether from the input. There's no need for it to be there if the `span` is placed over the input. Leaving it in will cause Safari to announce it without context (like tabbing into the input will cause it to say "Universe" without any context if that's what's in the placeholder).

sarayourfriend (changes requested):
This still breaks VoiceOver, but is an improvement.

The animation and text typing is janky in Safari generally as well (just let it run and you'll notice it jumping around, typing over itself, and then eventually just stop working altogether).

The major problem I see with this is that if you type something into the search box then completely clear it, VoiceOver will get stuck saying that Safari is busy and announcements and interactions stop altogether. Pressing `tab` to leave the text box fails as VoiceOver announces that Safari is still busy.

zackkrida:
Closing in favor of switching to a static homepage that shows a randomized featured search.

Issue #659, opened 2022-01-20
Executing a search from the homepage using a screen reader is confusing


sarayourfriend:
## Description

Executing a search from the homepage using a screen reader leads to a confusing and frustrating experience.

## Reproduction

1. Using Safari and VoiceOver, visit the homepage and execute a search
2. Note that VO will announce that Safari is busy. If you try to interact with the page, you'll end up focusing the logo button.

## Expectation

We should automatically focus the logo button and change the window title to indicate the current search term. Figure out what is causing "safari is busy" and fix it.

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

sarayourfriend:
The easiest fix for this would be to add a "Skip to content" button similar to what's used in WordPress.org and many other sites.

PR #660, opened 2022-01-20
Fix homepage accessibility

### LINKED ISSUE #652: Animating the placeholder text on homepage is bad for accessibility

sarayourfriend:
## Fixes

Fixes #652 by @dhruvkb 

## Description

Switches to a static version of the homepage that randomly selects a set of featured images to show. To avoid creating any accessibility problems with these images, they're wrapped in `aria-hidden`.

This removes the animated placeholder text. There is an `aria-label` on the search input that indicates it is for searching.

`VInputField`'s `sr-only` label element was causing a weird "Safari is busy" bug with VoiceOver. Replacing it with an `aria-label` has the same effect as the `sr-only` label element but fixes this bug (I don't know why this bug is happening but we can dig into that a later date).

## Testing Instructions

Checkout this branch and run `pnpm dev`. Visit the site in Safari with VoiceOver and verify that you can successfully navigate the homepage and execute a search.

Note: there are other problems with moving from the homepage to the search results, but I've opened a separate issue to solve those here: https://github.com/WordPress/openverse-frontend/issues/659

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -60,20 +60,28 @@
         class="homepage-images flex flex-row gap-4 lg:gap-0 items-center lg:grid lg:grid-cols-2 lg:grid-rows-4 lg:w-[57.143vh] lg:h-[114.286vh]"
         aria-hidden
       >
-        <NuxtLink
-          v-for="(image, index) in featuredSearch.images"
-          :key="image.identifier"
-          :to="image.url"
-          class="homepage-image block aspect-square h-30 w-30 lg:h-auto lg:w-auto lg:m-[2vh] rounded-full"
-          :style="{ '--transition-index': `${index * 0.05}s` }"
-        >
-          <img
-            class="object-cover h-full w-full rounded-full"
-            :src="image.src"
-            :alt="image.title"
-            :title="image.title"
-          />
-        </NuxtLink>
+        <ClientOnly>
+          <Transition
+            v-for="(image, index) in featuredSearch.images"
+            :key="image.identifier"
+            name="fade"
+            mode="out-in"
+            appear
+          >
+            <NuxtLink
+              :to="image.url"
+              class="homepage-image block aspect-square h-30 w-30 lg:h-auto lg:w-auto lg:m-[2vh] rounded-full"
+              :style="{ '--transition-index': `${index * 0.05}s` }"
+            >
+              <img
+                class="object-cover h-full w-full rounded-full"
+                :src="image.src"
+                :alt="image.title"
+                :title="image.title"
+              />
+            </NuxtLink>
+          </Transition>
+        </ClientOnly>
       </div>
     </div>
 
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -23,6 +23,7 @@
       <VSearchBar
         v-model.trim="searchTerm"
         class="max-w-[40rem] mt-4 lg:mt-8"
+        :placeholder="$t('hero.search.placeholder')"
         @submit="handleSearch"
       >
         <ClientOnly>
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -142,7 +142,7 @@ const HomePage = {
       })),
     }))
 
-    const featuredSearchIdx = Math.floor((Math.random() * 10) % 3)
+    const featuredSearchIdx = Math.floor(Math.random() * 3)
     const featuredSearch = featuredSearches[featuredSearchIdx]
 
     const isMinScreenMd = isMinScreen('md', { shouldPassInSSR: true })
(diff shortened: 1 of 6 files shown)


zackkrida (approved):
Thanks for addressing this so quickly. Perhaps we can eventually make an improvement where we show the 'featured search' term somewhere else, not related to the input, and get these images cycling again in the short future.

src/pages/index.vue
```diff
@@ -23,16 +23,17 @@
...
         @submit="handleSearch"
       >
-        <VContentSwitcherPopover
-          v-if="isMounted && isMinScreenMd"
-          ref="contentSwitcher"
-          class="mx-3"
-          :active-item="contentType"
-          @select="setContentType"
-        />
+        <ClientOnly>
```
dhruvkb, line 28: It might be hard to understand why this is client only (because I don't get it). Do you think an explanatory comment `<!-- -->` here might be justified?

src/pages/index.vue
```diff
@@ -57,28 +58,22 @@
...
+        <NuxtLink
+          v-for="(image, index) in featuredSearch.images"
           :key="image.identifier"
-          name="fade"
-          mode="out-in"
-          appear
+          :to="image.url"
+          class="homepage-image block aspect-square h-30 w-30 lg:h-auto lg:w-auto lg:m-[2vh] rounded-full"
+          :style="{ '--transition-index': `${index * 0.05}s` }"
         >
```
dhruvkb, line 69: Should we preserve the one-by-one fade-in animation? That might still look quite nice on the static page?

src/pages/index.vue
```diff
@@ -154,61 +142,8 @@ const HomePage = {
...
-    }
-
-    const isMounted = ref(false)
-    onMounted(() => {
-      isMounted.value = true
-    })
-    onBeforeUnmount(() => {
-      isMounted.value = false
-    })
+    const featuredSearchIdx = Math.floor((Math.random() * 10) % 3)
```
dhruvkb, line 145: Nit: Since `Math.random()` is `[0, 1)`, multiplying by 3 gives `[0, 3)`, which can be floored to give `{0, 1, 2}` directly.
```suggestion
    const featuredSearchIdx = Math.floor(Math.random() * 3)
```

src/pages/index.vue
```diff
@@ -225,7 +160,7 @@ const HomePage = {
     const searchTerm = ref('')
     const handleSearch = async () => {
       await store.dispatch(`${SEARCH}/${UPDATE_QUERY}`, {
-        q: searchTerm.value || featuredSearches[featuredSearchIdx.value].term,
+        q: searchTerm.value,
```
dhruvkb, line 163: We would still need the fallback string when the user hits search without typing anything?

src/pages/index.vue
```diff
@@ -23,16 +23,17 @@
       <VSearchBar
         v-model.trim="searchTerm"
         class="max-w-[40rem] mt-4 lg:mt-8"
-        :placeholder="featuredSearchText"
```
dhruvkb, line 26: In the absence of animation, I suppose placing the featured search term as the placeholder is justified?

dhruvkb (changes requested):

src/pages/index.vue
sarayourfriend: I don't know why it is client only, I was actually hoping you did! `ClientOnly` just removes the extra logic that we previously had for the `isMounted` ref that was being maintained.

src/pages/index.vue
sarayourfriend: Placeholders get read by the screen reader. Without the visual context of the opposite side of the screen, it would be a completely context-less placeholder and the screen reader would read "Olympic games" and the person relying on a screen reader would have no idea why.

src/pages/index.vue
sarayourfriend: Sure thing.

panchovm:
Can we add a static placeholder that says `Search for content`? An input without label nor placeholder looks odd.

src/pages/index.vue
dhruvkb: 🤷

dhruvkb (approved):

Issue #662, opened 2022-01-21
All results boxed audio a11y


zackkrida:
## Description

From @sarayourfriend 

Audio results need to be composite items with the play/pause button removed from the default tab order. Currently tabbing through each result is fine until you hit an audio result which will tab from the result link into the play/pause button before moving on to the next result. The tab order should just be each result. This happens both with and without VoiceOver activated.

With VoiceOver, the audio results are not announcing their title. I think this would be fixed with the composite items as I believe they require `aria-label` which would announce the title.

Audio links are also missing an `aria-label` so all audio links just read "link, Audio player" in VoiceOver.

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

zackkrida:
@WordPress/openverse-frontend I'm not sure what the best approach is here—I feel like I'm over thinking it or missing something obvious. I think the _behavior_ we want is:

1. Can only focus the outermost node of the boxed audio, not internally
	- Is the best way to do this to remove the tabindex on play-pause when the layout is `box`?
2. Pressing space or clicking the play button directly will toggle the audio playback
3. Clicking anywhere except the play button or pressing enter on the audiobox will navigate to the single result

sarayourfriend:
I think that sounds right Zack.

> Is the best way to do this to remove the tabindex on play-pause when the layout is box?

Setting the tabindex to `-1` specifically would do it I think, yes.

> Pressing space or clicking the play button directly will toggle the audio playback

The only thing with this is that we'll have to indicate somehow to both screen reader users and folks who only rely on keyboard navigation otherwise that they can play the audio using the space bar. Under most circumstances the spacebar, when focused on a link, will scroll the page. Screen readers are probably the easier case here as we could incorporate it in the `aria-label` of the link (which is currently missing, I've updated the issue description to include that as a requirement) using a format like `` `${audio.title} - Press spacebar to play and pause the audio preview` ``.

For users who are only using keyboard navigation without a screen reader, we'll need to figure out another way to inform them about this behavior. Maybe we can show a popover above the audio result that indicates that the spacebar can be used the play/pause the audio preview, but only when we've detected that someone is using the keyboard to navigate. We'll also need to hide this information from screen reader users to avoid duplicating the information. In this case we'll need to set the `autoFocusOnShow` prop to `false` on the popover.

PR #681, opened 2022-01-21
Notices for Jamendo tracks when seeking is impossible

### LINKED ISSUE #680: Jamendo tracks can't be seeked in Blink browsers
## Description

Jamendo does not send an `Accept-Ranges:` header with its audio results, which makes seeking not work in Blink-based browsers like Edge, Chrome, and Opera. This PR adds a notice for these browsers when they attempt to interact with an audio waveform.

We should add a notice when seeking is attempted on one of these tracks.

## Environment

 - Browser: Any blink/chromium based browser

## Resolution

zackkrida:
## Fixes

Fixes #680 by @zackkrida 

## Description

Adds a `featureNotices` feature to the audio track so custom messages can be displayed when any given feature isn't active. It's a bit of an extra abstraction but felt cleaner and simpler than writing some specific hack for this functionality.

## a11y info

When the waveform is disabled, it is no longer focusable, so when users tab from the play button they are sent straight to the next play button. This means they are not explicitly informed about the disabled seek bar, but they're also not incorrectly able to focus the waveform or anything worse like that. I'm not sure how to best approach the _ideal_ solution to this.

## Caveats

I couldn't get our UA detection to run server-only, which I would have liked. Perhaps this could be refactored in the future to add the parsed UA information to a Vuex store instead of injecting it into the Nuxt context, which would preserve the data on client-side renders after the initial server render, but alas, I tried that and wasn't able to get it to work either. For whatever reason it seemed my mutation wouldn't run.

## Testing Instructions

Visit an audio results page in Chrome or Edge and observe the hover effect on a Jamendo track.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VAudioTrack/VWaveform.vue
+++ b/src/components/VAudioTrack/VWaveform.vue
@@ -149,6 +149,14 @@
     >
       {{ message }}
     </div>
+
+    <!-- Seek disabled message overlay -->
+    <div
+      v-if="seekDisabledNotice"
+      class="invisible group-hover:visible group-focus:visible absolute w-full inset-0 flex items-center justify-center font-bold text-xsm bg-yellow/75 z-40"
+    >
+      {{ seekDisabledNotice }}
+    </div>
   </div>
 </template>
 
@@ -215,6 +223,14 @@ export default defineComponent({
     features: {
       type: Array,
       default: () => ['timestamps', 'seek'],
+    },
+    /**
+     * An object of notices to display when a feature is disabled.
+     * `'timestamp'`, `'duration'`, `'seek'`.
+     */
+    featureNotices: {
+      type: Object,
+      default: () => {},
     },
   },
   emits: [
@@ -296,6 +312,9 @@ export default defineComponent({
     const showDuration = computed(() => props.features.includes('duration'))
     const showTimestamps = computed(() => props.features.includes('timestamps'))
     const isSeekable = computed(() => props.features.includes('seek'))
+
+    /* Feature notices */
+    const seekDisabledNotice = computed(() => props.featureNotices?.seek)
 
     /* State */
 
@@ -486,6 +505,8 @@ export default defineComponent({
       showDuration,
       showTimestamps,
       isSeekable,
+
+      seekDisabledNotice,
 
       isReady,
       isInteractive,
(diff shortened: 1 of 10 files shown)


src/components/VAudioTrack/VWaveform.vue
```diff
@@ -149,6 +149,14 @@
...
       {{ message }}
     </div>
+
+    <!-- Seek disabled message overlay -->
+    <div
+      v-if="seekDisabledNotice"
+      class="invisible group-hover:visible group-focus:visible absolute w-full inset-0 flex items-center justify-center font-bold text-xsm bg-yellow/75 z-40"
+    >
+      {{ seekDisabledNotice }}
+    </div>
```
dhruvkb, line 159: Imo we use the existing `message` field, with a prop that allows it to be transient. I doubt the two will ever need to overlap.

dhruvkb (approved):
This is huge good, appreciate the features being popped with notices! Approving with a non-blocking suggestion.

src/components/VAudioTrack/VWaveform.vue
zackkrida: At a minimum they could definitely share some styles; good suggestion.

sarayourfriend:
@zackkrida I like this solution but just letting you know I'm going to take some extra time to review this on using JAWS on a Windows VM to make sure that it's a pleasant experience for screen reader users.

FWIW I think just skipping the seeking _might_ be okay for now? I considered suggesting that we added some kind of extra information into the label on the play/pause button to indicate that seeking would be disabled for that particular audio track, but I want to be wary of the fact that if you narrowed a search to just jamendo tracks then every single play/pause button would include this extra information and could be frustrating for the user.

zackkrida:
> I'm going to take some extra time to review this on using JAWS on a Windows VM 

@sarayourfriend please do, thank you 😄 Keep that VM hot for #662 later too 😆

panchovm:
I added a few mockups on the [Releases](https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=133%3A6869) file showing the alert message above grid results and in the wave area on both [desktop](https://cln.sh/stqwng) and [mobile](https://cln.sh/sNkPG0). The message is primarily static on mobile, yet it shows the playing state through the yellow background.

The `limited to Firefox and Safari` sounds intentional instead of a condition openverse is dealing with.

sarayourfriend:
Oh that's something I hadn't considered, how this will appear to Android Chrome users.

PR #695, opened 2022-01-24
Allow focusing specific element when opening modal

### LINKED ISSUE #641: Incorrect element is focused when opening mobile content switcher via keyboard
## Description

The openverse logo button is focused when opening the mobile content switcher via keyboard instead of the content types themselves.

Marking this as high because although for sighted keyboard users who don't rely on a screen-reader this won't be too big of an issue (more of an annoyance than something breaking the usability of the site), it does cause a loss of context for users who depend on screen-readers.

## Reproduction

1. Run `main` and shrink your window to a small enough width to trigger the mobile content switcher modal.
2. Using your keyboard, tab into the content switcher button in the header and press enter to open it.
3. Note that the openverse logo is focused instead of the content type buttons.

## Expectation

When opening the content switcher, the active content button should be focused.

This may require adding back some props that were eventually removed from the original `VModal` component for passing a specific node to focus. Currently the `useDialogContent` composable just focuses the first focusable item inside the dialog rather than allowing you to pass a specific node to focus. This works for most cases but breaks in cases like this.

## Resolution

sarayourfriend:
## Fixes

Fixes #641 by @sarayourfriend

## Description

In order to fix the issue where opening the mobile content type switcher was focusing the wrong element, we needed to update the modal to allow for this behavior to specify the element we want to focus when the content is shown.

Even though we only needed this for the mobile menu content modal, I went ahead and implemented it in the default modal component and added a story control to demonstrate it working in the story.

## Testing Instructions

Checkout the branch and run `pnpm dev`. Visit the search page with a narrow width viewport and open the content type switcher using the keyboard. It should focus the active content type button. Be sure to try different content types being selected. It should always focus the active one.

For good measure, run `pnpm storybook` and visit the modal story http://localhost:6006/?path=/story/components-vmodal--default&args=useCustomInitialFocus:true with `useCustomInitialFocus` control set to `true`. Ensure that it focuses the button in the middle of the modal content instead of the default of the first tabbable element (in this case the close button). Toggle the control to `false` and ensure that the close button is focused on that case.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VContentSwitcher/VMobileMenuModal.vue
+++ b/src/components/VContentSwitcher/VMobileMenuModal.vue
@@ -15,9 +15,11 @@
       :trigger-element="triggerRef"
       :hide="close"
       :aria-label="$t('header.filter-button.simple')"
+      :initial-focus-element="initialFocusElement"
     >
       <nav class="p-6" aria-labelledby="content-switcher-heading">
         <VContentTypes
+          ref="contentTypesNode"
           size="small"
           :active-item="content.activeType.value"
           @select="selectItem"
@@ -29,7 +31,13 @@
 </template>
 
 <script>
-import { onMounted, reactive, ref, watch } from '@nuxtjs/composition-api'
+import {
+  onMounted,
+  reactive,
+  ref,
+  watch,
+  computed,
+} from '@nuxtjs/composition-api'
 import { useBodyScrollLock } from '~/composables/use-body-scroll-lock'
 import useContentType from '~/composables/use-content-type'
 import usePages from '~/composables/use-pages'
@@ -61,6 +69,8 @@ export default {
     const content = useContentType()
     const pages = usePages()
 
+    /** @type {import('@nuxtjs/composition-api').Ref<import('vue/types/vue').Vue | null>} */
+    const contentTypesNode = ref(null)
     const modalRef = ref(null)
     const triggerContainerRef = ref(null)
 
@@ -76,6 +86,10 @@ export default {
 
     const triggerRef = ref()
     onMounted(() => (triggerRef.value = triggerContainerRef.value?.firstChild))
+
+    const initialFocusElement = computed(() =>
+      contentTypesNode.value?.$el?.querySelector('[aria-checked="true"]')
+    )
 
     watch([visibleRef], ([visible]) => {
       triggerA11yProps['aria-expanded'] = visible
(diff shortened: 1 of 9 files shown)


src/components/VContentSwitcher/VMobileMenuModal.vue
```diff
@@ -77,6 +87,10 @@ export default {
     const triggerRef = ref()
     onMounted(() => (triggerRef.value = triggerContainerRef.value?.firstChild))
 
+    const initialFocusElement = computed(() =>
+      contentTypesNode.value?.$el?.querySelector('[aria-checked="true"]')
```
obulat, line 91: Nice way of getting the element using `aria-checked`!

obulat (commented):
This is not the full focus of this PR, but I tried testing tabbing through the screen setting the width to 350px. Well, it's probably not used very often, but it is a mess :(
On Chrome it works well, you can open the mobile modal and tab from content types to the pages, and then to Openverse and `close` button.
On Firefox, you can only tab to the 'Content type item'. Tabbing again you leave the content of the browser, going to the browser controls (`back`, `forward`, `reload` etc). Openverse logo and close button are inaccessible.
On Safari, when you tab away from the search bar, you get to the audio boxes 🙈 And you can only access audio boxes (no images!) and the search box through tabbing.

obulat (approved):
This PR works well in Chrome. In the content switcher modal, the initially focused element is the selected content type. In the filters modal, not touched by this PR, it's the Openverse logo (and then, `close` button and filter checkboxes). 
However, in Firefox only the 'focusing of the selected content type' works well. 
After focusing on the selected type, you can only use arrow keys to select the content type, and cannot tab to the pages menu, or the Openverse logo or the close button. In the Filters modal, by the way, the tabbing order is: Openverse logo, close button, the body of the modal (scrolling element below the modal header), filters inputs.
Safari is even worse, I couldn't open the modals using keyboard.

obulat:
> Do you happen to have the keyboard navigation setting disabled in macOS? You can [follow these instructions](https://bugzilla.mozilla.org/show_bug.cgi?id=1587962#c0) to enable it and see if that fixes the issue with Firefox.

Thank you! I did turn on a different keyboard navigation setting in Firefox using the touchbar <kbd>F7</kbd> (the one on the external keyboard didn't work for some reason), and now I can navigate using <kbd>Tab</kbd>.
I didn't realize that Safari was focusing, just not showing it due to `focus-visible` problem!

zackkrida (approved):
This makes the modal much more robust. Nice work

PR #701, opened 2022-01-25
Switch to path-based i18n routing

### LINKED ISSUE #700: Translations aren't shown on localized homepages
## Description

If you visit a localized site, the homepage isn't translated even though translations exist.

## Reproduction

1. Visit https://es.wordpress.org/openverse
2. Observe the page is in English

## Screenshots

CleanShot 2022-01-24 at 22 26 42@2x

## Additional context

The current cookie-based translation mechanism is also the only response cookie sent by Openverse. We could eliminate this cookie based approach and switch to a path approach, which would _also_ enable more aggressive caching on the frontend, as we could cache each localized page without worrying about, for example, translations in one language polluting the cache. 

## Resolution

zackkrida:
## Fixes

Fixes #700 by @zackkrida 

## Description

This PR switches Openverse to use the url path instead of a cookie for setting translations. This has a few benefits:

1. Don't have to worry about cookie values being set correctly when switching from one locale to another
2. Makes it easy for developers to test different locales, just modify the url path
3. Allows for much more aggressive front-end caching (currently the site is 'dynamic' in the sense that the same page with the same url should show different contents based on a cookie). This makes each page infinitely cacheable by url.
4. Fixes the issue in #700 

To use this PR, we need an accompanying PR in the Openverse WordPress theme that prefixes the iframe url with the correct locale string. 

## Testing Instructions

1. Run the app locally
2. Try different locale-based paths like `localhost:8443/es`, `localhost:8443/ru` and so on.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


nuxt.config.js
```diff
@@ -172,20 +172,12 @@ export default {
...
      * */
-    detectBrowserLanguage: {
-      useCookie: true,
-      cookieKey: 'i18n_redirected',
-      alwaysRedirect: true,
-      cookieCrossOrigin: true,
-      cookieSecure: true,
-    },
-    baseUrl: 'http://localhost:8443',
+    detectBrowserLanguage: false,
```
zackkrida, line 180: Now we'd only look at the path supplied to the iframe, no other communication necessary.

dhruvkb:
Wouldn't exposing translations in the URL allow invalid combinations like https://ru.wordpress.org/openverse/es?

obulat (changes requested):
Works great, and will prevent the problems with mismatch with the outer frame we sometimes have (Openverse in French with outer frame in English).
I've scanned the repo for paths, and there is one place where we need to change path to `localePath`:
https://github.com/WordPress/openverse-frontend/blob/0f784f4df9fdf4b460dab387f0877229420efebc/src/components/VAllResultsGrid/VAllResultsGrid.vue#L10

I don't think this change should affect the `urlChange` message exchange with the outer frame in `sendWindowMessage`, but it would be nice to check.

obulat:
> Wouldn't having translations in the URL allow entering invalid combinations like https://ru.wordpress.org/openverse/es?

I guess we could add a redirect to URLs like this. Is it possible to detect the mismatch between the `ru` and `es` locales in the URL `https://ru.wordpress.org/openverse/es`, and then set the iframe locale to `ru`, in this case?

zackkrida:
@dhruvkb it what scenario would a url like ru.wordpress.org/openverse/es be produced? What would be adding the `/es` to the end of the path?

dhruvkb:
It wouldn't be occurring naturally, that's just my imagination of someone trying to mess with the system by intentionally entering such a URL into the searchbar.

zackkrida:
Got it! In this case, I think it would just 404 and that's okay.

zackkrida:
Thanks @obulat I also fixed the 'back to search results' links; I would've missed that without your suggestion!

obulat (approved):

zackkrida:
Oh also @dhruvkb, only folks using the non-english site who would know to inspect the iframe would even think to look at the url there, notice we're using a base path for locales, and then try to manipulate, where it would only effect their local browsing experience.

nuxt.config.js
sarayourfriend: @zackkrida can you update the comment for this variable? If it's still critical, we should describe why and how, especially now that it's simply disabled.

src/composables/use-match-routes.js
```diff
@@ -1,17 +1,26 @@
...
 export const useMatchRoute = (routes = []) => {
+  const { app } = useContext()
   const route = useRoute()
   const router = useRouter()
-  const matches = ref(routes.includes(route.value.name))
+  const localizedRoutes = routes.map(
+    (route) => app.localeRoute({ name: route }).name
+  )
+  const matches = ref(localizedRoutes.includes(route.value.name))
+  console.log(route.value.name)
```
sarayourfriend, line 21:
```suggestion
```

src/pages/image/_id.vue
```diff
@@ -61,7 +61,10 @@ const PhotoDetailPage = {
   },
   beforeRouteEnter(to, from, nextPage) {
     nextPage((_this) => {
-      if (from.path === '/search/' || from.path === '/search/image') {
+      if (
+        from.name === _this.localeRoute({ path: '/search/' }).name ||
+        _this.localeRoute({ path: '/search/image' }).name
+      ) {
```
sarayourfriend, line 67:
```suggestion
      if (
        from.name === _this.localeRoute({ path: '/search/' }).name ||
        from.name === _this.localeRoute({ path: '/search/image' }).name
      ) {
```

sarayourfriend (changes requested):
Just a couple of things/minor logic bugs.

Otherwise this LGTM. We'll still need one more fix to the homepage to fully fix the internationalization of it as #765 still needs to be addressed.

nuxt.config.js
zackkrida: Thanks for the suggestion, I updated the comment with more useful information!

sarayourfriend (approved):
LGTM! Thanks for the fixes 🎉

PR #707, opened 2022-01-25
Refactor the usage of i18n result count computation

### LINKED ISSUE #706: Incorrect text on the content links
## Description

Currently, the content links on the 'All' view both show the text for 'N results' instead of using 'N image results' for image link and 'N audio results' for the audio link:

Screen Shot 2022-01-25 at 3 07 48 PM

## Expectation

The content link should have a media-specific label (`image results` and `audio results`)

## Additional context

This is caused by using the default argument (`ALL_MEDIA`) when computing the label, and not passing the actual media type.

## Resolution

obulat:
## Fixes

Fixes #706 by @obulat

## Description

This PR refactors the function creating results count string in the `use-i18n-utilities` to use the static keys to ensure that the `po` files get the correct line number. It now also returns a blank string for an unsupported media type to make it more robust when we will add Video as `additional` media type.

This function is reused in the `VHeader`, and the i18n result computation is removed from the `VSearchGrid` which doesn't use the result string anymore. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb (approved):
I'm a huge fan of the `it.each` test tables. Great refactor!

test/unit/specs/composables/use-i18n-utilities-in-component.spec.js
```diff
@@ -0,0 +1,42 @@
...
+      mocks: { $nuxt: { context: { i18n } } },
+    }
+  })
+
+  it.each`
+    resultCount | mediaType | expectedText
+    ${0}        | ${IMAGE}  | ${'No image results'}
+    ${0}        | ${AUDIO}  | ${'No audio results'}
+    ${10}       | ${IMAGE}  | ${'10 image results'}
+    ${10}       | ${AUDIO}  | ${'10 audio results'}
```
sarayourfriend, line 28: Should we add `1` here as well given it usually creates a distinct branch for plurals in English?

```suggestion
    ${0}        | ${IMAGE}  | ${'No image results'}
    ${0}        | ${AUDIO}  | ${'No audio results'}
    ${1}        | ${IMAGE}  | ${'1 image result'}
    $10}        | ${AUDIO}  | ${'1 audio result'}
    ${10}       | ${IMAGE}  | ${'10 image results'}
    ${10}       | ${AUDIO}  | ${'10 audio results'}
```

zackkrida (changes requested):
The refactor looks great! But I don't think we actually want the content link changes, see more context [here](https://github.com/WordPress/openverse-frontend/issues/706#issuecomment-1021126439)

test/unit/specs/composables/use-i18n-utilities-in-component.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    ${10000}    | ${IMAGE}  | ${'Over 10,000 image results'}
+    ${10000}    | ${AUDIO}  | ${'Over 10,000 audio results'}
+  `(
+    'Should show correct result for $resultCount of type $mediaType',
```
sarayourfriend, line 32: Thanks for adding these tests! How do they differ from the other tests you added that test the composable in isolation? They seem to test the same things, just in different contexts. With this test in particular it would break if `VContentLink` stopped displaying the count for any reason, which seems weird.

If we need to test it in a component, then a test specific component created in this file might be a better option to avoid that kind of thing.

sarayourfriend (approved):
Nice refactor! Thank you 😁 

Just one non-blocking question/recommendation.

test/unit/specs/composables/use-i18n-utilities-in-component.spec.js
obulat: I guess after the comment on the original issue that we don't actually need the 'image/audio results' string in the `VContentLink`, I will simply remove it. I used this test to use the nuxt `app.i18n.tc` that is not available outside of a component, and using a component defined especially for test for testing Nuxt-specific things can be tricky.

sarayourfriend: Ahh gotcha, that makes sense!

obulat:
I fixed the composable utility function (got the `can only use this function in a setup function` error when using the `i18nResultCount` function, because it is also used in the watcher in the `VHeader` component).
Also, used the English number display style for RTL languages.

src/components/VContentLink/VContentLink.vue
```diff
@@ -17,17 +17,18 @@
...
 <script>
 import { computed, defineComponent } from '@nuxtjs/composition-api'
-import { resultsCount } from '~/composables/use-i18n-utilities'
-import {
-  AUDIO,
-  IMAGE,
-  supportedMediaTypes as mediaTypes,
-} from '~/constants/media'
+import { useI18nResultsCount } from '~/composables/use-i18n-utilities'
+import { AUDIO, IMAGE, supportedContentTypes } from '~/constants/media'
```
obulat, line 21:
```suggestion
import { AUDIO, IMAGE, supportedMediaTypes } from '~/constants/media'
```

src/components/VContentLink/VContentLink.vue
```diff
@@ -38,7 +39,7 @@ export default defineComponent({
     mediaType: {
       type: String,
       required: true,
-      validator: (val) => mediaTypes.includes(val),
+      validator: (val) => supportedContentTypes.includes(val),
```
obulat, line 42:
```suggestion
      validator: (val) => supportedMediaTypes.includes(val),
```

krysal (approved):
Nice refactor!

sarayourfriend:
This also fixes https://github.com/WordPress/openverse-frontend/issues/726 I think?

Issue #687, opened 2022-01-23
Header button small in mobile view


dhruvkb:
## Description

The homepage link in the header has reduced spacing around the logo on smaller screen sizes. Even more problematic is the complete lack of padding on the right side.

## Reproduction

0. Run the dev server on `main`.
1. Switch to a smaller screen size and open the header.
2. Hover over the logo to show the yellow background.

## Expectation

The logo padding should be the same across screens, or at the very least consistent on all four sides.

## Screenshots

![image](https://user-images.githubusercontent.com/16580576/150670318-be1e321c-16b6-4037-92da-ced862638709.png)

Contrast this with the space around the logo on larger screens.

![image](https://user-images.githubusercontent.com/16580576/150670340-b8649b99-31c7-4dbd-b930-1a9d25b816de.png)

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

PR #718, opened 2022-01-26
Fix header items not fitting in

### LINKED ISSUE #711: Mobile header spans multiple lines
## Description

The mobile header is meant to be a single row; on small devices it spans two lines.

## Reproduction

1. Visit https://wordpress.org/openverse/search?q=pumpkins on a small mobile device; alternatively resize your browser window as small as possible.

## Screenshots

> header occupies two lines, the filter button is pushed downward
signal-2022-01-25-091151_002

## Environment

 - Device: Tested in Firefox 96.2.0 on a Samsung Galaxy A12 

## Resolution

obulat:
## Fixes

Fixes #711 by @zackkirda

## Description

Currently, the mobile header uses `flex-wrap` to wrap the elements and push the search bar to the second row. This prevents us from using the normal `flex` mechanism to ensure that the other row fits into the screen.
This PR sets the width of the search bar using the `CSS` calc property to calculate the width that's left after buttons and paddings. It also removes the margins from the header menu buttons and handles the spacing from the parent (header) using `gap` property.
Screen Shot 2022-01-26 at 9 08 05 AM

This works fine until the width of 379px. Below that, because the width of the content switcher is hard-coded to 5rem (90px), there is not enough width in the header to fit all components, so the filter button goes to the second line:

Screen Shot 2022-01-26 at 9 08 20 AM

Another problem I have is with the inner pages on smaller widths, as there is no Figma mockup showing the header on the inner pages. This is what it currently looks like:
Screen Shot 2022-01-26 at 9 40 45 AM

@panchovm, I have 2 questions:
1. What should I do for the widths lower than 394px?
2. Does the inner page header look OK?

## Testing Instructions

Run the app, and try decreasing the screen size down to 350px (or lower?). Try different pages, and with the header scrolled/ not scrolled. All element should fit into one row, or two rows on a search page when the header hadn't been scrolled.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


panchovm:
Here my comments

## Header with content switcher
### Content switcher width
The label in content switcher has a `5 rem` width. It should be `auto` as the filter button keeps fixed and content switcher changes its width.

CleanShot 2022-01-26 at 10 41 27@2x

### Content switcher padding
Content switcher should have a padding-right of `12px` too look balanced compared to left padding where icon is.

CleanShot 2022-01-26 at 10 47 24@2x

### Filter button size
Filter button is not square and it should be square. When scrolling down it goes as square, but `48px` instead of `40px` as the mockup. Buttons size match the search bar height and we might need to review in the case that values do not have the visual effect expected. I'm open to review this from the design.

CleanShot 2022-01-26 at 10 42 52@2x

It seems that header does not fit well in smaller viewports, even with the above changes applied. I am testing `320px` as some devices are and actions move to a second row. I suggest that we review the header again once changes are done and see if hiding `openverse` label works well. 

## Header without content switcher (internal pages)

We do have a header for internal pages where content switcher is not visible. Go to see the [page version of modal](https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=1685%3A20886). Adding a screenshot.

CleanShot 2022-01-26 at 10 51 53@2x

obulat:
@panchovm , I think I've fixed the problems you listed :) Now, I need to adjust the header for under 370px...

zackkrida:
@obulat @panchovm Maybe we should hide the Openverse text on very small (less than around 350px seems to work for me) screens. That seems like an easy fix that looks okay:

CleanShot 2022-01-27 at 07 34 03@2x

panchovm:
I noticed the following:

### Buttons alignment
Based on @zackkrida screenshot and what I tested, the content switcher and filters should be fixed on the right side with a `8px` of spacing between **Content switcher** and **Filters** buttons. This to prevent uneven spaces between components on bigger displays. Here is how it should behave.



And agree with Zack about hiding Openverse text on small screens.

### Spacing between buttons and search bar

The space between buttons and search bar seems minor than the mockup. It should be `16px`. Adding a screenshot



### Logo position
When opening the menu, logo position changes and it shakes slightly. This might not be related to what this PR solves, feel free to dismiss the comment and I can create an issue.



### Keyboard navigation
I think @sarayourfriend commented this days ago, when navigating with the keyboard, the order does not follow the visual LTR logic. Instead of jumping from logo to content switcher, it goes to the search bar.

sarayourfriend:
> The label in content switcher has a 5 rem width. It should be auto as the filter button keeps fixed and content switcher changes its width.

This also changes from language to language by the way, and we shouldn't bank on it being 5 rem. In Portguese for example it's four full-width characters longer and visually ~40% longer.

> All content
> Todo o conteúdo

That might not be the translation we end up using the end, but just something to keep in mind that we can't rely on visible text labels to be a specific width.

Removing the Openverse word from the logo button gives back a lot of space, which is great 👍

zackkrida:
To @sarayourfriend's point we should likely add the `truncate` class to content switcher.

obulat:
Done:
- [x] Buttons alignment: buttons are aligned to the right in the below-md screens. Only on screens that are below 370px is the content switcher center-aligned.
- [x] Openverse text is hidden on screens smaller than 370px.
- [x] Spacing between the first and the second row on mobile header set to 16px.
- [x] Logo and Openverse text do not change position when opening mobile menu; I extracted the component combining `VLogoLoader` and `OpenverseText` to make this easier, and called it `VLogo`, please help me come up with a better name :)

Not done:
- [ ] Keyboard navigation. Unfortunately, fixing this would require more thought and effort. I was thinking that we could move all the complexity to the components of VHeader, and have the header as a simple `render function` component without a template that would place the SearchBar either second or last based on `isHeaderScrolled` value.
- [ ] Change the content switcher static width of 20rem. It is only set on desktop screens, and I believe @dhruvkb set it so that the popup doesn't jump either horizontally or vertically when scrolled. I also had thought about i18n problems with it, but what is a better solution for popover shifting?

zackkrida:
The extraction of the logo and text you added is relevant to @sarayourfriend in #687, cool!

sarayourfriend:
>  Keyboard navigation. Unfortunately, fixing this would require more thought and effort. I was thinking that we could move all the complexity to the components of VHeader, and have the header as a simple render function component without a template that would place the SearchBar either second or last based on isHeaderScrolled value.

Olga, I would suggest opening a separate issue for this. It will mainly affect sighted keyboard users on mobile (an extremely small cohort) who will be temporarily confused by the unexpected tab order... but it's likely to be something they've experienced on other websites before.

For folks relying on a screen reader, they won't notice the bad order because in all likelihood they're not relying on the visual context of the page to know what to expect from their reader. I'm not saying it's impossible, just saying it's unlikely.

Given the component is broadly accessible (and we've fixed the focus mangement issues with the modals) I'm less concerned about the wonky tab order being addressed in this PR and holding up the other fixes that will have a broader positive impact once merged.

zackkrida:
I made a small PR for the content switcher truncation issue: https://github.com/WordPress/openverse-frontend/pull/738

panchovm:
Here what I noticed:

The content switcher should never be center-aligned, both buttons are always fixed on the right side. Sorry for not being clear enough about this before.

The non-scrolled version of content switcher looks good, but it seems that it broke the scrolled version. The previous square version now has more padding on the right side.

CleanShot 2022-01-27 at 15 20 27@2x

obulat:
> The content switcher should never be center-aligned, both buttons are always fixed on the right side. Sorry for not being clear enough about this before.
> 
> The non-scrolled version of content switcher looks good, but it seems that it broke the scrolled version. The previous square version now has more padding on the right side.

Both issues fixed, thank you, @panchovm !

src/components/VHeader/VHeader.vue
```diff
@@ -1,44 +1,30 @@
 <template>
   <header
-    class="flex py-4 px-4 md:px-7 items-stretch z-40 w-full bg-white md:gap-x-2 gap-y-4"
+    class="flex px-4 md:px-7 items-stretch z-40 w-full bg-white gap-x-2 gap-y-4 flex-wrap md:flex-nowrap"
```
sarayourfriend, line 3:
```suggestion
    class="flex px-4 md:px-7 items-stretch z-40 w-full bg-white gap-x-2 gap-y-4 flex-wrap md:flex-nowrap py-3"
```

src/components/VHeader/VHeader.vue
```diff
@@ -1,44 +1,30 @@
...
     :class="{
-      'flex-wrap': isSearchRoute,
+      'py-3 md:py-4': !isHeaderScrolled,
+      'py-3': isHeaderScrolled,
```
sarayourfriend, line 6: `py-3` as being applied either way, so we can move it to the static `class` list.

```suggestion
      'md:py-4': !isHeaderScrolled,
```

src/components/VInputField/VInputField.vue
```diff
@@ -1,11 +1,12 @@
 <template>
   <div
-    class="input-field group flex flex-row items-center hover:bg-dark-charcoal-06 focus-within:bg-dark-charcoal-06 group-hover:bg-dark-charcoal-06 h-12 p-0.5px focus-within:p-0 border focus-within:border-1.5 border-dark-charcoal-20 rounded-sm overflow-hidden focus-within:border-pink"
+    class="input-field group flex flex-row items-center hover:bg-dark-charcoal-06 focus-within:bg-dark-charcoal-06 group-hover:bg-dark-charcoal-06 p-0.5px focus-within:p-0 border focus-within:border-1.5 border-dark-charcoal-20 rounded-sm overflow-hidden focus-within:border-pink"
     :class="[
       {
         // Padding is set to 1.5px to accommodate the border that will appear later.
         'border-s-0 ps-1.5px rounded-s-none': connectionSides.includes('start'),
         'border-e-0 pe-1.5px rounded-e-none': connectionSides.includes('end'),
+        [sizeClass]: true,
```
sarayourfriend, line 9: I learned reviewing one of @dhruvkb's PRs the other day that you can mix the array and object class binding syntax. To avoid having to do `[sizeClass]: true` you can just put it into the array next to the object instead of inside the object.

sarayourfriend (approved):
LGTM! Just some non-blocking minor de-duplication things, but the logo button looks like it works fine for me and the header items seem to fit now 🙂

src/components/VHeader/VHeader.vue
obulat: That was actually an incorrect implementation, thank you for catching it! It should be py-4 if the header isn't scrolled, and py-3 if it is, no matter the screen size.

src/components/VHeader/VLogoButton.vue
```diff
@@ -0,0 +1,51 @@
...
+  },
+})
+export default VLogoButton
+</script>
+<style scoped>
+@media (max-width: 370px) {
+  .logo-text {
+    display: none;
+  }
+}
```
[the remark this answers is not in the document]
obulat: We could add one more tailwind screen size for that, something like `xs`. Should I make an issue for it, @sarayourfriend ?

src/components/VInputField/VInputField.vue
obulat: This is nice, Thank you, @sarayourfriend !

PR #724, opened 2022-01-26
Fix focus-visible for Safari

### LINKED ISSUE #723: :focus-visible selector not supported in Safari but used across almost all of our focusable elements
## Description

[`:focus-visible` is unsupported in any current version of Safari](https://caniuse.com/?search=focus-visible)

## Reproduction

1. Open non tech-preview Safari (the current version or an iPad or iPhone safari)
2. Focus the header buttons using the keyboard
3. Note the lack of focus styles on any button

## Expectation

Buttons should have focus styles in all browsers

## Environment

 - Browser: Safari

## Additional context

Found while testing https://github.com/WordPress/openverse-frontend/pull/695

## Resolution

sarayourfriend:
## Fixes

Fixes #723  by @sarayourfriend 

## Description

Fixes focus-visible for Safari using a polyfill and a postcss plugin.

## Testing Instructions

Checkout this branch and run it in Safari and verify that focus rings appear on elements that need them when focused via the keyboard (like buttons in the header).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/layouts/blank.vue
```diff
@@ -1,3 +1,9 @@
 <template>
   <Nuxt />
 </template>
+
+<script>
+import '~/utils/safe-focus-visible-polyfill'
```
sarayourfriend, line 6: Is there a better way to import this into all layouts client-side only? I tried using a setting in `head` object in nuxt-config but I wasn't sure how to get the polyfill file referenced properly 🤔 Would we need to copy it out of `node_modules` during the build step or something into the static folder so it would get copied into the build output and we could reference it in a script element in the head?

src/utils/safe-focus-visible-polyfill.js
```diff
@@ -0,0 +1 @@
+if (process.client) require('focus-visible')
```
sarayourfriend, line 1: `focus-visible` breaks SSR so we need to guard it to only be imported on the client.

dhruvkb (commented):
I'm not well versed with Nuxt config but could we add this somewhere globally instead of importing in every layout?

sarayourfriend:
@dhruvkb Is there a global root file for Nuxt? I couldn't really find one that is accessible to edit, that's why I placed it in the layouts 🤔

dhruvkb:
There has to be some level above the layouts. I'll dig through the docs.

dhruvkb:
If we can do that, it'll be great. But considering the a11y implications of completely forgetting about a browser, we could also merge this as a quickfix for now.

sarayourfriend:
Well, either way we'll have to deploy the fix which probably won't happen for another couple hours anyway.

@zackkrida can you see any reason not to use a CDN version of the polyfill (provided we include all the appropriate integrity checks?)

We could use https://unpkg.com/focus-visible@5.2.0/dist/focus-visible.js?meta as we already rely on Cloudflare anyway.

zackkrida:
I'm comfortable with unpkg + integrity checks, I don't see any issue there!

sarayourfriend:
Went the plugin route that @obulat recommended. It's much more sensible!

zackkrida (approved):
This is working well on desktop Safari. A good way for others to test this is comparing the hover state of the homepage content type chooser on localhost vs. https://wordpress.org/openverse/

dhruvkb (approved):

PR #727, opened 2022-01-26
Style links globally

krysal:
## Description

This small change will save us from copying the styles for links several times in every place where one appears (I was doing this in the single result page for images). The default color is pink and they should have an underline on hover state.

## Testing Instructions

Run the project, everything should look the same. Writing a new link should have the default styles without requiring to do anything else.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VAllResultsGrid/VAllResultsGrid.vue
+++ b/src/components/VAllResultsGrid/VAllResultsGrid.vue
@@ -8,7 +8,7 @@
         v-for="[key, item] in results"
         :key="key"
         :to="{ path: `/search/${key}`, query: $route.query }"
-        class="lg:col-span-2 focus:bg-white focus:border-tx focus:ring focus:ring-pink focus:outline-none focus:shadow-ring focus:text-black rounded-sm"
+        class="lg:col-span-2 hover:no-underline"
       >
         <VContentLink :results-count="item.count" :media-type="key" />
       </NuxtLink>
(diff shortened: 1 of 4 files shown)


zackkrida (approved):
I love this idea! I'm wondering if there's a 'safer' way to implement it. For example, one idea would be a `<TextLink>` component that wraps a `<NuxtLink>` and applies the default text link styles. 

Despite that, I think your approach works well because it ensures plain links are accessible against our standard text color by default 👍

obulat (approved):
I like how much cleaner the class lists look now!
This approach might be a good way to get rid of the scss styles in the future, too.

dhruvkb (commented):
Instead of styling `a` directly on the base, could we create a new Tailwind component like so
```css
@layer components {
  .text-link {
    @apply text-pink hover:underline;
  }
}
```

so that it only applies when we explicitly call for it to be applied?

In my experience, it's easier to apply classes when needed rather than unapply them when they're applied by default.

But yeah, I do appreciate that links are no-longer colored the same as regular text.

krysal:
I precisely wanted to forget about repeating classes for normal links 😆 I pondered whether to create a new component or not but seemed not worth it for me. This is just a change in styles that certainly should be global. One can forget to use the customized link component, go directly with an `a` element and forgot to apply the styles. 

Although it can be a good idea if we add the function to vary between an internal and external link if Openverse continues to exist in an iframe. If this change turns out to be more problematic in the future I'll happily create the component.

PR #738, opened 2022-01-27
Truncate text in content switcher button, allow width to change

zackkrida:
## Description

This PR allows the content switcher to get _smaller_ horizontally depending on the length of the text, but gives a max-width matching the width of 'All content'. Strings longer than this, with localization in mind, are truncated.

The width is done with some magic numbers, but they're at least in REMs so should scale correctly.

CleanShot 2022-01-27 at 11 27 32@2x

CleanShot 2022-01-27 at 11 23 34@2x

CleanShot 2022-01-27 at 11 23 15@2x

CleanShot 2022-01-27 at 11 23 34@2x

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat:
From what I understand (@dhruvkb might know better and correct me), we need to limit the Content switcher for the popover not to jump. Since we use a modal instead of the popover on mobile screen, I don't really see the need for truncating on mobile. 
Unless there is such a long string that it takes up much more space than the width of the screen minus logo and filter button?? When the header is scrolled, we don't show the label anyways.

zackkrida:
> Unless there is such a long string that it takes up much more space than the width of the screen minus logo and filter button??

There are definitely strings that would do this, yes. For example the Spanish translation is `Todo el contenido`.

Now I'm wondering though...truncated text for a button isn't very useful to users.

dhruvkb:
The reason I added a fixed width was to avoid the button changing size based on what option was selected. It should always be the size of the biggest content. 

Behaviour without the `w-20` class:

https://user-images.githubusercontent.com/16580576/151528941-0d328707-24aa-4e74-8840-0953cb374bb9.mov

In retrospect, adding a fixed width was misguided considering i18n strings can be longer than English.

obulat:
> In retrospect, adding a fixed width was misguided considering i18n strings can be longer than English.

I wonder if setting a fixed width in character`ch` units using all of the content-type strings for current locale would be over-engineering and slow down the site?

sarayourfriend:
> I wonder if setting a fixed width in characterch units using all of the content-type strings for current locale would be over-engineering and slow down the site?

I don't think this would slow down the site, FWIW, but I'm hesitant to suggest it as the right answer, if only because `ch` is a rough estimate based on the `0` glyph, which is probably somewhere in the middle as far as glyph widths go?

mmm
000

We can see the difference there I suppose.

We could create a shadow element button using the longest visible string found in the dropdown items and then calculate the width from that? I'm not sure if an actual width would be calculated for an element that doesn't appear in the DOM though, and it could cause client-side jiggling of the button's width, which is really undesirable...

This is a tough one!

panchovm:
Dropping my comment here as in the [slack meeting](https://wordpress.slack.com/archives/C02012JB00N/p1643728421902149).

> The reason I added a fixed width was to avoid the button changing size based on what option was selected. It should always be the size of the biggest content.

That is a good approach @dhruvkb but it jeopardizes spelling and understanding.

I have no concern about the button's jiggling effect as long as it does not break the header or the area where is placed, but the problem also covers smaller breakpoints where there is lesser space. In that case, this input is crucial for the search input work (#594) I started days ago and I can come back with a new proposal. Here is a quick recap of where and how we are using the component.

### Desktop
CleanShot 2022-02-01 at 13 06 02@2x

### Mobile
CleanShot 2022-02-01 at 13 06 45@2x

I need to explore more carefully a solution and test it on all breakpoints with multiple languages in different use cases. So, as @sarayourfriend said, we can move forward with the current solution.

zackkrida:
We should likely develop a checklist to check everything we build in—all accessibility states, string length, and things like that.

zackkrida:
We really need _some fix_ for this, to prevent experiences like this:

CleanShot 2022-02-02 at 08 44 30@2x

sarayourfriend:
I think the actual Portuguese and at least some of the Spanish translations use `Todo`/`Tudo` instead of the full phrase.

krysal (approved):
Since the width shifting is not an issue anymore, this looks good to me. I rebased it to update it with main.

zackkrida:
@obulat can you give this a re-review, testing in a few languages and breakpoints?

sarayourfriend:
Rebased the branch.

I think we're planning to hold of merging this until next week after deploying, is that right @zackkrida ?

Issue #692, opened 2022-01-24
Audio waveform progress bar extends beyond the end of the waveform body


sarayourfriend:
## Description

When playing an audio file all the way through to the end, the progress bar will extend beyond the end of the waveform body.

I'm not sure what causes this, but it seeems like it would just be a small math issue or more likely we're ignoring something critical when pulling the [waveform dimensions from `clientWidth`](https://github.com/WordPress/openverse-frontend/blob/0ebfbe66c923c548954b38051d1dcad4fd54b539/src/components/VAudioTrack/VWaveform.vue#L294).

## Reproduction

1. Run main and go to an audio search that includes a short file (it's easiest and most prominent on short files) like http://localhost:8443/audio/5365bca5-7bd8-4344-b8e4-dad1cf219456
2. Play the audio file through to the end
3. Note the progress bar, timestamp, and focus bar (when focused) extending beyond the end of the waveform body.

## Expectation

The progress bar should end exactly at the end of the waveform body.

## Screenshots

- [ ] 🙋 I would be interested in resolving this bug.

demophoon:
:wave:

Issue #480, opened 2021-12-03
Single result page modal behavior


zackkrida:
## Description

[ticket work required, still in progress]

Single result pages will be styled as _modals_ when navigated to from the search results view, but will be full-width pages when visited server-side, direct from a shared url, for example. You can observe this behavior on Unsplash by searching for an image, viewing the result, and then refreshing the page:

https://user-images.githubusercontent.com/6351754/144637806-7b4fb2c4-96d1-41b0-a04e-de9fd89f2689.mp4

panchovm:
I have been playing with different ideas, and here is the version I like most for both single result views.

<img width="1958" alt="CleanShot 2021-12-14 at 15 36 22@2x" src="https://user-images.githubusercontent.com/895819/146059281-e959554e-63ce-4d52-b7d0-6dd191cfe684.png">

## Header ideas

For the header, I tried a more prominent search bar, an expanded internal menu, and hiding the search result controls (content switcher + filters) since both are content-related. But I also tried showing the openverse logo (symbol + name) as thinking of users opening an Openverse link without knowing the tool. In that scenario, adding the openverse name would help with adding more context.

That being said, the header used in mockups has a full-width grid of **Modal L**, as introduced in ticket #522. 

<img width="1280" alt="Header ideas" src="https://user-images.githubusercontent.com/895819/146060167-de2fe8ac-169a-44bc-ad0a-0bc6697fbd36.png">

panchovm:
Regarding **Full size M** audio component, I tested different widths and background colors.

## Audio component versions

A full-width grid with a gray background version looks centered and balanced, but when playing it, the yellow background will look odd since it does not start from the far-left beginning.

<img width="1365" alt="CleanShot 2021-12-14 at 15 57 10@2x" src="https://user-images.githubusercontent.com/895819/146062061-c68fc622-c7e5-4efe-862a-167563f973f7.png">

A full-width grid with white background version looks clean and it blends greatly with the header. However, by consistency, the image single result page looks floating and empty due to white space on the sides. Plus the yellow background problem when playing the audio.

<img width="1901" alt="CleanShot 2021-12-14 at 15 57 27@2x" src="https://user-images.githubusercontent.com/895819/146062112-eacda8fc-62da-48f4-b652-7cb74a5b8380.png">

Finally, a full-page width with white background also looks beautiful by the blending effect, and it works well with showing the yellow background. But it has the consistency problem with the image single result view.

[...]

PR #744, opened 2022-01-28
Audio single result refinements

### LINKED ISSUE #544: Audio result page responsiveness is messy
## Description

The audio result page's responsiveness is kind of messy at the moment.

## Reproduction

1. Go to an audio result page
2. Resize the browser window from large to small screen size
3. Notice several elements on the page moving at different intervals and some based on different text widths.

## Expectation

We should be more proactive/defensive against window resizing and rearrange elements all at once instead of some elements at the medium breakpoint and others at the small breakpoint. Of particular concern is the wrapping on the play/pause button, track info, and download button. Either we need to reserve that layout for only large or xlarge screens or we'll need to rethink how it works at smaller sizes.

Additionally, long titles for tracks will cause wrapping issues as well:
Captura de Tela 2021-12-20 às 14 28 48

Long artist names will also cause wrapping issues, but different ones!
Captura de Tela 2021-12-20 às 14 29 25

## Screenshots

Here's a video showing the current state of responsiveness of the screen:

https://user-images.githubusercontent.com/24264157/146840902-f127251e-59d5-4355-b75b-75a6d88f3b80.mov

## Environment

 - Device: 
 - OS: 
 - Browser: 
 - Version: 
 - Other info: 

## Resolution

cc @panchovm, we'll need your help thinking about how to handle some of these cases. I was digging around the Audio Integration figma and I noticed some places (for the row version) we truncate these informations. That seems less than ideal for the full result page though. Maybe we can take advantage of the extra space and break things up into separate isolated rows?

### LINKED ISSUE #671: Single audio view lacks 'return to results' link
## Description

Currently images have a nice 'return to results' link that effectively mimics the browser back button. Audio results do not.
This will be fixed by #480, I'm not sure if we need interim solution until then.

## Resolution

### LINKED ISSUE #740: Single audio result view cleanup
## Description

The single audio result view needs some cleanup:

|Live|Mockup|
|-|-|
|CleanShot 2022-01-27 at 14 15 24@2x|image|

## Todos

- [ ] Increase waveform height
- [ ] New tag style
- [ ] Hide provider in details section
- [ ] Should link to the source
- [ ] Remove pluralization from 'Related audios'
- [ ] Switch related audio layout

## Resolution

### LINKED ISSUE #741: Audio download button should have rounded right corners when there's no dropdown
Currently when there is no dropdown menu of additional files, the right corners of the download button are squared:

CleanShot 2022-01-27 at 14 45 13@2x

These corners should be rounded instead.

## Resolution

zackkrida:
## Fixes

Fixes #741 
Fixes #671 
Fixes #740 
Fixes #544 

## Description

- Increase waveform height
- New tag style
- Remove tag links
- Hide source when matches provider in details section
- Make provider link to external url
- Remove pluralization from 'Related audios'
- Switch to correct related audio layout
- play/pause/title alignment to rest of content
- Add 'Back to search link'

|Before|After|
|-|-|
|CleanShot 2022-01-27 at 21 14 40@2x | CleanShot 2022-01-27 at 21 14 24@2x

## Testing Instructions

Click around single audio results on multiple display sizes.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/AudioDetails/VAudioDetails.vue
```diff
@@ -1,12 +1,24 @@
 <template>
   <!--eslint-disable @intlify/vue-i18n/no-raw-text -->
```
obulat, line 2: Great catch with the untranslated string! I think we can remove the eslint-disable now?

obulat (approved):
So many great improvements.

panchovm (changes requested):
Comments posted above in a long message.

src/layouts/error.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <main class="bg-yellow h-screen relative page-404 overflow-x-hidden">
-    <NuxtLink to="/" class="relative z-10">
+    <NuxtLink to="/" class="relative z-10 text-dark-blue">
```
krysal, line 3: Why does this change to dark blue? The logo color is already fixed in any case.

krysal (commented):
I still see the horizontal scrolling due to some weird behavior of the header, maybe it can be fixed after updating this branch with `main`.

zackkrida:
I'll implement some further refinements and cleanup tomorrow, and ask for another round of reviews.

sarayourfriend:
Did some preliminary testing of this and it looks good so far to me. Will review again once the updates Francisco requested are up

zackkrida:
@panchovm Apologies for how long it took to get to your feedback! Some questions:

> Position of back action
> The back to search should be outside the wave area. It is not possible to click on it when playing area overlaps the link.

I've fixed the link to be clickable and also to have some additional padding so it's a larger click surface. I am not sure how to move it 'outside the wave area'. This would create a situation where there's a grey bar above the waveform, like this:

|current|possible|
|-|-|
| CleanShot 2022-02-11 at 14 36 44@2x|CleanShot 2022-02-11 at 14 37 21@2x|

This is also relevant to the height discussion. The timestamp placement relating to the back link, in particular:

|current|possible|
|-|-|
|CleanShot 2022-02-11 at 14 44 38@2x|CleanShot 2022-02-11 at 14 44 38@2x|


> Tag pill size

The version in the actual page is _smaller_ than what I see in the mockups: The mockups have a font size of 14px, vertical padding of 8px, and horizontal padding of 16px.

Figma: https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=1228%3A21496

The other styling issues should be resolved! I was able to find a way to do the 'safe area' around the waveform:

CleanShot 2022-02-11 at 15 14 58@2x

Two full-height yellow elements pop in at the start/end of the waveform while it's playing or when it's finished.

src/locales/en.json
```diff
@@ -528,6 +528,9 @@
       }
     }
   },
+  "single-result": {
+    "back": "Back to search results"
+  },
   "image-details": {
     "back": "Back to search results",
```
krysal, line 535: I think we can remove this key: `image-details.back`

krysal (commented):
Looks like a huge improvement! I'm seeing just a few of off:

I think we should hide a detail property when it's not available.
CleanShot 2022-02-11 at 17 35 13@2x

When the page is SSR a Vue warning appears in the browser console:

> [Vue warn]: The client-side rendered virtual DOM tree is not matching server-rendered content. This is likely caused by incorrect HTML markup, for example nesting block-level elements inside `<p>`, or missing `<tbody>`. Bailing hydration and performing full client-side render.

It's repeated x10 so I assume it's something with Related audio.

sarayourfriend:
@zackkrida is that bug #692?

src/components/AudioDetails/VAudioDetails.vue
```diff
@@ -41,18 +52,24 @@
...
+          <div v-if="audio.source && sourceName !== providerName">
             <dt>
               {{ $t('audio-details.table.source') }}
             </dt>
             <dd>
               {{ sourceName }}
             </dd>
           </div>
-          <div v-if="audio.genres">
+          <div v-if="audio.genres.length > 0">
```
sarayourfriend, line 72: Genres can sometimes be null. Try this on `/audio/632d0076-fdbd-4919-8b8f-b498ef246ee8` and it will crash on this line.

```suggestion
          <div v-if="audio.genres?.length > 0">
```

zackkrida: ah, thanks! assumed it was always an empty array 😢

zackkrida:
@sarayourfriend that definitely sounds like #692, thanks for flagging it!

sarayourfriend:
@zackkrida I guess optional chaining isn't actually allowed in Vue templates. This PR currently throws a syntax error when to try to compile it.

src/components/AudioDetails/VAudioDetails.vue
```diff
@@ -41,18 +52,24 @@
...
+          <div v-if="audio.source && sourceName !== providerName">
             <dt>
               {{ $t('audio-details.table.source') }}
             </dt>
             <dd>
               {{ sourceName }}
             </dd>
           </div>
-          <div v-if="audio.genres">
+          <div v-if="audio.genres?.length > 0">
```
sarayourfriend, line 72:
```suggestion
          <div v-if="audio.genres && audio.genres.length > 0">
```

sarayourfriend (changes requested):

sarayourfriend (approved):
LGTM! Really good improvement and looks good on mobile as well.

I think putting the waveform progress into the sides of the waveform on wider screens is a little awkward, especially when the sound finishes playing... the last bit just sort of "pops" into the yellow color, breaking the flow of the waveform gradually completing. I'd also prefer if we just didn't extend the waveform progress past the waveform itself, but just a personal preference :shrug: cc @panchovm on that one.

panchovm (changes requested):
Nice improvement ✨ Here my single comment for the desktop version.

> I am not sure how to move it 'outside the wave area'. This would create a situation where there's a grey bar above the waveform.

I do not see that as a problem at all. Having it outside matches the mobile design and, for desktop users, it will be visible when opening openverse links. Most visitors will interact with the modal solution. My concern with having the action inside is that visually looks like an interactive area for seeking time when it is not.

Regarding @sarayourfriend comment on the yellow areas on the side of the waveform part, I think it is a good solution. Time ago [I explored different ideas](https://github.com/WordPress/openverse-frontend/issues/480#issuecomment-993886814) and none of them looked better. I agree with the pop effect, but it is more noticeable for the ending part. The starting one is smoother/quicker I guess.

The main aspect of this change is the interactive part and the visual effect indicating the playing state. We can try reducing the width of both areas to make the gap more unnoticeable.

---

And comments for the mobile view.

### Back action & Timestamp spacings
I do not see the back to results action on mobile. I went back to the mockups and the area blends with the gray area where main content is. It might need more contrast to differentiate both areas since they belong to different actions.

The spacing between area and timestamp on top seems related to the back action point. And this effect reinforces my idea of moving the action outside and using a white background to add contrast.

CleanShot 2022-02-14 at 11 24 06@2x

### Button size and text styles
The audio control button looks bigger than the mockup, while text styles inside the main button are big as well. For the main button, all font sizes should be `12px` while font weights vary depending on the base style. I wonder if the font size change impacts the vertical layout height composed by both buttons.

Audio title is also big, it should follow the `description bold` style to be `14px` as the author field.

CleanShot 2022-02-14 at 11 29 48@2x

### Tag pill style
The pill is not like the `medium` style. It should have a top and bottom padding of `4px` and text style as `caption bold`.

CleanShot 2022-02-14 at 11 45 31@2x

### Metadata grid
I noticed the row space is still big. I tried `row-gap: 16px` for the flex property on the inspector and it worked well, but not sure how feasible it is.

### Spacing in Related audio
I still see the spacing between audio as `48px` instead of `32px`.

sarayourfriend:
> The main aspect of this change is the interactive part and the visual effect indicating the playing state. We can try reducing the width of both areas to make the gap more unnoticeable.

Another solution that would take some extra work is to extend the waveform animation beyond the end of the waveform to fill the last gap. It would take some careful math to get right but it probably doable. Not in this PR though! Definitely something for a future low-priority PR.

Issue #710, opened 2022-01-25
Horizontal scroll issues on single image result view


zackkrida:
## Description

The current single image result view currently has a horizontal scrollbar: it's small on desktop and much more noticeable on large displays:

![CleanShot 2022-01-25 at 08 37 06@2x](https://user-images.githubusercontent.com/6351754/150988232-514e285e-5910-4413-81fe-1e1ac40ef2ca.png)

![CleanShot 2022-01-25 at 08 37 16@2x](https://user-images.githubusercontent.com/6351754/150988229-b2c3e93c-0f80-440a-a202-e701c9be6523.png)

## Expectation

This will be fixed by #682, or we could push a more emergency fix if that needs time.

## Environment

 - Device: 
 - OS: 
 - Browser: 
 - Version: 
 - Other info: 

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

zackkrida:
![signal-2022-01-25-091151_001](https://user-images.githubusercontent.com/6351754/150995000-07bac3b5-403c-4267-8c6f-d270c26246e4.jpeg)

Here's a screenshot from a Galaxy A12 Android phone running Firefox 96.2.0

PR #682, opened 2022-01-21
New image details page

krysal:
## Fixes

Fixes #529 by @zackkrida
Fixes #710 by @zackkrida
Fixes #737 by @zackkrida

## Description

This PR applies the new styles for the image results and rewrites some components to generalize to "media" or use one specific for images.

### Pending after this PR
- ~SketchFab viewer~
- Content report
- Show image tags? Yes
- Survey of usage?

## Testing Instructions

Run the project, go to the details page of any image and compare with mockups (Figma links in the issue associated). 

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable). **Still, there is work pending here.**
- [N/A] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #750, opened 2022-01-28
Fix horizontal scroll

krysal:
## Fixes

Related to #710 by @zackkrida 

## Description

This PR attempts to fix the appearance of horizontal scrolling in the audio and image details pages on mobile. At first I thought the cause was just the logo text in the header but on the audio page the related audios were also causing the page to expand horizontally. I changed the related audios to use the 's' layout on mobile so they hide the waveform and fit in the page.

The current image page still has this problem even on desktop but it is fixed in #682, and the mobile view for the new page is fixed here then.

I also removed some classes from the default layout that were hiding the overflow, I'm not sure if this was necessary in order to the page look good on the iframe?

## Testing Instructions

Run the project and navigate to the audio details page on mobile viewport, see if there is any horizontal scroll. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/styles/vocabulary.scss
```diff
@@ -18,7 +18,6 @@
 // Components
 @import 'components/button';
 @import 'components/image';
-@import 'components/navbar';
```
obulat, line 21: 🔥

src/layouts/default.vue
```diff
@@ -1,12 +1,9 @@
 <template>
-  <div class="app grid h-screen overflow-hidden">
+  <div class="app grid h-screen">
```
obulat, line 2: This breaks the separate scrolling of the filters sidebar and the main page. Currently we have this outer `div` full height, and its children have `overflow-y` set to `scroll`.

obulat (commented):
I think we need a solution that doesn't set `overflow-hidden` to the outer div, otherwise we break other parts of the page.
Was this PR only supposed to fix Audio pages? Because I still see a scroll bar on image search page.

I love extracting the condition for showing the logo text to a separate computed!

src/layouts/default.vue
krysal: I had a feeling there was a purpose to this but I didn't remember the filter sidebar, duh! I think we still can do it with this classes then.

krysal:
@obulat This was supposed to fix the mobile version of both, audio and (new, as the current one will be replaced soon) image pages hiding the text of the logo mainly, but that was before you merged #718 so it was probably already fixed there :) 

I'll update this PR and check it does not break the scrollbar of the filter sidebar and probably will be limited to only change the hide the waveform of related audio on mobile.

PR #755, opened 2022-01-31
Enable source maps in production

### LINKED ISSUE #754: Publish source maps to Sentry on release
## Problem

Currently our stack traces in Sentry are awful because Sentry does not have access to our source maps.

There are two recommended ways of getting source maps to Sentry:

1. Publish the source maps publicly alongside the code
2. Upload the source maps directly to Sentry

Sentry "strongly recommends" the second, for whatever reason. The first seems preferrable/fine for our application as there's absolutely zero reason for us to obfuscate any of the client code.

I would prefer we just publish the sourcemaps alongside the code in production for a couple reasons:

1. No need to manage a Sentry API key
2. No need to add an additional step for creating a production release other than what already exists
3. Publishing our source maps could actually help give confidence that we're running the code in production that we say we are running (i.e., we're not applying hidden patches to the open source code that would violate privacy promises)

Additionally, there is no hit to performance for users of the site because source maps that are split from the bundles aren't loaded unless dev tools are opened. Casual navigation of the site would not take any hit.

Sentry so strongly recommends that you upload source maps directly that they have a Sentry webpack plugin for configuring this, but again, it wouldn't benefit security/privacy researchers looking at our site's code and would also require managing another secret, which would be nice to avoid.

Instead I believe we can just publish production source maps like we would normally and Sentry will still pick them up:

https://docs.sentry.io/platforms/javascript/sourcemaps/generating/#advanced-usage

## Description

Publishing the source maps would fix the issue of unhelpful stack traces in Sentry issues.

## Implementation

sarayourfriend:
## Fixes

Fixes #754 by @sarayourfriend 

## Description

Enables source maps in production so that Sentry can use them for better stack traces in issues.

Does not effect the dev experience at all as they were already enabled in development, this PR just removes the "isDev" check that was guarding source map generation.

This will slightly slow down the production build but IMO that's the least concern (I'm more concerned that we're using production quality source maps in development and we could get a dev performance win if we switched to a different source map generation technique in dev, but that's another story altogether and our dev build is still pretty fast anyway).

## Testing Instructions

Run `pnpm build && pnpm start` and verify that you can access the source maps in your browser.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -207,9 +207,7 @@ export default {
     },
     // Enables use of IDE debuggers
     extend(config, ctx) {
-      if (ctx.isDev) {
-        config.devtool = ctx.isClient ? 'source-map' : 'inline-source-map'
-      }
+      config.devtool = ctx.isClient ? 'source-map' : 'inline-source-map'
     },
   },
   storybook: {


zackkrida (approved):
LGTM

krysal (approved):
Every piece of extra info for debugging is welcome 👍

PR #770, opened 2022-02-01
Unify number localization strategy across the app.

### LINKED ISSUE #535: RTL languages still using eastern arabic numerals on search page
## Description

Some locales will cause eastern arabic numerals to be used for number formatting on the search results page (`ar` for example). In previous discussions with WP i18n folks, we decided to just use western arabic numerals everywhere (like on the sources page). We already have logc in the source page for this:

https://github.com/WordPress/openverse-frontend/blob/92240e443664edb2a9d0c25835d9e1cc1a1f2ee3/src/pages/sources.vue#L175-L184

We should refactor this into a util and then use it everywhere.

## Resolution

sarayourfriend:
## Fixes

Fixes #535 by @sarayourfriend

## Description

The issue actually was out-dated, but the solution being used on the search result page differed from the solution used on other parts of the app (namely the sources page).

This PR unifies both approaches and fixes some minor bugs in how the locale was being used for the search results page, and also uses the `en-gb` locale for locales that would have used Eastern arabic numerals.

This also matches the approach taken in Gutenberg for formatting input numbers.

I also managed to get VueI18n into the Nuxt context type so that `useContext`'s return actually includes it, meaning we can add more composables to type checking.

## Testing Instructions

Visit the search page using a locale that uses eastern arabic numerals by default, like `ar` and verify that it is using western arabic numerals instead, like the rest of the locales.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/composables/use-get-locale-formatted-number.js
+++ b/src/composables/use-get-locale-formatted-number.js
@@ -1,0 +1,27 @@
+import { useContext } from '@nuxtjs/composition-api'
+
+export const EASTERN_ARABIC_NUMERAL_LOCALES = ['ar', 'fa', 'ckb', 'ps']
+
+/**
+ * Guards number formatting to prevent using Eastern Arabic Numerals,
+ * instead always using Western Arabic Numerals but still respecting
+ * the locale preferences for delimiters.
+ *
+ * @returns {(n: number) => string}
+ */
+export const useGetLocaleFormattedNumber = () => {
+  const { i18n } = useContext()
+
+  /**
+   * @param {number} n The number to format
+   */
+  return (n) => {
+    let { locale } = i18n
+
+    if (EASTERN_ARABIC_NUMERAL_LOCALES.some((l) => locale.startsWith(l))) {
+      locale = 'en-gb'
+    }
+
+    return n.toLocaleString(locale)
+  }
+}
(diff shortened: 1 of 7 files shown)


sarayourfriend:
@krysal and @obulat would it be possible to get a review on this PR soon? Thanks!

zackkrida (approved):
This is a very thoughtful addition, love the attention to detail.

obulat (approved):
I love the unification, thank you for taking care of it!

I have a couple of questions about the locales:
1. How did we choose the locales for the `ARABIC_NUMERAL_LOCALES`?
One concern I have is about the locales that are variants of those locales (ie, for Arabic `ar` locale, there are also Moroccan Arabic `ary` and Algerian Arabic `arq` in our `valid-locales.json` list). I'm not sure how much practical importance that has considering that we don't have those translations.
The other concern is: are there other locales that also need to use a different formatting that we haven't identified?
2. [Nitpicking] Why not use `en` locale instead of `en-gb` for the fallback formatting?

sarayourfriend:
> How did we choose the locales for the ARABIC_NUMERAL_LOCALES?

I'm forgetting the name of the organization and am failing at finding it right now, but there's an org that collects locale metadata from countries all over the world and maps their numeral system (among other things) to the locale.

However, I suppose the _best_ most robust way to do this would be to just format the number and check if it uses anything other than Western Arabic numerals and if it does, then format it using `en`.

I chose `en-gb` becuase the impression I got was that the decimal and thousands delimeter rules followed by `en-gb` more closely matched the expectations from the locales we were intercepting, but I could have gotten that totally wrong. Switching it to just `en` seems fine to me.

obulat:
Thank you for clarifying, @sarayourfriend !

PR #771, opened 2022-02-01
Add skip to content button

### LINKED ISSUE #659: Executing a search from the homepage using a screen reader is confusing

sarayourfriend:
## Fixes

Fixes #659 by @sarayourfriend

## Description

Adds a "Skip to content" button on the search page that should be the _first_ thing tabbable on the page. This will make it the first thing that keyboard navigation users will tab into and therefore make it easy to skip to the search results.

This is a general purpose solution that can be applied to all pages that need it eventually (for example the information pages or single result pages). It tries to be pretty unassuming in how it does things while also making it easy to add to pages in the future. For now we can put it just on the search page and see how it works out. It's the most critical page that needs this anyway.

## Testing Instructions

Do the following both with and without a screen reader using keyboard navigation as well as using mouse navigation and verify that the interactions make sense:

1. Execute a search from the home page using the mouse and verify that the experience still works and makes sense
2. Execute a search from the home page using keyboard navigation and verify the when you hit tab on the search results page the "Skip to content" button is focused and that it works to focus you into the results
3. Execute a search from the home page using a screen reader and keyboard navigation and verify that the "Skip to content" button is either automatically focused by the screen reader or is the first element focused when using the screen reader controls to navigate the page

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VSkipToContentContainer.vue
```diff
@@ -0,0 +1,60 @@
...
+
+    return { containerNode, skipToContent }
+  },
+})
+</script>
+
+<style module>
+.skipButton:not(:focus) {
+  @apply sr-only;
+}
```
sarayourfriend, line 59: We want to remove the `sr-only` class when the button is focused so that it becomes visible for sighted keyboard users who are not using a screen reader. This seems like the easiest way to do that but maybe there's something built into Tailwind that would simplify it?

tsconfig.json
```diff
@@ -34,6 +34,7 @@
   "include": [
     "src/composables/types.js",
     "src/composables/use-event-listener-outside.js",
-    "src/composables/use-active-audio.js"
+    "src/composables/use-active-audio.js",
+    "src/utils/key-codes.js",
```
sarayourfriend, line 38: Might as well include this even though it's not related. Doesn't hurt anything and will make the `key-codes.js` module easier to use anyway.

src/components/VSkipToContentContainer.vue
zackkrida: I think this is ideal, although it would be cool (confusing perhaps) if tailwind had a `not` selector.

zackkrida (approved):
Looks great. It's a real testament to the base components and reusable logic you've added that this was so quick to implement!

src/components/VSkipToContentContainer.vue
sarayourfriend: FWIW I did try using `sr-only focus:not-sr-only` but `not-sr-only` doesn't remove all the `sr-only` styles, just the ones preventing the element from being visible, meaning the button margins and padding are all messed up. Unfortunately it seems like this really is the best approach for `not-focus:sr-only` 😭

krysal (approved):
It works great!

PR #796, opened 2022-02-03
Fix homepage RTL

### LINKED ISSUE #765: RTL broken on homepage
## Description

The homepage is using the `blank` layout which doesn't have basic i18n wired up for setting the `dir` attribute.

## Reproduction

1. Set your locale to an RTL language like `ar`
2. Visit the homepage
3. Note that nothing is in RTL

## Expectation

RTL should work on all pages.

## Resolution

sarayourfriend:
## Fixes

Fixes #765 by @sarayourfriend

## Description

Fixes the homepage RTL problems.

1. The logo and brand mark were in a single SVG making it impossible for them to re-order based on reading direction. Splitting them into two separate SVGs makes this possible. It does make using them a little bit more of a hassle, but this shouldn't be something we're using in a bespoke manner more than a few times, so I don't see this as a maintainability problem. If these logos or brand marks ever change it'll require careful manual positioning fixes anyway regardless of if it's a single or two SVGs.
2. Some non-RTL compliant positioning classes were being used, so this switches them to use the relative margin/padding classes.

## Testing Instructions

Load up this branch and run `pnpm dev`.

1. Visit `/` and verify that the homepage looks good at all screen widths
2. Visit `/ar` and verify that the homepage looks good at all screen widths and that text is aligned
3. Visit `/404` and verify that the logo and brand mark look good
4. Visit `/ar/404` and verify that the logo and brand mark look good

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/pages/index.vue
```diff
@@ -6,20 +6,29 @@
     <header
       class="flex-grow w-full lg:w-auto lg:min-w-[32rem] xl:min-w-[64rem] box-border flex flex-col justify-between lg:justify-center"
     >
-      <VLogoButton class="lg:hidden ms-3" :auto-resize-logo="false" />
+      <VLogoButton
```
obulat, line 9: Neither header nor `VLogoButton` have a color class, so the color of the logo becomes black instead of dark charcoal.

obulat:
Do we really need to change the logo direction? I couldn't find anything specifically on this, but Ahmad Shadeed's RTL-styling blog uses the same logomark for RTL, without changing direction, for Smashing Magazine: https://rtlstyling.com/posts/rtl-styling#3.-differences-in-word-sizes-between-languages .

sarayourfriend:
@obulat I don't think we necessary do, but in the case of the homepage, the LTR design looks very unbalanced and awkward with out a flipped logo mark. There's no sensible way to have the logo mark "hang" off the leading edge of the page like it does on the LTR version.

If @panchovm has any ideas of how to do this without flipping the logo and brand mark, then we can implement that. But as it stands I couldn't come up with anything that looked decent.

obulat:
> @obulat I don't think we necessary do, but in the case of the homepage, the LTR design looks very unbalanced and awkward with out a flipped logo mark. There's no sensible way to have the logo mark "hang" off the leading edge of the page like it does on the LTR version.

Oh, I didn't realize that this is mainly a solution to fix the "hanging" part of the large screen logo, thank you for explaining!

By the way, the color of the logo in the homepage header on screens smaller than `lg` is still black.

panchovm:
I love ❤️ @sarayourfriend's solution of flipping the logo. It embraces the RTL logic as the most prominent use of Openverse will be through the symbol rather than symbol and wordmark. Therefore, switching positions is a great idea.

obulat:
I *can* see the difference in colors. I guess working closely with @panchovm on the header fixes helped me learn to pay more attention to colors and spacings 😀

obulat (approved):

sarayourfriend:
I envy your eyes then! I looked very hard and close and really couldn't tell the difference 😆 Thanks for lending your eyes 🙏

panchovm:
Although let me say that @obulat's eagle eyes date longer than the header work.

zackkrida:
Monitor color profiles could also play a major role here 😆

zackkrida (changes requested):
This looks great, I do see one visual regression. Probably as a consequence of separating the logo text and icon, the logo text is no longer aligned with the rest of the content:

> after

CleanShot 2022-02-07 at 12 40 37@2x

> before

CleanShot 2022-02-07 at 12 45 42@2x

The simplest fix might be for the logo text to be in the same container as the other content, and for the logo icon to be absolutely positioned relative to that container.

sarayourfriend:
Ah, that's a much wider browser window configuration than I'm able to achieve with my current set up 😅 I definitely ran into similar problems but didn't realize the margins would keep moving as the view got even larger.

I'll try the absolute positioning trick you recomended.

sarayourfriend:
@zackkrida updated, please let me know if this works for you!

zackkrida (approved):
Looks perfect!

PR #798, opened 2022-02-03
Remove usage session cookie

### LINKED ISSUE #791: Move analytics server `sessionId` to localstorage; rename.
## Problem

Currently the frontend uses a session cookie (one that expires upon the user closing their browser) to store a persistent identifier for Openverse users. This is a randomly generated [v4 uuid](https://github.com/uuidjs/uuid#uuidv4options-buffer-offset).

https://github.com/WordPress/openverse-frontend/blob/4dc241dbdd8d1a9e103ea0742d9c17850db3c96c/src/utils/session-id.js#L26

We would like to remove cookies from Openverse, as this allows us to aggressively cache our frontend to reduce cost and traffic to our origin servers. 

## Description

We should consider using `localStorage` or `sessionStorage`. The later and lesser-known `sessionStorage` is perhaps the most accurate representation of a 'session' here—it persists for the lifespan of a _browser tab_. So switching to that could actually give us more useful data in the analytics server.

## Alternatives

We could keep this as a cookie and lose the ability to cache or pages, or consider removing session data from our analytics server.

## Additional Information

Doing this would also allow us to remove cookie related js dependencies.

## Implementation

sarayourfriend:
## Fixes

Fixes #791 by @zackkrida

## Description

I ended up choosing not to go with the `localStorage` or `sessionStorage` approaches primarily because those are confined to the _client_ side of the application, meaning that any events happening during the SSR rendering of a page would be impossible to tie to a specific session.

In the linked issue I proposed a solution for being able to tie session IDs together on the analytics server side, essentially by treated each new one generated by the server that is replacing an ID the client already held in session storage as a node in a linked list (the client would be able to create a link between the previous session ID and the new one). That would have the benefit of creating a _longer_ lived session ID than the ephemeral one we have in this PR while still maintaining the basic browser definition of session storage, because if the client doesn't have an ID in session storage then it wouldn't be able to create the link between the previous ID and the new one the server generates.

In any case, I don't know if that level of complexity is necessary yet as we're not even really using the analytics data in any detailed way at the moment as far as I know. The primary goal here, at least in my view, is to be able to continue gathering some analytics data (once we turn the feature back on in production) while still eliminating the cookie that is currently being used.

In the future if we decide we do want that broader full-session ID chain then we can work on an RFC for how to implement that. It will require coordination between the frontend and API to make the changes.

## Testing Instructions

Checkout the branch and run `pnpm dev`. Load the site and ensure analytics requests are sent as expected with a single session ID between the server and client sides of the application.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/store/user.js
```diff
@@ -1,5 +1,18 @@
...
+ * @returns {UserState}
+ */
 export const state = () => ({
-  usageSessionId: SessionId('usageSessionId'),
+  /**
+   * Default to `undefined` on the client and let client side
+   * SSR state hydration pull the SSR generated ID into the client.
+   * This way we avoid generating any spurious IDs on the client.
+   */
+  usageSessionId: process.server ? uuidv4() : undefined,
```
zackkrida, line 17: 🔥 🔥 🔥

zackkrida (approved):
This definition of a 'session' should be fine for some time and the new types are great.

src/data/usage-data-service.js
```diff
@@ -4,17 +4,35 @@ const baseUrl = process.env.apiUrl
...
+   * @param {string} props.sessionId
+   */
   sendSearchQueryEvent({ query, sessionId }) {
     return this.post('search_event', { query, session_uuid: sessionId })
   },
 
+  /**
+   * @param {object} props
+   * @param {string} props.query
+   * @param {unknown} props.resultRank
```
obulat, line 32: This type can be set to a number because it's a result of a `findIndex` call: https://cs.github.com/WordPress/openverse-frontend/blob/af82bcd215933032fd3c557495bd32d41866b44d/src/store/media.js#L170

test/unit/specs/store/user.spec.js
```diff
@@ -0,0 +1,28 @@
...
+jest.mock('uuid', () => ({ v4: jest.fn() }))
+
+describe('user store', () => {
+  beforeEach(() => {
+    process.server = false
+    process.client = false
+  })
+
+  afterAll(() => {
+    // avoid polluting other tests that might be affected by these flags
```
obulat, line 13: Nice cleanup and comment of why it's necessary, especially for future use and copying :)

obulat (approved):
Very interesting to read about the approach.

src/data/usage-data-service.js
sarayourfriend, line 32:
```suggestion
   * @param {number} props.resultRank
```

PR #800, opened 2022-02-04
Update color 'primary' to 'pink' after redesign

dhruvkb:
## Description

This PR updates the usages of the color primary to reference pink as it was planned to be done after the redesign launch.

## Testing Instructions

0. Check out the PR and run the Storybook & dev server.
1. Search for the usages of color 'primary'.
2. The usages should all be pink.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VCheckbox/VCheckbox.vue
+++ b/src/components/VCheckbox/VCheckbox.vue
@@ -128,7 +128,7 @@ export default VCheckbox
 }
 .checkbox {
   @apply appearance-none w-5 h-5 border-dark-charcoal border rounded-sm me-3 flex-shrink-0 relative;
-  @apply focus:outline-none focus:ring focus:ring-offset-2 focus:ring-primary;
+  @apply focus:outline-none focus:ring focus:ring-offset-2 focus:ring-pink;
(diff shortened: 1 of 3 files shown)


sarayourfriend (approved):
I thought we would keep the `primary` key and switch everything that is using `pink` to use `primary` instead? 🤔 Otherwise future brand color changes will require a big find-and-replace job that could just be a one-line configuration change.

dhruvkb:
I suppose any changes to the brand in the future will involve more than color. Currently the usage of 'pink' is much more prevalent as compared to 'primary' which is just used in 2 places.

obulat (approved):

sarayourfriend:
Gotcha. I think we should re-evaluate this and think about how we use design tokens. You're right that a re-design would include more than color, but it could be "just color". Likewise, using a token like "primary" would make things like a dark-mode based on tokens easier to reason about instead of "pink" which would likely have a different color entirely under a dark-mode (maybe).

Either way, it's good to get rid of the two different names for the token, but I'd be interested in getting clearer decision making around whether to use generic token names or to use the color-names for the tokens.

PR #818, opened 2022-02-06
(#692) Fix duration mismatch between audio and metadata

### LINKED ISSUE #692: Audio waveform progress bar extends beyond the end of the waveform body

demophoon:
When an audio file is played and the duration from the passed properties
does not match the duration of the actual audio file, e.g. the duration
was rounded, the progress bar drifts outside of the total width of the
VWaveform dimensions.

On smaller width screens or larger tracks it is less noticeable but when
the audio track is small and the screen width is large that percentage
adds up to 10s or more of pixels which is where the bug crops up.

This commit fixes this issue by by using the audio context's duration
parameter and falling back to less specific sources for duration.

## Fixes

Fixes #692 by @sarayourfriend 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -124,6 +124,7 @@ export default defineComponent({
 
     const status = ref('paused')
     const currentTime = ref(0)
+    const audioDuration = ref(null)
 
     const initLocalAudio = () => {
       // Preserve existing local audio if we plucked it from the global active audio
@@ -133,6 +134,7 @@ export default defineComponent({
       localAudio.addEventListener('pause', setPaused)
       localAudio.addEventListener('ended', setPlayed)
       localAudio.addEventListener('timeupdate', setTimeWhenPaused)
+      localAudio.addEventListener('durationchange', setDuration)
 
       /**
        * Similar to the behavior in the global audio track,
@@ -216,6 +218,9 @@ export default defineComponent({
         }
       }
     }
+    const setDuration = () => {
+      audioDuration.value = localAudio?.duration
+    }
 
     /**
      * If we're transforming the globally active audio
@@ -238,6 +243,7 @@ export default defineComponent({
       localAudio.removeEventListener('pause', setPaused)
       localAudio.removeEventListener('ended', setPlayed)
       localAudio.removeEventListener('timeupdate', setTimeWhenPaused)
+      localAudio.removeEventListener('durationchange', setDuration)
 
       if (
         route.value.params.id == props.audio.id ||
@@ -286,7 +292,7 @@ export default defineComponent({
     /* Timekeeping */
 
     const duration = computed(
-      () => (localAudio?.duration ?? props.audio?.duration ?? 0) / 1e3 // seconds
+      () => audioDuration.value ?? props.audio?.duration / 1e3 ?? 0 // seconds
     )
 
     const message = computed(() => store.state.active.message)
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -285,7 +285,9 @@ export default defineComponent({
 
     /* Timekeeping */
 
-    const duration = computed(() => (props.audio?.duration ?? 0) / 1e3) // seconds
+    const duration = computed(
+      () => (localAudio?.duration ?? props.audio?.duration ?? 0) / 1e3 // seconds
+    )
 
     const message = computed(() => store.state.active.message)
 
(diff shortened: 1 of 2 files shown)


src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -285,7 +285,9 @@ export default defineComponent({
 
     /* Timekeeping */
 
-    const duration = computed(() => (props.audio?.duration ?? 0) / 1e3) // seconds
+    const duration = computed(
+      () => (localAudio?.duration ?? props.audio?.duration ?? 0) / 1e3 // seconds
```
sarayourfriend, line 289: This looks great! There's also another component that will need this update: `VGlobalAudioTrack`.

You can see this component on the audio search results page when your viewport is small, like on a mobile device (or just a narrow desktop browser window).

demophoon: Updated in `f3aa9f2`.

Hope its okay I added this to the single commit. Figured it was the same exact fix.

demophoon:
Hmm.. this actually needs more investigation before being merged. I'm still seeing the bar fall off the end..

sarayourfriend:
I think you have the right fit, but because the `localAudio` object isn't a `ref`, it's not going to recalculate.

One potential solution I can think of is to create a `duration` ref to store it in and use the `props.audio.duration` as the default value. Then we can add a `load` event listener to the `localAudio` object to retrieve the `duration` of the file once it's loaded.

What do you think @demophoon?

zackkrida:
@demophoon do you think you'll have an opportunity to revisit this PR? If not, we can get another contributor to look at it, totally up to you. Thank you for your work so far!

zackkrida:
@dhruvkb's expertise here would be useful as well, as the primary author of the original audio track components.

demophoon:
@zackkrida Oh yes! It totally slipped out of my mind after the conference. I'm looking at this again this morning.

@sarayourfriend That makes total sense to me since the duration that I initially used is set before the browser has even had a chance to decode the audio track to determine the real duration.

demophoon:
That change appears to have worked!

If there is a better way to notify a computed property of this change please let me know. I'm still a new to vue so I might not be aware of all the features.

Screenshots below of fix!
image
image

sarayourfriend:
Looking good! Seems to work well on the single audio results, but for the global player there's a subtle bug.

First, to activate the global audio play, open a very narrow browser window on the audio search results page, something like this:

image

When you play an audio file, the global audio player will appear floating at the bottom of the screen. Make sure to narrow the results to just Jamendo provider (for example `/search/audio?q=birds&source=jamendo`).

Next, play one audio file that you know is shorter than another. Then play another audio file. Confirm that the duration was updated for the global player. Next, play the first audio again.

Notice that the duration will not be updated in the global player's context.

This is because the `durationchanged` event isn't firing, the audio objects have just been swapped out.

I think in the global audio track component we need to use the `durationchanged` event _as well as_ grabbing the duration when the `activeTrack` ref changes, similar to how we grab the `currentTime` of the active audio on this line:

https://github.com/WordPress/openverse-frontend/blob/fc35e75208b67ae022c41828997a6735f334af1f/src/components/VAudioTrack/VGlobalAudioTrack.vue#L157

In fact, I think just adding another line below that one that does this would work:

```js
audioDuration.value = audio.duration
```

I tried that locally and it seemed to work :+1:

demophoon:
Oh, great catch. I didn't think about the object still being around on a second play through especially with respect to the global audio component.

Updated in `9f228640`.

sarayourfriend (approved):
This is great! Thanks so much for the contribution :tada:

zackkrida (approved):
So nice to see this fixed. The new code for `durationchange` listening and storage in a ref seems very logical!

PR #821, opened 2022-02-07
Allow local network access from other devices when in development

### LINKED ISSUE #820: Allow access from hosts other than `localhost` during development
## Problem

To be able to access a locally running nuxt server from another device on the same network, you need to either tunnel the request through something like `caddy` or `ngrok` or update the `nuxt.config.js` `server.host` property to `0.0.0.0`. We should just set the the `host` property to `0.0.0.0` automatically in dev. It doesn't hurt the existing workflow and only makes it more flexible for testing across multiple devices.

## Alternatives

Recommend an official method of tunneling requests locally (like `caddy`) and add a `package.json` script for initiating the tunneling.

The idea is to just make this process easier and not require temporary code changes.

## Implementation

sarayourfriend:
## Fixes

Fixes #820 by @sarayourfriend

## Description

Nuxt documentation about this change here: https://nuxtjs.org/docs/features/configuration#edit-host-and-port

## Testing Instructions

Checkout the branch and run `pnpm dev` then visit the locally running site from another device like a mobile phone or another computer. It should work now. Also visit it at `localhost:8443` like we used to. That should also work.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -111,7 +111,9 @@ export default {
   },
   srcDir: 'src/',
   modern: 'client',
-  server: { port: process.env.PORT || 8443 },
+  server: {
+    port: process.env.PORT || 8443,
+  },
   router: {
     middleware: 'middleware',
   },
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -111,9 +111,7 @@ export default {
   },
   srcDir: 'src/',
   modern: 'client',
-  server: {
-    port: process.env.PORT || 8443,
-  },
+  server: { port: process.env.PORT || 8443 },
   router: {
     middleware: 'middleware',
   },
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "author": "sebworks",
   "private": true,
   "scripts": {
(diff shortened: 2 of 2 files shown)


zackkrida:
It's odd that they show the strategy they _don't_ recommend first, but from the docs:

> Although you can modify this in the nuxt.config.js file it is not advised to as it might cause you issues when hosting your site. It is much better to modify the host and port direct in the dev command.

They recommend adding `--hostname '0'` to the `dev` npm script instead.

sarayourfriend:
Oops, I should have kept reading 🙂 I've changed it to use the `hostname` flag. Thanks for finding that Zack!

nuxt.config.js
```diff
@@ -111,7 +111,9 @@ export default {
   },
   srcDir: 'src/',
   modern: 'client',
-  server: { port: process.env.PORT || 8443 },
+  server: {
+    port: process.env.PORT || 8443,
+  },
```
sarayourfriend, line 116:
```suggestion
  server: { port: process.env.PORT || 8443 },
```

krysal (approved):
TIL this is possible. Nice! 💯

PR #829, opened 2022-02-08
Add a check for blank search term before searching

### LINKED ISSUE #769: Disable homepage search submission without a query
## Description

Currently, the homepage allows users to submit a search with no search query. We should block that, as it's one of the only explicit user-facing paths to the raw, query-less /search/ endpoint which has been identified as a problem in #596 

## Resolution

obulat:
## Fixes

Fixes #769 by @zackkrida

## Description

This PR adds a check for search term before submitting a query from the search page. This prevents the user from getting to the blank `/search/` url.

## Testing Instructions

Go to the homepage and click the search button. You should not be redirected, or see any changes.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -201,8 +201,9 @@ const HomePage = {
 
     const searchTerm = ref('')
     const handleSearch = async () => {
+      if (!searchTerm.value) return
       await store.dispatch(`${SEARCH}/${UPDATE_QUERY}`, {
-        q: searchTerm.value || '',
+        q: searchTerm.value,
         searchType: contentType.value,
       })
       const newPath = app.localePath({


src/pages/index.vue
```diff
@@ -201,8 +201,9 @@ const HomePage = {
 
     const searchTerm = ref('')
     const handleSearch = async () => {
+      if (!searchTerm.value) return
```
sarayourfriend, line 204: I wonder if we should `.trim()` the search term as well? Like if it's all spaces, do we still want to submit it?

DuckDuckGo actually displays an error when you try to submit a string of spaces.

Captura de Tela 2022-02-08 às 05 32 18

sarayourfriend (approved):
I was checking out how DuckDuckGo and Google handle this, just to make sure we match in behavior and because I was wondering if it was worth adding a `minlength="1"` to the input.

DuckDuckGo allows you to submit the form but automatically redirects you back to their homepage.

Google ignores the search request if JavaScript is enabled. If you disabled JavaScript, they also just redirect you back to the homepage if you submit an empty query.

Given this PR makes our implementation match Google's homepage, it seems like that's probably safe to do. They also do _not_ include anything like `minlength="1"` on the form.

We also don't have a way of handling JavaScript-disabled clients currently anyway so the second version isn't as important.

I had one suggestion but take it or leave it, not a blocker! LGTM 😁

src/pages/index.vue
obulat: This issue is actually already handled by Vue with the `v-model`'s `.trim` modifier:
https://github.com/WordPress/openverse-frontend/blob/31ba45713247e31c9ae2c7bd251808d586b0a6e9/src/pages/index.vue#L45
I did go back and test it, and it does work, you can't search for ' '.

sarayourfriend: 😱 I should have tried it out before commenting!

obulat:
Very interesting investigations, @sarayourfriend !

> I was wondering if it was worth adding a `minlength="1"` to the input.

I also had this thought in the back of my head, so thank you for sharing your investigation results.

krysal (approved):
Nice!

PR #844, opened 2022-02-10
Use the correct list of filters for All content

obulat:
## Description

This PR fixes the mixup between the `mediaFilterKeys` and `mediaSpecificFilters` objects in the `search` store.

The first is used to keep all the filter types that are used for the content type, and the second has a list of filters that are only used for the specific content type, and not used in other content types (such as `duration` for audio).

One of the places where the `mediaSpecificFilters` was used was the `search/CLEAR_OTHER_MEDIA_TYPE_FILTERS`. This action is dispatched when changing a content type to clear filters that do not apply to current content type. So, when changing from "All Content" to "Images", it would set all the filters that are specific for "All Content" and "Audio" to not checked. With the bug that is fixed here, it would also uncheck the following filters: `['licenses', 'licenseTypes', 'searchBy', 'mature']`. And because `mature` filter is not an object, but a boolean, it would crash when iterated over, as the `state.filters['mature']` is not an array here, so we can't use `map` here:
```
    Object.keys(state.filters).forEach((filterType) => {
      if (filterKeysToClear.includes(filterType)) {
        state.filters[filterType] = state.filters[filterType].map((f) => ({
          ...f,
          checked: false,
        }))
      }
    })
```
This did not actually cause an crash because we were only clearing the `supportedMediaTypes`, and not the `supportedContentTypes` (i.e. did not include `ALL_MEDIA`).

## Testing Instructions

Simply running the unit and e2e tests should be enough.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/store/search.js
+++ b/src/store/search.js
@@ -11,7 +11,7 @@ import {
   ALL_MEDIA,
   AUDIO,
   IMAGE,
-  supportedMediaTypes,
+  supportedContentTypes,
   VIDEO,
 } from '~/constants/media'
 import {
@@ -468,7 +468,7 @@ const mutations = {
    * @param {'all'|'audio'|'image'|'video'} searchType
    */
   [CLEAR_OTHER_MEDIA_TYPE_FILTERS](state, { searchType }) {
-    const mediaTypesToClear = supportedMediaTypes.filter(
+    const mediaTypesToClear = supportedContentTypes.filter(
       (media) => media !== searchType
     )
 
--- a/src/store/search.js
+++ b/src/store/search.js
@@ -33,7 +33,10 @@ import {
   SET_SEARCH_TYPE,
 } from '~/constants/mutation-types'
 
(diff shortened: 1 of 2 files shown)


sarayourfriend (approved):
Nice! Thank you for the explanation and the detailed documentation on each object. It makes me wonder if we could rename `mediaFilterKeys` to something more descriptive 🤔 `sharedFilterKeys` maybe? 🤷

obulat:
> Nice! Thank you for the explanation and the detailed documentation on each object. It makes me wonder if we could rename `mediaFilterKeys` to something more descriptive 🤔 `sharedFilterKeys` maybe? 🤷

I was also wondering about a better name, but couldn't find it yet. `shared` wouldn't be correct for this, because the object actually contains all filter keys for all media types.

PR #846, opened 2022-02-10
Set fixed width for the content switcher based on the longest i18n string

obulat:
## Fixes

Another take on #738 by @zackkrida 

## Description

This PR sets a fixed width to the content switcher when it has a label (i.e. on at least `md` screen) calculated using the translated strings, so that it is not too small for languages other than English. I tried setting the width using the tailwind arbitrary values `w-[npx]`, but that didn't work for some reason (the class was there, but the styles for the class were not generated), so I used `style` declaration there.
It uses an invisible `p` element to compute the width based on a [SO question](https://stackoverflow.com/questions/118241/calculate-text-width-with-javascript/21015393#21015393).
content-switcher

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VContentSwitcher/VContentSwitcherButton.vue
+++ b/src/components/VContentSwitcher/VContentSwitcherButton.vue
@@ -9,9 +9,19 @@
     @click="$emit('click')"
   >
     <VIcon :icon-path="icon" />
-    <span v-show="showLabel" :class="{ 'ms-2 md:w-20 text-left': showLabel }">{{
-      buttonLabel
-    }}</span>
+    <p
+      v-show="showLabel"
+      :class="{ 'ms-2 text-left': showLabel }"
+      :style="labelFixedWidthStyle"
+    >
+      {{ buttonLabel }}
+    </p>
+    <span
+      v-if="showLabel"
+      ref="testLabel"
+      class="absolute invisible h-auto w-auto whitespace-nowrap"
+      >{{ longestLabel }}</span
+    >
     <VIcon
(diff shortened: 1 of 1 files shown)


src/components/VContentSwitcher/VContentSwitcherButton.vue
```diff
@@ -9,18 +9,34 @@
...
+    <p
+      v-show="showLabel"
+      :class="{ 'ms-2 text-left': showLabel }"
+      :style="labelFixedWidthStyle"
+    >
+      {{ buttonLabel }}
+    </p>
+    <span
+      v-if="showLabel"
+      ref="testLabel"
```
krysal, line 21: Perhaps a better name for this would be `longestLabelRef` or `hiddenLabel`?

zackkrida:
@panchovm can chime in on the design decisions, but I think I actually prefer when the chooser changes size to match the length of the current text. I find it distracting to see the 'Audio' label with so much white space after the label. My current thinking is that we should let the size change freely depending on the string, but have a max-width determined by what breaks our layout.

panchovm:
Agree with @zackkrida. I thought that was the outcome of #738 discussion, or it was the [idea I proposed](https://github.com/WordPress/openverse-frontend/pull/738#issuecomment-1027014764).

obulat:
Closing this PR as unnecessary in light of the decision to only set `max-width`, and let the width change based on the content string length. 
It was a fun exploration :)

Issue #759, opened 2022-01-31
Header size in tablet


panchovm:
## Description
On tablet, `md` and `lg` breakpoints, the header breaks when opening the content switcher by changing its height and moving buttons to a second row.

## Expectation
The header should not change its size but adapt its width to the viewport.

## Screenshots

https://user-images.githubusercontent.com/895819/151820437-417db89f-9839-471a-a8fa-6cdfdc873725.mp4

https://user-images.githubusercontent.com/895819/151820462-7d33eeb8-d793-4145-80ec-daace36f0f01.mp4

zackkrida:
I'm not able to reproduce this on the production site in Chrome, Firefox, or Safari on my Mac. I can get you specific version numbers on those, but where are you seeing this issue?

panchovm:
On Firefox and Safari, in an iPad Pro 11-inch. Sorry for not adding the device and browser details.

sarayourfriend:
@panchovm can you share what version if iPadOS your device is running? You can find it by following these steps: https://support.apple.com/en-us/HT201685

panchovm:
I update the OS yesterday but today I am still seeing the bug. I have iOS 15.3

I think it has to do with the WordPress header. In the video below, I was trying to change content, see filters, and type a new search. None of them worked.

https://user-images.githubusercontent.com/895819/152164966-185ad0e5-77d0-473c-968f-fa8464b203e6.mp4

sarayourfriend:
Sounds good. I have an iPad I can test this with back home. I'll look into this later today unless someone else gets to it before me.

zackkrida:
@panchovm or @sarayourfriend could you check the `main` branch for this again on an iPad? it might have been fixed in a recent PR.

obulat:
There is a new regression on the `main` branch due to addition of the `VLogoButton`. Originally, the search bar width in a single-row header was calculated based on the widths of other components, and the logo was assumed to be 40px wide. Now it takes up 48px, so the search bar doesn't fit, and the last button goes to the second row.

panchovm:
I do not know how to set up my iPad for running dev, I tried mirroring it from my MacBook but did not work.

sarayourfriend:
I ran it yesterday and didn't notice this particular issue. But there are indeed other problems with mobile browsing and the header that we need to fix (I'm working on it last night and today).

panchovm:
This is what I see on Firefox and Safari (iOS 15.3)

[...]

PR #850, opened 2022-02-10
Update license explanation tooltip

### LINKED ISSUE #423: License explanation popup should close on click outside and be placed correctly
## Description

#406 changes the way License explanation popup is opened. It can be opened by clicking on the button with a question mark, and can be closed by clicking on it, as well.
It should also be possible to close the popup by clicking outside. Also, the popup should be placed correctly even after the filter has been scrolled, and not shift up or down from the button that opened it.
## Reproduction

1. Checkout #406 branch (`add/checkbox`) and run it with `npm run dev`
2. Open the search page with the filter sidebar `http://localhost:8443/search/image?q=cat&license=by-nd&license_type=commercial`
3. Click on the license explanation button (question mark next to one of the license filters). You cannot close the popup that opens by clicking outside it, the only way to close it is to click on the same button.
4. Scroll the page down, and open a license explanation popup again. It probably will open in a different place than is expected.

## Expectation

The popup should always open to the right of the button on desktop, and should be closed by clicking outside of the popup.

## Screenshots

Screen Shot 2021-11-18 at 5 57 17 AM

## Additional context

The solution should probably use the recently added `VPopover` component.

## Resolution

### LINKED ISSUE #541: Add license definition in filter sidebar
## Problem
The current mockups do not have the license definition in the Licenses section, and we need to show a license explanation and a link to the license description per license.

## Alternatives
This proposal adds a right-aligned help icon showing the license definition in a box. The interaction goes as it follows:
1. On hover, the `help` icon turns `dark-charcoal` and a tooltip appears on its left.
2. On click, the license explanation box appears top-aligned to the license checkbox.
3. Clicking outside the license explanation box closes the box.
4. Clicking on a different `help` icon closes the current box and opens the new one.

Here is a video showing the interaction. As noticed below, the checkbox state (enabled, disabled) does not affect the license explanation interaction.

https://user-images.githubusercontent.com/895819/146779058-a98903b0-5ed6-496d-9c36-bcdb9415bb0c.mp4

## Additional context
The popover design has a **shadow popover** style from the [WordPress Design Library](https://www.figma.com/file/e4tLacmlPuZV47l7901FEs/WordPress-Design-Library).

### LINKED ISSUE #761: Horizontal scroll issue on Openverse main page, when viewing license information.
## Description
While surfing the Openverse filters/licenses, noticed a screen scroll towards the right, which is not ideal for a smooth and fluent experience.

## Reproduction
please, refer to the linked screen recording.
[Screen capture](https://drive.google.com/file/d/1E1dXAKyB3DG02xR-rb-fs9uu_EcwW4Fu/view?usp=sharing)

## Expectation
The pop-up should be visible, without having to scroll horizontally, and also the pop-ups should be visible without moving back to the left. 

## Screenshots
[Screen Capture](https://drive.google.com/file/d/1E1dXAKyB3DG02xR-rb-fs9uu_EcwW4Fu/view?usp=sharing) - attaching the link again, for reference.

## Environment

 - Device: Laptops, Desktops
 - OS: Windows 10
 - Browser: Chrome, Firefox

dhruvkb:
## Fixes

Fixes #761 by @Ad-Mk
Fixes #423 by @obulat
Fixes #541 by @panchovm 

## Description

This PR
- adds icons and explanation for deprecated CC licenses
- uses `VPopover` to render license explanations
- refactors old components using `defineComponent` and 'V-' prefix 
- replaces hardcoded data in store with calculated data

## Testing Instructions

1. Search for any term and open the filter sidebar.
2. Check that the license explanation appears inside the bounds of the window.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VFilters/VFilterChecklist.vue
```diff
@@ -20,40 +20,45 @@
...
-      :icon-dom-node="$refs[`${licenseExplanationCode}licenseIcon`][0]"
-    />
   </fieldset>
 </template>
 
 <script>
 import helpIcon from '~/assets/icons/help.svg'
 
-import VLicenseExplanationTooltip from '~/components/VFilters/VLicenseExplanationTooltip.vue'
+import VLicenseExplanation from '@/components/VFilters/VLicenseExplanation.vue'
```
obulat, line 48:
```suggestion
import VLicenseExplanation from '~/components/VFilters/VLicenseExplanation.vue'
```

src/components/VFilters/VLicenseExplanation.vue
```diff
@@ -0,0 +1,79 @@
...
+      </template>
+      <template #mark>{{ license.toUpperCase() }}</template>
+    </i18n>
+  </div>
+</template>
+
+<script>
+import { isLicense } from '~/utils/license'
+
+import VLicenseElements from '@/components/VLicenseElements.vue'
```
obulat, line 36:
```suggestion
import VLicenseElements from '~/components/VLicenseElements.vue'
```

src/components/VFilters/VLicenseExplanation.vue
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * the code of the license whose elements need to be explained
+     */
+    license: {
+      type: String,
+      required: true,
+    },
+  },
+  computed: {
+    isLicense() {
```
obulat, line 59: It's probably good to add some explanation for this, along the lines of:
"PDM and CC0 are not technically licenses, but instead they are marks of Public Domain dedication"

src/components/VFilters/VLicenseExplanation.vue
```diff
@@ -0,0 +1,79 @@
...
+      if (licenseTerm === 'cc0') {
+        fragment = 'publicdomain/zero/1.0'
+      } else if (licenseTerm === 'pdm') {
+        fragment = 'publicdomain/mark/1.0'
+      } else if (DEPRECATED_LICENSES.includes(licenseTerm)) {
+        fragment = `licenses/${licenseTerm}/1.0`
+      } else {
+        fragment = `licenses/${licenseTerm}/4.0`
+      }
+      return `https://creativecommons.org/${fragment}/?ref=openverse&atype=rich`
```
obulat, line 75: We can probably remove `atype=rich` parameter, as it refers to a link that is inside the rich type license description, and I'm not sure it's useful anyways.

src/store/search.js
```diff
@@ -75,16 +76,11 @@ export const mediaSpecificFilters = {
...
-    { code: 'cc0', name: 'filters.licenses.cc0', checked: false },
-    { code: 'pdm', name: 'filters.licenses.pdm', checked: false },
-    { code: 'by', name: 'filters.licenses.by', checked: false },
-    { code: 'by-sa', name: 'filters.licenses.by-sa', checked: false },
-    { code: 'by-nc', name: 'filters.licenses.by-nc', checked: false },
-    { code: 'by-nd', name: 'filters.licenses.by-nd', checked: false },
-    { code: 'by-nc-sa', name: 'filters.licenses.by-nc-sa', checked: false },
-    { code: 'by-nc-nd', name: 'filters.licenses.by-nc-nd', checked: false },
-  ],
+  licenses: ACTIVE_LICENSES.map((licenseCode) => ({
```
obulat, line 79: This is a great idea for cleaning up the `filterData` object!

obulat (changes requested):
This is a great improvement!

There are a couple of things that don't match the mockups:
- there is not 'Cross' close button at the top right corner
- the heading font size should be 14px

src/components/VFilters/meta/VLicenseExplanation.stories.mdx
```diff
@@ -0,0 +1,39 @@
+import {
+  ArgsTable,
+  Canvas,
+  Description,
+  Meta,
+  Story,
+} from '@storybook/addon-docs'
+
+import VLicenseExplanation from '@/components/VFilters/VLicenseExplanation.vue'
```
dhruvkb, line 9:
```suggestion
import VLicenseExplanation from '~/components/VFilters/VLicenseExplanation.vue'
```

src/components/VFilters/VFilterChecklist.vue
```diff
@@ -20,40 +20,45 @@
...
-      </button>
+
+      <!-- License explanation -->
+      <VPopover v-if="filterType === 'licenses'">
+        <template #trigger="{ a11yProps }">
+          <button
+            v-bind="a11yProps"
+            :ref="`${item.code}licenseIcon`"
+            :aria-label="$t('browse-page.aria.license-explanation')"
+            class="text-dark-charcoal-70 appearance-none"
```
sarayourfriend, line 31: Is `appearance-none` redundant? I think tailwind already applies that to `button` elements in its reset stylesheet.

src/components/VLicenseElements.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <ul>
+    <li
+      v-for="(element, index) in elements"
+      :key="index"
```
sarayourfriend, line 5: Are the elements unique? If so, can't we just use the `element` as the key rather than the index?

sarayourfriend (changes requested):
Thanks for adding the deprecated license icons! That's a great addition :tada: 

Navigating to an image result from the search page results in the app crashing. It does work if you reload the result page directly in SSR though. This appears to be an issue with `media.license` being undefined instead of a string.

In addition to what Olga pointed out, there's a couple other issues I noticed:

1. Keyboard navigation is mostly fine, but we should change the initial focus element to be the license explanation text rather than the Read more link
2. On mobile, opening the filters really messes with the layout of the filters on the page. It's also difficult to close the popover by tapping outside of it, which is probably more a Popover bug than it is an issue with this PR. I tested the popover component on mobile in Storybook and they worked fine, tapping outside of the popover closed it as expected. I'm wondering if there's some edge case with mobile Safari where tapping inside a fieldset doesn't register a regular click event or something? I can't reproduce it in Firefox for Android (neither the interaction issue nor the layout issue).

Here's a screenshot of the layout issue on mobile Safari. Note the license buttons are pushed all the way next to the license text.
IMG_1564

src/components/VFilters/VFilterChecklist.vue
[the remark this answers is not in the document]
dhruvkb: Already resolved in 09152b2.

src/components/VFilters/VLicenseExplanation.vue
dhruvkb: Add documentation comments in c302831.

src/utils/license.js
```diff
@@ -31,9 +31,8 @@ export const getFullLicenseName = (license, license_version) => {
 /**
  * CC licenses have different legal status from the public domain marks
  * such as CC0 and PDM, and need different wording.
- * Returns false if licenseOrMark is CC0 or PDM
  * @param {License|LicenseName|''} licenseOrMark
- * @return {boolean}
+ * @return {boolean} false if licenseOrMark is CC0 or PDM, true otherwise
```
sarayourfriend, line 35: Funny, I think we can actually do better with this return type, if you're interested :slightly_smiling_face: 

```suggestion
 * @return {asserts licenseOrMark is Exclude<'pdm' | 'cc0', License | LicenseName | ''>} false if licenseOrMark is CC0 or PDM, true otherwise
```

src/components/VFilters/VLicenseExplanation.vue
dhruvkb: Removed `atype` param in 17c8d26.

src/utils/license.js
dhruvkb: It's as if return types were their own little language!

src/components/VFilters/VFilterChecklist.vue
dhruvkb: Tailwind only applies these:
```css
-webkit-appearance: button;
background-color: transparent;
```
But your assessment that `appearance-none` doesn't do anything is correct. Removing or keeping it seems to have no effect on the UI.

src/components/VLicenseElements.vue
dhruvkb: Resolved in 0a25092.

src/components/VFilters/VFilterChecklist.vue
dhruvkb: Resolved in f69df84.

src/components/VFilters/VLicenseExplanation.vue
```diff
@@ -0,0 +1,83 @@
+<template>
+  <div class="license-explanation w-80 p-6">
```
zackkrida, line 2:
```suggestion
  <div class="license-explanation w-full max-w-80 p-6">
```

That's a bit friendlier to the 300-ish pixel screens we've recently improved. We should probably discuss and codify our device size support sometime!

zackkrida (approved):
What an accomplishment! This looks and behaves great. I added one non-blocking suggestion, but lgtm!

src/components/VFilters/VFilterChecklist.vue
sarayourfriend: Ah! I was thinking of `webkit-appearance`, by bad!

Did you test that in different browser by any chance? I wonder if `appearance-none` was doing any resets for Safari for example?

obulat (approved):
Great improvement!

dhruvkb:
@sarayourfriend the bug with click outside to hide was because of a `@click.stop` on the fieldset. Funnily enough, the goal of that handler was to hide the popover 😄. Removed that (and some more cruft) in a98e709.

src/components/VFilters/VFilterChecklist.vue
```diff
@@ -20,40 +20,54 @@
...
-      >
-        <VIcon :icon-path="helpIcon" />
-      </button>
+
+      <!-- License explanation -->
+      <VPopover v-if="filterType === 'licenses'">
+        <template #trigger="{ a11yProps }">
+          <button
+            v-bind="a11yProps"
+            :ref="`${item.code}licenseIcon`"
```
sarayourfriend, line 29: It doesn't look like we're using this ref anywhere anymore:

```suggestion
```

src/components/VFilters/VFilterChecklist.vue
```diff
@@ -20,40 +20,54 @@
...
+            :aria-label="$t('browse-page.aria.license-explanation')"
+            class="text-dark-charcoal-70"
```
sarayourfriend, line 31: This button is missing focus styles. We could add them here, or, alternatively just use the `VButton` with `variant="plain" size="disabled"`

sarayourfriend (changes requested):
Looking really good! There's just two bugs:

1. The iOS spacing issue, which I found a fix for in this PR: https://github.com/WordPress/openverse-frontend/pull/870
2. The close button focus styles

After that I think this will be good to go!

src/components/VFilters/VFilterChecklist.vue
dhruvkb: The focus interaction improved massively after this was changed in b13ea0c.

src/components/VFilters/VFilterChecklist.vue
dhruvkb: Good catch, fixed in b13ea0c.

sarayourfriend:
@dhruvkb the close button still needs focus styles applied. `VIconButton` unfortunately doesn't have them I guess (I'm not sure why `VIconButton` isn't implemented using `VButton` :thinking:)

src/components/VFilters/VLicenseExplanation.vue
```diff
@@ -0,0 +1,88 @@
...
+        $t('filters.license-explanation.license-definition')
+      }}</template>
+      <template v-else>{{
+        $t('filters.license-explanation.mark-definition', {
+          mark: license.toUpperCase(),
+        })
+      }}</template>
+    </h5>
+
+    <VLicenseElements
```
sarayourfriend, line 14: We should mark these as `aria-hidden` in this use case, otherwise the popover is extremely repetitive, saying the license name over and over again with no explanation why.

src/components/VPopover/VPopoverContent.vue
```diff
@@ -1,15 +1,20 @@
 <template>
   <div
     v-if="visible"
-    ref="popoverRef"
-    :class="[$style.popover, { [`z-[${zIndex}]`]: zIndex !== undefined }]"
+    class="w-0"
```
sarayourfriend, line 4: For whatever reason Safari positions the element very strangely when VoiceOver is open unless using v-show with also the height set to 0. No idea why. It does not appear to effect other browsers I tested.

```suggestion
    v-show="visible"
    class="w-0 h-0"
```

Edit: I'm not sure why this suggestion is rendering so broken. The only changes necessary are to use `v-show` and add `h-0`. The other changes showing in the suggestion please ignore.

src/components/VFilters/VFilterChecklist.vue
```diff
@@ -20,40 +20,55 @@
...
+            type="button"
+          >
+            <VIcon :icon-path="icons.help" />
+          </VButton>
+        </template>
+        <template #default="{ close }">
+          <div class="relative">
+            <VIconButton
```
sarayourfriend, line 40: This button needs an aria label describing it as a close button.

src/components/VFilters/VFilterChecklist.vue
```diff
@@ -20,40 +20,55 @@
...
-        :aria-label="$t('browse-page.aria.license-explanation')"
-        class="text-dark-charcoal-70 appearance-none"
-        type="button"
-        @click.stop="toggleLicenseExplanationVisibility(item.code)"
-      >
-        <VIcon :icon-path="helpIcon" />
-      </button>
+
+      <!-- License explanation -->
+      <VPopover v-if="filterType === 'licenses'">
```
sarayourfriend, line 25: Let's pass the `label` prop here with the value of the license definition. That way the popover label is read out when it is opened and we don't have to mess with moving the focus away from the default of the close button (which I think makes the most sense ergonomically to auto-focus when the popover is opened).

sarayourfriend (changes requested):
Okay I broke out my old macbook to test with VoiceOver. It would be helpful if other folks were also testing in this so that we caught bugs earlier, I feel bad holding up PRs this way finding last minute accessibility bugs.

zackkrida:
> I feel bad holding up PRs this way finding last minute accessibility bugs.

Please don't feel bad about this @sarayourfriend, moving forward this a11y testing will happen earlier and more thoroughly from everyone on @WordPress/openverse-frontend. Thank you so much for the time you've devoted to this PR.

src/components/VFilters/VFilterChecklist.vue
dhruvkb: I did not understand what you mean by license definition. A license can have several elements, each with their own definition (eg. BY-NC-SA with 3 elements & 3 definitions).

src/components/VFilters/VFilterChecklist.vue
dhruvkb: Added ARIA-label in [3f91c0e](https://github.com/WordPress/openverse-frontend/pull/850/commits/3f91c0e3354840c516c7026d2d8ee03e8ba99ae3).

src/components/VFilters/VLicenseExplanation.vue
dhruvkb: Made ARIA hidden in [3f91c0e](https://github.com/WordPress/openverse-frontend/pull/850/commits/3f91c0e3354840c516c7026d2d8ee03e8ba99ae3).

src/components/VFilters/VFilterChecklist.vue
sarayourfriend: Ah, you're right. We might need to find a compromise for screen reader folks to get that information quickly upon opening the popover. Maybe something to explore later on though :thinking: Otherwise the aria-label would be a huge wall of text for multi-element licenses.

dhruvkb: Yeah, we can separate this into a separate issue to at least move past the post-launch bugs milestone.

zackkrida:
@dhruvkb anything needed here to get this merged, or are you just looking for a final review from @sarayourfriend?

sarayourfriend:
`aria-hidden` is being applied on the entire license element explanation, not just the decorative icon.

The icon should have `role="presentation"` rather than `aria-hidden`, that's my bad for the incorrect recommendation.

Otherwise this is looking good.

src/components/VLicenseElements.vue
```diff
@@ -0,0 +1,67 @@
...
+      v-for="element in elements"
+      :key="element"
+      class="flex items-center gap-3 mt-2"
+    >
+      <VIcon
+        view-box="0 0 30 30"
+        :size="isSmall ? 5 : 6"
+        :icon-path="icons[element]"
+      />
+      <span class="sr-only">{{ element.toUpperCase() }}</span>
```
sarayourfriend, line 13: For `pdm` and `cc0` or any other "single element licenses/marks" this can be left out as it's redundant with the heading of the popover.

src/components/VFilters/VLicenseExplanation.vue
sarayourfriend: Sorry, when I made this comment I didn't realize that `VLicenseElements` rendered the explanation in addition to the icon. I totally misread this component, my fault.

We should remove `aria-hidden` from this. My other comment in this review addresses the specific concern I had.

zackkrida: @dhruvkb @sarayourfriend is this resolved?

dhruvkb:
@sarayourfriend please review this PR. If all's okay, I'll resolve the conflicts and merge it.

sarayourfriend:
I don't currently have access to VoiceOver or NVDA so if someone else can test that it would be ideal. Based on the code changes the problems I was concerned about should be addressed so I'm removing my request for changes.

panchovm:
I noticed two things:
1. I do not see the tooltip when pointer is over the question icon
2. The question icon should change to `dark-charcoal` in `hover`.

Here is [the mockup](https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=1109%3A23176) for reference and a screenshot of expected implementation. The component is from the [WordPress Design Library](https://www.figma.com/file/e4tLacmlPuZV47l7901FEs/WordPress-Design-Library?node-id=28524%3A24023).

CleanShot 2022-02-22 at 10 38 57@2x

sarayourfriend:
> I do not see the tooltip when pointer is over the question icon

We don't currently have a component to handle tooltips. Can we implement this feature in a separate PR? This one is already really long running. The icon does have an aria-label so it's accessible for screen readers at least.

The main reason I'd like to push the tooltip into a separate issue/PR is that tooltips have their own set of complicated a11y considerations that would probably delay this PR even more than it's already been. Whomever implements the tooltip component will need to do some research into the best practices for them.

panchovm:
Absolutely, that makes sense. Putting my comment aside, excellent work @dhruvkb, as always ✨

obulat:
We could add a tooltip like that by adding `title` attribute, but I'm not sure how accessible it is. Both the title and the button label seem to be read.

sarayourfriend:
> We could add a tooltip like that by adding title attribute, but I'm not sure how accessible it is. Both the title and the button label seem to be read.

IIRC SRs behave slightly differently when there's both `title` and `aria-label`, some read the title, some read the label, some do other funky things to I think.

sarayourfriend:
@panchovm can you verify if this PR happens to fix the issues you filed in https://github.com/WordPress/openverse-frontend/issues/759?

zackkrida:
Here's a video of VoiceOver in Safari, using keyboard navigation. Please turn the sound on for the full experience.
It seems that:

- the close button is focused by default in the popover
- the popover contents are not read to the user

https://user-images.githubusercontent.com/6351754/155174845-d6a7703b-129d-437d-a238-5a513801b4b6.mp4

sarayourfriend:
@zackkrida I think those are meant to be addressed in a separate issue as discussed in this conversation https://github.com/WordPress/openverse-frontend/pull/850#discussion_r809954854 (you need to expand the resolved conversation to see it).

zackkrida (approved):
LGTM. I've tried to read through the (extensive) history of updates since I approved this, and I believe everything is done. I created a new issue for the high-priority bug of screen readers not announcing the popover contents: #987

zackkrida:
@panchovm if you have an opportunity today, could you test this pull request and then open it on your iPad? You should see a permalink like `http://172.16.227.1:8443/` that will allow you to open the app running on your computer on your iPad. 

We have a theory that this pull request fixes the issue you were seeing in #759 

If @dhruvkb has merged this before you get to do this, you can test it with the `main` branch. Thank you!

dhruvkb:
@panchovm I merged this pretty quickly because it was repeatedly developing conflicts. You can test this on your iPad by directly opening search-staging.

panchovm:
I tested it on my iPad on Firefox and Safari and the header bug (#759) has gone. It works perfectly. Excellent work ✨

PR #852, opened 2022-02-10
Fix search bar button crashing index page

sarayourfriend:
## Fixes

Fixes #[issue number] by @[issue author]

## Description

The root cause of this was that the contents of the VButton and VIconButton were not both another element. The VButton was rendering a plain string and VIconButton rendered an SVG. When Vue tried to evaluate the tag name of the VButton when it was just the text, it would fail because it was just a text node which has no tag name.

This _could_ just be a one-line fix by wrapping the contents of the VButton in a `span` BUT I think the VSearchButton component is actually simpler this way, uses less markup and less CSS _and_ works better with SSR (no more flashing "Search" text or icon for most cases of mobile/desktop usage). The only people who still get a little flashing of the search/icon are users of non-mobile devices with narrow viewports.

Something is currently weird about the search button. I don't know why it doesn't have a pink background.

## Testing Instructions

Load the homepage on a mobile device browser and verify it doesn't crash. Also, verify going back to the homepage from the search page.

Also e2e and unit testss

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
In Desktop Safari I'm still seeing Teleport errors and the 404 page when going back or clicking the logo:

```
> Error: VTeleportTarget: skip-to-content beforeDestroy but still has children mounted
> Error: VTeleport: non-existent target skip-to-content on unmount
```

sarayourfriend:
> In Desktop Firefox I'm still seeing Teleport errors and the 404 page when going back or clicking the logo:

Oop, okay there's more crashing the back button, sorry!

src/constants/screens.js
```diff
@@ -1,6 +1,8 @@
+/** @typedef {'2xl' | 'xl' | 'lg' | 'md' | 'sm' | 'mob'} Breakpoints */
+
 /**
  * Mapping of a breakpoint name to the lower-bound of its screen-width range
- * @type {Map<string, number>}
+ * @type {Map<Exclude<Breakpoints, 'mob'>, number>}
```
obulat, line 5: !TIL we can declare types with Exclude!

typings/openverse/index.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module '*.svg' {
+    const SVG: string
+    export default SVG
+}
+
+declare module '*.svg!inline' {
+    const SVG: string
+    export default SVG
+}
```
obulat, line 9:
```suggestion
}

```

src/constants/screens.js
sarayourfriend: In the JS files all the great up-to-date TS features will be available: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

Inside the `{}` basically everything that would fit on the right hand side of a TypeScript `type X =` statement is valid. JSDoc has some extra goodies (if you can call them that) to make up for limitations in the syntax, like `const` which allows applying the `as const` notation to an object.

src/components/VTeleport/VTeleportTarget.vue
```diff
@@ -20,11 +21,17 @@ export default defineComponent({
     targets[this.name] = this
   },
   beforeDestroy() {
-    delete targets[this.name]
-    if (this.children.length > 0)
+    if (this.children.length > 0) {
+      if (this.forceDestroy) {
+        this.children.forEach((child) => child.$destroy())
```
obulat, line 26: I've read somewhere that if you call `$destroy()` yourself, you have to also manually delete the HTML element.

sarayourfriend: :open_mouth: Good call!

sarayourfriend:
For some reason, something I changed in this PR is inadvertently causing tailwind to insert its reset stylesheet _after_ the generated styles.

Screenshot_2022-02-11_10-51-59

Going to dig into why this might be happening... but the tailwind nuxt plugin is kind of a black box so I'm not sure how far I'll get :disappointed: If someone else could verify this is also happening for them that would be helpful though.

obulat:
> might be happening... but the tailwind nuxt plugin is kind of a black box so I'm not sure how far I'll get 😞 If someone else



> For some reason, something I changed in this PR is inadvertently causing tailwind to insert its reset stylesheet _after_ the generated styles.

Just a quick question: does that happen when you run `pnpm build`, too?

sarayourfriend:
@obulat good point. Just checked and it does not happen when doing a full production build.

src/components/VTeleport/VTeleportTarget.vue
```diff
@@ -8,6 +8,7 @@ export default defineComponent({
   props: {
     name: { type: String, required: true },
     element: { type: String, default: 'div' },
+    forceDestroy: { type: Boolean, default: false },
```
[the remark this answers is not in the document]
sarayourfriend: I think it was lost in the merge, yeah. I'll push a fix for it.

PR #867, opened 2022-02-15
Refactor media services

### LINKED ISSUE #866: Simplify media services
## Problem

Currently, we have two almost identical services for media searching: [ImageService](https://github.com/WordPress/openverse-frontend/blob/c31a971088762d730375ccbe872677d91e4cdb57/src/data/image-service.js) and [AudioService](https://github.com/WordPress/openverse-frontend/blob/c31a971088762d730375ccbe872677d91e4cdb57/src/data/audio-service.js), both of which also use [BaseMediaService](https://github.com/WordPress/openverse-frontend/blob/c31a971088762d730375ccbe872677d91e4cdb57/src/data/base-media-service.js). 

## Description

Using one generic `MediaService` with a `mediaType` parameter would simplify the code and make it more maintainable, as well as simplify the addition of new media types in the future.

## Alternatives

An alternative would be to leave services separate. This would allow us to add methods that only exist for one media type to a single service. However, it would cause some inconveniences because:
- adding a new media type would require to create a new media service.
- a change in the API would require to change all services.

## Implementation

obulat:
## Fixes

Fixes #866 by @obulat

## Description

This PR joins three separate modules for `media-service`s into one generic `MediaService` that can be created with a `mediaType` parameter. All of the methods for both `audio` and `image` are the same. This change will allow us to simplify media fetching, and make the store conversion to Pinia and addition of other media types easier.

## Testing Instructions

Run the app as usual, and check that you can search for all media types. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
This is great Olga! I haven't tested the code locally yet but I did take it for a spin to [do some fun TypeScript hoop-jumping](https://github.com/WordPress/openverse-frontend/pull/868) to make the MediaStore a generic class based on the `mediaType` in the constructor. Take it or leave it, up to you :slightly_smiling_face:

obulat:
> This is great Olga! I haven't tested the code locally yet but I did take it for a spin to [do some fun TypeScript hoop-jumping](https://github.com/WordPress/openverse-frontend/pull/868) to make the MediaStore a generic class based on the `mediaType` in the constructor. Take it or leave it, up to you 🙂

Thank you, Sara, your PR is amazing! I've left my [notes there](https://github.com/WordPress/openverse-frontend/pull/868#pullrequestreview-882930040), and added some changes from it here.

src/data/api-service.js
```diff
@@ -1,8 +1,20 @@
...
+import { AUDIO, IMAGE } from '~/constants/media'
 
 const DEFAULT_REQUEST_TIMEOUT = 30000
 
+/**
+ * Returns a slug with trailing slash for a given resource name.
+ * For media types, converts the name into resource slug when necessary (i.e. pluralizes 'image'),
+ * for other resources uses the resource name as the slug.
+ * @param {import('../constants/media').MediaType|string} resource
+ * @returns {*}
```
sarayourfriend, line 12: The `*` (existential type) is basically `any` and will swallow the actual return type, which here is always `string`. Also, something to note, doing `AnyType | AWiderType` will always reduce to just `AWiderType`, so the `param` annotation will just reduce to `string`, which isn't as helpful. We can either leave out the `MediaType` annotation or update the call sites of this function to specify that the input is always `MediaType` (for example update type of `resource` on the `query` function to be `MediaType`).

```suggestion
 * @param {import('../constants/media').MediaType} resource
 * @returns {string}
```

src/store/types.d.ts
```diff
@@ -129,24 +132,20 @@ export interface ActiveMediaState {
...
-}
+export interface MediaStoreResult<T extends FrontendMediaType>
+  extends MediaResult<Record<MediaDetail['id'], T>> {}
 
 export interface MediaState {
   results: {
-    audio: MediaStoreResult
-    image: MediaStoreResult
+    audio: MediaStoreResult<AudioDetail>
+    image: MediaStoreResult<ImageDetail>
```
sarayourfriend, line 141: These need to refer to a `FrontendMediaType` value.

```suggestion
    audio: MediaStoreResult<'audio'>
    image: MediaStoreResult<'image'>
```

sarayourfriend (approved):
LGTM! Tested locally and it works great. Left a couple more subtle TypeScript things but given the files aren't added to `tsconfig` yet anyway it's no big deal to leave them out, we'll just have to fix them later when we add them to `tsconfig`.

sarayourfriend:
@WordPress/openverse-frontend can we get a second review on this? :slightly_smiling_face:

src/data/api-service.js
```diff
@@ -1,8 +1,20 @@
...
+ * Returns a slug with trailing slash for a given resource name.
+ * For media types, converts the name into resource slug when necessary (i.e. pluralizes 'image'),
+ * for other resources uses the resource name as the slug.
+ * @param {string} resource
+ * @returns {string}
+ */
+const getResourceSlug = (resource) => {
+  const slug = { [AUDIO]: 'audio', [IMAGE]: 'images' }[resource] ?? resource
+  return `${slug}/`
+}
```
zackkrida, line 17: nice fix here.

zackkrida (approved):
The new abstraction is quite nice, and the types are fantastic. Some non-blocking questions:

- Should we add the `peaks` key here to the `AudioDetail` interface? Since it will be used for the waveforms soon and is already accessed in the `VWaveform` component implementations.
- How will we handle if a media type has a one-off endpoint, like the `/waveform` endpoint for audio or future endpoints that, say, 3D models might need?

sarayourfriend:
> How will we handle if a media type has a one-off endpoint, like the /waveform endpoint for audio or future endpoints that, say, 3D models might need?

Subclass the `MediaService`:

```js
class Model3DService extends MediaService<'3d models'> {
  constructor() {
    super('3d models')
  }

  getCustomData() {
    // impl
  }
}
```

PR #868, opened 2022-02-15
Make MediaService generic

sarayourfriend:
## Description

Just some optional type enhancements to @obulat's excellent MediaService refactor.

## Testing Instructions

Create a test media service in `media-service.js` like this and inspect the return types of the various functions. Ensure they are as expected!

```js
const testService = new MediaService('image')
testService.getRelatedMedia({ id: 'asdf' })
    .then(({ data }) =>
        data.map(r =>
            /** `r` should be either ImageDetail or AudioDetail depending on what you passed to the MediaService constructor :o */
        )
    )
```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/constants/license.js
```diff
@@ -1,22 +1,45 @@
...
   'by-nc-nd',
-]
+])
+
+/**
+ * @template T
+ * @typedef {T extends `${infer P}-${infer PP}-${infer PPP}` ? [P, PP, PPP] : T extends `${infer P}-${infer PP}` ? [P, PP] : [T]} PartitionLicense
+ */
+
+/** @typedef {PartitionLicense<CCLicense>[number]} LicensePart */
```
sarayourfriend, line 15: This is just me being kind of silly with TS string types. We could just as easily (and probably should just) use a type directly as `'by' | 'sa' | 'nd' | 'nc'` :stuck_out_tongue_closed_eyes:

obulat: I tend to agree with your comment 😆  I mean it looks very interesting, but it would probably take me years to understand what it means :)

src/constants/license.js
```diff
@@ -1,22 +1,45 @@
-export const CC_LICENSES = [
+export const CC_LICENSES = /** @type {const} */ ([
```
obulat, line 1: @sarayourfriend , can you give me a reference I can read about the `type {const}` declaration? I couldn't find anything searching for `typescript const type`.

The reason I'm asking this is the messages I see in my IDEs.
**WebStorm** is trying to find something in the `node_modules`, I guess?

Screen Shot 2022-02-15 at 3 14 24 PM

**VS Code** just shows it as unresolved, even when I set language mode to TypeScript (I don't know what other information can I get about VS Code TS hints)

Screen Shot 2022-02-15 at 3 14 45 PM

sarayourfriend: It's the JSDoc version of this feature: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

The TypeScript JSDoc documentation for it is here: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#casts (the second example in that section)

I'm not sure why TS isn't picking it up. In VSCode have you tried running the "Restart TS Service" command? I have to do that anytime `tsconfig.json` changes.

For WebStorm it might be the same. You will probably also have to enable the TS language server using the instructions described here: https://www.jetbrains.com/help/webstorm/typescript-support.html#ws_ts_use_ts_service_checkbox

obulat (commented):
Thank you for this PR, @sarayourfriend ! It's fantastic, and great for learning more about types, as always!

I have taken all the changes that don't create warnings in my IDE and pushed them to the original PR. There are 2 things I don't think we should tackle here:
1. licenses - Dhruv has made significant changes to the license code in #850, so it would create double work load to add typing here.
2. media type constant types - they have been changed a lot in #851, so the same here.
Also, I would really like to find out what's wrong with my setup with regards to the `{const}` type, but until that I would prefer not to add it. 
media-service raises errors when I run `pnpm run types`, probably because I have not added the `mediaType`-related constant types, so I didn't add  it and `licenses` to the list of files to check types in.

sarayourfriend:
Woohoo! :tada: Makes sense to leave that stuff out for now :+1: We definitely need to sort out why the const cast isn't working for you though, it's a fundamental tool in good TypeScript!

PR #871, opened 2022-02-15
Add VLink component for external links

### LINKED ISSUE #468: External links NOT opening in the Openverse iframe
## Description

On the single image result page, a click on a creator's link causes error in an iframe due to x-frame settings.

## Reproduction

1. Go to any single result page, for example [this one](https://wordpress.org/openverse/photos/20708f1c-0978-47ab-8855-e01f3a392ab3)
2. Click on the creator's name (pricklysarah)
3. See an error:
 - in Firefox the iframe shows a warning "Firefox Can’t Open This Page. To protect your security, www.flickr.com will not allow Firefox to display the page if another site has embedded it. To see this page, you need to open it in a new window.", with a button to open the page in a new window
 - in Chrome, the iframe is gray and seems to be stuck, but there is a faint text saying "www.flickr.com refused to connect."

## Expectation

The link should open. I think the best option would be to open it in a new tab. Another option is to set the target to the top window, so that the link opens in the same tab

## Screenshots

image

## Additional context

Because the iframe height was previously set to a high number (around 3000px), it is also hard to see the error message in Chrome, as it appears much lower than the fold.

## Resolution

obulat:
## Fixes

Fixes #468 by @obulat

## Description

This PR adds a simple `VLink` component that is a wrapper over an `a` anchor with the `target="_blank" rel="noopener noreferrer"` added to it, and replaces all external `a` links in the components with `VLink`.

This ensures that the links can be opened from the iframe, and reduces the cognitive load of having to remember to add the link properties every time.

An interesting thing I learned: [`target="_blank"` is not the same as `target="blank"`](https://css-tricks.com/targetblank/), with the latter opening all the links in a single new tab, and the former opening each new link in a new tab. I think we want the former behavior, but this is open for a discussion.
Also, in the modern browsers, `rel="noopenver"` is automatically added to all the links with `target="_blank"`, but it's nice to have it for older browsers.

I did not _initially_ replace the `DownloadButton`'s `<a>`. Trying out the DownloadButton on the production site, wordpress.org/openverse, with a sound from Freesound, I got a similar error screen as in the linked issue, so I used the `VLink` component for `DownloadButton` as well. But then I realized that the problem is caused by me not being signed in to Freesound, Jamendo audios _can_ be downloaded from the deployed site. But I reverted the change anyway because the tests fail. If we decide to change it, we can do it in a new PR.

**Unrelated bug-fix:** removes the bug-report state that was removed in a recent PR from the `feedback` page.

## Testing Instructions

Running the app, you should see all the external links working properly and opening in a new tab.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/pages/meta-search.vue
+++ b/src/pages/meta-search.vue
@@ -9,9 +9,9 @@
 
       <i18n path="meta-search-page.intro" tag="p" class="mb-4">
         <template #link>
-          <a aria-label="sources" href="/sources">{{
+          <NuxtLink aria-label="sources" :to="localePath('/sources')">{{
             $t('meta-search-page.link')
-          }}</a>
+          }}</NuxtLink>
         </template>
       </i18n>
       <p>{{ $t('meta-search-page.license') }}</p>
@@ -83,16 +83,16 @@
       </h2>
       <i18n path="meta-search-page.new.content" tag="p" class="mb-2">
         <template #issue>
-          <a
+          <VLink
             aria-label="issue"
             href="https://github.com/creativecommons/cccatalog/issues/new?assignees=&labels=awaiting+triage%2C+ticket+work+required%2C+providers&template=new-source-suggestion.md&title=%5BSource+Suggestion%5D+Insert+source+name+here"
-            >{{ $t('meta-search-page.new.issue') }}</a
+            >{{ $t('meta-search-page.new.issue') }}</VLink
           >
         </template>
         <template #email>
-          <a aria-label="email" href="mailto:openverse@wordpress.org">{{
+          <VLink aria-label="email" href="mailto:openverse@wordpress.org">{{
             $t('meta-search-page.new.email')
-          }}</a>
+          }}</VLink>
         </template>
       </i18n>
       <h2 class="mt-10 mb-4 text-2xl">
@@ -116,14 +116,18 @@
 </template>
 
 <script>
-export default {
+import { defineComponent } from '@nuxtjs/composition-api'
+import VLink from '~/components/VLink.vue'
+
+export default defineComponent({
   name: 'MetaSearchPage',
+  components: { VLink },
   head() {
     return {
       title: `${this.$t('meta-search-page.title')} | ${this.$t('hero.brand')}`,
     }
   },
-}
+})
 </script>
 
 <!-- Add "scoped" attribute to limit CSS to this component only -->
--- a/src/pages/meta-search.vue
+++ b/src/pages/meta-search.vue
@@ -9,9 +9,11 @@
 
       <i18n path="meta-search-page.intro" tag="p" class="mb-4">
         <template #link>
-          <NuxtLink aria-label="sources" :to="localePath('/sources')">{{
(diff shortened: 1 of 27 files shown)


src/pages/meta-search.vue
```diff
@@ -9,9 +9,9 @@
 
       <i18n path="meta-search-page.intro" tag="p" class="mb-4">
         <template #link>
-          <a aria-label="sources" href="/sources">{{
+          <NuxtLink aria-label="sources" :to="localePath('/sources')">{{
```
sarayourfriend, line 12: If we're already here, can we go ahead and translate the `sources` string in the aria-label if you don't mind?

src/pages/meta-search.vue
```diff
@@ -83,16 +83,16 @@
       </h2>
       <i18n path="meta-search-page.new.content" tag="p" class="mb-2">
         <template #issue>
-          <a
+          <VLink
             aria-label="issue"
             href="https://github.com/creativecommons/cccatalog/issues/new?assignees=&labels=awaiting+triage%2C+ticket+work+required%2C+providers&template=new-source-suggestion.md&title=%5BSource+Suggestion%5D+Insert+source+name+here"
```
sarayourfriend, line 88: Ooh, we should update this to Openverse!

sarayourfriend (approved):
Looks awesome! Love this kind of thing :tada: 

Left a couple of non-blocking suggestions of things we can fix up if you want.

Additionally: in VButton we log a warning for `a`'s that are missing `href` attribute. It'd be nice to transport that assurance to this component as well, just to catch accidental places where we bind it to a value and it turns up undefined, null, or as an empty string or something.

It might also be nice to throw a warning if we catch a non-external link being passed as those should all use `NuxtLink` anyway, right?

Finally: some simple unit tests would be nice as well :slightly_smiling_face: 

None of that are blockers though :+1:

krysal (commented):
Nice, looking at this we save several lines using a component. I wonder if it would be a good idea to make it more general and let it use an anchor or `NuxtLink` when it's an internal or external link, as it is already done elsewhere, but I don't remember exactly where.

If not, don't you think we should rename it to `ExternalLink` or something like that?

sarayourfriend:
@krysal do you think we should have it auto-detect when a relative path vs external path is passed in? It could automatically call to `localePath` and stuff like we're doing on all the `NuxtLink`s anyway without having to write that everywhere as well.

obulat:
Thank you for your great suggestions, @sarayourfriend  and @krysal ! I decided to open a new PR #879 for a larger VLink component for both external and internal links, as Krystle suggested.
I've moved the warning about missing `href` attribute from `VButton` to `VLink`, and added some unit tests (and shamelessly copied some from VButton)

> It might also be nice to throw a warning if we catch a non-external link being passed as those should all use NuxtLink anyway, right?
No need for a warning in a component that wraps both :)

PR #879, opened 2022-02-16
Make VLink component that wraps around both external and internal links

### LINKED ISSUE #468: External links NOT opening in the Openverse iframe

obulat:
## Fixes

Fixes #468 by @obulat

## Description

This PR supercedes #468, based on @krysal 's [comment about wrapping internal links, too](https://github.com/WordPress/openverse-frontend/pull/871#pullrequestreview-883719330).

Now, `VLink` is a wrapper component that should be used for all links. It determines whether the link is internal or external by checking if the `href` attr starts with `/`.

**External links:**
- use `a` element.
- add `target='_blank'` that allows the links to open in another tab, instead of causing errors in an iframe.
- adds `rel='noopener noreferrer' to prevent some referrer vulnerabilities in older browsers (`target='_blank'` is enough for modern browsers)

**Internal links:**
- use `NuxtLink` component.
- wrap the `href` in `app.localePath` to ensure that the correct locale is used, and pass it as the `to` prop to `NuxtLink`.

This PR moves the warning for missing `href` from `VButton` to `VLink`, so `VButton` doesn't have to handle the lack of `href` property.

## Testing Instructions

Run the app and test out internal links and external links. Also check that the links in the navbar (pages) work as expected. External links should always open in a new tab.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VLink.vue
```diff
@@ -0,0 +1,46 @@
...
+    const linkComponent = computed(() => (isInternal.value ? 'NuxtLink' : 'a'))
+    const { href, ...otherProperties } = attrs
+
+    // No need to declare `href` as an explicit prop as Vue preserves
+    // the `attrs` object reference between renders and updates the properties
+    // meaning we'll always have the latest values for the properties on the
+    // attrs object
+    if (!href || href === '#') {
+      warn('VLink is missing a valid `href` attribute.')
+    }
```
sarayourfriend, line 32: Just a nit, but this will only warn on the first render of the component. That _might_ be okay but it could be more useful to watch `attrs.href` and warn any time it's an invalid value.

src/components/VLink.vue
```diff
@@ -0,0 +1,46 @@
...
+import { computed, defineComponent, useContext } from '@nuxtjs/composition-api'
+import { warn } from '@/utils/warn'
+const defaultProps = { target: '_blank', rel: 'noopener noreferrer' }
+export default defineComponent({
+  name: 'VLink',
+  setup(_, { attrs }) {
+    const { app } = useContext()
+    const isInternal = computed(() => attrs.href?.startsWith('/'))
+    const linkComponent = computed(() => (isInternal.value ? 'NuxtLink' : 'a'))
+    const { href, ...otherProperties } = attrs
```
sarayourfriend, line 24: Destructuring these from `attrs` will make their values non-reactive in the `computed` for `linkProperties` below. To maintain reactivity they need to stay as properties on `props` or `attrs`.

test/unit/specs/components/v-link.spec.js
```diff
@@ -0,0 +1,60 @@
...
+import VLink from '~/components/VLink.vue'
+
+jest.mock('~/utils/warn', () => ({ warn: jest.fn() }))
+
+const nuxtLinkStub = {
+  template: '<a v-on="$listeners"><slot /></a>',
+  props: ['to'],
+}
+const nuxtContextMock = {
+  $nuxt: { context: { app: { localePath: (v) => v } } },
```
sarayourfriend, line 12: If we change `localePath` to a `jest.fn()` we could test that it get's called with the `to` value for internal links as well.

sarayourfriend (commented):
LGTM! Still need to test this locally but I'm still excited for this :tada: The only blocker right now is the `attrs` destructuring... I wish there was an eslint rule to warn against that, I feel like 99% of the time it's probably not the right thing to do. Maybe that's wrong though.

sarayourfriend:
I noticed three more `<a>`'s in the app that could use this:

1. DownloadButton (It's still using the `DropdownButton` component which doesn't use `VButton` underneath)
2. FooterSection: link to 4v licenses
3. FooterSection: link to FontAwesome

src/components/VLink.vue
```diff
@@ -0,0 +1,58 @@
...
+export default defineComponent({
+  name: 'VLink',
+  setup(_, { attrs }) {
+    const { app } = useContext()
+    const isInternal = computed(() => attrs.href?.startsWith('/'))
+    const linkComponent = computed(() => (isInternal.value ? 'NuxtLink' : 'a'))
+
+    let linkProperties = {}
+    watch(
+      [attrs],
```
sarayourfriend, line 34: Does watching `attrs` like this actually work? :thinking: When I test this locally it doesn't seem to fire the watcher when the attrs change (even though it does re-render). You can see in this example where clicking the button causes the re-render but doesn't fire the watcher to update the value: https://codesandbox.io/s/crazy-rosalind-y12vp4

I think attrs is probably the wrong tool for this anyway... `href` should just be a required string prop, and then we can add the `warn` in the validator function for it.

`watch` _will_ work on props, so we can successfully keep linkProperties updated (in fact, it can just be a `computed` at that point.

obulat: Right! Having href as a prop will ensure that it's not also passed as an attr to NuxtLink! So, there will be no need to destructure the attrs.

src/components/VLink.vue
```diff
@@ -6,50 +6,34 @@
...
 const defaultProps = { target: '_blank', rel: 'noopener noreferrer' }
 
 export default defineComponent({
   name: 'VLink',
-  setup(_, { attrs }) {
+  props: {
+    href: {
+      type: String,
+      required: true,
+      validator: (v) => !['', '#'].includes(v),
```
sarayourfriend, line 25: Nice :+1:

sarayourfriend (commented):
The MediaTag changes seem unrelated. Was it a faulty merge perhaps? :thinking:

src/components/VLink.vue
```diff
@@ -6,50 +6,34 @@
...
-          : {
-              ...otherProperties,
-              ...defaultProps,
-              href,
-            }
-      },
-      { immediate: true }
+    let linkProperties = computed(() =>
+      isInternal.value
+        ? { to: app?.localePath(props.href) ?? props.href }
```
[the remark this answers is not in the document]
obulat: When testing 😆 That was probably a lazy hack to ensure tests pass without mocking the nuxt context :)

obulat:
> The MediaTag changes seem unrelated. Was it a faulty merge perhaps? 🤔

Do you mean `src/components/VMediaTag/meta/VMediaTag.stories.mdx` changes, @sarayourfriend ?

This PR is kind of very small: adding one component, but it touches so many files!

src/components/VLink.vue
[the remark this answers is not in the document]
sarayourfriend: Oh dear! I wish it was easier to set up a globally enabled mock of the nuxt stuff :thinking:

krysal:
Relabelling as "priority: medium" as the linked issue is actually fixed in the main branch and this is more of an improvement.

sarayourfriend:
Oh sorry, I was looking specifically at the merge commit, not the full diff for the PR :facepalm: 

Anyway, it looks like making `props.href` required breaks client-side navigation of the result page :tada: I guess in the interest of not trying to fix too many things at once here we could just make the links wait until `href` is defined to render?

obulat:
> Relabelling as "priority: medium" as the linked issue is actually fixed in the main branch and this is more of an improvement.

I am ok with relabelling, but just want to note that the issue was about all the external `<a>` links, not just the homepage licenses. Try clicking on 'Go to source', or links in the pages menu on search-staging. The links will be opened in the same tab, and with the iframe on w.org, this will break.

src/components/VLink.vue
```diff
@@ -21,19 +23,26 @@ export default defineComponent({
   props: {
     href: {
       type: String,
-      required: true,
-      validator: (v) => !['', '#'].includes(v),
```
sarayourfriend: nit: I think we can keep the validator, otherwise accidental long-lived invalid `href` values won't ever get caught.

sarayourfriend (approved):
LGTM! Things no longer crash because of this (there are still _other_ crashes but unrelated to this change :+1:)

sarayourfriend:
@obulat a solution I've used in past projects to easily identify "bad links" is to add some kind of horrific inline-styles to them that make them obviously stand out. Something like `style="background-color: hotpink; border: 2rem solid green;"`

zackkrida (approved):
lgtm! I've clicked around pretty extensively and text searched for raw `<a>` tags, in the repo, and didn't find anything that shouldn't be there.

Question for @WordPress/openverse-frontend: Does anyone know of an eslint rule we could introduce to warn on raw `<a>` tag usage in our Vue templates?

sarayourfriend:
@zackkrida The [`no-restricted-syntax`](https://eslint.org/docs/rules/no-restricted-syntax) rule can probably do it.

krysal (approved):
@obulat You were right! I forgot about the untouched links on internal pages. It's great we can now be safe about external links with these changes :rocket:

zackkrida:
@obulat I made a small companion PR with an eslint rule to enforce `VLink` usage:

https://github.com/WordPress/openverse-frontend/pull/916

PR #880, opened 2022-02-16
Fix/back button handling

### LINKED ISSUE #663: Browser back button doesn't resubmit previous search
## Description

This is true of the live site and the development site...make any search, then change the search term, and hit the browser's back button. The 2nd search term will show but the images are from the 1st term results. This is a state management issue and not specific to any browser or OS.

## Resolution

### LINKED ISSUE #856: Using the back button to navigate from the images search to the all content search results in only images showing
## Description

In general the funcationality around the back button on the search page is very fragile and has lots of bugs. This is just one of them.

## Reproduction

1. Go to a search page like `/search/?q=galah`
2. Click on the "images" content type at the top
3.  Use the back button to navigate back to the "all' search
4. Note the lack of audio results or even the option

## Expectation

It should go back to the all search

## Resolution

sarayourfriend:
## Fixes

Fixes #663 by @zackkrida
Fixes #856 by @sarayourfriend

## Description

Basically queryToFilterData was getting passed the state.filters for the "default filters" right? So when it's working through the list of filter types there were two scenarios where it wasn't updating the filter value, basically any time it was getting set off based on just the URL (i.e., not explicitly set off by checking the filter's checkbox):

1. If you have a filter group and two of the filters in the group is checked, going back to when only one of the filters is checked doesn't toggle off the filter because it's using the existing state and queryToFilterData only ever toggled filters on if they existed in the query. Because it doesn't exist in the query, it was getting ignored completely and left on the "default" which should be the "off" state BUT because we're passing in state.filters instead of an actual default filter state, it was getting left in the "on" state.
2. If the the filter group only had a single filter applied and you go back to remove that single filter, then queryToFilterData completely ignores it because the filter group doesn't even exist in the query. Just like in the first case it assumes the default filter state is for everything to be off, but we're passing in state.filters so it's re-using the previous state where those filters are still on.

So there's two ways of fixing this I think, one maybe more correct than the other:

1. We could update queryToFilterData to set everything that isn't mentioned to checked = false. This is probably the "hacky" way.
2. We could fix calls to queryToFilterData to pass in the correct "default", i.e., the actual default filter state of everything being off, not the "previous" filter state of state.filters.

This PR implements the second "non-hacky" solution.

## Testing Instructions

Follow the reproduction steps in the two linked issues and verify that they no longer occur.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/store/search.js
```diff
@@ -400,7 +398,7 @@ const actions = {
     const newFilterData = queryToFilterData({
       query,
       searchType,
-      defaultFilters: state.filters,
+      defaultFilters: filterData,
```
sarayourfriend, line 401: This is the actual fix. The rest is all support/testing/validation.

src/store/search.js
```diff
@@ -81,7 +82,7 @@ export const mediaSpecificFilters = {
 }
 
 /** @type {import('./types').Filters} */
-export const filterData = {
+export const filterData = deepFreeze({
```
sarayourfriend, line 85: Adding this because (a) we treat this data as immutable and (b) this gives us assurances that it isn't being inadvertently changed.

sarayourfriend:
The e2e tests pass locally for me :thinking: Not sure why this is failing in the CI :disappointed:

src/store/search.js
obulat: The problem here is the provider filters. They are set after receiving a response from the API with all the provider data, and in the search state initialization they are blank.

src/store/search.js
obulat: We need to add provider filters to this object, because originally filterData doesn't have it, and it is downloaded in nuxtServerInit. Actually, since the number of providers doesn't change that often we could hard-code it, and update it whenever a new provider is added.

obulat (approved):

obulat (changes requested):
I accidentally clicked on _Approve_ instead of _Request changes_ after writing the previous comment, and now I can't select _Request changes_ without a comment ...

sarayourfriend:
@obulat do you have a way of verifying that that issue is fixed? I wasn't able to produce any bugs I could notice previously so I'm just guessing at the fix for it.

obulat (approved):

obulat:
It works perfectly, @sarayourfriend ! Sorry for late re-review

src/composables/use-load-more.js
```diff
@@ -28,7 +28,7 @@ export const useLoadMore = (props) => {
   })
 
   const onLoadMore = async () => {
-    if (canLoadMore) {
+    if (canLoadMore.value) {
```
obulat, line 31: Oh no! 🙈 
I wish there was a way of preventing this error (not adding `.value` to a check of a ref/computed). I have spent a couple of hours of my life debugging because of this.

src/store/search.js
zackkrida: I'm not seeing any issues related to setting providers, either client-side in the filters or when setting them by 'default' in the ssr-ed url.

src/store/search.js
```diff
@@ -130,6 +131,20 @@ export const filterData = {
...
+    return state.filters[`${mediaType}Providers`].map((provider) => ({
+      ...provider,
+      checked: false,
+    }))
+  }
+  return {
+    ...clonedeep(filterData),
+    audioProviders: resetProviders(AUDIO),
+    imageProviders: resetProviders(IMAGE),
+  }
```
zackkrida, line 147: nice

src/utils/deep-freeze.js
```diff
@@ -0,0 +1,24 @@
...
+ * @template {unknown} T
+ * @param {T} object
+ * @returns {import('@nuxtjs/composition-api').DeepReadonly<T>}
+ */
+export function deepFreeze(object) {
+  // Retrieve the property names defined on object
+  const propNames = Object.getOwnPropertyNames(object)
+
+  // Freeze properties before freezing self
+
```
zackkrida, line 12:
```suggestion
  // Freeze properties before freezing self
```

mega nit.

zackkrida (approved):
This behaves perfectly for me cross-browser (safari, firefox, and chrome) on mesktop and on mobile in Android FF and Chrome. I wouldn't expect cross-browser inconsistencies here anyway.

The deep freeze utility is a nice addition and I appreciate the new e2e tests. Do you want to make a separate issue for improving those with actual assertions that the image results have updated?

lgtm!

sarayourfriend:
> The deep freeze utility is a nice addition and I appreciate the new e2e tests. Do you want to make a separate issue for improving those with actual assertions that the image results have updated?

I thought about this but there's no way to do this that doesn't break the E2E tests as soon as we ingest new data and the results change. Once we have #881 merged maybe we can evaluate that as an option.

PR #881, opened 2022-02-16
Use talkback proxy to mock e2e API requests

### LINKED ISSUE #499: SSR request mocking on E2E tests
## Problem

We need o find a way to make the Playwright tests more robust without depending on the live API, as it's currently used (the app must be running in order to run the e2e tests).

Tests for the image detail page, for example, require an `id` of a specific image to load the page, solving this will allow to make them with more specific assertions and run faster.

## Alternatives

Continue to use the live API as currently, but this makes the test unstable as the network requests are not always reliable.

## Implementation

obulat:
## Fixes

Fixes #499 by @krysal

## Description

This PR adds a `talkback` proxy server that listens to the API requests during e2e testing. It saves the response data to `test/tapes` to be used on later requests.

This enables us to run e2e tests with real data, but without the internet access :)

I've also added a couple of pnpm scripts:
- `ci:e2e-test` runs a bash script that is to be run by the CI. It uses bash because our CI runs on Linux only. It runs the proxy server, builds the app for production, starts it and runs all the e2e tests.
- `update-tapes` removes all saved tapes, builds the server and runs the e2e tests, saving the responses to `test/tapes`.
There currently are no scripts that can be run with proxy server on Windows.
While testing on Windows I found that setting `--hostname=0` will not allow to build the app on Windows, so I changed it to cross-platform `0.0.0.0` value.

## Testing Instructions

1. Checkout the branch
2. `pnpm i` to install the new `talkback` dependency
3.  run `pnpm run ci:e2e-test`. You should see the proxy and build and test logs, and finally, a message that the tests were successful.
4. If you delete a couple of files in `test/tapes` and try to run again; some tests that use those tapes will fail.
5. Now, you can run `pnpm run update-tapes` and see that new tapes are added, and all the e2e tests pass.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


package.json
```diff
@@ -22,8 +22,9 @@
     "test:watch": "jest --collectCoverage=false --watch",
     "test:unit": "jest --collectCoverage=false",
     "test:e2e": "playwright test",
+    "test:e2e-ci": "start-server-and-test run-server http://localhost:8443 test:e2e",
     "run-server": "cross-env nuxt build && pnpm start",
-    "e2e:ci": "start-server-and-test run-server http://localhost:8443 test:e2e",
+    "e2e:ci": "cross-env API_URL=http://localhost:3000 node proxy.js & TALKBACK_PID=$! && sleep 1 && pnpm run test:e2e-ci ; kill -15 $TALKBACK_PID",
```
sarayourfriend, line 27: We might as well not use `cross-env` here, or really just use a proper bash script. This shouldn't get run outside of the CI so it doesn't need to be cross-platform compatible and the `&`, `&&` and `kill` commands I don't believe are cross-platform anyway.

proxy.js
```diff
@@ -0,0 +1,28 @@
...
+ * The talkback proxy for e2e tests. When making a request to the API during testing,
+ * it tries to use the responses it previously saved in `/test/tapes` folder. If no
+ * response is found there, it makes an actual request, and saves the response for
+ * future use.
+ * This makes it possible for the e2e tests to run without internet, and makes the
+ * tests less flaky due to changes in the API or API data.
+ */
+const talkback = require('talkback')
+const path = require('path')
+const host = 'https://api.openverse.engineering/v1'
```
sarayourfriend, line 11: Let's just catch everything going to the API, otherwise things like analytics (which isn't version-prefixed) will fall through.

```suggestion
const host = 'https://api.openverse.engineering'
```

obulat:
> Is it common not to save the tapes in the repository? If we don't then CI will never have them.

Oops, I was sure I'd pushed the tapes :) 

> It would also be nice to have an "update-tapes" script that clears the tapes directory then runs the full e2e suite.
Great idea!

test/proxy.js
```diff
@@ -0,0 +1,26 @@
...
+
+const opts = {
+  host,
+  port: 3000,
+  path: './test/tapes',
+  record: talkback.Options.RecordMode.NEW,
+  debug: false,
+  name: 'Openverse e2e proxy',
+  tapeNameGenerator,
+}
```
sarayourfriend, line 22: Should we add the `ignoreHeaders` option to ignore `user-agent`, `origin` and `referrer`? For each one, here's why I think we shouldn't match tapes against them:

* `user-agent`: This differs between platforms so even just running in CI or updating browser versions could cause a mismatch (though Chrome has frozen their UA string I think; but the tapes wouldn't match if we started testing multiple UAs like Firefox, Safari or mobile browsers)
* `origin`: This will differ between environments, in particular the plan for the visual regression testing RFC, we'll run those e2e tests inside a docker container (we could run _all_ playwright tests in docker if we wanted) and in that case the origin would either be `host.docker.internal` to allow for accessing the host's localhost; or if we use docker networking and run the frontend in docker as well as the e2e tests then it'd be the frontend container's hostname.
* `referrer`: Same as `origin`.

We'd also want to include the two defaults of `content-length` and `host` I guess.

test/proxy.js
```diff
@@ -0,0 +1,26 @@
...
+const talkback = require('talkback')
+const host = 'https://api.openverse.engineering/v1'
+
+const tapeNameGenerator = (tapeNumber) => `response-${tapeNumber}`
+
+const opts = {
+  host,
+  port: 3000,
+  path: './test/tapes',
+  record: talkback.Options.RecordMode.NEW,
```
sarayourfriend, line 18: FWIW this appears to be the default value for `record` so we can leave it out if we keep it as `NEW`.

However, I'm wondering if we should actually change this to `DISABLED` and set the fallback mode to `NOT_FOUND`. I'm leaning in this direction because ideally we'd prevent _any_ external request from hitting the actual API and get some noticeable errors happening if we've forgotten to add tapes for new routes, or if our matching rules aren't working as expected.

To make it easy to update tapes, it'd be convenient to look at `process.argv` for a `--update-tapes` param then only use `RecordMode.NEW` if that's present.

test/proxy.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * The talkback proxy for e2e tests. When making a request to the API during testing,
+ * it tries to use the responses it previously saved in `/test/tapes` folder. If no
+ * response is found there, it makes an actual request, and saves the response for
+ * future use.
+ * This makes it possible for the e2e tests to run without internet, and makes the
+ * tests less flaky due to changes in the API or API data.
+ */
+const talkback = require('talkback')
+const host = 'https://api.openverse.engineering/v1'
```
sarayourfriend, line 10: This comment https://github.com/WordPress/openverse-frontend/pull/881#discussion_r810079110 got resolved with this changing or explanation why we shouldn't. Was that just an oversight or was there a reason we need to keep the `v1` prefix for the routes?

test/e2e/utils.js
```diff
@@ -83,20 +38,14 @@ const currentContentType = async (page) => {
...
 
-const mockAllSearch = async (context) => {
-  await Promise.all([
-    mockAudioSearch(context),
-    mockImageSearch(context),
-    mockThumbnails(context),
-    mockImages(context),
-  ])
+const mockProviderApis = async (context) => {
+  await context.route('**.jamendo.com**', (route) => route.abort())
```
sarayourfriend, line 42: I kept meaning to ask about this, but why do we only mock the jamendo provider? Wouldn't we want to mock all providers to stop e2e from actually hitting them if we're mocking any of them at all?

test/proxy.js
obulat: The API service uses the `host` URL to construct the request URL, so for search it would use `${host}/search/?=galah`. If we don't add `v1`, then all the requests to the API go to the wrong URL (without `v1`), and the tests fail.

I think it's best to create an issue about it and either change the API service, adding `/v1` to each request function individually, or change the Analytics server (why isn't it using `/v1/` ?).

test/e2e/utils.js
obulat: I was mocking all the requests that end with `jpg`| `svg` | `png` before to make sure we stop hitting the providers. And it's not possible to mock the jamendo request by using an extension, because its URL doesn't have one. I think this should be another new issue: Configure mocking requests to providers for media.

test/proxy.js
sarayourfriend: :scream: I totally missed that that would be a problem when talkback actually has to forward the request :grimacing: 

I assumed the analytics endpoints aren't under `v1` because they're technically not part of the open API of Openverse, they're just internal metrics gathering tools? :shrug: 

Agreed that the frontend should probably be making the requests itself.

I wonder if a stop-gap is to have _two_ talkback servers running? Or is that overkill and we should just wait and fix it for real? :thinking:

zackkrida: > I assumed the analytics endpoints aren't under v1 because they're technically not part of the open API of Openverse, they're just internal metrics gathering tools? 🤷

that's right. it's a one line NGINX config change in the infra if we want to change this, btw!

TESTING_GUIDELINES.md
```diff
@@ -78,8 +80,29 @@ Buttons, for example, should have appropriate labels. If the visible text of the
...
+
+Openverse uses [Vue Testing Library](https://testing-library.com/docs/vue-testing-library/intro/) and [Vue Test Utils](https://vue-test-utils.vuejs.org/) for unit testing, and [Playwright](https://playwright.dev) for End-to-End (e2e) testing.
+
+### E2e tests
+
+Before running the e2e tests, install the browsers that Playwright needs:
+`pnpx install playwright`
+
+If you don't have the app running, start by running it in the dev mode:
+`pnpm run dev`
```
dhruvkb, line 99: Triple backticks ``` would put the code snippets on their own line. That'll look better.

test/proxy.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * The talkback proxy for e2e tests. When making a request to the API during testing,
+ * it tries to use the responses it previously saved in `/test/tapes` folder. If no
+ * response is found there, it:
+ * - by default, returns 'Not found'.
+ * - if you pass `--update-tapes` as a parameter, makes an actual request, and saves the response for
```
dhruvkb, line 6: We would want this to run periodically so that we can ensure that the tapes don't diverge too much from the API responses.

dhruvkb (commented):
We should ignore the tapes from Prettier and ESLint. I'm not requesting changes because these are not blockers. Appreciate the grammar fixes to documentation.

test/e2e/utils.js
obulat: Opened the issue: #914

test/proxy.js
obulat: Issue opened: #915

test/proxy.js
obulat: > Appreciate the grammar fixes to documentation.

It's all because of WebStorm complaining about them :)

obulat: Should we set up the weekly (monthly/daily?) action for this, @dhruvkb ?

.prettierignore
```diff
@@ -1,6 +1,7 @@
 .github/**/*.md
 dist
 test/unit/coverage
+test/tapes
```
sarayourfriend, line 4: We can add them to .dockerignore as well?

TESTING_GUIDELINES.md
```diff
@@ -78,8 +80,46 @@ Buttons, for example, should have appropriate labels. If the visible text of the
...
 
 ## Conclusion
 
 Please note that these are non-expert and non-exhaustive recommendations. Spend time reading the [WAI-ARIA spec](https://www.w3.org/TR/wai-aria/) and other web accessibility materials. Even just knowing about specific roles and interactions that exist and are meant to be developed in consistent ways is a good first step to learning what to look out for when testing.
+
+## Running the tests
+
+Openverse uses [Vue Testing Library](https://testing-library.com/docs/vue-testing-library/intro/) and [Vue Test Utils](https://vue-test-utils.vuejs.org/) for unit testing, and [Playwright](https://playwright.dev) for End-to-End (e2e) testing.
```
sarayourfriend, line 91: We do still have some tests written just in test-utils but given it's our aspiration to move entirely to testing library we can probably leave it out of the docs.

```suggestion
Openverse uses [Vue Testing Library](https://testing-library.com/docs/vue-testing-library/intro/) for unit testing, and [Playwright](https://playwright.dev) for End-to-End (e2e) testing.
```

Alternatively, we can make explicit mention that those are legacy tests that will eventually be re-written:

```suggestion
Openverse uses [Vue Testing Library](https://testing-library.com/docs/vue-testing-library/intro/) for unit testing, and [Playwright](https://playwright.dev) for End-to-End (e2e) testing.

There are also legacy unit tests written in [Vue Test Utils](https://vue-test-utils.vuejs.org/) but those are slated to be re-written using testing library.
```

.prettierignore
obulat: We already have the `/test` directory in `.dockerignore` already. That does mean all nested folders, doesn't it, @sarayourfriend ?

obulat:
I had to write the e2e script for GitHub actions because `concurrently` was killing the proxy and that returned 1, which caused e2e tests to look as if they failed: https://github.com/WordPress/openverse-frontend/runs/5269787229?check_suite_focus=true

package.json
```diff
@@ -23,7 +23,7 @@
     "test:e2e": "playwright test",
     "build-and-e2e": "start-server-and-test 'cross-env API_URL=http://localhost:3000 nuxt build && pnpm start' http://localhost:8443 test:e2e",
     "ci:unit-test": "jest --collectCoverage=false",
-    "ci:e2e-test": "concurrently \"node test/proxy.js\" \"pnpm run build-and-e2e\" -k",
+    "ci:e2e-test": "sh ./e2e.sh",
```
sarayourfriend, line 26: Does `update-tapes` need to be corrected for this as well?

sarayourfriend (approved):
LGTM! There's one last thing though, I noticed that if for any reason talkback fails to start (for example if there's something already running on port 3000) then the e2e tests still run; ideally talkback failing to start would stop the whole test run.

Adding the typical fast-fail mechanisms won't work because we're using `&` to capture the PID, we'll need specific error handling for that case :confused: 

I don't want to block this PR any longer though, so feel free to create a fast-follow issue to correct that if you want.

Otherwise, this is awesome! I'm so excited to start using this :tada:

obulat:
> @obulat Apologies, could you update this PR with testing instructions? It looks like the commands in the PR body don't exist. I'd love to review this for you tomorrow.

Sorry for not writing the testing instructions in the first place, @zackkrida . I've updated the issue description

zackkrida (approved):
This is very cool. Testing worked as-expected.

PR #898, opened 2022-02-17
Avoid translating brand names

### LINKED ISSUE #807: Strings with a value of "Openverse" should never be translated
## Description

The English string "Openverse" is a brand and should not be translated. We should remove this from our locale files.

```
❯ rg '"Openverse"' src/locales/en.json
5:    "link-title": "Openverse",
9:    "brand": "Openverse",
101:      "search": "Openverse",
125:      "search": "Openverse",
```

## Resolution

krysal:
## Fixes

Fixes #807 by @zackkrida

## Description

This PR removes brand keys (Openverse, Creative Commons, etc.) from the `en.json` file to avoid its translation. Also, removes some unnecessary aria-labels as they're were applied for anchor links with text inside and _ARIA attributes should be used sparingly_.

## Testing Instructions

Run the project and observe there are no errors on each page (including the error page), check titles and inspect the text of the about page.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/pages/about.vue
```diff
@@ -3,90 +3,71 @@
...
+              <NuxtLink :to="localePath('/meta-search')">{{
+                $t('about.planning.meta')
+              }}</NuxtLink>
             </template>
             <template #search>
-              <a
-                :aria-label="$t('about.aria.search')"
-                href="https://github.com/wordpress/openverse-frontend/"
-                >{{ $t('about.planning.search') }}</a
+              <!-- eslint-disable @intlify/vue-i18n/no-raw-text -->
```
sarayourfriend, line 25: I wonder if to avoid having to disable this error everywhere, we couldn't just create a top-level property on the translation object `"brand": "Openverse"` and then add a translators comment to ignore it?

It's just frustrating disabling eslint errors in the template because you also have to _re-enable_ them or they'll be disabled for the whole rest of the template, which could actually introduce accidental untranslated strings.

sarayourfriend (changes requested):
I guess I agree with this in principle... but having to disable and re-enable the ESLint rule in the templates every time is pretty tedious and error prone.

If we do this, can we also add https://mysticatea.github.io/eslint-plugin-eslint-comments/ to our ESLint config to prevent accidentally leaving disabled rules disabled for the entire file unnecessarily?

dhruvkb (commented):
I oppose this change. Imo, we should have the Openverse text in translations, only for it to be transliterated. Many people who do not speak English might benefit from being able to read the Openverse name written in the script of their own language.

src/pages/about.vue
krysal: Good catch! Totally forgot to close the exception comments here. Fixed in [8baeb20](https://github.com/WordPress/openverse-frontend/pull/898/commits/8baeb201b7a975e4b9df59ec6812a5fd552f2f34).

> I wonder if to avoid having to disable this error everywhere, we couldn't just create a top-level property on the translation object "brand": "Openverse" and then add a translators comment to ignore it?

Although it would help to concentrate the brand in one place I don't think this is an acceptable solution as it still leaves the possibility for the brand string will be translated and we will have to verify that the comment is actually honored.

krysal:
@sarayourfriend Nice! That plugin will surely help, I created a separate issue to add it.

@dhruvkb That is an interesting case, [Transliteration](https://en.wikipedia.org/wiki/Transliteration). I'm still not sure it applies to brand names, it's not the case in Russian for example, see https://search-staging.openverse.engineering/ru/about

zackkrida:
Most of the translators have left 'Openverse' as-is. 
Could we make a constants file that has these strings, so we're not using raw strings in the Vue files? That might be an okay way to suppress the warnings.

sarayourfriend:
We also don't have translations of the brand SVGs. If we decide to transliterate Openverse into other writing systems then we'd have some extra work to do there as well.

obulat:
I support the idea of not using transliteration.
We could remove the "Openverse" string from the English `.pot` file when we upload it to PolyGlot, and programmatically add it into all of the translated JSON files when converting them from `.po` files. So, in this function:
https://github.com/WordPress/openverse-frontend/blob/fd623a5d46984f1dd8f97f7848632693d59460c3/src/locales/scripts/get-translations.js#L61-L67
We could replace `const translations = replacePlaceholders(rawTranslations)` with `const translations = { ...replacePlaceholders(rawTranslations), "Openverse": "Openverse"}, and use the key `Openverse` anywhere we need that string.

The link that @krysal sent actually supports the fact that we should not send `Openverse` strings for translation, because it has `Openverse` in one place, and `OpenVerse` in another :)

zackkrida:
Could we do this simpler, and remove strings that are only "Openverse" but in longer sentence strings containing Openverse just leave it, maybe with a comment for translators?

obulat:
> Could we do this simpler, and remove strings that are only "Openverse" but in longer sentence strings containing Openverse just leave it, maybe with a comment for translators?

Wouldn't you then still have to surround the only "Openverse" strings with the eslint-disable comments?
The solution for longer "Openverse" string seems great!

zackkrida:
@obulat Could we do something like:


```vue
<script>
import { openverse } from '~/utils/brand-strings' # or whatever 
</script>

<template>
 <div>{{ openverse }}</div>
</template>
```

Would that suppress the warning?

krysal:
I think we are overcomplicating things a bit here.

@obulat In your proposed solution, brand keys do not reach the translators, right? In that case, sounds good, although passing the string in the template is even simpler. Like in:

https://github.com/WordPress/openverse-frontend/blob/1235b6d138eaa2ab1c80153ad7396d3a5e85b026/src/pages/about.vue#L5-L7

That line does not raise warnings.

@zackkrida That's two lines (including the `return { openverse }` of the setup function) the same as if we were to use ESlint block comments, and we could actually save one if it is the case of the `next-line` version. Seems only a matter of preferences and I really don't mind the eslint comments very much.

obulat:
That is so much simpler, @krysal ! 🎉

obulat (dismissed):
Waiting for the implementation of your idea :)

src/locales/en.json
```diff
@@ -2,17 +2,15 @@
...
   },
   "hero": {
-    "brand": "Openverse",
     "subtitle": "Browse through over 600 million items to reuse",
     "search": {
       "placeholder": "Search for content"
     },
     "disclaimer": {
-      "content": "All Openverse content is under a {license} or is in the public domain.",
+      "content": "All {openverse} content is under a {license} or is in the public domain.",
```
sarayourfriend, line 13: There's a second instance of this string that needs updating as well: 

https://github.com/WordPress/openverse-frontend/blob/bec564f99b986015a018269363e32860d28f20bd/src/pages/index.vue#L118

src/locales/en.json
```diff
@@ -65,44 +63,27 @@
...
-    "title": "About Openverse",
+    "title": "About {openverse}",
     "description": {
-      "content": "Openverse is a tool that allows openly licensed and public domain works to be discovered and used by everyone."
+      "content": "{openverse} is a tool that allows openly licensed and public domain works to be discovered and used by everyone."
     },
     "collection": {
-      "content": "Openverse searches across more than 300 million images from open APIs and the {common-crawl} dataset. It goes beyond simple search to aggregate results across multiple public repositories into a single catalog, and facilitates reuse through features like machine-generated tags and one-click attribution.",
-      "common-crawl": "Common Crawl"
+      "content": "{openverse} searches across more than 300 million images from open APIs and the {common-crawl} dataset. It goes beyond simple search to aggregate results across multiple public repositories into a single catalog, and facilitates reuse through features like machine-generated tags and one-click attribution."
```
sarayourfriend, line 71: This one also needs the variable added 

https://github.com/WordPress/openverse-frontend/blob/8a208a7e67dcfbcb40b57197782f07f240159aed/src/pages/about.vue#L13

src/locales/en.json
```diff
@@ -132,28 +111,29 @@
...
-    "content-types": "Meta Search on Openverse currently supports image, audio, and video search on external sites.",
+    "content-types": "Meta Search on {openverse} currently supports image, audio, and video search on external sites.",
     "use": " Use ",
     "images": {
       "title": "How does the Meta Search feature work for indexed media types?",
       "demo-label": "Layout of Meta Search for indexed media type, like images",
-      "content": "Openverse supports built-in search for some media types, such as images. When you enter a query on the Openverse homepage, a page of results will be displayed. Scrolling to the bottom of the page, you have two choices:",
-      "continue": "{load-more}, and continue browsing in Openverse.",
+      "content": "{openverse} supports built-in search for some media types, such as images. When you enter a query on the {openverse} homepage, a page of results will be displayed. Scrolling to the bottom of the page, you have two choices:",
+      "continue": "{load-more}, and continue browsing in {openverse}.",
```
sarayourfriend, line 129: Also missing the variable

https://github.com/WordPress/openverse-frontend/blob/15e9e2fd2015e45cdeb0e21e52e047a68e3da1a2/src/pages/meta-search.vue#L36

src/locales/en.json
```diff
@@ -165,14 +145,14 @@
...
+      "new": "For users of the legacy CC Meta Search site, the Meta Search feature on {openverse} will look familiar. The goal was to ensure that the functionality is not lost, but is updated and embedded within our new search engine for openly licensed content. In addition, the Meta Search feature builds on this functionality, allowing us to quickly add new sources to Meta Search as we discover them, and support new content types in the future, as we expand.",
       "aria-label": "feedback",
       "feedback-suggestions": "We hope you enjoy, and if you have suggestions for improvement, leave us {feedback}.",
       "feedback-link": "feedback"
     }
   },
   "search-guide": {
-    "title": "Openverse Syntax Guide",
+    "title": "{openverse} Syntax Guide",
```
sarayourfriend, line 155: Also missing variable

https://github.com/WordPress/openverse-frontend/blob/5bf180f0d5d2e14c4a58f21e5da99f3fbcb0bc3f/src/pages/search-help.vue#L265

src/locales/en.json
```diff
@@ -474,7 +450,7 @@
       "short": "Report",
       "long": "Report this content",
       "form": {
-        "disclaimer": "For security purposes, Openverse collects and retains anonymized IP addresses of those who complete and submit this form.",
+        "disclaimer": "For security purposes, {openverse} collects and retains anonymized IP addresses of those who complete and submit this form.",
```
sarayourfriend, line 453: https://github.com/WordPress/openverse-frontend/blob/26af972f90f3f9b8081131d8c61a2aefb8329991/src/components/VContentReport/VContentReportForm.vue#L40

sarayourfriend (commented):
Found a few places that I think are missing the variable for interpolation. Otherwise this is looking great! Wish there was a way to set a "default" value for specific variables names in the template, would make this a little easier :sweat_smile:

src/locales/en.json
```diff
@@ -603,7 +600,7 @@
     }
   },
   "migration-notice": {
-    "intro": "CC Search is now called Openverse and joins WordPress as an open source project.",
+    "intro": "CC Search is now called {openverse} and joins {wordpress} as an open source project.",
```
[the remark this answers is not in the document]
zackkrida: We should extract it, good call 👍

obulat:
What about creating a dummy string, something like:
`{ "Openverse" : "{openverse}" }`, 
and then using it wherever we need to use the simple _Openverse_ string:
`$t('openverse', { openverse: 'Openverse' })` ? This way we wouldn't need to use eslint-disable

zackkrida:
@krysal will you be able to finish this PR this week? It looks like there's only a few small changes to get it resolved and I don't want the translations of code to drift too much and make merge conflicts difficult before you have a chance to finish it.

src/locales/en.json
```diff
@@ -10,7 +10,7 @@
       "placeholder": "Search for content"
     },
     "disclaimer": {
-      "content": "All Openverse content is under a {license} or is in the public domain.",
+      "content": "All {openverse} content is under a {license} or is in the public domain.",
       "license": "Creative Commons license"
```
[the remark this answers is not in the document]
krysal: It's that way because that is a link so it needs the anchor to be applied in the template.

krysal:
This is ready for review again @zackkrida. I addressed all the @sarayourfriend's comments, thanks for checking!

src/locales/en.json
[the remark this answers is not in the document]
sarayourfriend: I meant in line 14, but if it's part of the anchor then I guess we can just leave it? Out of scope for this PR either way, no worries!

zackkrida (approved):
Excellent! There are fewer places where `<!-- eslint-disable-next-line @intlify/vue-i18n/no-raw-text -->` are needed than I anticipated.

zackkrida:
@WordPress/openverse-frontend Could someone give this PR a re-review this week?

sarayourfriend:
Reviewing now.

src/pages/about.vue
```diff
@@ -3,88 +3,82 @@
...
+            <!-- eslint-disable-next-line @intlify/vue-i18n/no-raw-text -->
+            <template #openverse>Openverse</template>
             <template #common-crawl>
-              <VLink
-                :aria-label="$t('about.aria.common-crawl')"
-                href="http://commoncrawl.org/"
-                >{{ $t('about.collection.common-crawl') }}</VLink
-              >
+              <!-- eslint-disable-next-line @intlify/vue-i18n/no-raw-text -->
+              <VLink href="http://commoncrawl.org/">Common Crawl</VLink>
```
sarayourfriend, line 18:
```suggestion
              <VLink href="https://commoncrawl.org/">Common Crawl</VLink>
```

src/pages/about.vue
```diff
@@ -3,88 +3,82 @@
...
-              <VLink
-                :aria-label="$t('about.aria.search')"
-                href="https://github.com/wordpress/openverse-frontend/"
-                >{{ $t('about.planning.search') }}</VLink
-              >
+              <!-- eslint-disable @intlify/vue-i18n/no-raw-text -->
+              <VLink href="https://github.com/wordpress/openverse-frontend/">{{
+                $t('about.planning.meta')
+              }}</VLink>
+              <!-- eslint-enable @intlify/vue-i18n/no-raw-text -->
```
sarayourfriend, line 33: I _think_ this eslint-disable is superfluous. Double check me on that though!

```suggestion
              <VLink href="https://github.com/wordpress/openverse-frontend/">{{
                $t('about.planning.meta')
              }}</VLink>
```

test/unit/specs/components/v-link.spec.js
```diff
@@ -49,6 +48,7 @@ describe('VLink', () => {
     ${'http://localhost'} | ${'A'}
   `('VLink as a $component handles click', async ({ href }) => {
     const createVLinkWrapper = (href) =>
+      // eslint-disable-next-line vue/one-component-per-file
```
sarayourfriend, line 51: :pray:

sarayourfriend (approved):
LGTM! Didn't notice anything show-stopping, just a couple minor things or stuff technically out of scope (like the `https` correction).

Nice working finding a good solution to this. I feel like we landed on the best solution in the end.

test/unit/specs/components/v-link.spec.js
krysal: The annotation in the GitHub UI caused by this was bothering me. Glad we can get rid of it.

Issue #831, opened 2022-02-08
Set up Pinia


obulat:
## Problem

Pinia setup needs several changes in the `nuxt.config.js` and the project folder structure.

## Description

Here is a list of what needs to be done to install Pinia and make sure it can run correctly:

- [ ] Run `pnpm install pinia @pinia/nuxt @pinia/testing`
- [ ] Add `@pinia/nuxt` to the `buildModules` in `nuxt.config.js`
- [ ] Add an mjs rule snippet to nuxt.config.js's `build.extend` to ensure mjs files are treated correctly (and prevent errors like: Can't import the named export 'watch' from non EcmaScript module (only default export is available)):
```
config.module.rules.push({
  test: /\.mjs$/,
  include: /node_modules/,
  type: 'javascript/auto',
})
```
- [ ] Create stores folder and add types.d.ts from store folder into it.

## Implementation

- [x] 🙋 I would be interested in implementing this feature.

Issue #756, opened 2022-01-31
Switch to Pinia


sarayourfriend:
## Problem

Currently we use Vuex for global state management. This works fine for Vue 2 but has some considerable disadvantages for Vue 3, especially when it comes to TypeScript support and composition-api usage.

## Description

We should switch to Pinia instead. It is the new official default recommendation of the Vue project for state management instead of Vuex: https://blog.vuejs.org/posts/vue-3-as-the-new-default.html

It also has better interoperability between Vue 2's composition-api and Vue 3, and would allow us to decouple our state-management API usage from Nuxt. Currently we rely on Nuxt's home-grown `useStore` composable from the `@nuxtjs/composition-api` extension on top of `@vue/composition-api` but there's no guarantee that that composable will stick around in Nuxt 3.

[`useStore` will not exist in Nuxt 3](https://v3.nuxtjs.org/getting-started/bridge-composition-api#usestore) and we'll need to remove usages of it in order to be able to use Nuxt bridge anyway.

Nuxt 3 also includes a new composable called [`useState` that acts as an SSR friendly verison of `ref`](https://v3.nuxtjs.org/docs/usage/state#state). There are probably several things that we could transition to use that instead of Pinia... but it would take a lot more care to design sensible, reactive APIs around it.

## Alternatives

We could just stick with the current Vuex approach and during the Nuxt bridge upgrade replace all usages of `useStore` with `useNuxtApp().$store` instead. This would introduce more code changes during the Nuxt bridge upgrade BUT would prevent us from having to re-write our entire state management strategy before migrating to Nuxt 3. Curious what folks think about this so cc @WordPress/openverse-frontend for input on which direction to go in.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

Marking this as `high` priority because whatever decision we make here informs how the Nuxt bridge upgrade will be approached. If we choose to go with Pinia, then the Nuxt bridge upgrade would probably be stalled until we switched to it. Otherwise, the Nuxt bridge upgrade could be attempted again sooner.

sarayourfriend:
I think the output of this issue should be a milestone with discrete tasks for converting specific stores to Pinia. The first issue of course would be to add Pinia to the frontend infrastructure and convert the smallest store possible to Pinia as a proof of concept.

We will also need to make sure to implement the correct SSR tooling for Pinia including client-side state hydration.

obulat:
[...]

PR #906, opened 2022-02-18
Create a proof-of-concept for Pinia migration

dhruvkb:
## Migration to Pinia

Feel free to respond to this PR/RFC/proposal with
- additional migration PRs (or commits)
- comments suggesting changes to the approach

### Goals

- To completely eliminate all uses of Vuex, gradually replacing them with Pinia
- To do the upgrade in a gradual way, entering a transition period where Vuex and Pinia coexist

### Side-goals

- Removal of the verbosity introduced by module and action constants
- To create helpers, utils and patterns for testing store usage consistently in unit tests
- Abandon the usage of the controversial `useStore` helper in favour of modular `useX` from Pinia stores

## Steps

0. Install Pinia and test compatibility (this PR)

1. Replace Vuex stores with their Pinia counterparts
  - state -> state
  - getters -> getters
  - mutations and actions -> actions
  
    In this stage we must determine if we must use mutations as actions. Pinia allows changing the state directly in the components (which was frowned upon by Vuex) so we could do away with mutations in theory. But having them and following the Vuex principle makes transition smoother and keeps state changes limited to one file instead of scattered around the codebase.

2. Update tests, tests involving changes to the Vuex store now need to be updated to work with Pinia.
  - beforeEach
    ```js
    setActivePinia(createPinia())
    ```

  - in the test case
    ```js
    const xyzStore = useXyz()
    expect(xyzStore.state).to // ...
    ```

3. Remove unused constants from 
  - `~/constants/mutation-types.js`
  - `~/constants/action-types.js`
  - `~/constants/store-modules.js`

-1. In the end, with no Vuex stores remaining, Vuex and all associated dependencies will be removed. The `store/index.js` file can also be removed after the complete migration to Pinia. The `disableVuex: false` flag should also be removed.

### Nomenclature

- In the transition, we name stores after their Vuex counterparts.
  - `active.js` &rarr; `activeStore` / `useActive`
- State is unchanged, except that it is now an arrow function that returns the dictionary.
- Mutations become the camelCasedVersions of their constant names
  - `SET_ACTIVE_MEDIA_ITEM` &rarr; `setActiveMediaItem`

### Type hinting

In the interest of a quick migration, we can continue to use the existing type definitions for the state and the various mutations/actions from the `types.d.ts` file. Some definitions in the file may be out of date (as was the case for `ActiveMediaState`), so a cursory glance to check that would be useful.

## Procedure

To minimise disruption, each store will be changed using a lock-based approach where the developer takes a lock on the store using a notification to the group, makes a PR updating a single store and its usages and this lock is released after the PR is merged.

This prevents PRs from diverging too much and allows a seamless transition for all involved.

## Fixes

Addresses #756 by @sarayourfriend
Addresses #831 by @obulat

## PR description

This PR 
- sets up Pinia alongside Vuex
- updates the smallest store, `active`, to Pinia
- updates associated tests

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb:
Oh yeah, this was just a test to see if things work and if this is the right direction to proceed in, before we start migrating some of the more massive modules

zackkrida:
@dhruvkb this looks excellent, great initial proof of concept. I'd love to see this followed up with a Pina RFC next week to outline further next steps in more detail.

src/store/attribution.js
```diff
@@ -1,12 +1,16 @@
...
 import { CopyAttribution } from '~/analytics/events'
-import { COPY_ATTRIBUTION } from '~/constants/action-types'
 
-export const createActions = (googleAnalytics) => ({
-  [COPY_ATTRIBUTION](_, params) {
-    const event = CopyAttribution(params.type, params.content)
-    googleAnalytics().sendEvent(event)
+export const useAttribution = defineStore(ATTRIBUTION, {
+  actions: {
+    [COPY_ATTRIBUTION](params) {
```
sarayourfriend, line 11: Why would we continue using these constants with pinia? It makes the devex frustrating and defeats one of the best parts of Pinia, which is that the store is a well-typed object. Compare to:

```
export const useAttribution = defineStore('attribution', {
  actions: {
    copyAttribution(params) {
      // ...
    }
  }
}

const attribution = useAttribution()

attribution.copyAttribution()
```

This just works, and best, TypeScript is able to give you the correct types and hints when you do `attribution.`, the `.` will trigger auto-complete with the whole store's object.

If we could completely get rid of the store constants altogether that would be stellar in my opinion. With Pinia are there any reasons to keep them?

dhruvkb: This is true, I was going for the path of least change to identify if there was a way to migrate to Pinia while keeping the DX identical. Admittedly that was misguided and I've updated the PR accordingly.

sarayourfriend:
> -1. In the end, with no Vuex stores remaining, Vuex and all associated dependencies will be removed. The store/index.js file can also be removed after the complete migration to Pinia.

We should also make a note to remove the `disableVuex: false` flag from the pinia configuration.

src/layouts/default.vue
```diff
@@ -60,8 +56,9 @@ const embeddedPage = {
   setup() {
     const mainContentRef = ref(null)
     const mainRef = ref(null)
-    const { store } = useContext()
-    const isReferredFromCc = store.state[NAV].isReferredFromCc
+
+    const navStore = useNav()
+    const isReferredFromCc = computed(() => navStore.isReferredFromCc)
```
sarayourfriend, line 61: Is this computed even necessary? It's just wrapping a ref in another ref as far as I can tell. Can `isReferredFromCc` be used directly off the store?

src/store/active.js
```diff
@@ -1,57 +1,44 @@
...
-   * @param {import('./types').ActiveMediaState} _state
-   */
-  [EJECT_ACTIVE_MEDIA_ITEM](_state) {
-    _state.type = null
-    _state.id = null
-    _state.status = 'ejected'
+  actions: {
+    /**
+     * @param {object} payload
+     * @param {'image' | 'audio' | null} payload.type
```
sarayourfriend, line 20: Should this really be allowed to be `null`?

src/store/active.js
```diff
@@ -1,57 +1,44 @@
...
+     * @param {string | null} payload.id
```
sarayourfriend, line 21: Same here, seems like it shouldn't be able to be `null`.

sarayourfriend (commented):
I'd love to see adding a requirement be TypeScript types for each re-written store. To include this we'd also have to add each re-written store to the `tsconfig.json` `includes` list.

This is great though!

> In this stage we must determine if we must use mutations as actions. Pinia allows changing the state directly in the components (which was frowned upon by Vuex) so we could do away with mutations in theory. But having them and following the Vuex principle makes transition smoother and keeps state changes limited to one file instead of scattered around the codebase.

I'd be nice to stick mutations in the Pinia store, if only because it'll make them significantly easier to test without having to rely on triggering them via some component interaction.

src/layouts/default.vue
dhruvkb: Yes, `navStore.isReferredFromCc` is not reactive unless you either put it inside `computed` or use `storeToRefs`.

src/store/active.js
dhruvkb: Yeah, when no media is active, this is set to `null`.

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -58,11 +58,9 @@ import VRowLayout from '~/components/VAudioTrack/layouts/VRowLayout.vue'
...
 import VGlobalLayout from '~/components/VAudioTrack/layouts/VGlobalLayout.vue'
 
-import { ACTIVE, MEDIA } from '~/constants/store-modules'
-import {
-  PAUSE_ACTIVE_MEDIA_ITEM,
-  SET_ACTIVE_MEDIA_ITEM,
-} from '~/constants/mutation-types'
+import { MEDIA } from '~/constants/store-modules'
+
+import { useActive } from '~/store/active'
```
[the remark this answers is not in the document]
dhruvkb: This makes sense. I'll make the change. (I didn't change the name because I wanted to minimise disruptions).

src/store/active.js
```diff
@@ -1,57 +1,44 @@
-import {
-  SET_ACTIVE_MEDIA_ITEM,
-  PAUSE_ACTIVE_MEDIA_ITEM,
-  EJECT_ACTIVE_MEDIA_ITEM,
-  SET_MESSAGE,
-} from '~/constants/mutation-types'
+import { defineStore } from 'pinia'
+
+export const ACTIVE = 'active'
```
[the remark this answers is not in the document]
dhruvkb: Yeah, this isn't necessary. Since we should name the store for dev-tools, I assumed using a constant (especially the old one for continuity) would be nice. But this doesn't need to be exported.

src/store/active.js
dhruvkb: This is null when there is no active media (such as when no music is playing or when no image is maximised).

src/store/active.js
sarayourfriend: But the _payload_ type shouldn't allow passing in `null`. The store value can be null for sure but there's no reason this function should allow it.

src/components/VAudioTrack/VAudioTrack.vue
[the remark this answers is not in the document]
dhruvkb: Renamed in 5173d09.

src/store/active.js
dhruvkb: You're right, I missed the fact that we have the `ejectActiveMediaItem` mutation specifically for this.

src/store/active.js
dhruvkb: You're right, I missed the fact that we have the `ejectActiveMediaItem` mutation specifically for this. Fixed in 46b1639.

sarayourfriend:
@dhruvkb for this project it might be helpful to lay out each of the stores, how they're related, which ones depend on each other and how best to migrate them.

I think there are stores that commit across each other (in particular IIRC search and media do this) and we'd need to convert them in the correct order for it to work (and confirm we can call Pinia store actions from within a Vuex mutation?)

obulat:
> I'd be nice to stick mutations in the Pinia store, if only because it'll make them significantly easier to test without having to rely on triggering them via some component interaction.
I think it would be better to approach this on a case-by-case basis, @sarayourfriend . Some mutations are very simple, created only because there is no way of changing the state from the actions in Vuex. One of them is `FETCH_START_MEDIA`, which simply does `state.fetchState = true`. I don't see a point of testing such a mutation.
But there are, of course, larger mutation functions that do need to be kept, and converted into actions.

sarayourfriend:
> One of them is FETCH_START_MEDIA, which simply does state.fetchState = true. I don't see a point of testing such a mutation.

Agreed that testing that directly wouldn't make sense but we would want to test that that does happen as part of a fetch media transction. Without being able to spy on the function that causes that state to change, i'm not sure how you'd write that test. If the action was like this:

```
async fetchMedia() {
  state.isFetching = true
  try {
    await doFetching()
    // ... etc
  } catch (e) {
    state.fetchError = // ...
    return // ...
  } finally {
    state.isFetching = false
  }
}
```

Without encapsulating the fetch-state mutations there's no (non-convoluted) way to unit test that they're happening. You can't make assertions after calling `fetchMedia` that `isFetching` is true because you'll be introducing a data race in the test against `doFetching` returning (which in the test should be immediate); or you'll be relying on very specific event loop behaviors that probably aren't guaranteed.

Compare to if you use functions for the fetch-state modifications:

```
async fetchMedia() {
  this.startFetch()
  try {
    await doFetch()
    // ...
  } catch (e) {
    this.failFetch(e)
    return // ...
  } finally {
    this.finishFetch()
  }
}
```

In the test you could `jest.spyOn(store, 'startFetch')` and so on to verify that they're called as expected.

In that case you _would_ want to test each of those functions as well to verify that those units work as expected, even if it's a single line :shrug:

sarayourfriend:
@dhruvkb 

Did you have any time to look into this?

> In your time researching, did you come across any opinions about whether to use the object or function callback syntax for defining a store? I strongly prefer the function callback syntax because I feel like it more closely mirrors the way state is created in setup (in fact, it's exactly the same) and it also makes having private helper functions possible (compared to the object syntax where all the functions on the definition object will be public). It also eliminates this usage which I've generally found to be a good thing for comprehensibility in JavaScript generally.

obulat:
Can you give an example of two ways of defining a store, @sarayourfriend ? I can only see one in Pinia docs: 
```
export const useStore = defineStore('main', { // other options... }) 
```

sarayourfriend:
@obulat just to clairfy, is the goal to merge this branch once the RFC is approved?

dhruvkb:
@sarayourfriend I would prefer that.

krysal:
I'd think we should decide to tackle the Pinia milestone before introducing the dependency, and then focus on migrating the whole store as it's not recommended to keep both Vuex and Pinia coexisting at the same time.

sarayourfriend:
> I'd think we should decide to tackle the Pinia milestone before introducing the dependency, and then focus on migrating the whole store as it's not recommended to keep both Vuex and Pinia coexisting at the same time.

@krysal can you add that comment to the RFC? There's some discussion in there about what that guidance in the docs actually means: https://github.com/WordPress/openverse/pull/174/files#r813094475

@dhruvkb did some good investigation that clarified the situation for me personally. Curious what you think about it.

dhruvkb:
My opinion is that big PRs that change a lot of things can 
- get stuck in reviews
- frequently get stale and need to be constantly updated
- block work in other PRs that might want to use the new code or change the old one

Since Vuex and Pinia can coexists (which although not-recommended, works quite well) we should not hesitate to use their coexistence to migrate stores in several small PRs that can be merged fairly quickly.

nuxt.config.ts
```diff
@@ -147,6 +147,7 @@ const config: NuxtConfig = {
     '@nuxtjs/style-resources',
     '@nuxtjs/svg',
     '@nuxtjs/eslint-module',
+    ['@pinia/nuxt', { disableVuex: false }],
```
zackkrida, line 150:
```suggestion
	  // @todo: Remove `disableVuex` once all stores are migrated to pinia
    ['@pinia/nuxt', { disableVuex: false }],
```

zackkrida (approved):
lgtm. My only remaining question is if we want to use the `useNav` naming convention or be extra redundant and clear with `useNavStore`, but that's a very minor concern we could change later.

obulat (approved):

src/stores/active-media.js
```diff
@@ -0,0 +1,68 @@
...
+    type: null,
+    id: null,
+    status: 'ejected',
+    message: null,
+  })
+  /**
+   * Only the properties used by components are exported as refs.
+   * `status` is not used anywhere in the components.
+   */
+  const { type, id, message } = toRefs(state)
```
sarayourfriend, line 26: I couldn't find any usages of `activeMediaStore.type`. `id` and `message` both have usages though.

src/stores/active-media.js
```diff
@@ -0,0 +1,68 @@
...
+   * @param {string} params.message
+   */
+  function setMessage(params) {
+    state.message = params.message
+  }
+
+  return {
+    type,
+    id,
+    message,
```
sarayourfriend, line 60: Should these all be `readonly` as well to prevent modifying them without going through the action functions?

src/stores/nav.js
```diff
@@ -0,0 +1,28 @@
+import { defineStore } from 'pinia'
+import { reactive, toRefs } from '@nuxtjs/composition-api'
+
+export const NAV = 'nav'
```
sarayourfriend, line 4: I think we decided we didn't need to export these.

```suggestion
const NAV = 'nav'
```

src/stores/nav.js
```diff
@@ -0,0 +1,28 @@
...
+  function setIsReferredFromCc(isReferredFromCc = true) {
+    state.isReferredFromCc = isReferredFromCc
+  }
+  return {
+    isEmbedded,
+    isReferredFromCc,
+    setIsEmbedded,
+    setIsReferredFromCc,
+  }
+})
```
sarayourfriend, line 28: It looks like we use a mix of updating the `isEmbedded` and `isReferredFromCc` values directly and also through the setter functions. Should we standardize on one approach for simple properties like this?

https://github.com/WordPress/openverse-frontend/blob/7c59473bdc4e0aaa86d0c5b9174a90808e914249/src/middleware/middleware.js#L33

https://github.com/WordPress/openverse-frontend/blob/7c59473bdc4e0aaa86d0c5b9174a90808e914249/src/plugins/migration-notice.js#L11

test/unit/specs/stores/active-media-store.spec.js
```diff
@@ -0,0 +1,56 @@
...
+import { AUDIO } from '~/constants/media'
+
+import { useActiveMediaStore } from '~/stores/active-media'
+
+const initialState = { type: null, id: null, status: 'ejected', message: null }
+const statuses = ['ejected', 'paused', 'playing']
+
+describe('Active Media Store', () => {
+  beforeEach(() => {
+    setActivePinia(createPinia())
```
sarayourfriend, line 12: Does this have the effect of resetting the store before each test?

sarayourfriend (commented):
As mentioned by @zackkrida elsewhere there are errors in the app when trying to run it locally. None of the comments I added are blocking I don't think.

One thing I'd like to see is adding the two new stores to `tsconfig.json` (or updating them to be `.ts`). I tried this locally (converting to `.ts`) and there were only three errors to fix in active-media.ts which just came down to moving JSDoc parameter type annotations into TS annotations.

.eslintrc.js
```diff
@@ -1,3 +1,6 @@
+// WebStorm fix for `~` alias not working:
+// https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000771544-ESLint-does-not-work-with-webpack-import-resolver-in-2017-3
+process.chdir(__dirname)
```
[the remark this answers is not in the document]
obulat: Yes, definitely. It seems that the bug reports are from 2018, but I'm having this problem now 🤷

src/stores/nav.js
obulat: I prefer setter functions because they are easier to observe.

test/unit/specs/stores/active-media-store.spec.js
obulat: Yes:
// creates a fresh pinia and make it active so it's automatically picked
// up by any useStore() call without having to pass it to it:
// `useStore(pinia)`

src/stores/active-media.js
obulat: I think `type` will be used when we have video, and we can have an `activeVideo` which will need to be treated differently from `activeAudio`

src/stores/nav.js
zackkrida: If we went with setter functions is there a way to enforce their usage, like to make `isReferredFromCc` readonly?

obulat: Yes! I'll do that

zackkrida:
Reconfirming that this looks good! LGTM.

PR #913, opened 2022-02-19
Use distroless and explore space saving measures for final production image

sarayourfriend:
## Description

Proof of concept build of the production docker image using `distroless` for the final base image. To be fair, `distroless` by itself would only cut the image size by like 4 MB compared to using alpine... but I've also applied some other space saving techniques like removing all the `node_modules` we don't need for running the app.

Kind of an annoying part of Nuxt deployments is that your "dev" dependencies and regular dependencies are pretty muddled. I bet we could do some more careful experimentation of what is actually literally needed for running the production application and move everything else to `devDependencies`. Then we could skip the step where we trash all the dependencies that don't have `nuxt` in their name (which is admittedly probably not the safest strategy :sweat_smile:)

In any case, it definitely runs locally. I'd be curious to see the image deployed to staging maybe just to test the whole app. There are CORS errors when you run it locally but those exist for the existing production image as well. Otherwise it appears to run fine.

Here's the stats:

* Before
```
openverse-frontend                              latest          7416fb178348   4 seconds ago        496MB
```

* After
```
openverse-frontend                              latest          ba673154e585   4 minutes ago    288MB
```

**Update:**

I realized I'd accidentally take the measurement using the `debug` base image. Here's the actual built image size when using the non-debug image:

```
openverse-frontend                              latest          5a38af745e8f   54 seconds ago   287MB
```

1 more MB saved :wink: 

So yeah, almost cuts the image in half. It's still huge... but that's node for you :confused: 

### Takeaways

Regardless of whether we decide to use `distroless` for the deployed image (which I think we should consider, it seems to be best practice along with running the application as a nonroot user) we can definitely do a better job at managing the dependencies we copy over and furthermore making sure to copy the `pnpm-store` between the base images so that we're not re-downloading dependencies on every `pnpm install` layer: that should be able to just happen a single time in the `builder` image.

We're also not using a non-root user in the current production image. This PR does do that and it appears to not cause any problems.

cc @rbadillap for all that; you're the best person to advise on what the positive takeaways from this experiment are :slightly_smiling_face: 

## Testing Instructions

Checkout this branch and run `docker build . --target app -t openverse-frontend:latest`. Wait for the build to happen, then run `docker run -it --rm openverse-frontend:latest` and confirm that the app starts and that you can hit it locally (albeit with CORS errors).

You can use `docker images | grep openverse-frontend` to check the resulting image size.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ COPY pnpm-lock.yaml .
 COPY .npmrc .
 
 # install dependencies including local development tools
-RUN pnpm install
+RUN pnpm install --frozen-lockfile --store-dir=./pnpm-store
 
 # copy the rest of the content
 COPY . /usr/app
@@ -27,48 +27,30 @@ ENV NUXT_TELEMETRY_DISABLED=1
 # build the application and generate a distribution package
 RUN pnpm run build
 
+
 # ==
-# development
+# pre-production
 # ==
-# application for development purposes
-FROM node:16 AS dev
+# trash existing deps and install only necessary deps to reduce final image size
+FROM builder AS postbuild
 
-WORKDIR /usr/app
+RUN rm -rf node_modules
 
-# Install pnpm
-RUN npm install -g pnpm
+ENV NODE_ENV=production
 
-ENV NODE_ENV=development
-ENV CYPRESS_INSTALL_BINARY=0
+RUN pnpm install --frozen-lockfile --store-dir=./pnpm-store
 
-# copy files from local machine
-COPY . /usr/app
+# Delete everything that isn't nuxt related
+RUN ls node_modules | grep -v nuxt | xargs rm -rf
 
-# disable telemetry when building the app
-ENV NUXT_TELEMETRY_DISABLED=1
-
-# install dependencies (development dependencies included)
-RUN pnpm install
-
-# expose port 8443
-EXPOSE 8443
-
-# run the application in development mode
-ENTRYPOINT [ "pnpm", "run", "dev" ]
 
 # ==
 # production
 # ==
 # application package (for production)
-FROM node:alpine AS app
+FROM gcr.io/distroless/nodejs-debian11:16-debug AS app
 
 WORKDIR /usr/app
-
-# Install pnpm
-RUN npm install -g pnpm
-
-ENV NODE_ENV=production
-ENV PLAYWRIGHT_SKIP_BROWSER_GC=1
 
 # copy package.json and package-lock.json files
 COPY package.json .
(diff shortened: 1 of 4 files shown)


dhruvkb:
Is the main value of `distroless` to provide a Debian base (and thus `glibc` over `musl`) for an extra 4MB over `alpine`?

sarayourfriend:
> Is the main value of distroless to provide a Debian base (and thus glibc over musl) for an extra 4MB over alpine? Calling it distroless seems misleading imo if it is clearly built from Debian.

The main value, as I understand it, is a reduced attack surface if someone is able to break into your container out of your app's runtime. Not having any shell whatsoever, for example, and just plain removing binaries that aren't necessary for the functioning of your application.

There's not much "debian" left once everything is stripped away, to be fair.

From the top of their README: 

> "Distroless" images contain only your application and its runtime dependencies. They do not contain package managers, shells or any other programs you would expect to find in a standard Linux distribution.

Here's their own description of distroless's value: https://github.com/GoogleContainerTools/distroless#why-should-i-use-distroless-images

In any case, I should probably rename this PR to "explore space saving measures in production docker image"; the distroless base image, like I said, only shaves 4 MB off the alpine base.

But now that I'm looking at the sizes of the base images, it looks like the node 16 distroless image is actually 7MB _larger_ than the alpine node. I'm not sure why that's the case.

```
gcr.io/distroless/nodejs-debian11               16              1fd6dfaf3abe   52 years ago    117MB
node                                            16-alpine       0e1547c0f4a4   10 days ago     110MB
node                                            alpine          c6a2764d974b   9 days ago      170MB
```

I'm not sure why there's a discrepancy between 16-alpine and plain alpine.

obulat:
When I run the commands from the 'Testing instructions', I get the usual output:
```
   │   Nuxt @ v2.15.8                        │
   │                                         │
   │   ▸ Environment: production             │
   │   ▸ Rendering:   server-side            │
   │   ▸ Target:      server                 │
   │                                         │
   │   Memory usage: 43.9 MB (RSS: 141 MB)   │
   │                                         │
   │   Listening: http://172.17.0.2:8443/   
```
But when I try to open 172.17.0.2:8443, I get timeout ("The server at 172.17.0.2 is taking too long to respond.") ...

sarayourfriend:
> But when I try to open 172.17.0.2:8443, I get timeout ("The server at 172.17.0.2 is taking too long to respond.") ...

I wonder if this is a docker networking bug/oddity on macOS? Does the same thing happen on the version in `main`?

obulat:
> I wonder if this is a docker networking bug/oddity on macOS? Does the same thing happen on the version in main?

Yes, I get the same on main branch.

>@obulat does localhost:8443 work?

No, it doesn't. Does it work for you?

obulat:
WebStorm is complaining about this line not being allowed:
https://github.com/WordPress/openverse-frontend/blob/6ee52199ee48aa51216c56142de438bf9448852e/docker-compose.yml#L6

I tried again, and got it to work at `localhost:8443`. 
When I look at the container in Docker Desktop's Inspect Tab, I see `8443/tcp Not bound`.
On a Container that I got by running `docker-compose up`, the same line was `8443/tcp 0.0.0.0:8443`.

sarayourfriend:
@obulat I accidentally removed the `dev` target from the Dockerfile :facepalm:

obulat:
> But note, docker-compose isn't set up to run the production image, so please use `docker run` after building the `app` target using `docker build . --target app -t openverse-frontend:latest`.

I understand that. I was just trying whatever I can to get it to work :) Do you know why I get `8443/tcp Not bound` when running the container build from this PR?

sarayourfriend:
> Do you know why I get 8443/tcp Not bound when running the container build from this PR?

I'm not sure... do you get this after running `docker run`? Or is it elsewhere?

I wonder if it could be something different in the way macOS docker requires you to expose ports; the Dockerfile already exposes the port but maybe using [the `-p` setting for run would fix it?](https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose)

Could you try running this:

```bash
docker build . --target app -t openverse-frontend:latest -p 127.0.0.1:8443:8443/tcp
```

And see if that lets you access the app running in the container?

zackkrida:
@obulat `docker run -it --rm -p 8443:8443 openverse-frontend:latest` should work for you to run the application at `localhost:8443`.

zackkrida (changes requested):
lgtm once we get the `dev` target working again! The `pnpm` module deleting and only installing once seem like the biggest innovations here. `Distroless`, while slightly larger, is probably more secure than `alpine`, so that seems like a win too.

sarayourfriend:
I don't think we should go with this solution either way. I'm adding back the dev target as an example but brute-force deleting node modules is not a good solution. I'll write an issue with some ideas I have of actually good solutions to this problem of identifying build vs runtime vs dev dependencies.

Dockerfile
```diff
@@ -27,70 +27,55 @@ ENV NUXT_TELEMETRY_DISABLED=1
...
+ENV NODE_ENV=production
 
-# install dependencies (development dependencies included)
-RUN pnpm install
+RUN pnpm install --frozen-lockfile --store-dir=./pnpm-store
 
-# expose port 8443
-EXPOSE 8443
+# Delete everything that isn't nuxt related
+RUN ls node_modules | grep -v nuxt | xargs rm -rf
```
rbadillap, line 44: If you want to avoid any possible error if the `node_modules` doesn't exists or there are no `nuxt` files within, you can use `find` command which doesn't throw any error code if there are no files during the lookup.

```bash
find node_modules -type f -not -name '*nuxt*' -delete
```

PR #918, opened 2022-02-21
Enable SSR for migration banner

### LINKED ISSUE #515: CC Migration banner doesn't SSR
## Description

The migration banner does not render server side so it "pops in" after the page is hydrated on the client.

## Reproduction

1. Visit https://wordpress.org/openverse/?referrer=creativecommons.org
2. Watch the top of the Openverse iframe and note the banner popping in shortly after the page finishes loading

## Expectation
The banner should be SSR rendered. We have the query param already (and we could check the referrer header anyway) so it's definitely possible to know that it should be present server side.

## Resolution

dhruvkb:
## Fixes

Fixes #515 by @sarayourfriend

## Description

This PR enables the migration banner to render via SSR.

## Cause
Nuxt middleware runs once on the server during the first render and subsequent times on the client during routing. On the server, the plugin `migration-notice.js` sets the Vuex state `isReferredByCc` to true, and the middleware immediately sets it to false. So the banner isn't part of the SSR page.

On the client the plugin again sets the Vuex state `isReferredByCc` to true. The middleware already having executed on the server is not run on the client for the first page, so the banner pops in. Further navigation causes the middleware to run and remove the banner (as expected).

By changing the middleware to not run the unset mutation on the server, we are able to ensure that it only has the intended banner clearing effect on the client.

This problem was made harder to debug because the layout was non-reactively using the store state.

## Testing Instructions

1. Open the URL http://localhost:8443/search/image?q=cat&referrer=creativecommons.org.
2. Observe that the migration banner is SSR-ed.
3. Navigate to any different page.
4. Observe that the banner goes away and never returns.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/middleware/middleware.js
+++ b/src/middleware/middleware.js
@@ -28,7 +28,7 @@ export default function ({ store, query, route }) {
     })
   }
 
-  if (!process.server && store.state.nav.isReferredFromCc) {
+  if (process.client && store.state.nav.isReferredFromCc) {
     store.commit(`${NAV}/${SET_REFERRED}`, { isReferredFromCc: false })
   }
 }
--- a/src/middleware/middleware.js
+++ b/src/middleware/middleware.js
@@ -28,7 +28,7 @@ export default function ({ store, query, route }) {
     })
   }
 
-  if (store.state.nav.isReferredFromCc) {
+  if (!process.server) {
     store.commit(`${NAV}/${SET_REFERRED}`, { isReferredFromCc: false })
   }
 }
--- a/src/middleware/middleware.js
+++ b/src/middleware/middleware.js
@@ -28,7 +28,7 @@ export default function ({ store, query, route }) {
     })
   }
 
-  if (!process.server) {
+  if (!process.server && store.state.nav.isReferredFromCc) {
     store.commit(`${NAV}/${SET_REFERRED}`, { isReferredFromCc: false })
   }
 }
--- a/src/components/VMigrationNotice.vue
+++ b/src/components/VMigrationNotice.vue
@@ -1,5 +1,5 @@
 <template>
-  <VNoticeBar>
+  <VNoticeBar class="cc-ov-migration">
     {{ $t('migration-notice.intro') }}
     <i18n tag="span" path="migration-notice.more">
       <template #read-more>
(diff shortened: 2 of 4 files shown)


src/middleware/middleware.js
```diff
@@ -28,7 +28,7 @@ export default function ({ store, query, route }) {
     })
   }
 
-  if (store.state.nav.isReferredFromCc) {
+  if (!process.server && store.state.nav.isReferredFromCc) {
```
sarayourfriend, line 31: nit: FWIW positive assertions always seem easier for me to quickly grok than a negation. Clearer for new folks or someone skimming as well who might not have the full context of `process.server` and `process.client`.

```suggestion
  if (process.client && store.state.nav.isReferredFromCc) {
```

sarayourfriend (approved):
LGTM! Nice and easy fix :tada: 

Can you add an E2E test for this? :thinking:

obulat:
If I understand correctly, the migration banner can only be displayed on the homepage, because that's the only page that CC Search redirects to. So, we can remove the migration banner from the default layout. In fact, we could even move the MigrationBanner into the HomePage component. This way we could prevent the vertical overflowing of the page that you can see now.

zackkrida:
@obulat No, their redirect preserves paths, so that image links or search result links out in the wild resolve correctly. You can test this with a URL like search.creativecommons.org/photos/19e2e417-a819-4761-b838-d648b92d48ce?referrer=creativecommons.org

krysal (approved):
+1 to Sara's suggestion about tests but glad it's working SSR.

obulat:
Oh, thank you for explaining, @zackkrida !

src/middleware/middleware.js
dhruvkb: This is classic half-knowledge stuff. I didn't even know there was a `process.client`!

dhruvkb (commented):
I'll add an E2E test-case for this. But I've never written one before so there might be some delay due to the learning involved.

PR #982, opened 2022-02-21
Fetch single image result in `asyncData` hook instead of `fetch`

### LINKED ISSUE #980: Navigate to the image detail page with an invalid id breaks it
## Description

The image details page does not handle invalid or null `id` well.

## Reproduction

1. Go to https://search-staging.openverse.engineering/image/foo
2. See a blank page.

or alternately:

1. Run the project and go to http://localhost:8443/image/foo
2. See the error page but inspect the browser console and the terminal.

## Expectation

When the page gets an invalid or null `id` it should show the error page and return a 404 response status code.

### On staging
CleanShot 2022-02-21 at 14 53 00@2x

### Locally
CleanShot 2022-02-21 at 14 52 35@2x

## Additional context

Seems to be a residue or a more general case of #496.

## Resolution

krysal:
## Fixes

Fixes #980 by @krysal

## Description

This page was using the `fetch` hook when it should be [`asyncData`](https://nuxtjs.org/docs/features/data-fetching#async-data) hook of Nuxt as in the audio page, the last one is the correct for fetching async data on the server-side.

Managing invalid or null ids this way we no longer need the `index.vue` pages for /image and /audio, and these were also returning inaccurate 200 status codes with the layout for a 404 error page.

## Testing Instructions

Reproduce steps in the mentioned issue and observe the expected behavior: the error is handled correctly showing the error page without fetching and a 404 status code is returned.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/pages/audio/index.vue
```diff
@@ -1,11 +0,0 @@
...
-
-<script>
-import { defineComponent } from '@vue/composition-api'
-import VFourOhFour from '~/components/VFourOhFour.vue'
-
-export default defineComponent({
-  components: { VFourOhFour },
-  layout: 'blank',
-})
-</script>
```
sarayourfriend, line 11: I _knew_ there had to be a better way to do this. Thank you for removing the need for these!

sarayourfriend (approved):
LGTM! Verified that this fixes the issue compared to staging.

Seriously so glad you were able to get rid of those two `index.vue` pages :rocket:

obulat:
This is a very good solution to the 404 problem.
If I remember correctly, it was done as this (with fetch and thumbnail) to make the navigation seamless: you would see the thumbnail at first, and then, after the actual image loads, you would see (but probably not notice the change to) the actual image.
Now, when you navigate to the detail page, you (or at least I) can see the blank image section, and then the image pops up into the view. 
@zackkrida , was my explanation correct?

sarayourfriend:
>  If I remember correctly, it was done as this (with fetch and thumbnail) to make the navigation seamless: you would see the thumbnail at first, and then, after the actual image loads, you would see (but probably not notice the change to) the actual image.
> Now, when you navigate to the detail page, you (or at least I) can see the blank image section, and then the image pops up into the view.

This was my understanding of the _intended_ behavior but honestly I haven't seen that working in months.

If I go to https://search-staging.openverse.engineering/search/?q=galah and click any of the image results, I see the same behavior you described, Olga, of the space for the image being blank and then popping in afterwards.

krysal:
@obulat that makes sense, I can see that behavior in production. Fortunately, that's easy to add! :)

zackkrida (approved):
@krysal ideally, we would only show the thumbnail first _if_ the user is navigating from the search result page, where the thumbnail has already been loaded. With the current approach, any user visiting a server-rendered image url directly, ends up downloading two images, the thumbnail and the full image, when it would be better to only download the full image.

I think this is okay to merge without fixing this, since the current solution optimizes for the most common use case, but it would be good to fix in a separate issue.

krysal:
@zackkrida That optimization sounds possible and good for a following "PR", here I was thinking of just keeping the old behavior after fixing the bug. Thanks for reviewing!

PR #984, opened 2022-02-22
Rename media store getters

### LINKED ISSUE #830: Rename getters that have the same name as state properties in Vuex stores
## Problem

In Pinia, `getters` are available as a property of `this` inside the store, or as a property of an object returned from `useSomeStore` inside the components. This means that we cannot use the same name for `getters`, state properties, and even actions.
We do, however, have some getters that use the same name as the state property, as here in `media.js` store:
https://github.com/WordPress/openverse-frontend/blob/9924de7183c4f8742540384bf4010e4d289bee1b/src/store/media.js#L262-L269
https://github.com/WordPress/openverse-frontend/blob/9924de7183c4f8742540384bf4010e4d289bee1b/src/store/media.js#L40-L42

## Description

We need to rename all `getters` that have the same name as the state properties (or rename the state properties instead).

## Implementation

obulat:
## Fixes

Fixes #830 by @obulat

## Description

This PR renames some of the `media` store `getters` so that there are no `getters` with the same name as a state property. This is necessary because both `getters` and state properties will be available as properties of `this` in Pinia.

While renaming, I did a couple of simplifications, too:
- all media items get assigned `frontendMediaType` property by the media service, so I removed the addition  of `frontendMediaType` from the `VAllResultsGrid` component.
- `results` getter was only used by the `VContentLink` component to access the result counts for each media type, so I rewrote the getter as `mediaResultCounts` that only returns `{ mediaType: N }` instead of passing all of the data.
- reverted the conversion of `items` from an Array to an Object. We only look up an item by its id in the Global Audio Player. Everywhere else we use items as an array, and because of that had to convert the items object into an array in at least 3-4 places.
- 
## Testing Instructions

Run the app, try selecting different search types, clicking the 'Load more items' button, select various filters. Also, try to load each kind of pages from the server (i.e. go after searching for something or opening a single result page, refresh the page). There should be no errors or warnings.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat:
We have decided to use the _setup store_ for Pinia stores, not the _options store_, so this PR is no longer relevant.

PR #993, opened 2022-02-22
Fix CSS import ordering

### LINKED ISSUE #992: Focused image result license icons are wrong colors
## Description

License icon color on the search result pages is correct in development, but incorrect in the built application due to the CSS ordering issue.

- Pink here: https://search-staging.openverse.engineering/search/?q=mars
- White here: https://search-staging.openverse.engineering/search/image?q=mars

## Expectation

These license icons should be the default text color in both places.

## Screenshots

CleanShot 2022-02-22 at 14 09 19@2x

CleanShot 2022-02-22 at 14 09 13@2x

sarayourfriend:
## Fixes

Fixes #992 by @zackkrida 

## Description

Using the [templates](https://nuxtjs.org/docs/configuration-glossary/configuration-build#templates) option in Nuxt config we can override the problematic templates to fix the import order. I've opened a PR upstream to address this but it hasn't gotten any attention from maintainers: https://github.com/nuxt/nuxt.js/pull/10275

The idea here is to ensure that the CSS imports happen before any other styles are defined, especially `@apply` Tailwind styles.

We shouldn't stop investigating ways of fixing this after this PR; maintaining these template overrides is far from ideal. I wasn't able to reproduce the behavior we're seeing in our application [in codesandbox](https://codesandbox.io/s/eager-glade-gewem) so I feel like there's something very subtly wrong with our configuration that is somehow causing this CSS misordering to occur. I just don't know what it could be and we need a stop-gap for this problem in the meantime.

## Testing Instructions

Checkout `main` and run `pnpm build` and take note of the existing import ordering in `./.nuxt/index.js` and `./.nuxt/App.js`.

Now checkout this branch and run `pnpm build` and verify that the overridden templates are applied in the same files. Compare them to the previous versions to confirm that they've changed _only in the ways described in the README.md added in this PR_.

Run `pnpm dev` for good measure and make sure the template changes are also present.

Next try to reproduce the bug in the linked issue, verify that it is fixed. Also verify that there are no more "ghost buttons" that we were seeing previously (like the skip to content button, which can be accessed by pressing `tab` immediately after loading a search results page).

This fix also introduced a regression of the double focus rings on the search type switcher; turns out we were fighting windmills there too. Verify that these buttons have a single focus ring. Also test the page menu on desktop and mobile widths.

Please test in Firefox, Chrome and Safari if possible and both with desktop and mobile viewport widths, with and without keyboard navigation.

Lastly, smoke test the app for styles and make sure that things "make sense" and that nothing is visually broken (in particular Buttons as that was the component that was being imported "too early"). I'd suggest using keyboard navigation to highlight all the focus rings and such around the application and make sure they still work.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
To reviewers, in particular I think @zackkrida and @obulat worked on fixes for this directly, am I forgetting anything that we should remove/test that we were doing for this CSS ordering issue before?

zackkrida (dismissed):
LGTM. I think you caught all of the band-aid fixes but @obulat might know of more I am forgetting.

Wow, it sucks to have to dupe these huge files, but it's all we can do. I would suggest a different approach where we actually modify the Nuxt templates themselves with [patch-package](https://github.com/ds300/patch-package), but it seems like `patch-package` might have some compat issues with `pnpm`.

sarayourfriend:
I did think about using patch-package but... considering nuxt supports changing the templates without that I figured it'd be nice to avoid having to maintain patches (and yes pnpm has some trouble with patches but they're not impossible to solve, I had to deal with that in Gutenberg's pnpm PR)

sarayourfriend:
I noticed that this introduces a regression in Chromium on Windows. I'll push a fix for that soon. Can someone with Safari test that as well, in particular the search type switcher popover focus rings.

zackkrida:
Apologies for not testing this in Safari! Seeing the following bugs:

In dev, I'm not seeing _any_ pink focus rings anywhere. It might be an issue with `focus-visible` then? 

In prod:

- Double pink focus rings in the content switcher, the internal and external
- No pink focus ring on the homepage search bar
- No focus state of any kind on the http://localhost:8443/search/image?q=dogs results

sarayourfriend:
> In dev, I'm not seeing any pink focus rings anywhere. It might be an issue with focus-visible then?

That does indeed sound like an issue with `focus-visible` polyfill. I'll work on fixing the Chromium issue and see if I can reproduce those problems with iPad Safari.

sarayourfriend:
@zackkrida @obulat This is ready for re-review at your leisure  :slightly_smiling_face:

krysal (approved):
All looks good to me! Thanks for adding the README also, this change can be quite confusing without knowing the context.

PR #999, opened 2022-02-23
Add eslint rules for imports and eslint comments

### LINKED ISSUE #855: Add eslint-plugin-import to enforce import order and extension consistency
## Problem

There is no enforced consistency in how we import modules in JavaScript. Sometimes we use extensions for `.js` files and other times not.

In https://github.com/WordPress/openverse-frontend/discussions/465 we decided that it'd be good to enforce consistency here using eslint.

`eslint-plugin-import` has tons of helpful rules for this.

## Description

Add `esling-plugin-import` to the eslint configuration.

## Implementation

### LINKED ISSUE #901: Add eslint-plugin-eslint-comments to prevent accidentally leaving disabled rules for entire files
## Problem

Some ESlint rules can be disabled in block comments for specific parts of a file, we use this for disabled some texts that should _not_ be translated for example. The problem is that this approach is prone to forgetting to close the exception as happened in the `VImageDetails` component.
https://github.com/WordPress/openverse-frontend/blob/8e50891d71f437733835753f3975d3d4da0d759d/src/components/VImageDetails/VImageDetails.vue#L41-L50

## Description

Adding [eslint-plugin-eslint-comments](https://mysticatea.github.io/eslint-plugin-eslint-comments/) can help offload this burden from the developer who writes the exception comment as well as those who review the changes.

## Alternatives

Be aware in code reviews that ESlint comments are not applied without their corresponding closure.

## Additional context

Recommended in #898 by @sarayourfriend.

## Implementation

sarayourfriend:
## Fixes

Fixes #901 by @krysal 
Fixes #855 by @sarayourfriend 

## Description

Adds two new ESLint plugins, `eslint-comments` and `import` and configures them and fixes any errors they surfaced.

The biggest change here is of course the import order linting. I tried to create an ordering scheme that more or less matched what we'd been doing on our own. I'm not sure it's perfect and we can bikeshed a little bit about it in this PR (in particular whether treating assets and components as separate groups makes sense).

Overall just hoping to get some consistency :slightly_smiling_face: 

## Testing Instructions

Checkout the branch and try some of the new rules out. Almost all the import ones are auto-fixable which is really nice. The eslint-comments ones take a little more manual work.

Also review each of the plugins and see if there's any rules in their recommended sets that we want to disable. Likewise, see if there are any excluded from the recommended set that we'd like to use.

Please also let me know if there're any additional comments that would be helpful to add to the rule configurations.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/License/VLicense.vue
```diff
@@ -43,6 +41,8 @@ import nd from '~/assets/licenses/nd.svg'
 import pdm from '~/assets/licenses/pdm.svg'
 import sa from '~/assets/licenses/sa.svg'
 
+import VIcon from '~/components/VIcon/VIcon.vue'
```
sarayourfriend, line 44: Hmmm, I'd like components to come _before_ assets. I'm not sure why this one is after them :confused: Maybe my partial matching approach isn't working as expected.

.eslintrc.js
```diff
@@ -73,12 +75,89 @@ module.exports = {
     ],
     'unicorn/filename-case': ['error', { case: 'kebabCase' }],
     '@typescript-eslint/ban-ts-comment': ['warn'],
-    '@typescript-eslint/no-var-requires': ['warn'],
+    '@typescript-eslint/no-var-requires': ['off'],
```
krysal, line 78: This change seems unrelated, I'd think we have to discuss whether disabling this rule or not.

krysal (commented):
The codebase will look much more organized! I'll look into it further, but I did notice a small detail that maybe belongs in a separate issue/PR.

.eslintrc.js
sarayourfriend: Sure thing. I don't really know why those files are being imported using requires.

FWIW, I turned it off because we didn't previously have this rule until the introduction of the TypeScript eslint package, so it seemed safe to turn off and add back if we found it was causing issues with TypeScript stuff _in particular_. There's apparently no issue with it for other things?

src/components/License/VLicense.vue
sarayourfriend: Fixed in https://github.com/WordPress/openverse-frontend/pull/999/commits/66f9cd3f42f7767fcae5948072fe24ef94332651 . Indeed, I have no idea how minimatch's `partial` option is meant to work :shrug:

.eslintrc.js
sarayourfriend: Also, based on the description that the ESLint plugin gives, I think we can safely disable it anyway: https://typescript-eslint.io/rules/no-var-requires/#when-not-to-use-it

src/components/VContentSwitcher/VSearchTypeButton.vue
```diff
@@ -22,20 +22,22 @@
...
   inject,
   useContext,
 } from '@nuxtjs/composition-api'
+
+import { ALL_MEDIA, supportedSearchTypes } from '~/constants/media'
 import useSearchType from '~/composables/use-search-type'
 import { isMinScreen } from '~/composables/use-media-query'
 
-import caretDownIcon from '~/assets/icons/caret-down.svg'
+import VIcon from '~/components/VIcon/VIcon.vue'
```
obulat, line 36: It only reorders the imports, and does not add empty lines between groups, right? The blank line between two components # 37 could be removed, too.

src/components/VContentSwitcher/VSearchTypeRadio.vue
```diff
@@ -13,17 +13,19 @@
...
   IMAGE,
   supportedSearchTypes,
 } from '~/constants/media'
+
 import VIcon from '~/components/VIcon/VIcon.vue'
 
-import audioIcon from 'assets/icons/audio-wave.svg'
-import imageIcon from 'assets/icons/image.svg'
-import allIcon from 'assets/icons/all-content.svg'
+import audioIcon from '~/assets/icons/audio-wave.svg'
```
obulat, line 26: How did this get in 🤦 . I think it raises an error in tests when there is no leading `~/`

src/components/VContentSwitcher/VSearchTypes.vue
```diff
@@ -20,12 +20,15 @@
...
 </template>
 <script>
+import { computed, defineComponent } from '@nuxtjs/composition-api'
+
 import { supportedSearchTypes } from '~/constants/media'
+
 import useSearchType from '~/composables/use-search-type'
 
 import VItemGroup from '~/components/VItemGroup/VItemGroup.vue'
+
```
obulat, line 30: Oh, it even adds blank lines between components... 🤔

obulat (approved):
Thank you **so much** for adding this! It would make reviewing so much easier on the eyes!

src/components/VContentSwitcher/VSearchTypeRadio.vue
sarayourfriend: Interesting! Well our linter will prevent it from now :slightly_smiling_face:

zackkrida (approved):
I apologize for the delay—I had actually reviewed this previously, last week, and I guess I never submit my review for some reason. 

I had no idea import order rules existed, and I'm so excited that we're able to enforce standards we've been conventionally following but sometimes miss. The app is working as expected, the comments on the import order rules are great, and lgtm!

PR #1011, opened 2022-02-27
Use props instead of store for searchTerm (query.q)

### LINKED ISSUE #1010: Single type search pages (`search/<image|audio>`) should not use the store
## Problem

Currently the image and audio search result pages get the `searchTerm` (`store.search.state.query.q`) as a prop, but they don't use it when setting the page title. Instead, they are calling the store again, and getting the value from the store.

## Description

We should replace the store access with props:
https://github.com/WordPress/openverse-frontend/blob/3ab03476a378768d329e3a21ec0f3a402bb4b560/src/pages/search/audio.vue#L49-L54
```
  setup(props) {
    const { i18n } = useContext()
    useMeta({ title: `${props.searchTerm} - Openverse` })
```

## Alternatives

We could remove all props, and just use the store in these pages. This would make the component more difficult to test.

## Additional context

It would be nice to somehow document which components are 'presentational', i.e. do not use the store values, only using the `props` passed from the parent component. Also, having the props definition in a separate file hides what props the component has, which might cause repeated access to the same piece of data.

## Implementation

obulat:
## Fixes

Fixes #1010 by @obulat

## Description

This PR removes store access from the image and audio search result pages, and turns them again into `presentational` components, i.e., depending only on the props they get, and not the store values.
It's nice to add this change before converting the stores to Pinia so that the number of changes to review in that PR is at least a little bit lower.

## Testing Instructions

Run the app and search for images and audio. You should see the correct page titles:
**All content:**
Screen Shot 2022-02-27 at 9 32 01 AM
**Images:**
Screen Shot 2022-02-27 at 9 27 22 AM

BTW, why is it 'Cat | Openverse' for  'All content' and 'Cat - Openverse' for 'Images' and 'Audio'?

There is also an unrelated change to speed up the e2e test runs: when checking that an element does not exist with `await expect(element).toHaveCount(0)`, Playwright does not detect an element, so it waits for a default time period (probably 5 seconds), which makes the test take longer. I added a timout to `toHaveCount(0)` calls to speed the tests up.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
> BTW, why is it 'Cat | Openverse' for 'All content' and 'Cat - Openverse' for 'Images' and 'Audio'?

I think it's just a mistake, feel free to correct them both to use the pipe (which I think is what most pages use) or open an issue for it.

sarayourfriend:
> There is also an unrelated change to speed up the e2e test runs: when checking that an element does not exist with await expect(element).toHaveCount(0), Playwright does not detect an element, so it waits for a default time period (probably 5 seconds), which makes the test take longer. I added a timout to toHaveCount(0) calls to speed the tests up.

Nice! This totally makes sense and but would never have occurred to me. It seems obvious now though that you've pointed it. I wonder if we could add an eslint warning to help prevent `toHaveCount(0)` from slowing down tests?

sarayourfriend (approved):
LGTM! Tested SSR and client side rendering and it works perfectly. Nice to remove unnecessary usage of the store.

src/pages/search/audio.vue
```diff
@@ -47,11 +46,9 @@ const AudioSearch = defineComponent({
   },
   props: propTypes,
   setup(props) {
-    const store = useStore()
     const { i18n } = useContext()
 
-    const query = computed(() => store.state.search.query.q)
-    useMeta({ title: `${query.value} - Openverse` })
+    useMeta({ title: `${props.searchTerm} - Openverse` })
```
zackkrida, line 51:
```suggestion
    useMeta({ title: `${props.searchTerm} | Openverse` })
```

While we're touching this it might be good to fix the separator.

test/e2e/search-types.spec.js
```diff
@@ -74,10 +74,21 @@ async function checkSearchMetadata(page, searchType) {
   }
 }
 
+async function checkPageMeta(page, searchType) {
+  const sep = searchType.id === 'all' ? '|' : '-'
```
zackkrida, line 78: Okay, let's definitely change the separator, as identified above, so we can eliminate this logic.

zackkrida (approved):
Lets update the separator to `|` everywhere so we can simplify the code, and then LGTM.

PR #1023, opened 2022-02-28
Convert 6 utils to TypeScript

### LINKED ISSUE #922: Add types to `utils/decode-data.js`
## Description

Add type checking to `decode-data.js`. This module has zero dependencies. Make sure to add it to `tsconfig.json`'s `include` list.

## Additional context

Part of an ongoing effort to add type checking to parts of the project that can be type checked.

## Implementation

### LINKED ISSUE #924: Add types to `utils/dev.js`
## Description

Add type checking to `utils/dev.js`. This module has zero dependencies. Make sure to add it to `tsconfig.json`'s `include` list.

## Additional context

Part of an ongoing effort to add type checking to parts of the project that can be type checked.

## Implementation

### LINKED ISSUE #925: Add types to `utils/string-to-boolean.js`
## Description

Add type checking to `string-to-boolean.js`. This module has zero dependencies. Make sure to add it to `tsconfig.json`'s `include` list.

## Additional context

Part of an ongoing effort to add type checking to parts of the project that can be type checked.

## Implementation

### LINKED ISSUE #926: Add types to `utils/env.js`
## Description

Add type checking to `env.js`. This module has one dependency which will be typed in #925. Make sure to add it to `tsconfig.json`'s `include` list.

Blocked by #925

## Additional context

Part of an ongoing effort to add type checking to parts of the project that can be type checked.

## Implementation

### LINKED ISSUE #936: Add types to `utils/resampling.js`
## Description

Add type checking to `resampling.js`. This module has zero dependencies. Make sure to add it to `tsconfig.json`'s `include` list.

## Additional context

Part of an ongoing effort to add type checking to parts of the project that can be type checked.

## Implementation

### LINKED ISSUE #938: Add types to `utils/sentry-config.js`
## Description

Add type checking to `sentry-config.js`. This module has zero dependencies. Make sure to add it to `tsconfig.json`'s `include` list.

## Additional context

Part of an ongoing effort to add type checking to parts of the project that can be type checked.

## Implementation

dhruvkb:
## Fixes

Works on [milestone 10 i.e. TypeScriptification](https://github.com/WordPress/openverse-frontend/milestone/10)

Fixes #922
Fixes #924
Fixes #925
Fixes #926
Fixes #936
Fixes #938

## Description

This PR
- converts a number of utilities to TypeScript
- updates usages of TS-ified utilities
- adds a lint rule for proper TSDoc comments
- converts the Nuxt config to TS and fixes type mismatches
- refactors code as needed

## Testing Instructions

Passing CI should be sufficient imo. You can try checking out the PR, running the dev server and playing with it for a bit. Repeat with the Storybook.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (commented):
Looking great so far! Will happily approve once un-drafted :tada:

dhruvkb:
@sarayourfriend if that's okay I'm planning to hit all the utils in one go and reduce the overhead of approvals, merges and conflicts.

src/utils/resampling.ts
```diff
@@ -2,47 +2,46 @@
...
+ * Resizes a given array to consist of fewer elements than provided. This uses
  * the Largest Triangle Three Buckets algorithm by Sveinn Steinarsson.
  *
  * @see {@link https://github.com/sveinn-steinarsson/flot-downsample}
  *
- * @param data {number[]} the list of data points to interpolate
- * @param threshold {number} the number of expected data points from the array
- * @returns {number[]} the array with the required number of points
+ * @param data - the list of data points to interpolate
+ * @param threshold - the number of expected data points from the array
```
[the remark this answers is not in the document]
dhruvkb: Added this in 6a3b521. Have currently restricted this rule to only apply to `.ts` files.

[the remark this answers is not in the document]
sarayourfriend: If you don't mind, it'd probably be better to add it to #999 or just put it into a separate PR entirely. I'd like to go ahead and apply the rules to JS, TS and Vue SFCs as well in one go.

sarayourfriend:
@dhruvkb That might be fine for most of the utils, but I'd encourage splitting anything that has any runtime changes or substantial type additions into a separate PR, especially if the module isn't well unit tested. In my experience in adding types to various Gutenberg packages it was actually much faster go module by module mostly because it made it far easier for reviewers. If converting to TS just means renaming the file and adding parameter and return types to a function and nothing else then it's probably safe to batch those PRs. Otherwise the burden for the reviewer goes up a lot and I'd worry that you'd get stalled just waiting for reviews on a big PR.

src/utils/warn.ts
```diff
@@ -0,0 +1,9 @@
+export const getLogger = (level: 'log' | 'warn' | 'error') =>
+  process.env.NODE_ENV !== 'production'
+    ? console[level]
+    : () => {
+        // do nothing
+      }
```
sarayourfriend, line 8: Unfortunately this makes the types slightly worse than they were before. To get the right types we'll need to do something like this:

```ts
import { isNotProd } from '~/utils/dev'

export const getLogger = <T extends keyof typeof console>(level: T): typeof window.console[T] =>
  isNotProd
    ? console[level]
    : (() => {
        // do nothing
      }) as typeof console[T]

export const warn = getLogger('warn')
export const log = getLogger('log')
```

sarayourfriend: We can probably also rename the module to `console` if it's going to include more than just `warn` :grin:

dhruvkb:
@sarayourfriend you're right. In that case, I'll stop this PR here and open it to review.

Dockerfile
```diff
@@ -76,7 +76,7 @@ COPY pnpm-lock.yaml .
 COPY .npmrc .
 
 # copy the nuxt configuration file
-COPY --from=builder /usr/app/nuxt.config.js .
+COPY --from=builder /usr/app/nuxt.config.ts .
```
sarayourfriend, line 79: :scream: For some reason I didn't actually think we could use TypeScript for the nuxt config until Nuxt bridge/Nuxt 3 but this is great!!

src/locales/scripts/types.d.ts
```diff
@@ -1,3 +1,9 @@
-/**
- * @typedef {{ code: string, wpLocale?: string, file?: string, iso?: string, name: string, dir?: string, translated?: number }} i18nLocaleProps
- */
+interface i18nLocaleProps {
```
sarayourfriend, line 1: Nit (and not even really part of the changes you made but would be good to change if we can): Using uppercase letters to start interfaces and types is considered good practice as a way to easily distinguish them from instances.

We can also use just a regular `.ts` file for this rather than a `.d.ts`. It's also necessary to add an `export` to this declaration (JSDoc `@typedef` types are implicitly exported where as in regular TypeScript you have to be explicit about which types to export).

```suggestion
export interface I18nLocaleProps {
```

src/utils/resampling.ts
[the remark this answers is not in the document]
dhruvkb: Adding this to JS and Vue is leading to a lot of errors because of the `typedef` comments. It might be good to re-enable it for JS after a major chunk of the migration is complete.

src/utils/dev.ts
```diff
@@ -0,0 +1,3 @@
+export const isDev: boolean = process.env.NODE_ENV === 'development'
+export const isTest: boolean = process.env.NODE_ENV === 'test'
+export const isNotProd: boolean = process.env.NODE_ENV !== 'production'
```
sarayourfriend, line 3: Should we just name this `isProd` and inverse the condition? I'm worried we'll end up with usages of this that look like `if (!isNotProd)` which introduces a confusing double negative boolean, always good to ignore those in my experience :sweat_smile: 

```suggestion
export const isDev: boolean = process.env.NODE_ENV === 'development'
export const isTest: boolean = process.env.NODE_ENV === 'test'
export const isProd: boolean = process.env.NODE_ENV === 'production'
```

Of course we'll need to update where it's used as well.

src/utils/warn.ts
dhruvkb: That makes sense, updated in 3ad59c2e.

src/utils/env.ts
```diff
@@ -1,15 +1,10 @@
...
-export const env = {
+export const env: Record<string, string> = {
   apiUrl: process.env.API_URL ?? 'https://api.openverse.engineering/v1/',
   filterStorageKey: 'openverse-filter-visibility',
   notificationStorageKey: 'openverse-show-notification',
-  enableInternalAnalytics: stringToBoolean(
-    process.env.ENABLE_INTERNAL_ANALYTICS
-  ),
+  enableInternalAnalytics: process.env.ENABLE_INTERNAL_ANALYTICS ?? 'false',
 }
```
sarayourfriend, line 10: Let's use `as const` here: it's perfect for this kind of thing. `Record<string, string>` is actually even worse than what TypeScript infers as it swallows all the key names. `as const` is the best though as it will turn constant strings into constant string types instead of just `string`. Changing `enableInternalAnalytics` to a `string` boolean representation instead of just a plain boolean also strikes me as a step backwards, what was the reason for that?

tsconfig.json
```diff
@@ -46,12 +46,10 @@
     "src/data/api-service.js",
     "src/data/usage-data-service.js",
     "src/store/user.js",
-    "src/utils/decode-data.js",
```
sarayourfriend, line 49: To get `nuxt.config.ts` included in the type checking we'll need to change the first entry in `includes` to just `**/*.ts` I think. Right now it's only type checking stuff inside the `src` file (which isn't good considering we'll be adding TypeScript to the `test` directory as well with the visual regression stuff.

Dockerfile
ronny-badilla-SWO: Question: if the previous stage is the builder and its entrypoint is `npm run build`, does that means that we shouldn't have any `.ts` file at this point?

rbadillap: Question: if the previous stage is the `builder` and its entrypoint is `npm run build`, does that means that we shouldn't have any `.ts` file at this point?

src/utils/resampling.ts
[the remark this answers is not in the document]
sarayourfriend: Yeah, I think it'd be better to use a generic JSDoc one than a TSDoc thing as TSDoc is actually a specific _separate_ standard than JSDoc with different rules in some cases.

Dockerfile
sarayourfriend: That's a good question :thinking: The Nuxt config is used both to generate the Webpack config but also to configure some runtime properties for Nuxt, so it's used both for building _and_ for runtime. Nuxt's runtime doesn't use TypeScript though, at least not in Nuxt 2, I don't think.

rbadillap: Ok I understand now, I went to the docs, and `npm start` can differ its behavior depending the way you build the application (`npm run build`).

Since the [deployment target](https://nuxtjs.org/docs/features/deployment-targets) is not being specified in the [config file](https://github.com/WordPress/openverse-frontend/blob/main/nuxt.config.js), then the default value is [`server`](https://nuxtjs.org/docs/get-started/commands/#target-server-default-value) which means, we need Node.js to run the application from the generated `.nuxt` folder.

> nuxt start - Start the production server (after running nuxt build). Use it for Node.js hosting like Heroku, Digital Ocean, etc.

That means, at this point, if we want to keep using typescript (after building the application from the previous stage) we will require `ts-node`. Or, my preferred option, as you said Sara, don't convert the config file so it can be read by Node.js runtime without any external dependency.

sarayourfriend: Yeah, I'd prefer if we kept it as TS for now until we move to Nuxt bridge. To use TS for this properly it looks like we'd have to bring in another module for it: https://typescript.nuxtjs.org/guide/runtime/

We can always add type-checking to the file even if it stays as JSDoc though; it just can't consume actual TS files (which means anything we use in the `nuxt.config.js` needs to stay as regular JS with JSDoc types instead of getting converted to TypeScript).

src/utils/dev.ts
dhruvkb: Thanks for noticing this, resolved in [ad3b2b1](https://github.com/WordPress/openverse-frontend/pull/1023/commits/ad3b2b18b38e1868715e792009bf0c33be0bcaab).

src/locales/scripts/types.d.ts
dhruvkb: Fixed in [16cd2c9](https://github.com/WordPress/openverse-frontend/pull/1023/commits/16cd2c9d63c064e515d50842fe2912a9195c7b07).

src/utils/env.ts
sarayourfriend: Makes sense! I wonder what a good way to enforce that would be :thinking: I guess `Record<string, string>` would be it but having the keys swallowed is a bit of a bummer :confused:

Dockerfile
dhruvkb: If we don't use TS for `nuxt-config`, we cannot import the TypeScript versions of the utils `node-env`, `env` and `sentry-config` in it. I'm not entirely sure why making the `nuxt-config` file TS works but if I change it to JS, it can no longer import any TS files.

src/utils/env.ts
dhruvkb: We wouldn't need to enforce this, `tsc` will raise an error if we use anything other than a string as the value. Also using `as const` with it works.

src/utils/warn.ts
sarayourfriend: @dhruvkb did you miss https://github.com/WordPress/openverse-frontend/pull/1023#discussion_r816127505 or was it something you think we shouldn't do?

Dockerfile
sarayourfriend: JS files can't import TS files without the JS files being compiled first. Because a `.js` nuxt.config doesn't get compiled, it won't be able to import TS directly I guess (because it's not importing anything from the compiled bundle).

Considering the e2e tests are passing (they run on the `pnpm build && pnpm start` version of the app, for @rbadillap's context) I guess it's safe.

Also don't know why it's working though (which scares me a little bit :sweat_smile:)

src/utils/warn.ts
dhruvkb: This is slightly confusing because I'm still getting the right types. If I import `log` or `warn` from this module they show the same type as `console.log` and `console.warn`.

sarayourfriend: This is the difference I see:

https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3gSzQBRiWAC54AjJJCEKFAbgCgHl0sBzEDAGSTY5ngBeeAAoA1iACepAOQReM+AB94MgO6wUilTIxQy1GQEohAPgbxL8AIS4CReAH5EqNFRABtCZIC68UiIiJoKm8ADeAL5GTCyY8PJsQvAc3Lz8InIK0QD02VbwAHqOMa5YGjAoSSk8fCAwGeVaOXlWRSWs8HoGCMLVaXUZXYbN+W3tcQBm2AAeIMAA4pw1-EkAPAAq8CDTGCAowGjw3kgT8ADCru6m4lKk60akF+juHut+IRZWtviEwE4uz2oXikfgCQTM4Si8CghyebiBbyYzFK8V4ACYklNZgslv16pk2MYGLl8oVisiOo0McIsXNFqlavjGkSSa1ybEsEMQNT4LScQz0rp9MNiS1LG0gA

I'm not sure exactly why it squashes everything to `(...data: any[]) => void`. I think because it's merging the empty function and the actual console function's type? But it doesn't fix it if you cast the empty function either so I don't think that's why really:

https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3gSzQBRiWAC54AjJJCEKFAbgCgHl0sBzEDAGSTY5ngBeeAAoA1iACepAOQReM+AB94MgO6wUilTIxQy1GQEohAPgbxL8AIS4CReAH5EqNFRABtCZIC68UiIiJoKm8ADeAL4mUGjwGJIADiBIAGYu6O4e8Ump8N4+TCyY8PJsQvAc3Lz8InIKRgwA9I1W8AB6joWuWBowKOWVPHwgMLW9Wg3NrR1drHH61AOcQzW6CyDGTS1WM7PFKdgAHiDAAOLL1SPlADwAKvAghxggKMCx3rkAwq7upuJSpFuRlI3wy1A8tz8IQsVls+EIwCc6Tc4Py-lEQTM4Si8Bi8FBKM8kKYzG6JV4ACZygdjmcLsNRnU2JsptNOqS5uMqcIaSdzlUGWNNCztpZdkUsHoDCBufBeXSBaspYZJqL2p0gA

Anyway, it's not a big deal.

.eslintrc.js
```diff
@@ -18,7 +18,13 @@ module.exports = {
     'plugin:@intlify/vue-i18n/recommended',
   ],
   // required to lint *.vue files
-  plugins: ['@typescript-eslint', 'vue', 'vuejs-accessibility', 'unicorn'],
+  plugins: [
+    '@typescript-eslint',
+    'eslint-plugin-tsdoc',
```
sarayourfriend, line 23: I still think we might want to use https://www.npmjs.com/package/eslint-plugin-jsdoc instead of a TSDoc one. We have no real idea of how long Vue files will be excluded from TS and it'd be nice to have consistency across the stack with our JSDoc comments.

src/components/VIcon/VIcon.vue
```diff
@@ -33,7 +35,7 @@ export default {
       /**
        * In `jest` our icons get transformed to Vue components
        */
-      type: process.env.NODE_ENV === 'test' ? Object : String,
+      type: isTest ? Object : String,
```
sarayourfriend, line 38: I would like to check that dead code elimination still works when you don't inline the NODE_ENV check. I don't think we have environment based dead code currently but it would be good to add that. We might end up having to revert these checks back to the inline but I'm not sure.

sarayourfriend (approved):
LGTM. Still not sure about the TSDoc linting instead of just JSDoc, would like others to give input there. The console types I'd personally like changed but are also not a blocker.

obulat (approved):
Exciting changes! Thank you for keeping this PR small. With all of the `console` changes it already looks kind of intimidating :)

src/components/VIcon/VIcon.vue
dhruvkb: I'll make a new issue for this, it seems that this should work because it'll evaluate to 
```js
if (false) {
  // ...
}
```
in the compiled JS.

dhruvkb: Opened #1030.

PR #1029, opened 2022-03-01
Add linting for JSDoc

dhruvkb:
## Description

This PR tests the possibility and impact of linting JSDoc comments in our JS and Vue files. The main argument against the addition of this ESLInt plugin would be that it will raise a lot of warnings (never errors).

## Testing Instructions

0. Check out the PR.
1. Run the `lint` script.
```bash
$ pnpm run lint
```
2. Observe the errors and warnings.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
FWIW Gutenberg also uses this plugin.

zackkrida:
I support using this plugin; great idea! Agree with @sarayourfriend regarding making it stricter and more useful.

.eslintrc.js
```diff
@@ -75,6 +75,16 @@ module.exports = {
     '@typescript-eslint/ban-ts-comment': ['warn'],
     '@typescript-eslint/no-var-requires': ['warn'],
   },
+  overrides: [
+    {
+      files: ['*.js', '*.vue'],
+      extends: ['plugin:jsdoc/recommended'],
+      rules: {
+        'jsdoc/check-tag-names': ['warn', { definedTags: ['values'] }],
+        'jsdoc/require-hyphen-before-param-description': ['warn'],
```
sarayourfriend, line 84:
```suggestion
        'jsdoc/check-tag-names': ['error', { definedTags: ['values'] }],
        'jsdoc/require-hyphen-before-param-description': ['error'],
```

For something that is purely style I don't see the value in warnings at all. At most they'll just be annoying and still get ignored and we still won't have consistency (but every PR will be plagued by 20 additional ESLint comments added by the CI).

.eslintrc.js
```diff
@@ -75,6 +75,16 @@ module.exports = {
     '@typescript-eslint/ban-ts-comment': ['warn'],
     '@typescript-eslint/no-var-requires': ['warn'],
   },
+  overrides: [
+    {
+      files: ['*.js', '*.vue'],
+      extends: ['plugin:jsdoc/recommended'],
+      rules: {
+        'jsdoc/check-tag-names': ['warn', { definedTags: ['values'] }],
```
sarayourfriend, line 83: What is the `values` tag?

.eslintrc.js
dhruvkb: I wanted these to be warnings till we fix these issues and then escalate them to errors.

sarayourfriend: I guess I don't fully see what the plan is. Is it to merge this then open separate PRs to fix the various issues? Or a single follow up PR to fix them all? Why not just fix them in this PR, considering they're not runtime changes and present no risk to the app's functionality? It'd also give us a sense of the scope of the rules being proposed and whether they follow the existing style of the repository or if we'd be imposing a new style (either of which can be okay, it's just something we should considering and be aware of as we're merging in new linting rules).

.eslintrc.js
sarayourfriend: Nice, for anyone else curious about that, documentation is here: https://vue-styleguidist.github.io/docs/Documenting.html#available-tags

Should we add a link to that in the rule configuration so future Openers won't have the same question?

.eslintrc.js
dhruvkb: All the rules defined in `plugin:jsdoc/recommended` are also warnings by default.

dhruvkb:
Even at this stage, this plugin doesn't work all that well with all the TypeScript types in the docs so I propose to leave this PR as is and revisit after the TypeScript milestone.

.eslintrc.js
sarayourfriend: That's a shame, they don't even have a `recommended-error` or something :disappointed: 

Here would be my recommendation:

```json
    "rules": {
        "jsdoc/check-access": ['off'], // The `@access` tag isn't useful with modules
        "jsdoc/check-alignment": ['error'], // This prevents misaligned astriks
        "jsdoc/check-examples": ['off'], // This rule doesn't work with eslint 8
        "jsdoc/check-indentation": ['error'], // Prevents extra indentation outside of `@example`
        "jsdoc/check-line-alignment": ['error'], // I don't personally like this rule but Gutenberg uses it and aligning with them seems good if others are okay with it. I find space aligned text annoying because I use a proportional font and they look very wonky and bad to me, but for most folks it would be easier to read.
        "jsdoc/check-param-names": ['error'], // Keeps the documentation up-to-date with implementation
        "jsdoc/check-property-names": ['error'], // Hopefully this one becomes irrelevant as we move to TypeScript but doesn't hurt
        "jsdoc/check-syntax": ['off'], TypeScript compiler will catch this when it actually matters
        "jsdoc/check-tag-names": ['error'], // With the additional `@values` tag you added for Vue
        "jsdoc/check-types": ['off'], // Again, TS will catch this for us
        "jsdoc/check-values": ['off'], // This doesn't check anything that would matter to us
        "jsdoc/empty-tags": ['error'], // Ensure tags that aren't meant to have values don't have them
        "jsdoc/implements-on-classes": ['off'], // Avoid using JSDoc as a replacement for class-based semantics that don't really exist in JavaScript anyway
        "jsdoc/match-description": ['error'], // Sure why not, we do this in Python anyway.
        "jsdoc/multiline-blocks": ['error'], // Prevents wonky formatting
        "jsdoc/newline-after-description": ['error'], // More whitespace is good
        "jsdoc/no-bad-blocks": ['error'], // Keeps two asterisks on the opening line
        "jsdoc/no-defaults": ['off'], // Allows us to document default values
        "jsdoc/no-missing-syntax": ['off'], // I'm not sure how this one works, I don't think we need it
        "jsdoc/no-multi-asterisks": ['error', { allowWhitespace: true }], // Seems sensible to me
        "jsdoc/no-restricted-syntax": ['off'], // I don't think we need this and we'd have to configure specifics for it anyway
        "jsdoc/no-types": ['off'], // I guess because we're using the TSDoc linter for TS we don't need this
        "jsdoc/no-undefined-types": ['off'], // This rule is broken with TypeScript type params
        "jsdoc/require-asterisk-prefix": ['error'], // No comment
        "jsdoc/require-description": ['off'],
        "jsdoc/require-description-complete-sentence": ['error'], // Matches our Python style (I think)
        "jsdoc/require-example": ['off'],
        "jsdoc/require-file-overview": ['off'],
        "jsdoc/require-hyphen-before-param-description": ['error', "never"], // My own preference is to not use the hyphen, I don't think we use it currently, but feel free to change this to 'always' if that's what folks want
        "jsdoc/require-jsdoc": ['off'], // Not sure we need this
        "jsdoc/require-param": ['error'],
        "jsdoc/require-param-description": ['error'],
        "jsdoc/require-param-name": ['error'],
        "jsdoc/require-param-type": ['error'], // TS might catch this too, not sure, but it might also just use implicit any which we don't want. Might as well enforce adding types.
        "jsdoc/require-property": ['errpr'], // Prevent `object` and `any` aliases
        "jsdoc/require-property-description": ['error'],
        "jsdoc/require-property-name": ['error'],
        "jsdoc/require-property-type": ['error'],
        "jsdoc/require-returns": ['error'],
        "jsdoc/require-returns-check": ['error'], // This one seems sensible, prevents adding a `return` annotation when the function doesn't return anything
        "jsdoc/require-returns-description": ['error'],
        "jsdoc/require-returns-type": ['error'],
        "jsdoc/require-throws": ['off'], // We're not writing Java :slightly_smiling_face: 
        "jsdoc/require-yields": ['error'], // Don't think we have any generators but might as well document them
        "jsdoc/require-yields-check": ['error'],
        "jsdoc/tag-lines": ['error'], // Stop adding newlines between tags... Not sure about this one TBH
        "jsdoc/valid-types": ['off'] // TS will catch this when it matters
    }
```

sarayourfriend:
> Even at this stage, this plugin doesn't work all that well with all the TypeScript types in the docs so I propose to leave this PR as is and revisit after the TypeScript milestone. We already have good linting for TSDoc so this PR might not be as necessary after the migration.

It would be necessary for Vue SFCs where TypeScript can't be used.

We can also just disable that type syntax rule.

Are we expecting that everything gets converted to TypeScript? There are some things we can just leave as JSDoc with types (why spend time converting them if they're already checked?) unless we actually want to get rid of plain JS entirely.

Either way, Vue SFCs won't be able to use TS until the Nuxt bridge upgrade I don't think.

sarayourfriend:
Cross posting this comment here as well: https://github.com/WordPress/openverse-frontend/pull/1052#discussion_r819547321

The TSDoc plugin doesn't cover some of the things the JSDoc plugin covers, so it'd be nice to have both.

src/components/VContentSwitcher/VSearchTypePopover.vue
```diff
@@ -67,6 +67,7 @@ export default defineComponent({
     /**
      * When in the searchbar, content switcher button has a border when the
      * search bar group is hovered on.
+     *
      * @type {ComputedRef<boolean>}
```
obulat, line 71: I guess this type annotation is incorrect because it did not import `ComputedRef` from `@nuxt/composition-api`... WebStorm automatically inserts types without the imports.

src/components/VButton.vue
```diff
@@ -158,7 +158,8 @@ const VButton = defineComponent({
   },
   /**
    * @param {Props} props
-   * @param {import('@nuxtjs/composition-api').SetupContext}
+   * @param root0
```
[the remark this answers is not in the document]
sarayourfriend: It's because the context wasn't given a name. It should accept this though:

```suggestion
   * @param {import('@nuxtjs/composition-api').SetupContext} context
```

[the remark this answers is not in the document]
dhruvkb: It seems to be two different steps
- removed the `@param {...}` because there isn't a variable after it
- adds the `root0` and `root0.attrs` to document the `{ attrs }` param of the function

src/data/media-provider-service.js
```diff
@@ -2,8 +2,9 @@ import ApiService from './api-service'
 
 /**
  * Service that calls API to get Media Provider stats
+ *
  * @param {('image'|'audio')} mediaType
- * @constructor
+ * @function Object() { [native code] }
```
obulat, line 7: Is this type annotation correct?

src/plugins/url-change.js
```diff
@@ -7,6 +7,10 @@ import { useNavStore } from '~/stores/nav'
  * Normally, the `src/middleware/embed.js` middleware does this on every
  * route change. However, it does not run on the initial render.
  * So, for the initial render this plugin runs when router is ready.
+ *
+ * @param root0
```
obulat, line 11: Is it possible to remove these changes in the `.js` files? I guess we are going to move to TS in them soon, so it seems like it adds more work to rename `root0` annotations, and then to replace them with TS

src/plugins/vue-i18n.js
```diff
@@ -4,8 +4,8 @@ export default () => {
     silentFallbackWarn: true,
     pluralizationRules: {
       /**
-       * @param choice {number} a choice index given by the input to $tc: `$tc('path.to.rule', choiceIndex)`
-       * @param choicesLength {number} an overall amount of available choices
+       * @param choice - {number} a choice index given by the input to $tc: `$tc('path.to.rule', choiceIndex)`
```
obulat, line 7: Hm, intersting. It doesn't recognize that there is a type annotation that should be switched with the param name here, and simply adds a dash in front of it.

obulat, line 7:
```suggestion
       * @param {number} choice - a choice index given by the input to $tc: `$tc('path.to.rule', choiceIndex)`
```

src/plugins/vue-i18n.js
```diff
@@ -4,8 +4,8 @@ export default () => {
...
+       * @param choicesLength - {number} an overall amount of available choices
```
obulat, line 8:
```suggestion
       * @param {number} choicesLength - an overall amount of available choices
```

src/plugins/vue-i18n.js
sarayourfriend: It's because the order is wrong, the type annotation comes before the name in JSDoc :slightly_smiling_face:

src/plugins/url-change.js
sarayourfriend: If we kept param documentation in the TS docstrings then we'd need these anyway, FWIW. It uses `root0` as the default name (and `root1` and so on if there are multiple destructured arguments), ideally we'd replace these with more sensible names.

obulat (approved):
I love the consistency that linting brings! I think it's better to lint all of the code once, and add any missing parameter descriptions gradually.

sarayourfriend:
@dhruvkb any thoughts on how to finish up this PR? Should we close this an add a new one?

dhruvkb:
I haven't been actively working on it, and I don't think I am capable of completing it.

PR #1035, opened 2022-03-03
Add width and height properties to images

zackkrida:
## Fixes

Related to #1034 by @zackkrida 

## Description

Adds width and height properties to images, as well as setting the background to grey while they are loading. We _could_ and maybe _should_ wait to display images until they are loaded, and show the skeleton components while they are loading, but this still makes an improvement to the current loading appearance via the grey backgrounds. 

## Testing Instructions

Observe the appearance of the all results grid and image results views while loading.

> before

CleanShot 2022-03-02 at 18 46 14@2x

CleanShot 2022-03-02 at 18 46 35@2x

> after (loading appearance while on a throttled 4g connection)

CleanShot 2022-03-02 at 19 21 26@2x

CleanShot 2022-03-02 at 19 19 11@2x

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VAllResultsGrid/VImageCellSquare.vue
+++ b/src/components/VAllResultsGrid/VImageCellSquare.vue
@@ -13,12 +13,12 @@
     >
       <img
         ref="img"
-        class="w-full h-full object-cover rounded-sm"
+        class="w-full h-full object-cover rounded-sm bg-dark-charcoal-10 text-dark-charcoal-10"
         loading="lazy"
         :alt="image.title"
         :src="getImageUrl(image)"
-        :width="image.width"
-        :height="image.height"
+        :width="250"
+        :height="250"
         itemprop="thumbnailUrl"
         @error="onImageLoadError($event, image)"
       />
--- a/src/components/VImageGrid/VImageCell.vue
+++ b/src/components/VImageGrid/VImageCell.vue
@@ -19,7 +19,7 @@
         :alt="image.title"
         :src="getImageUrl(image)"
         :width="imgWidth"
-        :height="imageight"
+        :height="imgHeight"
         @load="getImgDimension"
         @error="onImageLoadError($event, image)"
       />
--- a/src/components/VImageGrid/VImageCell.vue
+++ b/src/components/VImageGrid/VImageCell.vue
@@ -1,7 +1,7 @@
 <template>
   <VLink
     :href="'/image/' + image.id"
-    class="w-full block group relative overflow-hidden rounded-sm focus:ring-[3px] focus:ring-pink focus:ring-offset-[3px] focus:outline-none"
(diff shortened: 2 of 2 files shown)


src/components/VAllResultsGrid/VImageCellSquare.vue
```diff
@@ -13,12 +13,12 @@
     >
       <img
         ref="img"
-        class="w-full h-full object-cover rounded-sm"
+        class="w-full h-full object-cover rounded-sm bg-dark-charcoal-10 text-dark-charcoal-10"
         loading="lazy"
         :alt="image.title"
         :src="getImageUrl(image)"
-        :width="image.width"
-        :height="image.height"
```
zackkrida, line 21: these properties aren't actually defined, so it's better to use a default with the correct aspect ratio, for now.

obulat: Are these properties never defined and will never be used? Or could we have something like `"image.width ?? 250"`?

obulat (approved):
This creates a much nicer experience!

Could we also maybe add skeleton loaders for when we click on 'Load more', before the images are loaded? Now it takes some time for the gray squares to appear, and then some more time for the images to show.

src/components/VAllResultsGrid/VImageCellSquare.vue
dhruvkb: I second @obulat's comment, that'll allow us to send the dimensions from the backend without needed a frontend change/deploy.

zackkrida: > absolutely should have dimensions coming from the API.

Yes, but none of our provider scripts set these fields. This suggestion is well-taken though! I'll do it this way.

sarayourfriend (approved):
It's not perfect (especially for the images-only search) but definitely an improvement! Getting the dimensions into the catalog would do wonders to fix this problem altogether.

src/components/VAllResultsGrid/VImageCellSquare.vue
krysal: In fact, the StockSnap and the WordPress Photos Directory scripts do assign these values, for the rest it is a matter of updating them.

https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/stocksnap.py#L141-L142

https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/providers/provider_api_scripts/wordpress.py#L259-L260

zackkrida: That's great! Unfortunately those properties aren't exposed on the list view, just in the single result view:

https://api.openverse.engineering/v1/images/?q=dog&source=stocksnap

PR #1040, opened 2022-03-03
Convert the search store to Pinia

### LINKED ISSUE #1037: Convert `search` store from Vuex to Pinia
## Problem

As part of the conversion of the Vuex store modules to Pinia, we need to convert the [search store](https://github.com/WordPress/openverse-frontend/blob/e0ac241dfe8b2b4b7eaff8eaee7e06eda0dce197/src/store/search.js) to Pinia.

This is a fairly large store, so the conversion should have 3 stages:
1. Extracting the filter initialization to a constants file.
2. Extracting everything related to filters to a new Pinia store.
3. Converting what's left in the search store (`searchTerm` and `query`) to Pinia.

## Description

-  convert state to reactive

## Additional context

## Implementation

### LINKED ISSUE #1121: Reduce to a single source of truth for search filters
## Problem

Currently there's a relatively complex relationship between the search and filter store. The search is intended to represent data that is explicitly derivative of the filter store. In particular it creates an interface for the filter store appropriate for URL query params that are used for updating the frontend URL as well as for making requests to the API. The search store also manages deriving updated filter state from changed URL query params (for example, during backwards and forwards navigation when the route changes or when the page first loads, the filter state needs to be derived from the URL query params).

## Description

This discussion originated in this PR review: https://github.com/WordPress/openverse-frontend/pull/1040#pullrequestreview-909942347

Here's the 1000 meter view as far as I can tell. @obulat feel free to correct me if I'm missing anything here.

The filter store encodes filters in a way that makes it easy to manage the filter UI, as fundamentally a list of checkboxes with on and off states. The filter store also is responsible for deciding what filters are available for a given search type (images, audio, and all-content searches all have different types and levels of filters available to them).

Since https://github.com/WordPress/openverse-frontend/pull/1040 and before, the search store is used exclusively by the code to update the search term, the search type, and as an interface to update individual filters through the methods `updateQuery` and `toggleFilter` among others. The gist of all of these functions goes something like this:

1. Accept some values to update
2. Update the values in the filter store
3. Call a function that recalculates the search store values based on the updated filter store

This follows a pattern that implies that the search store is actually dependent on the filter store state. The only reason the search store is being used to mutate the filter store state is essentially to "notify" it of the updates so that it can recalculate its state.

Vue already provides a mechanism for that: `computed` and `watch`.

Rather than expose confusing duplicate functions on the filter store and the search store, we could treat the `filter` store as the single source of truth with the search store being composed completely of state derived directly from (using `computed` and `watch`) the filter store.

The filter store would need to be updated with a method that accepts a URL query params string and parses it to the filter store state. This would replace the single independent activity of the search store which is to derive filter state from URL query params.

Then the search store could literally just be derived state and expose no functions for manipulating its state directly. Doing this would prevent the following class of bugs:

1. Updating filters in one place and not the other (which we already experienced before with the back and forward browser buttons updating filter checkbox state but not the actual search state)
2. New contributors (or lets face it, experienced contributors too) being unsure of whether to call `toggleFilter` on the filter store or the search store.

## Alternatives

We could treat the URL query params as the single source of truth and have the filter store interface directly with it via `useRouter`. I'm not sure if this would be better or worse to reason about, but it could be worth exploring both options to see which makes more sense and is easier to work with.

## Implementation

### LINKED ISSUE #1128: Europeana and SoundCloud don't support search filters
## Description

The meta search URL for Europeana (audio and video) and SoundCloud (audio) do not update based on the commercial/modify filters.

## Reproduction

1. Search for a strange string.
2. See the URL for Europeana and SoundCloud.
3. Toggle the commercial/modify fiiters.
4. Observe no change to the URLs.

## Expectation

The code for the URLs to change exists (but it has a bug). So by fixing it, we should have the URLs updating based on the filters.

## Additional context

The code references `search.filters` which is always undefined as there is no `filters` field on the `search` object.

Europeana audio:
https://github.com/WordPress/openverse-frontend/blob/bccbd1333fd0deb51fdd90a391d65e4666caf554/src/utils/get-legacy-source-url.js#L44

Europeana video:
https://github.com/WordPress/openverse-frontend/blob/bccbd1333fd0deb51fdd90a391d65e4666caf554/src/utils/get-legacy-source-url.js#L61

SoundCloud audio:
https://github.com/WordPress/openverse-frontend/blob/bccbd1333fd0deb51fdd90a391d65e4666caf554/src/utils/get-legacy-source-url.js#L133

## Resolution

obulat:
## Fixes

Fixes #1037 by @obulat
Fixes #1121 by @sarayourfriend
Fixes #1128 by @dhruvkb 

## Description

This PR converts the search store from Vuex to Pinia and uses TypeScript.

For the search store state, we have 3 main pieces of data:
- search term (`q` parameter in the API request query)
- search type (is converted to path segment after `/search/`), and used throughout the codebase
- filters - these are lists of objects in each filter category (in the filter sidebar / filter modal) that have the code of the filter, the name (used to get the i18n string for it) and `checked` parameter. The set of filters that is displayed is dependent on the currently selected search type.

There is also a `query` object that holds the data necessary to set the URL for the API request. This was a separate object that was updated manually before. But in light of #1121, this PR converted it into a computed.
The search store uses the filters object as the single source of truth, and uses that to compute the `searchQueryParams` to be used for the API request.

There are currently 37 files that changed, but most of the changes in the components are renaming. The most significant changes are in the `stores/search.ts` and `VHeader.vue` and `search-query-transform.ts` files.

## Testing Instructions

Try out running the app and:
- searching for different words using the homepage and the searchbar in the header
- try checking and unchecking filters
- changing the search type using both the 'Content link' and the search type switcher at the top of the page
- clearing the filters using 'Clear filters' button.
- going to the detail page and back to the search page.
- going to the detail page and searching from it (it should go to the All content view with no filters applied)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (approved):
This looks really good. Most importantly I'm seeing everything behave correctly on the frontend, with no behavioral regressions.

There's some things I would have loved to be able to do, like look at the logic in some of the [search store methods](https://github.com/WordPress/openverse-frontend/blob/1966c7e7cdbcdaba2e52504f384172451f7afdcd/src/stores/search.js) more closely, but there were a lot of changes to view here.

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+   * drops all parameters with blank values.
+   *
+   * @returns {import('@nuxtjs/composition-api').ComputedRef<import('../store/types').ApiQueryParams>}
+   */
+  const searchQueryParams = computed(() => {
+    // Ensure that q filter always comes first
+    return Object.keys(state.query).reduce(
+      (obj, key) => {
+        return !['q'].includes(key) && state.query[key].length
+          ? { ...obj, [key]: state.query[key] }
```
sarayourfriend, line 64: I read a blog post recently that changed my perspective on how `reduce` is used. https://www.richsnapp.com/article/2019/06-09-reduce-spread-anti-pattern

I think the post is a _little_ dramatic (I wouldn't call this an "anti-pattern" personally) but the gist is that because the accumulator object is local to the reduce function, it's actually fine to mutate it directly instead of creating a new copy per iteration.

Mostly boils down to just using `Object.assign` if you want a psuedo-functional approach still or plain old property assignment to the accumulator.

```suggestion
          ? Object.assign(obj, { [key]: state.query[key] })
```

I wonder what you think about this? Is it something you've considered before? I personally still like the ergonomics of the spread operator here and find it more readable than `Object.assign`. I'm also skeptical that the performance impact of this is that significant in most use cases (hard to tell for this one though! We have a variable amount of filters to work through). My understanding is that JavaScript engines (or at least V8, I haven't read this about Spidermonkey or CoreJS but I'm assuming they work similarly) create dynamically defined classes under the hood to represent individual JavaScript `Object` instances and that they have to recreate the class definition whenever you add another property.

This makes me wonder what the performance difference between something like `Object.fromEntries(...)` with a `reduce` to generate the array of key/value pairs is vs `reduce`ing to an Object is. Presumably the `fromEntries` version gives the runtime more up-front information about what the resulting object is going to be, but does the engine actually use that information?

Anyway, just some food for thought, not any kind of concrete recommendation :slightly_smiling_face:

src/components/VHeader/VHeader.vue
```diff
@@ -149,12 +147,12 @@ const VHeader = defineComponent({
...
     const searchStatus = computed(() => {
-      if (!isSearchRoute.value || store.state.search.query.q === '') return ''
+      if (!isSearchRoute.value || searchStore.query.q === '') return ''
       if (isFetching.value) return i18n.t('header.loading')
       return getI18nCount(resultsCount.value)
     })
 
-    const localSearchTerm = ref(store.state.search.query.q)
+    const localSearchTerm = ref(searchStore.query.q)
     const searchTerm = computed({
```
sarayourfriend, line 156: We don't appear to need the distinction between `localSearchTerm` and `searchTerm` any more, do we? `searchTerm` just gets and sets `localSearchTerm`'s value anyway.

src/components/VHeader/VHeader.vue
```diff
@@ -163,7 +161,7 @@ const VHeader = defineComponent({
     })
 
     watch(
-      () => store.state.search.query.q,
+      () => searchStore.query.q,
       (newSearchTerm) => {
         if (newSearchTerm !== localSearchTerm.value) {
           localSearchTerm.value = newSearchTerm
```
sarayourfriend, line 167: What is this watcher meant to accomplish? How would the search store query change without the search term in the text box changing? :thinking:

src/pages/search.vue
```diff
@@ -56,11 +54,25 @@ const BrowsePage = {
...
     const { isVisible } = useFilterSidebarVisibility()
     const showScrollButton = inject('showScrollButton')
+    const searchStore = useSearchStore()
+
+    const query = computed(() => searchStore.query)
+    const searchType = computed(() => searchStore.searchType)
+    const searchQueryParams = computed(() => searchStore.searchQueryParams)
+    const supported = computed(() =>
+      supportedSearchTypes.includes(searchType.value)
+    )
```
sarayourfriend, line 64: This `supported` value appears to be calculated redundantly in a few places. Should we create a getter for it instead? `searchStore.isSelectedSearchTypeSupported`?

src/store/provider.js
```diff
@@ -37,8 +38,9 @@ export const state = () => ({
 })
 
 export const getters = {
-  getProviderName: (state, getters, rootState) => (providerCode) => {
-    const mediaType = rootState.search.mediaType
+  getProviderName: (state) => (providerCode) => {
+    const searchStore = useSearchStore()
+    const mediaType = searchStore.mediaType
```
sarayourfriend, line 43: `mediaType` doesn't appear to be a property of the `searchStore` :thinking: Should this be `searchType`?

src/stores/filter.js
```diff
@@ -17,11 +17,12 @@ import {
...
-  /** @type {{ filters: import('../store/types').Filters}} */
+  /** @type {import('../store/types').Filters} */
   const filters = reactive(clonedeep(filterData))
-  const searchType = /** @type {import('../store/types').SearchType} */ (
-    ref(ALL_MEDIA)
-  )
+  const searchType =
+    /** @type {import('@nuxtjs/composition-api').Ref<import('../store/types').SearchType>} */ (
+      ref(ALL_MEDIA)
+    )
```
sarayourfriend, line 25: It would be more concise to cast `ALL_MEDIA` than the whole ref (can avoid importing `Ref`). That would match more of what you'd do in actual TS too:

```suggestion
  const searchType = ref(/** @type {import('~/store/types').SearchType} */ (ALL_MEDIA))
```

tsconfig.json
```diff
@@ -43,6 +43,7 @@
     "src/composables/window.js",
     "src/constants/media.js",
     "src/constants/screens.js",
+    "src/constants/filters",
```
sarayourfriend, line 46: You could probably add the filters store here too with all the type updates you made to it!

src/stores/filter.js
```diff
@@ -37,12 +38,15 @@ export const useFilterStore = defineStore('filter', () => {
...
-    return mediaTypeFilters
+    return filterKeys.reduce((obj, filterKey) => {
+      return { ...obj, [filterKey]: filters[filterKey] }
+    }, /** @type {Partial<import('../store/types').Filters>} */ ({}))
   }
+
+  const allFilterCategories =
+    /** @type {import('../store/types').FilterCategory[]} */ (
+      Object.keys(filters)
+    )
```
sarayourfriend, line 49: It may be possible to avoid the cast by setting the type on allFilterCategories directly (though I'm not sure).

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+ * - parameter names are the names that the API expects.
+ * - some parameter names correspond to different filter names for different media types:
+ * `categories` can mean `imageCategories` if the current `searchType` is `IMAGE`, or `audioCategories`
+ * if it is `AUDIO`.
+ * - query parameter values are the values that are used in the API query.
+ */
+export const useSearchStore = defineStore('search', () => {
+  const filterStore = useFilterStore()
+  /**
+   * @type {import('../store/types').SearchState}}
```
sarayourfriend, line 24:
```suggestion
   * @type {import('../store/types').SearchState}
```

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+  })
+  const { searchType, query } = toRefs(state)
+
+  /**
+   * @param {import('../store/types').SupportedSearchType} type
+   */
+  const setSearchType = (type) => {
+    state.searchType = type
+    filterStore.setSearchType(type)
+  }
```
sarayourfriend, line 51: I wonder if instead of having two public functions for setting the search type, we shouldn't just treat the filterStore search type as the single source of truth and add a computed to be able to reference it in this store? It'd be nice to not expose it in both places.

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+
+  /**
+   * Returns the search query parameters for API request:
+   * drops all parameters with blank values.
+   *
+   * @returns {import('@nuxtjs/composition-api').ComputedRef<import('../store/types').ApiQueryParams>}
```
sarayourfriend, line 57:
```suggestion
   * @type {import('@nuxtjs/composition-api').ComputedRef<import('../store/types').ApiQueryParams>}
```

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+   */
+  const searchQueryParams = computed(() => {
+    // Ensure that q filter always comes first
```
sarayourfriend, line 60: Should we move this comment down to the line it's referencing (the initializer?)

```suggestion
```

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+    return Object.keys(state.query).reduce(
+      (obj, key) => {
+        return !['q'].includes(key) && state.query[key].length
+          ? { ...obj, [key]: state.query[key] }
+          : obj
+      },
+      { q: state.query.q.trim() }
```
sarayourfriend, line 67:
```suggestion
      // Ensure that q filter always comes first
      { q: state.query.q.trim() }
```

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+    filterStore.updateFiltersFromUrl(query, searchType)
+    updateQueryFromFilters()
+  }
+  /**
+   * After a change in filters, updates the query. If the filter is not selected, query should be set to blank string "".
+   * @param {object} [params]
+   * @param {import('../store/types').SupportedSearchType} [params.mediaType]
+   * @param {string} [params.q]
+   */
+  function updateQueryFromFilters(params = {}) {
```
sarayourfriend, line 109: It does not appear that an argument is ever passed to this function when it is called. Can we remove the params if that's the case?

```suggestion
   */
  function updateQueryFromFilters() {
```

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+    const queryFromFilters = filtersToQueryData(
+      filterStore.filters,
+      params.mediaType || state.searchType,
```
sarayourfriend, line 112: Change related to removing the unused parameter.

```suggestion
      state.searchType,
```

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+    // So we check if the key exists, not if the value is not falsy.
+    const changedKeys = /** @type {import('../store/types').QueryKey[]} */ (
+      Object.keys(queryFromFilters)
+    )
+    const queryKeys = /** @type {import('../store/types').QueryKey[]} */ (
+      Object.keys(state.query)
+    )
+    state.query = queryKeys.reduce((obj, key) => {
+      if (key === 'q') {
+        obj[key] = params?.q || state.query.q
```
sarayourfriend, line 126:
```suggestion
        obj[key] = state.query.q
```

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  /**
+   * Toggles a filter's checked parameter
+   * @param {{ filterType: import('../store/types').FilterCategory, codeIdx?: number, code?: string}} params
+   */
+  function toggleFilter(params) {
+    filterStore.toggleFilter(params)
+    updateQueryFromFilters()
+  }
```
sarayourfriend, line 141: Same deal here as the search type. Should we just have a watcher to update the search store whenever the filter store sees a change? It would eliminate the confusing duplicated API.

src/stores/search.js
```diff
@@ -0,0 +1,172 @@
...
+   * handled separately.
+   */
+  function clearFilters() {
+    filterStore.clearFilters()
+    updateQueryFromFilters()
+  }
+
+  /**
+   * Returns the object with filters for selected search type, with codes, names for i18n labels, and checked status.
+   * @returns {import('@nuxtjs/composition-api').ComputedRef<Partial<import('../store/types').Filters>>}
```
sarayourfriend, line 155:
```suggestion
   * @type {import('@nuxtjs/composition-api').ComputedRef<Partial<import('../store/types').Filters>>}
```

src/utils/search-query-transform.js
```diff
@@ -47,40 +47,28 @@ const filterToString = (filterItem) => {
...
 
-  return queryDataObject
+  return mediaFilterTypes.reduce((query, filterCategory) => {
+    const queryKey = filterPropertyMappings[filterCategory]
+    const queryValue = filterToString(filters[filterCategory])
+    if (queryValue || !hideEmpty) {
+      query[queryKey] = queryValue
+    }
+    return query
+  }, /** @type {import('../store/types').ApiQueryFilters} */ ({}))
```
sarayourfriend, line 71: What a nice improvement! :clap:

sarayourfriend (commented):
Thank you for adding so many nice new tests as well! This PR is looking great so far. A few suggestions for how we could further improve things. Some of them I think we should address in this PR (like removing the duplicated filter store functions from the search store) and others can be pushed off to separate issues as you see fit.

src/store/provider.js
obulat: Great catch!

src/components/VHeader/VHeader.vue
obulat: Yes, I updated the handling of the `searchTerm`

src/components/VHeader/VHeader.vue
obulat: Removed :)

src/pages/search.vue
obulat: I removed the `supported` computed from VSearchGrid with the prop that is passed anyways, and also extracted the `isValidSearchType` prop validator. So, now `supported` value is only calculated once (except for prop validation), and there is probably no need for a getter.

src/store/media.js
```diff
@@ -266,8 +267,9 @@ export const getters = {
       )
     }
   },
-  searchType(state, getters, rootState) {
-    return rootState.search.searchType
+  searchType() {
+    const searchStore = useSearchStore()
+    return searchStore.searchType
   },
```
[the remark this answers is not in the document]
obulat: Refactored to use a computed in the `filter` store.

src/stores/filter.js
```diff
@@ -16,15 +11,19 @@ import {
...
+import { useSearchStore } from '~/stores/search'
+
 export const useFilterStore = defineStore('filter', () => {
   /** @type {import('../store/types').Filters} */
   const filters = reactive(clonedeep(filterData))
-  const searchType = ref(/** @type {import('~/store/types').SearchType} */ (ALL_MEDIA))
 
   /**
-   * @param {import('../store/types').SearchType} type
+   * Search store state can only be accessed inside a function to prevent circular imports.
```
sarayourfriend, line 21: Is it possible to treat the filter store as the source of truth for the search type rather than the search store? Would that allow us to avoid the circular import problem?

sarayourfriend (commented):
This PR has me wondering if the current way we're organizing the flow of data around the search and filter stores makes that much sense. I'm curious what you think of the following (not necessarily to implement in this PR but something for us to consider if we decide we have time to do some refactoring in this area). I think it could simplify things.

I think because we were using Vuex before, it was tempting to do everything in sort of this imperative way, so we end up with patterns in our code like this:

```js
function updateSomeSearchValue() {
  updateFilterStoreValue()
  recalculateSearchStoreValues()
}
```

This pattern implies a dependency, right? Well, Vue has a way of declaring these dependencies, `computed` and `watch`.

What if the entire search store could be refactored into being just a couple of `computed` or `watch` callbacks based on the filter store? That way we truly have a single source of truth and completely avoid duplicating state. There's no risk of unnecessarily recalculating the search store values because the `computed` ref (or whatever underlying `ref`/`reactive` a `watch` would update if `watch` is necessary) is efficient and only updates when the dependencies change. 

The code I shared above wouldn't exist any more. Instead of calling `updateSomeSearchValue` the consumer of that function would just update the filter store directly. Then the search store would reactively update in a way that feels more in line with how Vue and especially composition-api wants us to think about the flow of data.

What do you think about that? Is it possible or is there a piece of the puzzle I'm missing that makes this impractical for our use case? The comment that you added at the top of the search store more or less inspired this line of thinking. It really implies that the search store is "just" a transformation of the filter store or _derived state_ which fits the use-case for `computed` or a `ref`-updating `watch` perfectly.

Again, not suggesting we implement this refactor right now, but if it's something you think is possible I'll cut a ticket for us to further explore this simplification.

sarayourfriend:
Also this tests well locally for me btw. I want to spend a little more time reviewing the code changes but anticipate approving this soon :+1:

src/stores/search.js
obulat: This is very interesting, and great to know! I have simplified some `reduce`s, and will be careful with them in the future.

src/stores/search.js
obulat: I tried using pinia $subscribe to replace the calls for update:
`filterStore.$subscribe(() => updateQueryFromFilters())` inside the searchStore, but it did not call the `update` function on the server, and did not update the query (and the path) ...
Do you think opening a separate issue and solving it in a different PR would be a good idea?

obulat:
> What do you think about that? Is it possible or is there a piece of the puzzle I'm missing that makes this impractical for our use case? The comment that you added at the top of the search store more or less inspired this line of thinking. It really implies that the search store is "just" a transformation of the filter store or _derived state_ which fits the use-case for `computed` or a `ref`-updating `watch` perfectly.

This makes a lot of sense! Well, at least in theory it should be possible and a much more elegant solution. I do think it should be a separate ticket for refactor, but a really nice one :)

sarayourfriend:
I'll open the issue then :+1:

src/stores/search.ts
```diff
@@ -0,0 +1,341 @@
...
+   * audio/image provider filters are handled correctly.
+   */
+  function replaceFilters(newFilterData: Filters) {
+    const providerFilters: FilterCategory[] = [
+      'audioProviders',
+      'imageProviders',
+    ]
+
+    allFilterCategories.forEach((filterCategory) => {
+      if (providerFilters.includes(filterCategory)) {
```
sarayourfriend, line 164: When would this be false :thinking:

src/stores/search.ts
```diff
@@ -0,0 +1,341 @@
...
+        name: provider.display_name,
+        checked,
+      }
+    })
+  }
+  /**
+   * Toggles a filter's checked parameter. Requires either codeIdx or code.
+   */
+  function toggleFilter(params: {
+    filterType: import('../store/types').FilterCategory
```
sarayourfriend, line 214: This is already imported anyway.

```suggestion
    filterType: FilterCategory
```

src/stores/search.ts
```diff
@@ -0,0 +1,341 @@
...
+    codeIdx?: number
+    code?: string
+  }) {
+    if (
+      typeof params.codeIdx === 'undefined' &&
+      typeof params.code === 'undefined'
+    ) {
+      warn(
```
sarayourfriend, line 222: Should this be a full on Error rather than a warning?

src/stores/search.ts
```diff
@@ -0,0 +1,341 @@
...
+      if (item.code.includes('nc')) {
+        dependentFilters.push('commercial')
+      }
+      if (item.code.includes('nd')) {
+        dependentFilters.push('modification')
+      }
+      return (
+        state.filters.licenseTypes.filter(
+          (item) => dependentFilters.includes(item.code) && item.checked
+        ).length > 0
```
sarayourfriend, line 275: nit: filtering for existence could be `some` and save iterating the full list (if this is simpler, YMMV).

```suggestion
        state.filters.licenseTypes.some(
          (item) => dependentFilters.includes(item.code) && item.checked
        )
```

src/stores/search.ts
```diff
@@ -0,0 +1,341 @@
...
+    }
+    state.searchType = queryStringToSearchType(path)
+    // When setting filters from URL query, 'mature' has a value of 'true',
+    // but we need the 'mature' code. Creating a local shallow copy to prevent mutation.
+    const query = { ...urlQuery }
+    if (query.mature === 'true') {
+      query.mature = 'mature'
+    } else {
+      delete query.mature
+    }
```
sarayourfriend, line 313: It seems like in the API if mature is set to anything at all that is truthy it gets evaluated as true, not just if it's explicitly `"true"`.

https://github.com/WordPress/openverse-api/blob/f2335988b1ef5d030f3e3bca9cadbed4a012ba9b/api/catalog/api/controllers/search_controller.py#L202-L205

src/utils/search-query-transform.ts
```diff
@@ -1,11 +1,21 @@
+// @ts-ignore
```
sarayourfriend, line 1: Idk why this file doesn't have a Typescriptification issue but I feel like you should get credit for converting it still! :tada:

test/unit/specs/components/v-search-grid-filter.spec.js
```diff
@@ -73,7 +60,7 @@ describe('VSearchGridFilter', () => {
         },
       },
     }
-    filterStore = useFilterStore(pinia)
+    filterStore = useSearchStore(pinia)
```
sarayourfriend, line 63: Should we rename this to `searchStore` just for clarity's sake? :grinning:

test/unit/specs/stores/filter-store.spec.js
```diff
@@ -0,0 +1,464 @@
+import { setActivePinia, createPinia } from 'pinia'
```
sarayourfriend, line 1: Can we rename this file to `search-store.spec.js` to match the module it tests? I guess it really would just be `stores/search.spec.js`, either way though.

sarayourfriend (commented):
Looking great! Gonna test it locally now after reviewing the code but this is super exciting! Great work untangling those two stores' logic!

sarayourfriend:
It looks like this PR fixes #1047 now :tada:

sarayourfriend:
I'm noticing one very subtle difference between this branch and `main`. I think it would be a regression to change this but I'm not sure:

Do the following in staging.
1. Go to https://search-staging.openverse.engineering/search/image?q=galah&extension=jpg,png
2. Click on a result
3. Focus the search bar on the single result page and send the search -- note that it sends you specifically back to the image search with the image specific filters still applied 
4. Click on another result
5. Focus the search bar and change the search to `birds` -- note that it sends to you still to the image search, also with the image specific filters still applied

Do it again on this branch.

In the branch the resubmit (step 3) works the same way, but step 5, where you submit a new search term, sends you to the all search page and clears the image specific filters.

Also, we need to debounce updating the search term on the search page. Currently it submits a new query per keystroke... to very strange effect:

https://user-images.githubusercontent.com/24264157/158789479-ae5d3ae7-a85e-4b67-ad62-5ae8ee11ceee.mp4

There's another regression as well.

1. Go to `/search/?q=galah&license=by,by-sa`
2. Clear the search term by putting the cursor at the end of the term and holding backspace until it clears

On `main` this retains the existing search results. On this branch it changes to the search results for no query or (depending on the race condition described in #1132) one of the other sets of results from executing either `g` `ga` `gal` or `gala`... and so on, as you backspace through.

I think debouncing the search would fix this.

Alternative to debouncing could be to just wait until the user actually hits enter and submits the query in the search box when changing the term.

obulat:
> Do the following in staging.

For the first issue, I specifically reset the search type to "All content" and the filters when you search from a non-search page in this PR. But when doing so, I was thinking about the content pages (e.g.  'About us').
Do you think that the behavior for single media page should be different? It seems more logical to go to the same search type at least, even if the filter types are reset (should they? I don't know)

obulat:
As for the second problem, let me just implement the "Wait until the user hits the `Search` button to send the search request" solution in this PR. We can convert it to debouncing later.

src/stores/search.ts
obulat: Nice! Thank you for the link.

src/stores/search.ts
```diff
@@ -0,0 +1,341 @@
...
+    )
+  })
+
+  // Actions
+  /**
+   * After a search type is changed, unchecks all the filters that are not
+   * applicable for this Media type.
+   */
+  function clearOtherMediaTypeFilters(searchType: SupportedSearchType) {
+    const mediaTypesToClear = supportedSearchTypes.filter(
```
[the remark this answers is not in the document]
obulat: Yes, this is meant to include `ALL_MEDIA`, but I'm not yet sure if it works for non-supported `search` types such as `video` or maybe `3d_model`, so I just left it as `supportedSearchTypes` for now. We should audit how `additional` search/media types works, and fix any issues that we currently have. Let me create an issue for that.

src/store/provider.js
```diff
@@ -37,8 +37,10 @@ export const state = () => ({
 })
 
 export const getters = {
-  getProviderName: (state, getters, rootState) => (providerCode) => {
-    const mediaType = rootState.search.mediaType
+  getProviderName: (state) => (providerCode) => {
+    const searchStore = useSearchStore()
+    const mediaType =
+      searchStore.searchType === ALL_MEDIA ? IMAGE : searchStore.searchType
```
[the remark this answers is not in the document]
obulat: Agree. I think this might get fixed when we convert the provider store to Pinia. It actually needs to be refactored just like the other media files instead of hard-coding `audio` and `image`.

src/pages/index.vue
```diff
@@ -197,32 +200,27 @@ const HomePage = {
...
-      await store.dispatch(`${SEARCH}/${UPDATE_QUERY}`, {
-        q: searchTerm.value,
-        searchType: searchType.value,
-      })
+      searchStore.setSearchTerm(searchTerm.value)
+      searchStore.setSearchType(searchType.value)
+      const query = searchStore.searchQueryParams
       const newPath = app.localePath({
         path: `/search/${
           searchType.value === ALL_MEDIA ? '' : searchType.value
```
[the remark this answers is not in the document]
obulat: Agree, thank you for making the issue! It's been bothering me but not as much yet to open the issue myself :)

sarayourfriend:
>  Do you think that the behavior for single media page should be different? It seems more logical to go to the same search type at least, even if the filter types are reset (should they? I don't know)

For me, I agree, I think it makes more sense to keep the search type the same when the user has navigated to the result from a search page.

However, I suppose if someone lands on a result page directly, it maybe would make more sense for the header to search all content :thinking:

src/stores/search.ts
obulat: This was originally written to prevent the provider codes that do not exist in the filters for current search type from breaking this function. For example, going to `/search/audio?source=imagesource` would cause a problem because `imagesource` existed in the `imageProviders` but did not exist in `audioProviders`.
I've improved the `queryToFilterData` to prevent that, and added more documentation there in the last commit.

sarayourfriend:
Everything is looking great in this PR.

There's still the redirect to all content search from a result page if the person navigated to the result from a specific media type search, but I suppose we can open a separate issue to decide the specific behavior we want to see there, for the sake of being able to cut down the scope of this PR and merge it before more conflicts arise.

Great work @obulat! This was a big undertaking.

src/stores/search.ts
```diff
@@ -0,0 +1,339 @@
+import { defineStore } from 'pinia'
+import {
+  computed,
+  ComputedRef,
+  reactive,
+  toRefs,
+  watch,
+} from '@nuxtjs/composition-api'
+// @ts-ignore
```
sarayourfriend, line 9: nit: Can you silence this ESLint warning and add a Disable reason to note that we're removing this dependency soon and shouldn't worry about the missing types for now? Otherwise the warning will show on every subsequent PR. I opened an issue for us to discuss the warnings #1160

src/utils/search-query-transform.ts
sarayourfriend, line 1: Same here re: disabling the eslint warning.

test/e2e/filters.spec.js
```diff
@@ -83,7 +86,7 @@ test('common filters are retained when media type changes from all media to sing
   }
 })
 
-test('common filters are retained when media type changes from single type to all media', async ({
+test.skip('common filters are retained when media type changes from single type to all media', async ({
```
sarayourfriend, line 89: Why is it necessary to skip these?

test/unit/specs/store/media-store.spec.js
```diff
@@ -53,6 +57,7 @@ describe('Search Store', () => {
 
     beforeEach(() => {
       state = store.state()
+      setActivePinia(createPinia())
```
sarayourfriend, line 60: I'm unclear of when it's necessary to use `setActivePinia`. Could you add a blurb in the TESTING_GUIDELINES explaining that (or else open an issue to do it in a separate PR)?

sarayourfriend (approved):
LGTM! Just some requests minor non-blocking stuff.

Great work!

PR #1045, opened 2022-03-03
Use a non-versioned API URL to mock analytics requests, too

### LINKED ISSUE #915: Mock e2e testing analytics network requests
## Problem

The requests to Openverse API made in the e2e tests during the CI action will be proxied to use the pre-recorded responses in `test/tapes` folder. This, however, excludes the requests to the analytics server, because the proxy is set to use the `/v1/` API url, and analytics is using the non-versioned URL.
We need to have the same base URL to proxy all the requests through a single talkback proxy.

## Description

2. Set the API URL to `api.openverse.engineering/` (without the `v1`), in the front-end environment, and add the `v1` in the front-end data services that need it. So, instead of the UsageDataService removing `v1` from the baseURL, other services would be adding it:
https://github.com/WordPress/openverse-frontend/blob/99a803542c46b20b7cac36bc7d421ffcaa8398af/src/data/usage-data-service.js#L7

Using this API service constructor:
https://github.com/WordPress/openverse-frontend/blob/99a803542c46b20b7cac36bc7d421ffcaa8398af/src/data/api-service.js#L121
To add the version to the URL in the media services:
`const imageService = createApiService(baseUrl = `${process.env.apiUrl}/v1/)`

## Alternatives

We could also move the analytics API to `api.openverse.engineering/v1` to match the other endpoints, or use 2 talkback proxies set to 2 base URLs.

## Implementation

obulat:
## Fixes

Fixes #915 by @obulat

## Description

This PR creates two ApiService objects: one is versioned, for most endpoints, and an unversioned one for the analytics endpoints. 
Some data services were also refactored to reuse code from other services (`getMediaSlug`, for example), and the `bug-report` was removed as it is unused.
The proxy port was also changed from `3000` to `3322` because when I was testing, the API was running in Docker, and `3000` port was busy.
It also renames the `update-tapes` to `recreate-tapes`, and adds an `update-tapes` script that does not completely delete all the tapes, but simply adds any tapes that are missing. This way, we can update a single tape when any e2e test changes, and don't have too many changes in a PR.

## Testing Instructions

Enable internal analytics:
https://github.com/WordPress/openverse-frontend/blob/690f52360a6d7389f4015ed384b62d36b29b65e6/src/utils/env.ts#L9
Run `pnpm run update-tapes`. This will delete all the tapes files, and, more importantly, it will save any response that isn't found in `test/tapes` directory. You should see a lot of tapes for analytics responses. They are all 404, but they **_are_** there.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/data/api-service.js
```diff
@@ -35,6 +35,7 @@ Please check the url: ${config.baseURL}${config.url}`
  * @param {string} [baseUrl]
  */
 export const createApiService = (baseUrl = process.env.apiUrl) => {
+  console.log('will create api service with ', baseUrl)
```
sarayourfriend, line 38: Is this left over from debugging?

test/proxy.js
```diff
@@ -12,18 +12,18 @@ const process = require('process')
...
-const host = 'https://api.openverse.engineering/v1'
+const host = 'https://api.openverse.engineering'
 
 const tapeNameGenerator = (tapeNumber) => `response-${tapeNumber}`
 
 const updatingTapes = process.argv.includes('--update-tapes')
 const recordMode = updatingTapes
   ? talkback.Options.RecordMode.NEW
-  : talkback.Options.RecordMode.DISABLED
+  : talkback.Options.RecordMode.NEW
```
sarayourfriend, line 22: Doesn't this make `recordMode` always `NEW`? :thinking:

test/proxy.js
```diff
@@ -12,18 +12,18 @@ const process = require('process')
...
 console.log('Record mode: ', recordMode)
 const opts = {
   host,
-  port: 3000,
+  port: 3322,
```
sarayourfriend, line 26: :sweat_smile: This seems like a better choice considering how many other things live on port 3000 by default anyway (I know a few that are missing from this list too, like Flask I think?)

https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

In fact, I kind of wonder if we shouldn't switch all of our stuff to use ports in the completely unregistered "temporary services" range of 49152–65535.

sarayourfriend (commented):
Is there any way to run talkback with the normal `pnpm test:e2e` script? Or is it only possible with the CI version?

Looking good so far, just testing locally but will probably be able to approve soon.

sarayourfriend (approved):
This LGTM. I noticed that in addition to #985 talkback also didn't stop running when I ran `pnpm ci:e2e-test` locally even though the tests passed and didn't crash or anything :confused: Did you notice that happening as well?

sarayourfriend:
Oh I just noticed another tape was created when I ran `pnpm ci:e2e-test` locally.

src/data/api-service.js
obulat: 🙊

test/proxy.js
obulat: 49152 it is, then :)

obulat:
> Oh I just noticed another tape was created when I ran pnpm ci:e2e-test locally.

I just spent 15 minutes chasing these new tapes. There is often 1 image or audio request that's new. I think it's one of the related images  or audio that sometimes change. I guess we should open an issue to mock related image / audio requests...

sarayourfriend:
> I just spent 15 minutes chasing these new tapes. There is often 1 image or audio request that's new. I think it's one of the related images or audio that sometimes change. I guess we should open an issue to mock related image / audio requests...

I think the fix I have in the proxy to replace API results that reference the upstream server with proxy references will address this, so I wouldn't sweat it.

dhruvkb (approved):
Works well, LGTM 👍

PR #1046, opened 2022-03-03
handling null and undefined value for userAgent

### LINKED ISSUE #903: Non-string UserAgent can crash the app
## Description

https://sentry.io/share/issue/e01331b2eb71498080208da9d62de7f1/

If either `context.req.headers['user-agent']` or `navigator.userAgent` return a non-string value, the app crashes, because `useragent.parse` in `'express-useragent' expects a string argument and calls `replace` on it.
This is the error you would see: 
```
TypeError
Cannot read properties of undefined (reading 'replace')
```
Here is the code that causes the error:
https://github.com/WordPress/openverse-frontend/blob/5a3bed0dbab754e88749accc5a59477f7337583b/src/plugins/ua-parse.js#L3-L12

## Reproduction

1. Add `userAgent = null` or `userAgent = undefined` to line 11 in the code above.
2. Run the app and try to open http://localhost:8443
3. See error.

## Expectation

The app should not crash even if the user agent is undefined.
 
## Additional context

I wonder why the user agent can be non-string?

## Resolution

kb-0311:
## Fixes

Fixes #903  by @obulat 

## Description

explicitly declared the userAgent type as string 

Converted the `context.req.headers['user-agent']` and `navigator.userAgent` to string type. 

This will make sure the `user.parse(userAgent)`  always gets a string value as an argument.
`
## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/composables/use-browser-detection.js
+++ b/src/composables/use-browser-detection.js
@@ -7,18 +7,16 @@ export const useBrowserDetection = () => {
 
 export const useBrowserIsBlink = () => {
   const browser = useBrowserDetection()
-  if (browser!==null) {
+  if (browser !== null) {
     return browser.isChrome || browser.isEdge || browser.isOpera
   } else {
     return false
-  }
-  
-  
+  } 
 }
 
 export const useBrowserIsMobile = () => {
   const mobile = useBrowserDetection().isMobile
-  if (mobile!==null) {
+  if (mobile !== 'null') {
     return mobile
   } else {
     return false
--- a/src/composables/use-browser-detection.js
+++ b/src/composables/use-browser-detection.js
@@ -15,11 +15,10 @@ export const useBrowserIsBlink = () => {
 }
 
 export const useBrowserIsMobile = () => {
-  const mobile = useBrowserDetection().isMobile
+  const mobile = useBrowserDetection()
   if (mobile === null) {
     return false
-  }
-  if (mobile) {
-    return mobile
+  } else {
+    return mobile.isMobile
   }
 }
--- a/src/composables/use-browser-detection.js
+++ b/src/composables/use-browser-detection.js
@@ -11,14 +11,14 @@ export const useBrowserIsBlink = () => {
     return browser.isChrome || browser.isEdge || browser.isOpera
   } else {
     return false
-  } 
+  }
 }
 
 export const useBrowserIsMobile = () => {
   const mobile = useBrowserDetection().isMobile
-  if (mobile !== 'null') {
-    return mobile
-  } else {
+  if (mobile === null) {
     return false
+  } else {
+    return mobile
   }
 }
--- a/src/composables/use-browser-detection.js
+++ b/src/composables/use-browser-detection.js
@@ -7,9 +7,20 @@ export const useBrowserDetection = () => {
 
 export const useBrowserIsBlink = () => {
   const browser = useBrowserDetection()
-  return browser.isChrome || browser.isEdge || browser.isOpera
+  if (browser!==null) {
+    return browser.isChrome || browser.isEdge || browser.isOpera
+  } else {
+    return null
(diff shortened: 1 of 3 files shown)


kb-0311:
I cannot really add labels to a PR can someone please help me out with the labels ?

zackkrida:
Hm, I'm not sure we want to use a solution like this, for this reason:

```
String(null) === 'null'
String(undefined) === 'undefined'
```

kb-0311:
@zackkrida 
What exactly do you want to parse as userAgent in case it is null or undefined?

sarayourfriend:
I guess the more correct thing to do would be to set `$ua` to `null`  and force downstream users of that context variable to decide how to handle when it's not present?

kb-0311:
@sarayourfriend 
Hey so I tried to do what I thought I understood from your comment please suggest the changes to be made

obulat (approved):

src/plugins/ua-parse.js
```diff
@@ -8,7 +8,12 @@ export default function (context, inject) {
   } else if (typeof navigator !== 'undefined') {
     userAgent = navigator.userAgent
   }
-  const ua = useragent.parse(userAgent)
+  var ua
```
sarayourfriend, line 11: I'm not sure why our linter isn't complaining about this line, we have the `no-var` rule enabled :confused: 

```suggestion
  let ua
```

sarayourfriend (changes requested):
This actually becomes a bigger change once we do this, we'll need to update the places that consume this context to check for `null`:

https://github.com/WordPress/openverse-frontend/blob/6c938c1c6bb7fca75b937749f8e14acddd69eb2c/src/composables/use-browser-detection.js#L3-L15

I think the `useBrowserIsBlink` and `useBrowserIsMobile` functions can just return `false` if the return type from `useBrowserDetection` is `null`.

We'll also need to update the types here to include `null`, `$ua: UADetails | null`:

https://github.com/WordPress/openverse-frontend/blob/6c938c1c6bb7fca75b937749f8e14acddd69eb2c/typings/nuxt__types/index.d.ts#L7

src/plugins/ua-parse.js
kb-0311: And I will make the necessary changes to `use-browser-detection.js` and index.d.ts too. Thanks for the heads up.

kb-0311:
@sarayourfriend Hey so, I am not sure why I am getting the null type error in my recent commit. Are there changes to be made in some other file that accepts a string value of `useBrowserIsMobile()` or some other way null handling should have been done ?

sarayourfriend (approved):
This LGTM! Please do change the `mobile` variable to `browser` if you decide to keep it instead of the optional chaining approach, but otherwise this is great! Thanks so much for the contribution :tada:

src/composables/use-browser-detection.js
```diff
@@ -7,9 +7,18 @@ export const useBrowserDetection = () => {
...
+  if (browser !== null) {
+    return browser.isChrome || browser.isEdge || browser.isOpera
+  } else {
+    return false
+  }
 }
 
 export const useBrowserIsMobile = () => {
-  return useBrowserDetection().isMobile
+  const mobile = useBrowserDetection()
```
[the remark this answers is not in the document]
kb-0311: thanks for the suggestion however I would say more beginners like myself would find the if-else statements easier to understand rather than ternary operators. if there are some optimization issues I will do it.

Yes for sure the `mobile` variable was just a placeholder for the logic. I will change it to `browser` .

obulat:
Thank you for your contribution, @kb-0311 !

PR #1053, opened 2022-03-04
Convert `usage-data` store to an options Pinia store

### LINKED ISSUE #1026: Convert `usage-data` store from Vuex to Pinia
## Problem

As part of the conversion of the Vuex store modules to Pinia, we need to convert the [usage-data store](https://github.com/WordPress/openverse-frontend/blob/e0ac241dfe8b2b4b7eaff8eaee7e06eda0dce197/src/store/user.js) to Pinia. This store uses the `usageDataService` which needs to be passed as a parameter to the function that creates a store definition function, something like:
```js
import UsageDataService from '...'
const createUseUsageDataStore = (usageDataService=UsageDataService) => defineStore('usage-data', () => {})
export const useUsageDataStore = createUseUsageDataStore(UsageDataService)
```

## Description

-  convert state to reactive

## Additional context

## Implementation

obulat:
## Fixes

Fixes #1026 by @obulat

## Description

This PR converts `usage-data` store to Pinia. This store handles sending of the analytics data points to the API, and doesn't contain any state, so it might be a good idea to refactor it into a composable in the future.
For now, I converted it from Vuex to Pinia so that we can keep converting other stores that depend on it. 

This store needs to access `$sentryReady` from the Nuxt context. Because the Pinia options store has  the context available on `this.$nuxt`, I used an `options` store (with actions, similar to Vuex), and not the `setup` store (similar to composable functions).

## Testing Instructions

Enable Sentry and run the app: `ENABLE_INTERNAL_ANALYTICS=true pnpm run dev`

Revert the fix for analytics endpoint API bug that results in double slashes at the end:
https://github.com/WordPress/openverse-frontend/blob/7d886d34eb84a76cf8e8c9c58e0b319b6dc0b348/src/data/usage-data-service.js#L17
This will cause analytics requests to throw a network error. When searching, opening the detail page, clicking on the 'Copy attribution', you should see the sentry post requests in the console. You will have to disable ad blocking extensions if you have any.
The unit tests have almost 100% coverage (the internal of Sentry are not covered).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/ImageDetails/ReuseSurvey.vue
```diff
@@ -29,6 +27,18 @@ export default {
...
   components: { VLink },
   props: ['image'],
+  setup(props) {
+    const usageDataStore = useUsageDataStore()
+    const onReuseSurveyClick = async () => {
+      await usageDataStore.sendDetailPageEvent({
+        eventType: DETAIL_PAGE_EVENTS.REUSE_SURVEY,
+        resultUuid: props.image.id,
+      })
+    }
```
sarayourfriend, line 37: Analytics requests shouldn't block; they don't have any bearing on the UI and blocking on them will make the UI less responsive.

```suggestion
    const onReuseSurveyClick = () => {
      usageDataStore.sendDetailPageEvent({
        eventType: DETAIL_PAGE_EVENTS.REUSE_SURVEY,
        resultUuid: props.image.id,
      })
    }
```

src/stores/usage-data.js
```diff
@@ -0,0 +1,69 @@
...
+          scope.setTag('event_name', eventName)
+          scope.setTag('request_url', error.config.url)
+        })
+      )
+  )
+
+export const createUseUsageDataStore = (usageDataService = UsageDataService) =>
+  defineStore('usage-data', {
+    actions: {
+      async sendSearchQueryEvent(params) {
```
sarayourfriend, line 39: We can unable these as `async`, they don't actually return `Promise`s and there's nothing to await in or outside of them.

src/components/ImageDetails/ReuseSurvey.vue
zackkrida: We should add a comment about this so folks don't forget in the future.

obulat:
Closing this PR as it's replaced with #1084.

PR #1064, opened 2022-03-06
Replace `usage-data` store module with utils file

### LINKED ISSUE #1026: Convert `usage-data` store from Vuex to Pinia

krysal:
## Fixes

Fixes #1026 by @obulat (Alternative solution)

## Description

In this PR I propose to move actions of the `usage-data` out of Vuex and therefore delete this module, it does not have a state per se so we can save maintaining another module (in Vuex or Pinia). I created a utils file instead of a composable because it doesn't involve any state logic about the view.

This PR also removes the unused component `ReuseSurvey` and triggers the `sendDetailPageEvent` when clicking the source link on the **Image information** section.

## Testing Instructions

1. Check out this branch and enable Sentry temporarily in the [sentry-config.ts](https://github.com/WordPress/openverse-frontend/blob/usage-data-utils/src/utils/sentry-config.ts) file:
```diff
export const sentryConfig: ModuleConfiguration = {
  dsn:
    process.env.SENTRY_DSN || ...
-  disabled: !isProd,
+  disabled: false, 
  lazy: true,
}
```
2. Run the app with `ENABLE_INTERNAL_ANALYTICS=true pnpm dev`
3. Test the places from where events are sent:
- search results page: when searching for a term and/or clicking on a search result)
- image details page: click on author name or source, click on copy license info)
4. Observe if events are being sent

## Note
When we turn on analytics we should send these events from the audio page as well.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VImageDetails/VImageDetails.vue
```diff
@@ -26,9 +26,12 @@
       <div>
         <dt>{{ $t('image-details.information.source') }}</dt>
         <dd>
-          <VLink :href="image.foreign_landing_url" class="text-pink">{{
-            sourceName
-          }}</VLink>
+          <VLink
+            :href="image.foreign_landing_url"
+            @click="onSourceLinkClicked"
+            @keyup.enter="onSourceLinkClicked"
```
sarayourfriend, line 32: Would this not send the event twice?

https://codesandbox.io/s/upbeat-wind-qllc8c?file=/src/App.vue

`@click` on a link should trigger both with <kbd>Enter</kbd> and a click.

src/data/usage-data-service.js
```diff
@@ -14,7 +14,7 @@ const UsageDataService = {
    * @param {Parameters<typeof ApiService['post']>[1]} data
    */
   post(endpoint, data) {
-    return ApiService.post(`analytics/${endpoint}/`, data)
+    return ApiService.post(`analytics/${endpoint}`, data)
```
sarayourfriend, line 17: :sweat_smile: At some point I think I added this trailing slash to silence a console warning about it missing. Does that console warning not happen any more?

src/store/media.js
```diff
@@ -166,16 +158,13 @@ export const createActions = (services = mediaServices) => ({
...
-      },
-      { root: true }
-    )
+    const eventData = {
+      query: rootState.search.query.q,
+      resultUuid: id,
+      resultRank,
+      sessionId: rootState.user.usageSessionId,
+    }
+    usageData.sendResultClickedEvent(eventData, this)
```
sarayourfriend, line 167: I _really_ dislike how Nuxt plugins lock things like Sentry inside the Nuxt context instead of just being available via an import. It doesn't make any sense for us to have to pass around the context to utility methods :frowning_face:

sarayourfriend (commented):
Excited to review this! I'm in favor of this solution for the usage data stuff considering it doesn't have any relationship to the application state.

src/components/VImageDetails/VImageDetails.vue
krysal: That seems to be the case! I had the same question because I remember seeing something about that once but I don't know exactly where and couldn't find it. Also, we are repeating these events in other places so I will leave only the clicks in those. Thank you!

src/data/usage-data-service.js
krysal: The warning now was because it was duplicated.

src/store/media.js
krysal: I agree, it feels somehow wrong but still is how have been working til now.

src/pages/image/_id.vue
```diff
@@ -31,12 +31,9 @@
       <VButton
         as="VLink"
         :href="image.foreign_landing_url"
```
obulat, line 33: Nice cleanup! VLink default properties, and the redundant`keyup` handling for `click` events.

test/unit/specs/store/media-store.spec.js
```diff
@@ -269,22 +270,15 @@ describe('Search Store', () => {
...
-          query: params.q,
-          sessionId: context.rootState.user.usageSessionId,
-        },
-        { root: true }
-      )
+      expect(usageData.sendSearchQueryEvent).toHaveBeenCalled()
     })
 
-    it('does not dispatch SEND_SEARCH_QUERY_EVENT if page param is available', async () => {
+    it.skip('does not dispatch SEND_SEARCH_QUERY_EVENT if page param is available', async () => {
```
obulat, line 281: I think it's OK to keep this skipped, the handling of page param should be refactored anyway.

obulat (commented):
Nice, I like the move away from the store for usage-data.
I will approve when the tests are ready. Can you use the testing approach from my usage-data PR?

krysal:
@obulat I'll check that out! If you want to specify more what parts that would also be of help.

test/unit/specs/store/media-store.spec.js
krysal: The negation for the expected value was not working here but I found the reason, just have to reset the mock after each test, ha!

krysal:
This PR is ready for review folks.

obulat:
... And we can probably close it because of #1084, @krysal ?

krysal:
Yes, closing this as we decided to drop the analytics code until figuring out a better solution for that module.

PR #1071, opened 2022-03-08
Convert the `user` store to an `ssrRef`

### LINKED ISSUE #1025: Convert `user` store from Vuex to Pinia
## Problem

As part of the conversion of the Vuex store modules to Pinia, we need to convert the [user store](https://github.com/WordPress/openverse-frontend/blob/e0ac241dfe8b2b4b7eaff8eaee7e06eda0dce197/src/store/user.js) to Pinia.

**_The tricky part_** in the user store will be repeating the user store uses a workaround:  to make sure that the state that is generated on the server is used when hydrating the client-side state 
https://github.com/WordPress/openverse-frontend/blob/e8cab9fa310002f175f50705551e6c020a3a91d0/src/store/user.js#L17

## Description

-  convert state to reactive

## Additional context

## Implementation

obulat:
## Fixes

Fixes #1025 by @obulat

## Description

User store only contained one piece of data: the uuid that identified the user request. This ID should not be shared between requests, but should keep the SSR-set value on the client side, too.

I've replaced the `user` store with a composable that creates a `ssrRef` with a uuid on the server. `ssrRef` is set once, on the server, keeps the same value on the client side, and is not shared between requests. However, it can only be set from the `setup` function (setting it in global scope leads to sharing of the id between requests).

To ensure that the id is created in a component setup function, I've added conditional creation in two outer-most, user-entry components: `default` and `blank` layouts. And to make sure that it's not created twice, the id is saved to the media store after it's created. So, the id is created on SSR by one of the layout components if it is not available in the store. The user can enter the site from the index page, the search page, or other pages, and would still get an id. A limitation is that this id would be replaced if the user hard-refreshes the page.

Note: Nuxt 3 does not have `ssrRef`s, they are replaced with `useState`: https://v3.nuxtjs.org/getting-started/bridge-composition-api/#ssrref-and-shallowssrref

## Testing Instructions

Run the app, and open the Vue devtools media store. Go to localhost:8443 or localhost:8443/search/q=cat, and navigate after that. You should see that the `usageSessionId` is the same. If you click on refresh, you should see the id changing.
You can also access the app from a local network (e.g. from a phone). If you add logging, you will see that there are 2 `usageSessionId` values, one for each request.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/composables/use-usage-session-id.js
```diff
@@ -0,0 +1,22 @@
...
+ * in one of the layout components, and is saved to the store to be used when outside
+ * the component context.
+ * TODO: refactor to use Pinia store during the media store refactor.
+ *
+ * @param store
+ * @returns {{ usageSessionId: string }}
+ */
+export function useUsageSessionId(store) {
+  let usageSessionId = ssrRef('')
+  if (!store.state.media.usageSessionId) {
```
sarayourfriend, line 15: :grimacing: Would it be better to just create a separate store altogether for it? Throwing unrelated data into the media store doesn't feel right to me.

src/composables/use-usage-session-id.js
```diff
@@ -0,0 +1,22 @@
...
+ * Sets the usageSessionId as a per-request ssrRef. The value is created on the server
+ * in one of the layout components, and is saved to the store to be used when outside
+ * the component context.
+ * TODO: refactor to use Pinia store during the media store refactor.
+ *
+ * @param store
+ * @returns {{ usageSessionId: string }}
+ */
+export function useUsageSessionId(store) {
+  let usageSessionId = ssrRef('')
```
sarayourfriend, line 14: I think this also needs a key for us to be able to use multiple `ssrRef`s, doesn't it? Something like 

```suggestion
  const usageSessionId = ssrRef('', 'usage-session-id')
```

The documentation for this function is so unclear it's really frustrating.

Also we can update it to a const instead of let, right?

sarayourfriend (commented):
So did it end up not being possible to just use a pinia store for this? Having to use `ssrRef` _and_ a supplemental store value feels like a lot of work just to transport a value from the server to the client... and also completely goes contrary to my understanding of how pinia SSR is supposed to work.

src/composables/use-usage-session-id.js
obulat: `usageSessionId` is only used in the media store, when sending the `search_query` and `result_clicked` events, so I felt that it was more appropriate to put it to the media store instead of creating a whole store simply for one value. That said, I wonder why `usageSessionId` is not used for other `usage-data` events.

src/composables/use-usage-session-id.js
obulat: You need a key for multiple `ssrRef`'s , but I don't think we are going to use any more `ssrRef`'s anywhere? And when we go to Nuxt bridge/3, we will replace `ssrRef` with `useState` (keyed).

I agree about documentation, I added the key at first, but could not find how to use the keyed ssrRef, so I removed the key in the end. And yes, it should be `const`.

obulat:
> So did it end up not being possible to just use a pinia store for this? Having to use `ssrRef` _and_ a supplemental store value feels like a lot of work just to transport a value from the server to the client... and also completely goes contrary to my understanding of how pinia SSR is supposed to work.

The main problem was using this value outside the setup function (in the Vuex store). We might be able to simplify this after we convert the media store to Pinia: then `usageSessionId` will be used by functions inside the store functions.

So, I guess this is an interim solution that will let us convert the `media` store to Pinia, and may be changed after `media` store is using Pinia.

And all this is making me think we should explore other third-party solutions for analytics sooner.

src/composables/use-usage-session-id.js
sarayourfriend: > That said, I wonder why usageSessionId is not used for other usage-data events

:thinking: you would think it should be eh?

sarayourfriend:
> And all this is making me think we should explore other third-party solutions for analytics sooner.

If we do that we could also just put it into a session cookie. We'll need a consent banner for that I think (non-essential) but thirty party solutions will probably all use cookies for this kind of thing.

obulat:
Closing this PR as it's replaced with #1084.

PR #1093, opened 2022-03-09
Create component to wrap content pages

### LINKED ISSUE #1090: Create `VContentPage` component
## Problem

When building text-heavy content pages, it's difficult to make sure the styles are correct on every page. To fix this, we could build a set of typography primitives: `<Heading level="1">` type components, and so on. Until we get to that stage, which would require a solid amount of work, it would be easier to create a `VContentPage` component that wraps raw html tags and applies styles to them using https://tailwindcss.com/docs/typography-plugin

This will allow us to make copy edits without also having to write many complex styles, and allow our content pages to be consistent by default. Then, if and when we make standalone small typography components, we can 'turn off' features of the tailwind plugin and switch to those incrementally.

## Implementation

zackkrida:
## Fixes

Fixes #1090 by @zackkrida 

## Description

This PR creates a `VContentPage` component, that simply applies some styles to raw html tags that are its children. It also matches the body font size of these pages and the content width to better match the mockups.

## Testing Instructions

View the standalone content pages, like `/about` and `/sources`, and observe that they look correct. This is a little subjective since the styles for these pages aren't the _most_ defined, so use your judgement and please point out if something doesn't look good.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VContentPage.vue
+++ b/src/components/VContentPage.vue
@@ -1,0 +1,9 @@
+<template>
+  <div class="pt-5 md:pt-10" dir="ltr">
+    <div
+      class="px-6 lg:px-0 mb-10 lg:mb-30 md:max-w-4xl xl:max-w-5xl prose prose-sm md:prose-base mx-auto max-w-none prose-a:text-pink prose-headings:font-bold lg:prose-headings:text-3xl lg:prose-h1:text-6xl"
+    >
+      <slot default />
+    </div>
+  </div>
+</template>
--- a/src/styles/text-only-page.scss
+++ b/src/styles/text-only-page.scss
@@ -11,11 +11,4 @@ a > em {
   font-weight: 600;
 }
 
-.table {
-  table-layout: fixed;
-  width: 100%;
-}
 
-.table.is-bordered td {
-  word-break: break-all;
-}
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "@pinia/nuxt": "^0.1.8",
     "@popperjs/core": "^2.11.2",
     "@tailwindcss/line-clamp": "^0.3.1",
+    "@tailwindcss/typography": "^0.5.2",
     "@wordpress/is-shallow-equal": "^4.3.1",
     "axios": "^0.21.2",
     "axios-mock-adapter": "^1.20.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -23,6 +23,7 @@ specifiers:
   '@playwright/test': ^1.16.2
   '@popperjs/core': ^2.11.2
   '@tailwindcss/line-clamp': ^0.3.1
+  '@tailwindcss/typography': ^0.5.2
   '@testing-library/jest-dom': ^5.16.1
   '@testing-library/user-event': ^13.2.1
   '@testing-library/vue': ^5.8.2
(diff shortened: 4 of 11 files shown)


src/components/VContentPage.vue
```diff
@@ -0,0 +1,9 @@
+<template>
+  <div class="pt-5 md:pt-10" dir="ltr">
+    <div
+      class="px-6 lg:px-0 mb-10 lg:mb-30 md:max-w-4xl xl:max-w-5xl prose prose-sm md:prose-base mx-auto max-w-none prose-a:text-pink prose-headings:font-bold lg:prose-headings:text-3xl lg:prose-h1:text-6xl"
```
obulat, line 4: Can we set the width a little smaller? Something closer to the '70 to 80 characters per line (about 8 to 10 words in English)' that is usually suggested for comfortable reading?

src/components/VContentPage.vue
```diff
@@ -0,0 +1,9 @@
+<template>
+  <div class="pt-5 md:pt-10" dir="ltr">
```
obulat, line 2: What is the reason for setting direction here? It is probably breaking 'About' page for RTL.

zackkrida: Good question: all of those pages had `dir="ltr"` set so I was just preserving the current behavior. I'll test if it should be removed.

src/styles/text-only-page.scss
```diff
@@ -11,11 +11,4 @@ a > em {
   font-weight: 600;
 }
 
-.table {
```
obulat, line 14: Would it make sense to move the rest of this file to `tailwind.css` base layer?

zackkrida: I think the `em` tag is being used incorrectly here, so perhaps better to just remove these. I'll look into it!

src/components/VContentPage.vue
sarayourfriend: In discussion with WP i18n folks we decided to set the dir on these pages because they're largely untranslated and also not RTL compatible to begin with. If they're translated _and_ they've been fixed to support RTL then it can be removed.

zackkrida: Ah, that's right. Thanks Sara! I'll test the status of translations _and_ how they look in RTL.

obulat: What do you think is the best time to flip the switch is? When any RTL language translation is available? Or when all have been translated to some percentage? I guess it's not possible to do it per-language...

sarayourfriend: I think we could do per language as long as translation percentage is available at run time (which I think the banner uses, so it must be, right?)

src/components/VContentPage.vue
zackkrida: It's much smaller than the current width of these pages, but maybe we could go even smaller. @panchovm could you look at this PR this week?

src/components/VContentPage.vue
zackkrida: @obulat  I made a new issue for this. None of these RTL languages have any translations, still. 

https://github.com/WordPress/openverse-frontend/issues/1118

panchovm:
The main styles look correct on desktop, but not sure how deep is the comparison I need to do. Nonetheless, I noted the following:

### On desktop
1. Margin bottom of `h1` looks too small, it should be `40px`.
2. Margin top of `h3` looks too small, it should be `40px`.
3. Links in main content have an underline style in resting state. It should show the underline in `hover`, as the source table items.
4. In the table, labels in header cells should be `description bold` style.
5. In the table, texts in cells should be `description regular` style.
6. External icon in main CTA is incorrect, it should be the WordPress Design Library one.

### On mobile
1. `h1`, `h3`, and paragraphs do not follow the style change as defined in mockups.
2. The table proposed has a different layout where table header is in a left column and data in a right column.

None of the observations noted above are blockers. Feel free to narrow the ticket scope and work on the suggestions in a different PR.

zackkrida:
Closing in favor of #1172

Issue #1088, opened 2022-03-09
Create table component; convert styles to tailwind


zackkrida:
## Problem

Currently, the table on the [sources page](https://wordpress.org/openverse/sources) relies on custom Sass styles. Additionally, all the logic for this table is included directly in the page. As part of our [milestone to remove legacy styles](https://github.com/WordPress/openverse-frontend/milestone/13), we should convert this component to use Tailwind for styles: inline in the templates where possible, but using `@apply {}` declarations where needed. 

We should also make a simple `VTable` component that extracts the current table and styles into a standalone component. From personal experience, table data structures and prop API design can be very controversial 😆 . Whoever implements this should discuss their proposed design prior to building.

**Edit:** Alternatively, this table could, at first, be hardcoded with only the necessary data for this page, and be called the `VSourcesTable`. We could then wait until tables appear again in our designs to write a full table abstraction. I think this is actually a better approach.

PR #1096, opened 2022-03-09
Delete external Sass stylesheets

zackkrida:
Proof of concept to delete all external stylesheets except `app.css`, previously `styles/tailwind.css`. This PR has known visual errors, allowed for the purposes of testing. All edited files besides the nuxt config, and the deleted style files will be reverted prior to undrafting this PR. This PR will be the final piece of the legacy style removal milestone.

I really wanted to test this PR for performance, but it seems difficult. I'll have to use the bundle analyzer and compare the css in our main application chunk. I _tried_ extracting all CSS to a single, cacheable CSS file here, because Tailwind's optimizations and purge rules _should_ result in a pretty lean file, but it sadly didn't work. Nuxt doesn't officially support it and the logic I tried to create a single css chunk in `nuxt.config.ts` resulted in odd errors. Here's the config I added:

```ts
{
build: {
  optimization: {
      // Extract CSS to single named file
      splitChunks: {
        cacheGroups: {
          styles: {
            name: 'styles',
            test: /\.(css|vue)$/,
            chunks: 'all',
            enforce: true,
          },
        },
      },
    },
  }
}
```


nuxt.config.ts
```diff
@@ -121,22 +121,14 @@ const config: NuxtConfig = {
   router: {
     middleware: 'middleware',
   },
-  components: [
-    { path: '~/components', extensions: ['vue'], pathPrefix: false },
-  ],
+  components: true,
```
zackkrida, line 124: will undo this.

zackkrida:
I'm going to close this. Once we implement #1088 this will be quite easy, that's the last instance of this we need to resolve.

PR #1100, opened 2022-03-11
Handle waveform with `peaks` prop as a blank array

### LINKED ISSUE #1074: Empty waveform peaks data renders an empty waveform; should render placeholder instead.
## Description

The API returns an empty array when the waveform peaks have not been generated. The frontend currently assumes that _any_ array returned will include waveform data. However, the frontend should also check the length of the array and use the placeholder if the peaks array is empty.

Alternatively we could update the API to return `null` for the `peaks` when they haven't been generated yet.

Marked as critical because we can't deploy without this fix or we'll have empty waveforms for most audio.

## Resolution

dhruvkb:
## Fixes

Fixes #1074 by @sarayourfriend 

## Description

This PR updates the `peaks` prop to use the default value even when given a truthy value of `[]`. This prevents the peaks from showing up as blank.

## Testing Instructions

0. Check out the PR and run the Storybook.
1. Visit a Waveform story and change the control for peaks to a blank array.
2. See the waveform become random.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VAudioTrack/VWaveform.vue
+++ b/src/components/VAudioTrack/VWaveform.vue
@@ -186,7 +186,6 @@ export default defineComponent({
     peaks: {
       type: Array,
       required: false,
-      default: () => Array.from({ length: 100 }, () => Math.random()),
       validator: (val) => val.every((item) => item >= 0 && item <= 1),
     },
     /**
@@ -332,8 +331,13 @@ export default defineComponent({
     const peakCount = computed(() =>
       getPeaksInWidth(waveformDimens.value.width)
     )
+    const peaks = computed(() =>
+      props.peaks?.length
+        ? props.peaks
+        : Array.from({ length: 1000 }, () => Math.random())
+    )
     const normalizedPeaks = computed(() => {
-      let samples = props.peaks
+      let samples = peaks.value
 
       const givenLength = samples.length
       const required = peakCount.value
--- a/src/components/VAudioTrack/VWaveform.vue
+++ b/src/components/VAudioTrack/VWaveform.vue
@@ -334,7 +334,7 @@ export default defineComponent({
     const peaks = computed(() =>
       props.peaks?.length
         ? props.peaks
-        : Array.from({ length: 1000 }, () => Math.random())
+        : Array.from({ length: 100 }, () => Math.random())
     )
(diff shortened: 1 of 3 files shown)


obulat (approved):
Nice solution 👌

src/components/VAudioTrack/VWaveform.vue
```diff
@@ -332,8 +331,13 @@ export default defineComponent({
     const peakCount = computed(() =>
       getPeaksInWidth(waveformDimens.value.width)
     )
+    const peaks = computed(() =>
+      props.peaks?.length
+        ? props.peaks
+        : Array.from({ length: 100 }, () => Math.random())
+    )
```
sarayourfriend, line 338: In a comment elsewhere you mentioned moving the logic up into `decode*`. I liked that suggestion because it alleviated the component from having to handle edge cases, essentially moving the edge-case handling "up the chain". Other places that rely on the waveform in the future would need to also handle this edge-case.

sarayourfriend (approved):
LGTM! I did like the suggestion you had before and am curious what your thought process was evaluating the trade offs between this solution and the other you suggested :slightly_smiling_face:

krysal (commented):
Can some tests be included here? I noticed that the waveform component has none and it is important.

zackkrida:
I'm okay with tests being added in a separate high-priority issue, in case we want to deploy this change this week or next.

krysal:
There is no set date for a release the next week so I don't see a reason to rush here, as long as it is handled before the release as mentioned in the linked issue.

src/components/VAudioTrack/VWaveform.vue
dhruvkb: I picked this because decode currently has no media-type specific logic.

dhruvkb:
I can add a couple of tests today and then merge it. Won't delay any releases!

sarayourfriend:
> I'm okay with tests being added in a separate high-priority issue, in case we want to deploy this change this week or next.

I don't think we can/should deploy this week anyway, it's Friday after all :slightly_smiling_face:

PR #1129, opened 2022-03-16
Fix audio track null duration and add defaultRef

### LINKED ISSUE #1076: Audio track current time sometimes being set to non-real number
## Description

https://sentry.io/share/issue/96afadf57a7e40acbd0fdb52a7d7d9c2/

I think this is being caused by a discrete math order of operations bug on this line: https://github.com/WordPress/openverse-frontend/blob/29bf0a0075548014a3d1280f0b56ae1f7e5f0b77/src/components/VAudioTrack/VAudioTrack.vue#L299

```
> const audioDuration = { value: null }
undefined
> const props = { audio: null }
undefined
> audioDuration.value ?? props.audio?.duration / 1e3 ?? 0
NaN
```

As you can see, if `props.audio` is null, it is still dividing `null / 1000` which is NaN. NaN isn't affected by `??` either so it doesn't default to 0.

It'd probably be best to just split this expression into multiple lines rather than keeping it a one-liner.

Also needs to be fixed in `VGlobalAudioTrack` as well.

## Reproduction
I'm not sure how to actually reproduce this

## Resolution

sarayourfriend:
## Fixes

Fixes #1076 by @sarayourfriend 

## Description

`defaultRef` is just a computed that uses some default computed value until another value is explicitly set.

This allows us to get rid of the confusing `duration` and `audioDuration` ref duplication that was happening in the audio track components by hiding what was the effective "default" in the original computed.

It also fixes the `null / 0` error.

## Testing Instructions

Check the unit tests to make sure I didn't miss any edge case.

Test the audio track, global, search results, related audio, and single result tracks, specifically from Jamendo to make sure there isn't a regression on #692.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,8 +11,8 @@ module.exports = {
     parser: '@typescript-eslint/parser',
   },
   extends: [
-    'plugin:@typescript-eslint/recommended',
     'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
     'plugin:vue/recommended',
     'plugin:prettier/recommended',
     'plugin:vuejs-accessibility/recommended',
@@ -143,9 +143,6 @@ module.exports = {
         ],
       },
     ],
-    // https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unused-vars.md
-    'no-unused-vars': 'off',
-    '@typescript-eslint/no-unused-vars': ['error'],
   },
   overrides: [
     {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -143,6 +143,9 @@ module.exports = {
         ],
       },
     ],
+    // https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unused-vars.md
+    'no-unused-vars': 'off',
+    '@typescript-eslint/no-unused-vars': ['error'],
   },
   overrides: [
     {
--- a/src/components/VAudioTrack/VGlobalAudioTrack.vue
+++ b/src/components/VAudioTrack/VGlobalAudioTrack.vue
@@ -32,6 +32,7 @@
 import { computed, defineComponent, ref, watch } from '@nuxtjs/composition-api'
 
 import { useActiveAudio } from '~/composables/use-active-audio'
+import { defaultRef } from '~/composables/default-ref'
 
 import { useActiveMediaStore } from '~/stores/active-media'
 
@@ -107,7 +108,11 @@ export default defineComponent({
 
     const status = ref('paused')
     const currentTime = ref(0)
-    const audioDuration = ref(null)
+    const duration = defaultRef(() => {
+      if (typeof props.audio?.duration === 'number')
+        return props.audio.duration / 1e3
+      return 0
+    })
 
     const setPlaying = () => {
       status.value = 'playing'
@@ -129,7 +134,7 @@ export default defineComponent({
       }
     }
     const setDuration = () => {
-      audioDuration.value = activeAudio.obj.value?.duration
+      if (activeAudio.obj.value) duration.value = activeAudio.obj.value.duration
     }
 
     const updateTimeLoop = () => {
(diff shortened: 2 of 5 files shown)


src/components/VAudioTrack/VGlobalAudioTrack.vue
```diff
@@ -107,7 +108,11 @@ export default defineComponent({
 
     const status = ref('paused')
     const currentTime = ref(0)
-    const audioDuration = ref(null)
+    const duration = defaultRef(() => {
+      if (typeof props.audio?.duration === 'number')
```
obulat, line 112: Can this ever be `NaN`? Would it be safer to add a check for `NaN`, too?

test/unit/specs/composables/default-ref.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { defaultRef } from '~/composables/default-ref'
```
obulat, line 1: Nice addition, and great to have tests for it!

obulat (approved):
Great fix, and useful addition of the `defaultRef`

src/components/VAudioTrack/VGlobalAudioTrack.vue
sarayourfriend: `props.audio` is the Audio result object which doesn't get modified by the frontend once it's decoded.

NaN isn't a valid JSON value so I don't think it could ever end up there:

```
> JSON.stringify({'a': NaN})
'{"a":null}'
```

src/composables/default-ref.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Produces a `computed` that returns a default value until
+ * an explicit value is set.
+ *
+ * @param getDefault - A function returning the default value
+ */
+export const defaultRef = <T>(getDefault: () => T) => {
+  const explicitlySet = ref<T | typeof NotSet>(NotSet)
+  return computed<T>({
+    get() {
+      if (explicitlySet.value !== NotSet) return explicitlySet.value as T
```
dhruvkb, line 15: When reading this condition "not equal to not set" feels like a double negative. Can we invert the `if`?

src/composables/default-ref.ts
```diff
@@ -0,0 +1,22 @@
+import { computed, ref, UnwrapRef } from '@nuxtjs/composition-api'
+
+const NotSet = Symbol('NotSet')
+
+/**
+ * Produces a `computed` that returns a default value until
+ * an explicit value is set.
+ *
+ * @param getDefault - A function returning the default value
```
dhruvkb, line 9: We should also document the `@returns`.

.eslintrc.js
```diff
@@ -143,6 +143,9 @@ module.exports = {
         ],
       },
     ],
+    // https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unused-vars.md
+    'no-unused-vars': 'off',
+    '@typescript-eslint/no-unused-vars': ['error'],
```
dhruvkb, line 148: I believe this can also be fixed by switching the order of
```js
extends: [
  'plugin:@typescript-eslint/recommended',
  'eslint:recommended',
  // ...
]
```
to
```js
extends: [
  'eslint:recommended',
  'plugin:@typescript-eslint/recommended',
  // ...
]
```
in the ESLint conf.

dhruvkb (changes requested):

src/composables/default-ref.ts
sarayourfriend: Done in https://github.com/WordPress/openverse-frontend/pull/1129/commits/bd5d8bb8d7568cf93a94ea62316401bef8b180aa

.eslintrc.js
sarayourfriend: Good call. Done in https://github.com/WordPress/openverse-frontend/pull/1129/commits/38ee96867d8dee5856a1b43702402e30742aafb2

src/composables/default-ref.ts
dhruvkb: That's true. I guess it's not really important when the description is so detailed, but I always write it (even if twice).

dhruvkb (approved):
Concerns addressed, LGTM

PR #1150, opened 2022-03-21
Typescriptify `api-service`

### LINKED ISSUE #1149: Add types to `data/api-service`
## Problem

The media store relies on this services, so it's nice to typescriptify them before converting it to TS.

## Implementation

obulat:
## Fixes

Fixes #1149 by @obulat

## Description

This PR typescriptifies the api and media services. 
There are still 2 issues left that I don't know how to solve:
```
 ERROR  ERROR in src/data/media-service.ts:32:7                                                                                                                                                                                  13:11:16
TS2322: Type 'Record<string, DetailFromMediaType<T>>' is not assignable to type 'T extends "image" | "audio" ? Record<string, DetailFromMediaType<T>> : never'.
    30 |     return {
    31 |       ...data,
  > 32 |       results: mediaResults.reduce((acc, item) => {
       |       ^^^^^^^
    33 |         acc[item.id] = decodeMediaData<T>(item, this.mediaType)
    34 |         return acc
    35 |       }, {} as Record<string, DetailFromMediaType<T>>),

ERROR in src/data/media-service.ts:92:7
TS2322: Type '(ImageDetail | AudioDetail)[]' is not assignable to type 'T extends "image" | "audio" ? DetailFromMediaType<T>[] : never'.
    90 |     return {
    91 |       ...res.data,
  > 92 |       results: res.data.results.map((item) =>
       |       ^^^^^^^
    93 |         decodeMediaData(item, this.mediaType)
    94 |       ),
    95 |     }

```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
@obulat Those errors are pretty unfortunate... I think for now we can fix them with a cast:

```diff
diff --git a/src/data/media-service.ts b/src/data/media-service.ts
index 8fc2a12c..d5016cbf 100644
--- a/src/data/media-service.ts
+++ b/src/data/media-service.ts
@@ -32,7 +32,7 @@ class MediaService<T extends FrontendMediaType> {
       results: mediaResults.reduce((acc, item) => {
         acc[item.id] = decodeMediaData<T>(item, this.mediaType)
         return acc
-      }, {} as Record<string, DetailFromMediaType<T>>),
+      }, {} as T extends 'audio' | 'image' ? Record<string, DetailFromMediaType<T>> : never),
     }
   }
 
@@ -91,7 +91,7 @@ class MediaService<T extends FrontendMediaType> {
       ...res.data,
       results: res.data.results.map((item) =>
         decodeMediaData(item, this.mediaType)
-      ),
+      ) as T extends 'image' | 'audio' ? DetailFromMediaType<T>[] : never,
     }
   }
 }
```

Once we've updated the media store to TypeScript after the Pinia conversion and are able to pass regular generic parameters, we won't have to rely on a string parameter to deduce the media type and should be able to convert `MediaService<T extends FrontendMediaType>` to `MediaService<T extends MediaDetail>` and also get rid of the generic parameter on MediaDetal (currently called BaseMediaDetail but the `Base` prefix is probably unnecessary).

src/data/api-service.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+  )
+
+  return {
+    /**
+     * @param resource - The endpoint of the resource
+     * @param params - Url parameter object
+     * @returns response  The API response object
+     */
+    query<T>(
```
sarayourfriend, line 79: `T` defaulted to `unknown` before to make it easier to call from JavaScript (because in JS you can't pass a generic type parameter). It'd be nice to keep it that way if possible to allow for casting the inner response time later in the code rather than immediately when calling the methods.

src/data/media-service.ts
```diff
@@ -0,0 +1,99 @@
...
+    id: string
+  }): Promise<DetailFromMediaType<T>> {
+    if (!params.id) {
+      throw new Error(
+        `MediaService.getMediaDetail() id parameter required to retrieve ${this.mediaType} details.`
+      )
+    }
+
+    const res = await VersionedApiService.get<DetailFromMediaType<T>>(
+      <SupportedMediaType>this.mediaType,
```
sarayourfriend, line 69: This style of casting isn't really used that much any more (I think because it's easy to confuse with JSX).

Why is the cast necessary though? `mediaType` already extends `'image' | 'audio'` which should match `SupportedMediaType` :thinking: 

```suggestion
      this.mediaType as SupportedMediaType,
```

src/data/media-service.ts
```diff
@@ -0,0 +1,99 @@
...
+   * @param params - API search query parameters
+   */
+  async search(
+    params: ApiQueryParams
+  ): Promise<MediaResult<Record<string, T>>> {
+    const res = await VersionedApiService.query<MediaResult<T[]>>(
+      <SupportedMediaType>this.mediaType,
+      params
+    )
+    console.log('res in search: ', res)
```
sarayourfriend, line 50: Nit: can we remove this console log?

src/utils/decode-media-data.ts
```diff
@@ -3,15 +3,15 @@ import { title } from 'case'
...
 /**
  * This interface is a subset of `Media` that types dictionaries sent by the API
  * being decoded in the `decodeMediaData` function.
  */
-interface ApiMedia extends Omit<Media, 'frontendMediaType' | 'title'> {
-  title?: string
-}
+// interface ApiMedia extends Omit<Media, 'frontendMediaType' | 'title'> {
+//   title?: string
+// }
```
sarayourfriend, line 14: Should we remove this code?

src/data/media-service.ts
```diff
@@ -0,0 +1,98 @@
...
+   * @param data - search result data
+   */
+  transformResults(data: MediaResult<T[]>): MediaStoreResult<T> {
+    const mediaResults = <DetailFromMediaType<T>[]>data.results
+    return {
+      ...data,
+      results: mediaResults.reduce((acc, item) => {
+        acc[item.id] = decodeMediaData<T>(item, this.mediaType)
+        return acc
+      }, {} as T extends SupportedMediaType ? Record<string, DetailFromMediaType<T>> : never),
```
sarayourfriend, line 35: This is incorrect, technically. It should just be `FrontendMediaType` as there are only two possible values for `T`, `image` and `audio` (unless there's also an ALL_CONTENT MediaService but it would fail to initialize as `this.mediaType` wouldn't match the `FrontendMediaType` constraint.

The only reason this is passing is because FrontendMediaType is a subset of SupportedMediaType, so it's saying, sure, it will extend SupportedMediaType.

obulat: I have read and re-read your [first comment](https://github.com/WordPress/openverse-frontend/pull/1150#issuecomment-1073797171), and then just tried implementing it in the media conversion PR. That PR is quite large, but I suspect it's better to tackle these two things together because media depends on types in the services so much ...

sarayourfriend: Could we split the PR to just typescriptify `api-service` maybe?

sarayourfriend (approved):
LGTM!

krysal (approved):
LGTM!

Issue #1020, opened 2022-02-28
Configure Storybook visual regression tests


sarayourfriend:
## Problem

We've added visual regression testing to our existing e2e suite to test against the running application. We should also add visual regression testing capability for Storybook where components are rendered in isolation.

## Description

**Copied from the RFC**

- [ ] Copy the existing `visual-regression` folder's configuration and create a new test folder called `storybook-visual-regression`.
- [ ] Create corresponding local and CI scripts in `package.json` to run these tests, using `storybook` instead of `run-server` and use the correct port for Storybook
- [ ] Write basic state and variation tests for each button variant (or some other straight component with similar easy to configure props).
    - This will also establish a pattern for writing e2e tests against Storybook.
- [ ] Copy the visual regression test CI and swap the appropriate scripts to run the Storybook ones instead.
- [ ] If possible, create a shared action that can be re-used between all the Playwright based test workflows.

Blocked by #1017

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

PR #1140, opened 2022-03-18
Fix hero search button layout error

t-hamano:
## Fixes
Fixes #1136

## Description
This PR fixes two problems with the hero search button:

- Text protrudes from the button
- Gallery images are placed above the search box at certain browser width
- The letters on the buttons are vertically aligned

## Testing Instructions
When the search button text is displayed (screens larger than 768px), check the following:

- Text must not be wrapped.
- The width of the search button should be variable depending on the length of the text.
- The search button should be positioned above the gallery image.

## ScreenShot

I have tested with `ja`, `fr`, `fa` locale.

https://user-images.githubusercontent.com/54422211/159016913-5e573d5b-8354-4e4f-a2e7-ad523ed86dff.mp4

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
[...]

Issue #974, opened 2022-02-21
Create `useUiStateCookie` composable


sarayourfriend:
## Problem

As part of https://github.com/WordPress/openverse/pull/164 we need a `useUiStateCookie` that reveals a readonly `reactive` object representing the UI state cookie.

We should implement this as a Pinia store to avoid global Ref issues with SSR (extremely important for this one otherwise cookie state could easily be leaked across requests). The store should grab default state variables from the cookie if it exists. Otherwise, it should create the cookie and set the appropriate defaults. Use UA sniffing to set the correct default breakpoint (`sm` for mobile UAs, `md` otherwise).

The store should use `useEventListener` to watch for resize changes and update the cookie breakpoint value appropriately. (Please debounce this function so that resizes aren't updating the cookie until resizing has stopped).

## Description

The store should expose `set` handles for updating the store that also update the stored cookie value.

Blocked by Pinia infrastructure being added.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

sarayourfriend:
This is blocked by pinia infrastructure getting merged in, likely as part of https://github.com/WordPress/openverse-frontend/pull/906

PR #1153, opened 2022-03-21
Add local visual regression infrastructure

### LINKED ISSUE #1017: Configure local visual regression testing
## Problem

We're adding visual regression testing and need to set up the local infrastructure for it.

## Description

**Copied from the RFC**

- [ ] Create a `playwright.config.ts` file in the new directory with the following code:
```ts
// /test/visual-regression/playwright.config.ts
import { PlaywrightTestConfig } from '@playwright/test'

export default {
    testDir: '.',
    use: {
        baseURL: 'http://localhost:8443',
    },
    timeout: 60 * 1000,
} as PlaywrightTestConfig
```
- [ ] Add the two `bin/visual-regression.{sh,js}` files described in the "Cross platform rendering" section above.
- [ ] Add three new scripts to the root `package.json` (**Note: in light of https://github.com/WordPress/openverse-frontend/pull/881 these scripts will also need to be updated to use talkback to mock API calls during e2e!**):
```
"test:visual-regression": "node ./bin/visual-regression.js",
"test:visual-regression:local": "playwright test -c test/visual-regression",
"ci:visual-regression": "start-server-and-test run-server http://localhost:8443 test:visual-regression"
```
- [ ] Install and configure `jest-image-snapshot` and `@types/jest-image-snapshot`:
    - Note: It may be necessary to add additional TypeScript typings in the root level `typings` folder for TypeScript to know that `expect()` includes the `toMatchImageSnapshot` function. Follow the existing examples extending `@types/nuxt` to extend the `@playwright/test` library's `Matchers` interface.
    - Additional note: As mentioned in the section about `jest-image-snapshot` above it may not be possible to get this library to play nicely with Playwright's specific `expect.extend` context. I'd say time box trying to get this to work to an hour or so and then just move on to the next thing. There's an additional task at the end of this list for writing the `pixelmatch` diff generator script in case we need it.
    - **Extra important note:** This library might not work with Playwright, in which case we'll just leave this out; there's a contingency plan for this case below.
- [ ] Write basic tests for the homepage; a full-page snapshot for `ltr` and `rtl` versions of the page using `testEachBreakpoint` should do just as a start and proof-of-concept.
    - Note: It will be necessary to remove the featured images from the page before taking the snapshot or the diff will fail 2/3 of the time due to them being different. The following code can be used to accomplish this:
    ```ts
    const deleteImageCarousel = async (page: Page) => {
        const element = await page.$('[data-testid="image-carousel"]')
        await element.evaluate((node) => node.remove())
        element.dispose()
    }
    ```

## Implementation

### LINKED ISSUE #1019: Configure CI to run visual regression tests
## Problem

We'll have added visual regression tests in #1017 but still need to run it in CI

## Description

Add visual regression tests to the CI workflow. You should be able to copy the existing e2e workflow and replace the call to `ci:e2e` to `ci:visual-regression`.

Blocked by #1017

## Implementation

sarayourfriend:
## Fixes

Fixes #1017 by @sarayourfriend
Fixes #1019 by @sarayourfriend

## Description

Changes in this PR:
* Renames the `e2e` test suite to `playwright`
* Adds a folder of visual regression tests to the playwright test suite
* Splits the existing playwright test utils into two separate modules and adds a new `breakpoints` utils for simplifying running a single scenario against multiple breakpoints
* Removes the half-baked cross-platform support and in the process removes playwright runner script duplication ~allowing us to rely entirely on the package.json scripts~
    * We still have to use a `bin/playwright.sh` in order to easily support passing arguments directly to `playwright` in the docker container.
* Fixes the docker tapes permission issue (thanks to dropping native Windows support it is possible to use `id -u` to get the user ID and have the correct permissions from the get go)
* Updates CI to run the full playwright suite
* Adds alias module resolution to Playwright. This prevents us from having to use relative imports both from `src` and from the root directory. **Should we enable the import rule to not allow relative imports?** :thinking: 
* Moves the Playwright testing information out of TESTING_GUIDELINES.md into a `test/playwright/README.md` and expands the documentation to include information about visual regression testing as well as additional information about the Talkback proxy and how it is configured.
* Uploads failed test-result artifacts. This not only includes the `actual` `expected` and `diff` snapshots but also the `trace.zip` for _any_ failed Playwright test, meaning it should be easier to debug all flaky Playwright tests.

Notably:
* Does not use `jest-image-snapshot`. This library doesn't support playwright/test, we'd have to switch to playwright-jest which we previously decided against using (which I think is fine). It also turned out to be unnecessary as Playwright does actually generate the `expected` `actual` and `diff` versions of the snapshot for failed tests in `test-results` whenever you pass a file extension to the snapshot name.
* Paves the way to setup storybook playwright tests (#1020). For that PR we'll move `playwright.config.ts` up one level and rename it `playwright.config.base.ts` and be able to just change a few properties to fit running the storybook in the playwright container instead of the Nuxt app

For the storybook PR, there's two approaches I think that follow naturally from this PR:
1. Either setup storybook in a separate folder (`test/storybook/playwright.config.ts`);
2. or else sub-folder the `playwright` suite further (`test/playwright/{storybook,nuxt}/playwright.config.ts`)

I think I like the second one better. Either way we'd want to update the package.json scripts to be clearer... `test:playwright` -> `test:playwright:nuxt`? It gets verbose if we do it that way. We could shorten playwright to `pw`, `test:pw:nuxt`, `test:pw:storybook` isn't so bad.

## Testing Instructions

Checkout this branch and run `pnpm install`.

Run `pnpm test:playwright` and check that the tests pass. Run the other `test:playwright:*` scripts and verify they do all the things we'd expect them to do and work as expected.

Also update something on the homepage (add a new bit of text or change the background color) and run `pnpm test:playwright -- visual-regression` and observe how failures are handled, including the `diff` generation in `test-results`. Likewise run `pnpm test:playwright -- visual-regression -u` to test snapshot updating.

~NOTE: Currently the snapshot tests aren't passing and I still need to debug what's going on there.~

It turns out I forgot to add the `data-testid` attribute that was necessary for deleting the image carousel from the page during snapshot tests. The image carousel is random from page-load to page-load so it introduces variations that cause tests to be flaky. It'd be nice if we could configure it to be non-random (perhaps using a feature flag? :eyes: https://github.com/WordPress/openverse/issues/178) but for now just deleting it from the page simplifies it.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
Opening this for review even though the snapshot tests aren't fully passing yet. I don't expect that whatever it takes to fix them will result in any changes to the infrastructure side of things so this should be safe to review for the most part already.

package.json
```diff
@@ -12,27 +12,25 @@
...
     "talkback": "node ./test/proxy.js",
-    "pretest:e2e": "pnpm install",
-    "test:e2e": "cross-env PLAYWRIGHT_VERSION=1.17.1 TEST_TYPE=e2e docker-compose -f docker-compose.playwright.yml up --abort-on-container-exit --exit-code-from playwright",
-    "test:e2e:local": "playwright test -c test/e2e",
-    "test:e2e:debug": "cross-env PWDEBUG=1 pnpm test:e2e:local",
-    "test:e2e:recreate-tapes": "npm-run-all delete-tapes test:e2e:update-tapes --continue-on-error",
-    "test:e2e:update-tapes": "npm-run-all update-tapes fix-tape-permissions --continue-on-error",
+    "playwright-version": "pnpm ls --dev --depth=0 | grep playwright | awk '{print $2}'",
+    "pretest:playwright": "pnpm install",
+    "test:playwright": "USER_ID=${USER_ID:-$(id -u)}; PLAYWRIGHT_VERSION=$(pnpm playwright-version --silent) docker-compose -f docker-compose.playwright.yml up --abort-on-container-exit --exit-code-from playwright",
```
sarayourfriend, line 28: The semicolon is necessary for being able to programatically set the `USER_ID` variable including with a default. the `${}` syntax with a default value doesn't work if you do the `VAR=[value] [command]` syntax directly.

package.json
```diff
@@ -12,27 +12,25 @@
...
+    "test:playwright:local": "playwright test -c test/playwright",
+    "test:playwright:debug": "PWDEBUG=1 pnpm test:playwright:local",
+    "test:playwright:recreate-tapes": "rimraf test/tapes && test:playwright:update-tapes",
+    "test:playwright:update-tapes": "UPDATE_TAPES=true pnpm test:playwright",
     "ci:unit-test": "jest --collectCoverage=false",
-    "delete-tapes": "rimraf test/tapes",
-    "update-tapes": "cross-env UPDATE_TAPES=true pnpm test:e2e",
-    "fix-tape-permissions": "docker run -it --rm --mount type=bind,source=$(pwd),target=/app node:alpine chmod -R 0777 /app/test/tapes",
```
sarayourfriend, line 35: Fixing tape permissions is no longer necessary with the `USER_ID` improvements to the docker configuration.

package.json
```diff
@@ -125,6 +123,7 @@
     "jest": "^26.6.3",
     "jest-transform-stub": "^2.0.0",
     "lint-staged": "^11.1.1",
+    "module-alias": "^2.2.2",
```
sarayourfriend, line 126: https://github.com/ilearnio/module-alias/

0 dependencies and seems solid. Hasn't been updated in a while but I guess it's not really something that needs to change very often. Supports CJS and ESM (though at this point we're just using ESM anyway :shrug:)

test/playwright/playwright.config.ts
```diff
@@ -1,7 +1,15 @@
...
+
 import type { PlaywrightTestConfig } from '@playwright/test'
 
+addAliases({
+  '~': path.resolve(process.cwd(), 'src'),
+  '~~': process.cwd(),
+})
+
 const config: PlaywrightTestConfig = {
-  testDir: '.',
```
sarayourfriend, line 4: By default playwright recursively checks the directory that the configuration file is in, so this setting was redundant with the default behavior.

sarayourfriend:
Okay, additional flakiness occurring now that is confusing. For some reason the `search` button renders wider in CI docker container than locally :thinking: Stumped by this!

index-ltr-md-actual
index-ltr-md-diff
index-ltr-md-expected

You can download the full `test-results` artifact at the bottom of this page: https://github.com/WordPress/openverse-frontend/actions/runs/2018793479

obulat:
> Okay, additional flakiness occurring now that is confusing. For some reason the search button renders wider in CI docker container than locally 🤔 Stumped by this!

I think this is caused by merging https://github.com/WordPress/openverse-frontend/pull/1140 that changed the search button rendering to allow for longer strings in other languages

sarayourfriend:
@obulat thanks for catching that, rebasing fixed the issue.

@dhruvkb and @obulat this is officially ready for review now.

I opened an issue to address the flaky test: #1162

package.json
```diff
@@ -12,27 +12,25 @@
   "author": "sebworks",
   "private": true,
   "scripts": {
-    "dev": "pnpm install && pnpm i18n:create-locales-list && pnpm i18n:update-locales && cross-env nuxt --hostname 0.0.0.0",
+    "dev": "pnpm install && pnpm i18n:create-locales-list && pnpm i18n:update-locales && nuxt --hostname 0.0.0.0",
```
obulat, line 15: Thank you for removing unnecessary `cross-env`! I did notice it and think of removing it a couple of times, but got distracted every time...

test/playwright/e2e/search-navigation.spec.ts
```diff
@@ -1,6 +1,6 @@
-const { expect, test } = require('@playwright/test')
+import { expect, test } from '@playwright/test'
 
-const { openFilters } = require('./utils')
+import { openFilters } from '~~/test/playwright/utils/navigation'
```
obulat, line 3: This file should probably also just use `mockProviderApis` on lines 5-6

test/playwright/playwright.config.ts
```diff
@@ -1,7 +1,15 @@
+import path from 'path'
+
+import { addAliases } from 'module-alias'
+
 import type { PlaywrightTestConfig } from '@playwright/test'
 
+addAliases({
+  '~': path.resolve(process.cwd(), 'src'),
+  '~~': process.cwd(),
```
obulat, line 9: Nice!

test/playwright/utils/breakpoints.ts
```diff
@@ -0,0 +1,87 @@
...
+          bp.replace('describe', '').toLowerCase() as Breakpoint
+        )
+      )
+        describe(block)
+    })
+  }
+
+const describeEvery = describeEachBreakpoint(Array.from(SCREEN_SIZES.keys()))
+const describeEachDesktop = describeEachBreakpoint(['2xl', 'xl', 'lg', 'md'])
+const describeEachMobile = describeEachBreakpoint(['sm', 'xs'])
```
obulat, line 79: Do we also need to pass the `ua-string`? Because we do check for it in a couple of places to render mobile views.

sarayourfriend: Oh right! Good call! I totally forgot about that. It will be used for #974.

sarayourfriend: Done in https://github.com/WordPress/openverse-frontend/pull/1153/commits/a2f19abe3bdcf8023a071a3f479145043a637ca0

test/playwright/e2e/search-navigation.spec.ts
sarayourfriend: Done in https://github.com/WordPress/openverse-frontend/pull/1153/commits/a3323af8af9d769e4beb112498a57aaccbd6cc6f

sarayourfriend:
@dhruvkb can you review this soon? I'd like to avoid further merge conflicts if possible. Thanks!

dhruvkb (approved):
I had gone through the PR earlier and everything seemed OK (except [this non-blocking comment](https://github.com/WordPress/openverse-frontend/pull/1153#discussion_r833306669)). I was actually looking for the discussion we had about Git LFS because I was under the impression we had voted to use that but turns out we didn't! 

I would love if we do switch to that soon though, considering that VRT snapshots (which range from ~5kb to ~50kb per image) will quickly increase the repo size. Just for the homepage we have about 2 dozen snaps.

test/playwright/playwright.config.ts
dhruvkb: Nuxt already uses this alias to point to `<rootDir>`. Can we pick something else here?

sarayourfriend: It is pointing to the root dir by using `process.cwd()`, these aliases match nuxt and ts configuration.

sarayourfriend:
@obulat if you have time to :heavy_check_mark: this PR soon I would appreciate it. Thanks.

test/playwright/e2e/search-types.spec.ts
```diff
@@ -23,34 +23,38 @@ test.beforeEach(async ({ context }) => {
...
+  results: /Over 10,000 results/,
+} as const
+
+const audioConfig = {
+  id: 'audio',
+  name: 'Audio',
+  url: '/search/audio?q=birds',
+  canLoadMore: true,
+  metaSourceCount: 3,
+  results: /93 results/,
```
obulat, line 49: These numbers make the tests flaky, but I'm not sure how to fix it. With all the hard work on the Catalog, there now are many more than 93 results for this search :)

obulat (approved):
I thought I had already approved this PR before :)
Such a relief that we can be confident that there will be no terrible visual regressions.

test/playwright/e2e/search-types.spec.ts
sarayourfriend: Shouldn't the tapes take care of that?

PR #1158, opened 2022-03-22
Convert `provider` store from Vuex to Pinia

### LINKED ISSUE #967: Add types to `data/media-provider-service.js`
## Description

Add type checking to `media-provider-service.js`. This module has zero untyped dependencies. Make sure to add it to `tsconfig.json`'s `include` list.

## Additional context

Part of an ongoing effort to add type checking to parts of the project that can be type checked.

## Implementation

### LINKED ISSUE #1027: Convert `provider` store from Vuex to Pinia
## Problem

As part of the conversion of the Vuex store modules to Pinia, we need to convert the [provider store](https://github.com/WordPress/openverse-frontend/blob/e0ac241dfe8b2b4b7eaff8eaee7e06eda0dce197/src/store/provider.js) to Pinia.

## Description

This store uses `nuxtServerInit` to fetch the providers on store initialization, and Pinia does not have an equivalent for that. We need to figure out the best way to initialize the provider data.
-  convert state to reactive

## Additional context

## Implementation

obulat:
## Fixes

Fixes #1027 by @obulat
Fixes #967 by @sarayourfriend 

**This PR was edited to convert the setup store to an options store to fix the server state hydration issues. Please, re-review :)**

## Description

This PR converts the `provider` store to Pinia. Provider stats include provider short name and display name, URL, as well as the media count per each provider. These data are used in the **provider filters for image/audio**, and on the **sources** page.

The state is simplified similar to the `media` store state to enable simple access to properties for various media types without string manipulation (e.g., `fetchState[mediaType].isFetching` instead of `isFetching${titleCase(mediaType)}Providers`).

Previously, we fetched provider data once, during the server initialization, using `nuxtServerInit` hook. Pinia does not have a direct analog.  Here, I used a `middleware` to call `fetchMediaProviders` before every route (found in [a Pinia discussion](https://github.com/vuejs/pinia/discussions/574)).
To minimize the number of unnecessary requests, I set a `lastUpdated` timestamp in the store as a `ssrRef`. When calling `fetchMediaProviders`, we only actually fetch data from the API if the provider array is empty, or more than 1 hour has passed since `lastUpdated`.
This way `provider` store loads all the provider stats for each request, and updates them if necessary once an hour. 

## Testing Instructions

Run the app, check the provider filters in the sidebar, they should work for images and audio.
Sources pages should display the image providers (note that audio providers were not yet added to this page).
Try disabling the network, you should see a warning in the console that providers were not loaded.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/middleware/middleware.js
```diff
@@ -31,4 +32,8 @@ export default function ({ query, route, $pinia }) {
   if (process.client && navStore.isReferredFromCc) {
     navStore.setIsReferredFromCc(false)
   }
+  if (process.server) {
+    const providerStore = useProviderStore($pinia)
+    await providerStore.nuxtServerInit()
```
sarayourfriend, line 37: What is this for? :thinking:

src/stores/provider.js
```diff
@@ -0,0 +1,112 @@
...
+ * @param {import('../store/types').MediaProvider[]} data
+ * @returns {import('../store/types').MediaProvider[]}
+ */
+const sortProviders = (data) => {
+  return data.sort((sourceObjectA, sourceObjectB) => {
+    const nameA = sourceObjectA.source_name.toUpperCase()
+    const nameB = sourceObjectB.source_name.toUpperCase()
+    return nameA.localeCompare(nameB)
+  })
+}
```
sarayourfriend, line 26: `sort` is not pure in JavaScript :disappointed: Should we spread `data` to avoid mutating it?

Adding `readonly` to array parameters would prevent these kinds of things I think.

```suggestion
/**
 * Sorts providers by their source_name property.
 * @param {readonly import('../store/types').MediaProvider[]} data
 * @returns {import('../store/types').MediaProvider[]}
 */
const sortProviders = (data) => {
  return [...data].sort((sourceObjectA, sourceObjectB) => {
    const nameA = sourceObjectA.source_name.toUpperCase()
    const nameB = sourceObjectB.source_name.toUpperCase()
    return nameA.localeCompare(nameB)
  })
}
```

src/stores/provider.js
```diff
@@ -0,0 +1,112 @@
...
+        })
+      }
+    }
+    return {
+      state,
+      nuxtServerInit: fetchMediaProviders,
+      getProviderName,
+    }
+  })
+export const useProviderStore = createUseProviderStore()
```
sarayourfriend, line 112: Why this `create...` pattern?

src/middleware/middleware.js
obulat: We were loading the array of providers with their stats on every server load by calling `nuxtServerInit` hook. This is the Pinia equivalent for it.
I guess we could just load a json file and update it regularly, similar to how we treat the i18n locale files.

src/stores/provider.js
obulat: When testing, we mock the `services`, and create the `provider` store using `const providerStore = createUseProviderStore(mockServices)`.
For the components, we can initialize by simply calling `useProviderStore`, which re-uses the store that has already been created using the default services.

src/middleware/middleware.js
sarayourfriend: Ohhh, so this shares the store state between requests?

src/stores/provider.js
sarayourfriend: Could we mock the services in the tests using `jest.mock`?

src/middleware/middleware.js
obulat: Well, I tested it, and it doesn't  :( It calls the API on each page reload I make.

sarayourfriend: Yeah, I wouldn't think that it would work this way.

If we need to share data between requests then we can do so with a singleton object in a module that gets initialized by the first request that hits it. We'd also need an accompanying accessible promise to act as an async lock for subsequent requests to await if they come in while the provider data is still being fetched.

src/stores/provider.ts
```diff
@@ -0,0 +1,150 @@
...
+    const noData = supportedMediaTypes.some(
+      (mediaType) => !state.providers[mediaType].length
+    )
+    if (noData || !lastUpdated.value) {
+      return true
+    }
+
+    const timeSinceLastUpdate =
+      new Date().getTime() - new Date(lastUpdated.value).getTime()
+    return timeSinceLastUpdate > 3600000
```
krysal, line 94: I believe that number could be an environment variable. Just a thought, is not really a requirement right now.

src/store/index.js
```diff
@@ -1,14 +0,0 @@
-import { FETCH_MEDIA_PROVIDERS } from '~/constants/action-types'
```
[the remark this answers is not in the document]
obulat: Great catch! I removed the folder, and moved the`types.ts` file to `/stores/` (and also moved some types from that file to where they are used).

test/unit/specs/stores/provider.spec.js
```diff
@@ -0,0 +1,149 @@
...
+  providerServices[mediaType].getProviderStats.mockImplementation(() =>
+    Promise.resolve({ data: [...mockData] })
+  )
+}
+
+const initialFetchState = {
+  canFetch: true,
+  fetchingError: null,
+  hasStarted: false,
+  isFetching: false,
```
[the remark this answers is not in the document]
obulat: When `hasStarted` is false, it means that the fetching for this specific query has never been done. This should be helpful for blank pages (for example, a start page for a provider): when `hasStarted` is false, we can show some preliminary information about the provider (collection), and when it's true, we either show the results, or the loading spinner.

src/stores/provider.ts
obulat: Updated in [150b8f8](https://github.com/WordPress/openverse-frontend/pull/1158/commits/150b8f81239e492a01136a80ab603aa001515316)

src/middleware/middleware.ts
```diff
@@ -11,11 +13,19 @@ import { useNavStore } from '~/stores/nav'
...
-export default function ({ query, route, $pinia }) {
+export default async function ({
+  query,
+  route,
+  $pinia,
+}: {
+  query: Record<string, string>
+  route: Route
+  $pinia: Pinia
+}): Promise<void> {
```
sarayourfriend, line 28: I guess this conflicts with #1228, should we merge that first then update this afterwards to use the standard implementation for the types?

src/stores/provider.ts
```diff
@@ -0,0 +1,154 @@
...
+  })
+}
+/**
+ * Timestamp is used to limit the update frequency to one every 60 minutes per request.
+ */
+const lastUpdated: Ref<Date | null> = ssrRef(null)
+
+const updateFrequency = process.env.providerUpdateFrequency
+  ? parseInt(process.env.providerUpdateFrequency, 10)
+  : 60 * 60 * 1000
```
sarayourfriend, line 47: Was this necessary in the end? I thought it was agreed to just make the request for the providers list on every request to SSR?

obulat: Yes, we've decided to make the request for the providers list on every request to SSR. However, without this check, it was making more than one provider request per SSR. It was updating on the server and on the client, too.

sarayourfriend: > it was making more than one provider request per SSR

Is this because multiple places request it? Is there somehow we could centralize it per request :thinking: Like adding it to the request context or something I guess?

src/stores/provider.ts
```diff
@@ -0,0 +1,154 @@
...
+  const needsUpdate = computed(() => {
+    const noData = supportedMediaTypes.some(
+      (mediaType) => !state.providers[mediaType].length
+    )
+    if (noData || !lastUpdated.value) {
+      return true
+    }
+
+    const timeSinceLastUpdate =
+      new Date().getTime() - new Date(lastUpdated.value).getTime()
```
sarayourfriend, line 97: TIL `getTime` and `valueOf` on `Date` are equivalent!

sarayourfriend (approved):
LGTM! Some minor typescript stuff but tested locally and it works great.

src/stores/provider.ts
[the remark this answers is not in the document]
obulat: Because I misunderstood how `env` variable checks are done :) Fixed!

src/models/media.ts
```diff
@@ -28,6 +28,8 @@ export interface Media {
 
   provider: string
   source?: string
+  providerName?: string
```
sarayourfriend, line 31: How does `providerName` get populated? And should it be snake case if it's coming from the API?

I see the API mapping `provider_name` field to `display_name` in the provider serializer: https://github.com/WordPress/openverse-api/blob/a7955c86d43bff504e8d41454f68717d79dd3a44/api/catalog/api/serializers/provider_serializers.py#L11-L14

But not on the media serializer :thinking:

src/utils/pinia-state.ts
```diff
@@ -0,0 +1,60 @@
+import { isClient } from '~/composables/window'
```
sarayourfriend, line 1: We can remove this module now right?

test/unit/specs/stores/provider.spec.js
```diff
@@ -0,0 +1,140 @@
...
+  beforeEach(() => {
+    setActivePinia(createPinia())
+    providerStore = useProviderStore()
+  })
+  afterEach(() => {
+    afterEach(() => {
+      providerServices.audio.getProviderStats.mockClear()
+      providerServices.image.getProviderStats.mockClear()
+    })
+  })
```
sarayourfriend, line 72:
```suggestion
  afterEach(() => {
    providerServices.audio.getProviderStats.mockClear()
    providerServices.image.getProviderStats.mockClear()
  })
```

src/stores/provider.ts
```diff
@@ -0,0 +1,166 @@
...
+    audio: MediaProvider[]
+    image: MediaProvider[]
+  }
+  fetchState: {
+    audio: FetchState
+    image: FetchState
+  }
+}
+
+export const initialProviderState: ProviderState = Object.freeze({
```
sarayourfriend, line 34: Should we use `deepFreeze` here since it's a nested structure?

src/stores/provider.ts
```diff
@@ -0,0 +1,166 @@
...
+  state: (): ProviderState => ({
+    providers: {
+      [AUDIO]: [],
+      [IMAGE]: [],
+    },
+    fetchState: {
+      [AUDIO]: { ...initialFetchState },
+      [IMAGE]: { ...initialFetchState },
+    },
+  }),
```
sarayourfriend, line 73: This can use the `initialProviderState` object, right? Just need to `deepclone` it if it's frozen.

sarayourfriend (approved):
Looks great! Tested well locally :grin:

src/models/media.ts
krysal: Good question, I wonder the same ❓

krysal (approved):
I tested it with changes of `main` and everything seems to work well :) Nice work!

src/models/media.ts
obulat: `providerName` is generated when fetching the media item (only when fetching the single item, because I guess search views do not require it?):
https://github.com/WordPress/openverse-frontend/blob/73d0dac020d9b495326109abf7b4dfdd42ab41d2/src/stores/media/index.ts#L343-L351
I guess we could also split the interface into `API Media` and `Local Media` if that makes it clearer in the future.

src/stores/provider.ts
obulat: I removed the `initialProviderState` object altogether, because it was overcomplicating things (deepFreeze/deepclone), and not really serving much purpose.

PR #1159, opened 2022-03-22
Convert `media` store to Pinia

### LINKED ISSUE #1042: Convert `media` store from Vuex to Pinia
## Problem

As part of the conversion of the Vuex store modules to Pinia, we need to convert the [media store](https://github.com/WordPress/openverse-frontend/blob/e0ac241dfe8b2b4b7eaff8eaee7e06eda0dce197/src/store/media.js) to Pinia.

This is a fairly large store, so this change should be split into several PRs.

## Description

-  convert state to reactive

## Additional context

## Implementation

obulat:
## Fixes

Fixes #1042 by @obulat

## Description

This PR converts the Media store from Vuex to Pinia, and typescriptifies it.
It replaces the types that were previously set to be used from JS with TS generics. There are still a couple of places where I couldn't fix types, and added `ts-ignore` to be able to commit and push the changes.

## Testing Instructions

Run the app. Searching all media types, filtering and loading more and viewing the detail pages should all work. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/stores/media.ts
```diff
@@ -0,0 +1,431 @@
...
+
+      // Elements exported only for testing, should not be used by components
+      test: {
+        fetchStartMedia,
+        fetchEndMedia,
+        fetchMediaError,
+        setMedia,
+        mediaNotFound,
+        handleMediaError,
+      },
```
sarayourfriend, line 427: The fetch state stuff seems like it could be it's own composable btw, in fact, it _is_ through `useFetch` and `useAsyncData` from Nuxt, right?

sarayourfriend (commented):
I opened a PR with some suggested additional simplifications to the media service alongside the ones you've already made: https://github.com/WordPress/openverse-frontend/pull/1173

This also removes the redundant `MediaDetail` type (it duplicates the `Media` type and does so at the cost of type resolution performance as well).

src/stores/media.ts
```diff
@@ -0,0 +1,431 @@
...
+          isFinished: false,
+        },
+        image: {
+          isFetching: false,
+          fetchingError: null,
+          isFinished: false,
+        },
+      },
+      audio: null,
+      image: null,
```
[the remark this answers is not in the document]
obulat: This looks like a great idea, @sarayourfriend ! I will have to think more about it.
I was planning to extract something like `singleMediaStore` for the detail pages, and the `relatedMediaStore`(converted from composable), after Pinia conversion is done.

src/stores/media.ts
obulat: I am hesitant to use `useFetch`  because it's not used in the same way in Nuxt 3.

I like the idea of creating a new `useFetchState` composable. I was thinking about using a status enum (idle/success/error/fetching), and error ref (null/error message), and a computed that returns the `isFetching/fetchingError` object.

sarayourfriend: That sounds good to me! Could that be in a separate PR that blocks this one?

src/models/media.ts
```diff
@@ -21,8 +21,38 @@ export interface Media {
...
+
+  thumbnail?: string
+  audio_set?: string
+  genres?: string[]
+  duration?: number
+  bit_rate?: number
+  sample_rate?: number
+  alt_files?: { provider: string; filetype: string }[]
+  filetype?: string
+  peaks?: number[]
```
sarayourfriend, line 54: I think some of these are inaccurate, either their not optional properties (but actually required and nullable instead) or they're always present but sometimes just empty (like `peaks` is a required property but is sometimes an empty array).

It's not a big deal though, in particular with the nullability, it'll just make writing code using these types more annoying becuase we'll have to add null checks in places that we don't need them (like on the array types).

`thumbnail` can be moved up to `Media` too as it's on both Audio and Images.

sarayourfriend (commented):
Looking great!

src/stores/media.ts
zackkrida: Resolved by the creation of `useFetchState` in #1178

sarayourfriend:
@obulat I opened https://github.com/WordPress/openverse-frontend/pull/1186 with suggestions for how to avoid the `create*` pattern using mocks. Sorry I didn't get that to you yesterday.

src/stores/media.ts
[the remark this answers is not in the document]
obulat: I opened a new issue for splitting the store based on @sarayourfriend 's comment, #1192 .

dhruvkb (approved):
This is a huge refactor! Thanks for adding/updating tests for the additional confidence that things are running as they should.

src/composables/use-fetch-state.ts
```diff
@@ -24,7 +24,6 @@ const statuses = Object.freeze({
   FETCHING: 'fetching',
   SUCCESS: 'success',
   ERROR: 'error',
-  FINISHED: 'finished',
```
zackkrida, line 27: Can we remove the comment above describing the `finished` state, if it no longer exists?

```
 * - `finished`: for multi-page requests, this is true when no more pages are left.
```

src/data/media-service.ts
```diff
@@ -0,0 +1,86 @@
...
+  private readonly mediaType: T['frontendMediaType']
+
+  constructor(mediaType: T['frontendMediaType']) {
+    this.mediaType = mediaType
+  }
+
+  /**
+   * Decodes the text data to avoid encoding problems.
+   * Also, converts the results from an array of media objects into an object with
+   * media id as keys.
```
zackkrida, line 19:
```suggestion
   * Decodes the text data to avoid encoding problems.
   * Also, converts the results from an array of media
   * objects into an object with media id as keys.
```

minor, minor nit to clean up the line breaks here

src/data/media-service.ts
```diff
@@ -0,0 +1,86 @@
...
+   * Retrieve related media
+   * @param params - object with id of the main media, for which to fetch related media
+   */
+  async getRelatedMedia(params: { id: string }): Promise<MediaResult<T[]>> {
+    if (!params.id) {
+      throw new Error(
+        `MediaService.getRelatedMedia() id parameter required to retrieve related media.`
+      )
+    }
+
```
zackkrida, line 72:
```suggestion
    }
```

for consistency with the method above it

zackkrida (approved):
I am sure a lot of the complexity came much earlier in the process, but I actually found this quite easy to review! It's great to see such a significant piece of code moved over to Pinia.

PR #1172, opened 2022-03-23
Content page component design fixes

### LINKED ISSUE #1090: Create `VContentPage` component

obulat:
## Fixes

Fixes #1090 by @zackkrida

## Description

This PR attempts to finish up the work started in #1093 to get it merged :)
It merges main and applies the following #1093 code review suggestions:

**On desktop**
- Margin bottom of h1 set to 40px.
- Margin top of h3 set to 40px.
- Links in main content only have an underline style on hover.
- In the table, labels in header cells are description bold style.
- In the table, texts in cells are description regular style.
- External icon in main CTA in sources uses correct icon.

**On mobile**
- h1, h3 styles updated

#### The following suggestions were not implemented:
- Mobile paragraph styles not updated.
- The table layout on mobile is unchanged.

#### Other changes:
- `scss` removed from the pages styles
- `VContentPage` name added so that it can be imported in the components.

## Testing Instructions

Check that the content pages look similar to the Sources page mockup, and are not broken.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VContentPage.vue
```diff
@@ -0,0 +1,15 @@
...
+      <slot default />
+    </div>
+  </div>
+</template>
+<script>
+const VContentPage = {
+  name: 'VContentPage',
+}
+export default VContentPage
+</script>
```
sarayourfriend, line 15: Take it or leave it :slightly_smiling_face: 

```suggestion

<script lang="ts">
import { defineComponent } from '@nuxtjs/composition-api'
export default defineComponent({
  name: 'VContentPage',
})
</script>
```

tailwind.config.js
```diff
@@ -202,7 +202,26 @@ module.exports = {
...
+            '--tw-prose-links': theme('colors.pink'),
+            a: {
+              textDecoration: 'none',
+              '&:hover': {
+                'text-decoration': 'underline',
+              },
+            },
+          },
+        },
+      }),
```
sarayourfriend, line 219: Isn't this redundant with what we have in `tailwind.css` in the `@base` layer extension?

sarayourfriend (commented):
Tested locally and looks good to me. One question before approving but otherwise looks good.

src/components/VContentPage.vue
zackkrida: Nice catch @sarayourfriend, I found myself staring at this in my review thinking "hm, something isn't right here" 😆

tailwind.config.js
obulat: With the new tailwind plugin, typography, the texts and links on the content pages have different classes applied to them, so it becomes necessary.

package.json
```diff
@@ -58,6 +58,7 @@
     "@pinia/nuxt": "^0.1.8",
     "@popperjs/core": "^2.11.2",
     "@tailwindcss/line-clamp": "^0.3.1",
+    "@tailwindcss/typography": "^0.5.2",
```
sarayourfriend, line 61: I guess I don't fully understand why we're using this plugin and why if we _do_ use it it doesn't render the code in `tailwind.css` unnecessary (should we remove it if we use this plugin?).

src/components/VContentPage.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <div class="pt-5 md:pt-10" dir="ltr">
+    <div
+      class="px-6 lg:px-0 mb-10 lg:mb-30 md:max-w-4xl xl:max-w-5xl prose prose-sm md:prose-base mx-auto max-w-none prose-headings:font-bold lg:prose-headings:text-3xl prose-h1:text-4xl prose-h1:text-bold lg:prose-h1:text-6xl md:prose-h1:mb-10 prose-h3:text-2xl prose-h3:font-semibold md:prose-h3:mt-10"
```
sarayourfriend, line 4: This massive list of `prose-*` classes also scares me. It has all the features of the worst parts of Tailwind IMO (namely, a super long and inscrutable list of class names that are difficult to edit, read, and get a full picture of).

Adding our own `prose` class to `tailwind.css` with these styles using the `@apply` rule would probably be easier to read and maintain :shrug: Just my own feeling about this plugin though.

package.json
sarayourfriend: Makes sense, thanks!

zackkrida:
@sarayourfriend the primary reason I'd like to use the plugin is so we have more raw html tags in use and can easily replace things, like `<h1>` ~> `<Heading level="1">` or whatever if and when we make our design system more robust, instead of having to do a bunch of manual work or, god forbid, regex matching on html tags with classes and other attributes. Your thoughts are well-received!

Issue #1132, opened 2022-03-17
Allow media store to cancel previous media-fetch requests if a new one is sent before the previous is completed


sarayourfriend:
## Problem

It's currently possible to cause a race condition in search. If you submit query A and then immediately submit query B, if the API takes longer to return A than it takes to send and return B, then B will come back before A, sent to the store, and then be replaced by the response to A.

## Description

[Axios supports `AbortController`](https://axios-http.com/docs/cancellation) so it shouldn't be too bad to use an abort controller entry in the media store as a "lock" on the search execution that is forcefully released each time a new search is executed.

## Alternatives

Prevent sending requests in the UI by blocking new requests while existing ones are still evaluating... This seems like a bad UX though.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

PR #1178, opened 2022-03-24
Add useFetchState composable

obulat:
## Fixes

Related to #1042 by @obulat

## Description

This PR adds a useFetchState composable that will be [used in the media store ](https://github.com/WordPress/openverse-frontend/pull/1159#discussion_r833459871)converted to Pinia. It will also be reused in the `related media` store, and media detail (or `single media item`) store. 

In addition to the `fetchState` properties that were used in the media store (`isFetching`, `fetchingError` and `isFinished`), this PR adds 2 more:
- `canFetchAgain`: this prop will help us avoid race conditions in fetching described in #1132. We can only send requests if `canFetchAgain` property is `true`
- `hasStarted`: this prop is false when the current search has not been executed yet (after the change of filters, or when the `q` search term hasn't been set). This will allow us to create provider landing pages. If the fetch state `hasStarted` is false, but the provider filter is selected, we can display the 'landing page' for the provider.

## Testing Instructions

Currently, this composable is not used anywhere, so we only need to agree on the properties that it exposes.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/composables/use-fetch-state.ts
+++ b/src/composables/use-fetch-state.ts
@@ -1,0 +1,111 @@
+import { computed, reactive, Ref, ref, watch } from '@nuxtjs/composition-api'
+
+import type { FetchState } from '~/store/types'
+
+/* Constants */
+
+/**
+ * Statuses of requests:
+ * - `idle`: the current request has never been sent yet, for example,
+ * on the first app load, or after a change in search filters or search term.
+ * - `fetching`: the request was sent, but no response was received yet.
+ * - `success`: a successful response was received for the current request.
+ * - `finished`: for multi-page requests, this is true when no more pages are left.
+ * - `error`: an error response was received.
+ */
+const statuses = {
+  IDLE: 'idle',
+  FETCHING: 'fetching',
+  SUCCESS: 'success',
+  ERROR: 'error',
+  FINISHED: 'finished',
+} as const
+
+type Status = typeof statuses[keyof typeof statuses]
+/**
+ * When one of these statuses is received, the fetchError is set to `null` in a watcher.
+ */
+const nonErrorStatuses: Status[] = [
+  statuses.IDLE,
+  statuses.FETCHING,
+  statuses.SUCCESS,
+]
+/**
+ * With these statuses, it is possible to send the same request for a new page.
+ * This can help debounce requests and prevent racing states.
+ */
+const canFetchAgainStatuses: Status[] = [statuses.IDLE, statuses.SUCCESS]
+
+/* Composable */
+
+export const useFetchState = (initialState = statuses.IDLE) => {
+  const fetchStatus: Ref<Status> = ref(initialState)
+  const fetchError: Ref<string | null> = ref(null)
+
+  watch(fetchStatus, () => {
+    if (nonErrorStatuses.includes(fetchStatus.value)) {
+      fetchError.value = null
+    }
+  })
+  const reset = () => {
+    fetchStatus.value = statuses.IDLE
+  }
+  const startFetching = () => {
+    fetchStatus.value = statuses.FETCHING
+  }
+  /**
+   * Called when any response is received. For an error response,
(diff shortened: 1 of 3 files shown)


src/composables/use-fetch-state.ts
```diff
@@ -0,0 +1,111 @@
...
+    fetchStatus.value = statuses.FINISHED
+  }
+
+  /**
+   * Whether the current request has been sent. This can be useful for displaying
+   * initial search page with a blank search term, such as for a provider search page.
+   */
+  const hasStarted = computed(() => fetchStatus.value !== statuses.IDLE)
+  const isFetching = computed(() => fetchStatus.value === statuses.FETCHING)
+  const canFetchAgain = computed(() =>
```
sarayourfriend, line 81: Should this just be `canFetch`? The "again" kind of implies that it's already been fetched which isn't true for the `idle` status :thinking:

src/composables/use-fetch-state.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  return {
+    isFetching,
+    fetchingError,
+    canFetchAgain,
+    hasStarted,
+    isFinished,
+    isError,
+
+    fetchingState,
```
sarayourfriend, line 104: What's the reason for the `fetchingState` reactive when all the data included in it is already present on the top level of the store's API?

src/composables/use-fetch-state.ts
```diff
@@ -0,0 +1,111 @@
...
+    fetchStatus.value = statuses.FETCHING
+  }
+  /**
+   * Called when any response is received. For an error response,
+   * pass the errorMessage string as a parameter.
+   * @param errorMessage - the message to show for response error.
+   */
+  const endFetching = ({
+    errorMessage = undefined,
+  }: { errorMessage?: string } = {}) => {
```
sarayourfriend, line 63: Just wondering why this is an object parameter rather than a plain string param. It seems like it'd be simpler as a string param:

```suggestion
  const endFetching = (errorMessage?: string) => {
```

Do we anticipate there being other things needing to be passed here?

sarayourfriend (approved):
LGTM!

src/composables/use-fetch-state.ts
obulat: True :)

src/composables/use-fetch-state.ts
```diff
@@ -0,0 +1,111 @@
...
+ * - `finished`: for multi-page requests, this is true when no more pages are left.
+ * - `error`: an error response was received.
+ */
+const statuses = {
+  IDLE: 'idle',
+  FETCHING: 'fetching',
+  SUCCESS: 'success',
+  ERROR: 'error',
+  FINISHED: 'finished',
+} as const
```
dhruvkb, line 22: Can `Object.freeze` this.

dhruvkb (approved):
LGTM, except for a non-blocking suggestion.

src/composables/use-fetch-state.ts
obulat: I meant to ask for everyone's opinion on what the best option is: 
- add `fetchState` as a sub-item to be used in components and the store as it is used now,
or
- to have everything in the top level. 
I guess having it in the top level should be enough, don't you think?

sarayourfriend: I think having it all on the top level makes sense too. Otherwise watching any of them requires watching either the full state object _or_ using `toRefs`, which feels cumbersome.

obulat: Ok, I removed the `fetchingState`, and then I realized that using all those single pieces separately as computed means that we would have to add `.value` to each of them every time we consume them, which isn't very user-friendly. Do you think having it as a reactive would really affect the performance, @sarayourfriend ?

PR #1163, opened 2022-03-22
Add support for TypeScript in Vue SFCs.

sarayourfriend:
## Description

Adds support for TypeScript in Vue SFCs.

Notable changes:
* Usage of `vue-tsc` and addition of some modules that Vue 2 doesn't automatically pull in necessary for template types
* Applied fixes from https://github.com/johnsoncodehk/volar/discussions/592 due to Storybook pulling `@types/react` into the chain causing templates to have the incorrect types applied to them by TypeScript compiler

I chose a simple `VLink` component to be the test candidate.

## Testing Instructions

Checkout this branch and make changes in VLink and run `pnpm types` test compilation.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #1182, opened 2022-03-24
Fix missing nuxt types

sarayourfriend:
## Description

In the Vue SFC typescript PR #1163 I added the `types` compiler option to our tsconfig and this removed some critical nuxt types from global scope. Adding them back also revealed that I'd accidentally duplicated the work that the Nuxt vue-i18n module does to make the types work so I was able to remove some code as well.

I also added an ESLint rule to prevent accidentally destructuring the `i18n` object as it will result in a failure if you do and try to call one of the methods destructured due to the dependency on `this` in the method.

## Testing Instructions

Go to `use-i18n.ts` and call the function, make sure it has the correct types when you inspect the type information in your editor (try to use some of the properties like `.t` for example). This tests the removal of the code in the nuxt__types/index.d.ts module and also tests that the types from the module are being added.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VLink.vue
```diff
@@ -33,9 +33,11 @@ export default defineComponent({
   },
   setup(props) {
     const { app } = useContext()
-    const hasHref = computed(
-      () => typeof props.href === 'string' && !['', '#'].includes(props.href)
-    )
+    function checkHref(p: typeof props): p is { href: string } {
+      return typeof p.href === 'string' && !['', '#'].includes(p.href)
+    }
```
krysal, line 38: This looks curious to me 😯 Is `p is { href: string }` the returning type or an extension of the params type?

krysal: That makes sense for the part of the code using the `checkHref` function but my doubt is actually with the definition highlighted here 😅

dhruvkb (approved):
This is great.

src/components/VLink.vue
sarayourfriend: I see, sorry.

The `p is { href: string }` is the return type yes. Any return type with the `is` form is taken to be a predicate and the return value is interpreted as a `boolean`. It refines the type whatever the caller passes for `p` at the call site.

obulat: Thank you for the explanation!

obulat (approved):
Tested using the instructions, works as you describe. LGTM!

src/components/VLink.vue
krysal: Now I see! Thank you for this explanation, the example in the link made it crystal clear. The way Typescript works is interesting.

PR #1184, opened 2022-03-25
Update banner styles

### LINKED ISSUE #1123: Style update of translation and news banners
## Problem
As noticed in a #904's [comment](https://github.com/WordPress/openverse-frontend/pull/904#pullrequestreview-887570089) by @zackkrida, the translation and News banners have styles that look clunky and broken on mobile. The yellow background of News banner makes the page feel cut by the translation one. 

## Description

|Desktop|Mobile|
|:---|:---|
|image|image|

The proposal above has the following changes:
* Background change of News banner, from brand yellow to `Blue 1` (#3E58E1) from [News redesign](https://wordpress.org/news/).
* New close action on the far-right area on both banners.
* Both banners have `16px 24px` for paddings on both desktop and mobile.
* On desktop, texts are `label regular`.
* On mobile, the close action is center aligned and texts are `caption regular`.

Regarding banners behavior, I think both elements should show over the page instead of pushing downward the homepage main content. In that way, closing the banners do not move the page upwards.

sarayourfriend:
## Fixes

Fixes #1123 by @panchovm

## Description

Primarily focused on updating the banner styles and unifying the implementations so that both have a close button. The unification effort is where most of the changes come from as well as some on-the-fly TypeScript conversion.

I'd like to split this PR into multiple parts but will be on vacation for the next week so just wanted to get my WIP up in case someone else wanted to pick it apart while I'm gone.

Changes proposed:
* Convert `use-storage` and `use-event-listener` to TypeScript and simplify the initial value handling in use-storage.
* Update `local.ts` to implement the full `Storage` API 1:1 instead of using a bespoke API. Also simplify the error handling in it.
* Convert NotificationBanner to TypeScript and update it to be able to handle dismissal on its own rather than relying on the parent to handle it. This makes it easy to implement new dismissable banners without having to duplicate logic.
* Update TranslationBanner to use the new NotificationBanner API and remove the unnecessary stuff from `use-i18n-sync` (which can probably be renamed now that nothing is being "synced")

## Testing Instructions

Checkout the branch and run `pnpm dev`. Visit the site in English and verify there is no banner. Visit the site in Russian  (`ru`) and verify there is a banner and that it can be dismissed and that it stays dismissed between requests (aside from the already existing problem that the banner is rendered SSR, to be fixed by the UI state cookie milestone).

Visit the site in another low-translation percent language like Portuguese (`pt`). Verify that a new banner shows with the correct information and that it is dismissable and stays dismissed. Verify that you can switch between Russian and Portuguese and get no banner.

Now do the same with two more low-translation-percent languages and do not dismiss the banners and ensure that both show the appropriate banner between page requests.

Note: All of the above could be encoded in e2e tests and I plan to do that when I get back from vacation and have time to implement additional tests for this on top of the already existing ones.

The CC referral banner is not yet implemented but testing it will go as follows:

Visit the site with the `referrer=creativecommons.org` query param and ensure that the banner shows and that it is dismissable. Ensure that it stays dismissed on subsequent requests to the site with the query param. (note again that this has not yet been implemented on this branch!).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
Blocked by #1241 for the tapes fix.

src/components/DownloadButton.vue
```diff
@@ -91,7 +91,7 @@ export default {
   },
   data() {
     const savedFormatExtension =
-      local.get(LS_DOWNLOAD_FORMAT_EXTENSION_KEY) ?? null
+      local.getItem(LS_DOWNLOAD_FORMAT_EXTENSION_KEY) ?? null
```
obulat, line 94: Would using `useStorage` composable be an overkill here?

src/components/VTranslationStatusBanner.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <VNotificationBanner v-show="!shouldHideBanner" @close="dismissBanner">
+  <VNotificationBanner
+    v-if="needsTranslationBanner"
+    :id="bannerKey"
+    variant="informational"
+  >
     {{
       // eslint-disable-next-line @intlify/vue-i18n/no-raw-text
       '⚠️'
```
obulat, line 9: We could add this symbol to ignored strings that @krysal created.

src/components/DownloadButton.vue
sarayourfriend: I guess not, I'd just like to avoid doing that refactor in this PR as well if that's okay? We can open a separate issue for it :slightly_smiling_face:

src/components/VTranslationStatusBanner.vue
sarayourfriend: Should we use that list for _anything_ you think? The ones Krystle added made sense to me because they're used in a couple of places in the app, I think... Hopefully raw emoji aren't that common in the app :thinking:

obulat: Now that you say so, I guess it does make sense to use that list for repeated strings, and not for single emojis :)

src/components/DownloadButton.vue
obulat: Sure, I just remembered looking at this PR previously and trying to understand the difference between the two. Then I saw that `useStorage` is actually using `local` internally

obulat (changes requested):
The button on the mobile version is positioned incorrectly: it should be at the end horizontally, but it shows at the bottom.

Figma mockup: https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Audio-Release?node-id=953%3A17185

sarayourfriend:
@obulat thanks for catching that. I think it is due to the `md:` modifiers in `VNotificationBanner`. They're leftover from the previous, specific implementation. Removing them now to see if it fixes it.

src/layouts/blank.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <VMigrationNotice v-show="isReferredFromCc" />
+    <VMigrationNotice v-if="isReferredFromCc" />
```
obulat, line 3: Would this cause client-server mismatch? Often when you use `v-if` and the server state is different from the client state (i.e., something depends on the client state such as localStorage), the mismatch will cause 'hydration bail' and full re-rendering on the client.

sarayourfriend: Ah, I thought I remembered there being a good reason for it being `v-show`. I'll put it back.

sarayourfriend: We can move it to `v-if` if we want when we implement the ui state cookie.

src/components/VTranslationStatusBanner.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <VNotificationBanner v-show="!shouldHideBanner" @close="dismissBanner">
+  <VNotificationBanner
+    v-show="needsTranslationBanner"
```
obulat, line 3: The `v-show` logic here only uses the translation percentage, and doesn't check for the value in localStorage. This is why if you dismiss a banner, and refresh the page, you will see the banner again, and will not be able to dismiss it anymore.

obulat (changes requested):
There needs to be some more logic checking the value in `localStorage` to show/hide the banner _**after**_ it's been dismissed and the page is reloaded.

src/components/VTranslationStatusBanner.vue
sarayourfriend: That logic is handled in the `VNotificationBanner` component itself, with the local storage. But it seems like stacking `v-show` or something is messing with it. Thanks for catching this, I will debug it.

sarayourfriend: So I think the issue is that because we have a `v-show` on the external render of the component, and because `v-show` just uses CSS to show/hide the element, it's overriding the local storage handling inside the component because `v-show="true"` in the component rendering it, even though inside (I've confirmed) `v-show="false"`.

The only thing I can think of is to forward the "v-show" variable as a new prop to be combined with the local storage variable. Hacky, but I guess it is necessary for `v-show` because of the way it works.

obulat: There are two layers of `v-show` now, in the `TranslationBanner` and in the `NotificationBanner` (one checks for local storage, the other - for translation percentage). Maybe they are cancelling each other out? It would be nice to combine them, use the outer `v-show` value as a prop to `NotificationBanner`, maybe?

obulat: The comment was too late, you've already implemented the fixes 🤣

obulat (approved):
Works! Thank you for lots of added tests.

sarayourfriend:
@dhruvkb can you take a look at this if you get a chance today? Thanks!

dhruvkb (approved):
LGTM, love the `tryUse` function.

sarayourfriend:
It must be the button variations. Looking into it now. Thanks for the review!

PR #1191, opened 2022-03-26
Fix buttons focus in old browsers

### LINKED ISSUE #1081: Focus-visible in Safari is not working
## Description

The `focus-visible` polyfill we added in #724 is not working.

## Reproduction

1. Run the app and open http://localhost:8443/search/?q=cat in a desktop Safari browser
2. Tab through the search bar to the page menu and content switcher buttons.
3. See error: You are supposed to see pink ring around those buttons when you focus on them using <kbd>Tab</kbd>, but there is no indication of where the focus is.

## Expectation

Elements that have keyboard focus should have a pink ring around them.

## Environment

 - Device:  laptop
 - OS:  MacOS
 - Browser:  Safari

## Additional context

It is not working on staging and in production, too.
## Resolution

krysal:
## Fixes

Fixes #1081 by @obulat

## Description

~Adds required styles to the base `VButton` component for browsers that don't support `focus-visible`, as suggested in the polyfill docs: [Backwards compatibility](https://github.com/WICG/focus-visible#backwards-compatibility).~

This PR adds the focus-visible styles in the template section of components where was missing. Also bumps the versions of `postcss-focus-visible` and `postcss` packages.

## Testing Instructions

Run the project and browse the search page tabbing on Safari version `15.2 (17612.3.6.1.6)` or `15.3 (17612.4.9.1.8)` for example, header buttons should show focus. Behavior in other browsers should not be affected.

Might be required in other custom components.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (changes requested):
This is definitely an improvement over current situation because it adds focus ring to some buttons in Safari.
However, I believe that normally all the styles were added automatically by the postcss plugin, not manually. 
Some of the buttons such as the homepage content switcher buttons (which probably use `<a>`) have a blue outline instead of the pink ring.

zackkrida (approved):
I actually like manually focusing the focus style for elements without `focus-visible` styles. I think that's a sensible default.

One thing I would like to add as part of this PR: Currently the `VContentLink` buttons have a blue outline in Safari. I believe this is because they don't actually have a focus and/or focus-visible style applied in tailwind correctly, rather they inherit a style from a global sass stylesheet:

CleanShot 2022-03-28 at 12 45 34@2x

Could you add focus-visible styles to the VContentLink component in this PR, following the example of one of our buttons or the VImageCellSquare component for example? Then I think we can merge.

zackkrida:
@krysal I actually meant to 'request changes', I'll leave it approved though.

krysal:
The previous solution was avoiding the polyfill, which I thought wasn´t working well but is indeed adding the `focus-visible` class and the `data-focus-visible-added` attribute when using the keyboard in safari. The actual problem is Tailwind styles for `focus-visible` aren't being applied when declared in a **CSS module**. You can see this behavior in a simple component like the `VLoadMore` button, which also didn't show a ring when focused on using the keyboard.

I found another workaround that is still not ideal because is repeating these utilities in higher-order components instead of just placing them in the `VButton` but seems better than the current situation.

I fixed all the focus on the search page but I noticed the Content Switcher on the homepage in desktop view needs fixing, I will address that today.

obulat:
There is an issue in the TailwindCSS Repository about `focus-visible` not working CSS Modules using `@apply`: https://github.com/tailwindlabs/tailwindcss/issues/6194 . It has a PR that's been merged, but I tried updating the `tailwind` package to the latest version, and it still doesn't fix the header buttons `focus-visible` issue in Safari.

I guess we could solve the problem with header buttons either by reconsidering using CSS Modules in the `VButton` component, or adding `focus-visible` classes to the button itself, not the classes in `<style>` declaration. In this case, we should also monitor all the content type switcher items to not have a double focus ring (in both mobile and desktop states).

obulat:
There is an issue in the TailwindCSS Repository about `focus-visible` not working CSS Modules using `@apply`: https://github.com/tailwindlabs/tailwindcss/issues/6194 . It has a PR that's been merged, but I tried updating the `tailwind` package to the latest version, and it still doesn't fix the header buttons `focus-visible` issue in Safari.

I guess we could solve the problem with header buttons either by reconsidering using CSS Modules in the `VButton` component, or adding `focus-visible` classes to the button itself, not the classes in `<style>` declaration. In this case, we should also monitor all the content type switcher items to not have a double focus ring (in both mobile and desktop states).

krysal:
Good call! I forgot about header components in mobile so it quickly become annoying to repeat the classes in each component and of course, the sensible solution was to add them inline, I don't know how it didn't occur to me before 😆 Thanks for the suggestion.

@obulat  could you review this again? Buttons on the homepage and the header should have their focus visible in Safari and there shouldn't be any with double rings.

zackkrida:
Re-reviewed; looks good!

PR #1203, opened 2022-03-29
Add Sketchfab and Thingiverse as additional sources for 3D models

dhruvkb:
## Fixes

Fixes #1194 by @zackkrida 
Fixes #1119 by @zackkrida 
Fixes #928 by @sarayourfriend 

<s>Parts of this PR conflict with #1190, which is unfortunate but unavoidable given the nature of the work.</s> Resolved in #1204.

## Description

This PR
- adds a 3D models media type
- renames legacy sources to additional sources
- adds Sketchfab and Thingiverse as sources for 3D models
- converts the util and 4 other components to TypeScript
- refactors to code to move most of the logic from components to the util
- updates tests

### Dev server

0. Check out the PR and run the dev server.
1. Search for a term with no results.
2. See the meta-search links.
3. Change the media type.
4. Observe the meta-search links change.

### Storybook

To see all media types and their search results, use the stories for VMetaSourceList in the Storybook.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

[...]

Issue #1197, opened 2022-03-28
Rename "Meta Search" to "External Sources"


zackkrida:
## Problem

We currently use two names "Meta Search" and "Additional Sources" to refer to the same feature: letting users search a list of external search engines that also contain CC-licensed content. I think "Additional Sources" is much more sensible for non-technical users. Additionally, using the word "Meta" in the context of our name "Openverse", as it relates to the "Metaverse" is also quite unfortunate.

## Description

Let's remove all references to 'meta search' and replace them with 'additional sources', including internal naming of code, methods, string keys, and so on.

We need to make sure a redirect is in place to preserve url so that `https://wordpress.org/openverse/meta-search/` redirects to `https://wordpress.org/openverse/additional-sources/`.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

panchovm:
The word "additional sources" is confusing term as we are not adding anything to Openverse. 

@sarayourfriend suggested in an internal conversation that “external sources” sounds more explicit about what will happen. And I agree with that point.

This term also relates to the "additional types" designed for the switcher popover to show non-supported-yet content types, like video. In that vein, it sounds reasonable revisiting both.

Based on the current pages and designs we made, here are the places that @obulat identified where we are using one or both terms.

## Content switcher

Ticket #739 proposed creating an "additional types" section for the upcoming content types. The feature is not live yet but I think we are on time to revisit it.

<img src="https://user-images.githubusercontent.com/895819/158441163-f7b2ac22-34a9-444b-b458-08aaf9909d65.gif" width="350">

Replacing "additional types" with "coming soon", as you suggested internally @zackkrida, is very accurate and explicit about what this content type belongs to.

## Meta search section

For this section, the current implementation is outdated and ticket #734 addresses this change. It is currently blocked since it waits for this copy definition. Here is how it looks on desktop.

<img src="https://user-images.githubusercontent.com/895819/151384759-80cc75d9-a131-45f8-b1d0-be6c2d098f81.png" width="1400">

In this case, I would replace the “Try another source” with “Try an external source” to follow the idea shared initially.

## Non-supported content type (video)

In previous conversations, we decided to create a page for the upcoming content types to show the existing external sources. We started with video for now, but this page applies to any future content type we decide to support.

Here is a screenshot of a design I am working on and plan to share once we find a path for this discussion.

[...]

PR #1190, opened 2022-03-26
Add minimum 3d model constants and frontend code for additional sources in the content switcher

zackkrida:
## Fixes

Related to #739 by @zackkrida 

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [ ] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

Issue #734, opened 2022-01-27
Update the meta search section with the new design


panchovm:
## Problem
The current external sources area in search results does not follow the redesign.
**_Updated issue:_** only the mobile popup needs to be implemented for this issue.

<img width="800" alt="CleanShot 2022-01-27 at 11 59 18@2x" src="https://user-images.githubusercontent.com/895819/151384406-b4d3dffa-74f6-4b62-a0c1-d507c7f5ace1.png">

## Description
We need to update the section with the design made on [the mockups](https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=133%3A6869).

### ~~Desktop version~~ Done
<img width="1387" alt="Meta search on desktop" src="https://user-images.githubusercontent.com/895819/151384759-80cc75d9-a131-45f8-b1d0-be6c2d098f81.png">

### Mobile version
Instead of the current implementation in the popover, the external sources should open in a modal:
<img width="300" alt="Meta search on mobile" src="https://user-images.githubusercontent.com/895819/151384963-7252b7c4-5ef1-4dcd-8bd8-ef9062a8ac9c.png">

dhruvkb:
Currently we have two ways in which the meta-search feature is reached.

- a search has zero results

<img width="610" alt="Screenshot 2022-06-28 at 11 35 35 AM" src="https://user-images.githubusercontent.com/16580576/176120958-56cd2e1e-3d75-45a8-b5d2-82e62e25a7b5.png">

- below the load more button of the search results

<img width="585" alt="Screenshot 2022-06-28 at 11 37 50 AM" src="https://user-images.githubusercontent.com/16580576/176121403-2bb70d2c-c65b-4fe3-b70d-9d4decb0bcc7.png">

@panchovm does this redesign only affect the second appearance?

panchovm:
Yes @dhruvkb, only the second appearance. Although I would pause this work until we have a decision regarding keeping using the term "Additional."

dhruvkb:
Good call @panchovm. I'll triage this issue as blocked for now.

obulat:
@panchovm, we have already decided on renaming the feature to "External sources". Can we update the status of this issue?

Currently, meta search section uses a popover on all layouts: desktop and mobile, the bottom of the search results and the page shown when nothing could be found:
### Desktop
[...]

PR #1205, opened 2022-03-30
Renamed Meta search to additional sources

HILAYTRIVEDI:
## Fixes
Fixes #[1197] by @[zackkrida]

## Description
This PR will replace all the instances of the meta-search with additional sources available in the GitHub files.

## Checklist


src/composables/use-pages.js
```diff
@@ -25,9 +25,9 @@ export default function usePages() {
       link: app.localePath('/search-help'),
     },
     {
-      id: 'meta-search',
-      name: 'header.meta-search-nav-item',
-      link: app.localePath('/meta-search'),
+      id: 'additional-sources',
+      name: 'header.additional-sources-nav-item',
+      link: app.localePath('/additional-sources'),
```
dhruvkb, line 30: Since you've changed the link, you will need to rename `src/pages/meta-search.vue` to `src/pages/additional-sources.vue`. Nuxt uses file names to map to URLs.

src/locales/en.json
```diff
@@ -110,38 +110,38 @@
...
-  "meta-search-page": {
-    "title": "Meta Search",
+  "additional-sources-page": {
+    "title": "Additional Sources",
     "intro": "{openverse} is built on top of a catalog that indexes CC-licensed and public domain content from selected sources. Learn more about our {link}.",
     "link": "sources",
     "license": "The internet is a big place! There are plenty of search portals and collections, hosting CC-licensed and public domain content, which we aren’t able to index or surface results from within the {openverse} interface. These are important sources of CC-licensed and public domain content, and we want to make sure that you are able to find the best openly licensed materials possible, regardless of where they are located.",
-    "content": "What content types can I search for using the Meta Search feature?",
-    "content-types": "Meta Search on {openverse} currently supports image, audio, and video search on external sites.",
+    "content": "What content types can I search for using the Additional Sources feature?",
```
dhruvkb, line 118: Example of a grammatical error introduced because "Additional Sources" is not the proper-noun of a feature name but rather an adjective-qualified common-noun.
```suggestion
    "content": "What content types are supported by additional sources?",
```

dhruvkb (changes requested):
I have some preliminary feedback before diving into the full review. This is good and covers most of the usages. There are a few issues, however:

- `package-lock.json` should not be modified, please revert it to the same version as on `main`.
- Since "Additional sources" is not the name of a feature as "Meta search" was, it's grammatically incorrect in some places. But those wordings will need to be changed anyway as a part of the Meta search redesign #734 so it's not a blocking change.

This PR will conflict with both #1190 and #1203. I would highly recommend waiting for those to get merged before working on this.

HILAYTRIVEDI:
Hi @dhruvkb, Thank you for the info. I will wait till the PRs will merge

HILAYTRIVEDI (commented):
I did the required changes But waiting for the PRs to be merged. once they are being pushed I will push my code too.

sarayourfriend:
@HILAYTRIVEDI please note that we use `pnpm`, not `npm` to install dependencies. While things may work with `npm install` you won't be running with the exact versions of packages that we've locked to and tested with. Please do follow the instructions in the [README's "Local Development" section](https://github.com/WordPress/openverse-frontend/blob/main/README.md#local-development) to correctly set up your local environment (and avoid introducing things like `package-lock.json` for example).

Thanks for the contribution!

src/composables/use-pages.js
sarayourfriend: Do we also want to map a redirect for this link?

krysal (commented):
@HILAYTRIVEDI are you still interested in completing this PR? Try resolving conflicts with `main` in that case, we can help if you have questions, or let us know if you can't continue working on this. Thanks!

obulat (commented):
Thank you for your work on this issue, @HILAYTRIVEDI !
I've merged main into your PR, and fixed the 'additional sources' page by renaming it to `additional-sources.vue`. It's still necessary to proofread the text of that page.

src/locales/en.json
sarayourfriend: Just a nit here, but this isn't technically a grammatical error, and in fact could maybe even be clearer than what you suggested in context. At least in this case the phrase 'Additional Sources feature' groups it as a single block. In other strings you're right that it reads awkwardly to just replace "Meta Search" directly with "Additional Sources", but only because it's not qualifying "feature".

In general I think the wording is probably bad here because we're treating "Additional Sources" (even if lower cased) as a feature and also the name of a collection of external resources. I think it could be a good idea to revisit this group of strings altogether instead of doing a 1-to-1 term replacement. For example, this particular string is really confusing to read as it could be asking either "what content types do the websites listed in the additional sources feature support" or it could read "what content types does Openverse integrate 'additional sources' into"... both of which expose what I think is still a really unclear name we've chosen in "additional sources" (it's maybe even more unclear than the very obscure "meta search" if only because "meta search" is almost semantically meaningless the way we used it but "additional sources" is rife with ways of misunderstanding/misinterpreting).

zackkrida:
Hi @HILAYTRIVEDI we have the final language this PR needs. You can see it here: https://github.com/WordPress/openverse-frontend/issues/1197#issuecomment-1195901505

Would you like to continue working on this or have someone else take over?

PR #1207, opened 2022-03-30
Clean up of `package.json` scripts

dhruvkb:
## Description

This PR
- cleans up the scripts section of the `package.json` file for improved consistency and clearer hierarchy
- increments the talkback port by one to prevent conflict with a macOS 12.3 service
- corrects some documentation

## Changes

- `prepare` moved to the top (considering it's run first)
- `ver` added as a generic replacement for `playwright-version`
---
- `dev` split into `predev` and `dev`
- `dev:only` added like `build:only`
- `build` split into `prebuild` and `build` 
- `build-only` renamed to `build:only` (for consistency with the `:` notation)
- <s>`prod` removed (purpose unclear)</s> restored, purpose discovered
---
- `storybook:build` and `storybook:smoke` added as they are common actions
---
- `test:all` renamed to `test`
- `test` renamed to `test:unit`
- `test:watch` renamed to `test:unit:watch`
- `ci:unit-test` removed (can just call `test:unit`)
- `playwright-version` removed (use `ver`)
- `pretest:playwright` changed to `pretest`
- `test:playwright:recreate-tapes` fixed by adding missing `pnpm`
- `generate-playwright-test` renamed to `test:playwright:gen`
---
- `lintfix` renamed to `lint:fix`
- `format:test` renamed to `format`
- `format` renamed to `format:fix` to match lint
---
- `i18n:get-translations` reduced to just the middle portion (with `i18n` replacing it)
- `i18n` added to run all i18n commands `i18n:create-locales-list`, `i18n:get-translations` and `i18n:update-locales`
- `i18n:no-get` added to run `i18n` except for downloading the 200 translations, used by `dev`
- `vue-i18n-check` renamed to `i18n:check` and updated to include `ts` files

## Testing Instructions

Play around with some stuff you usually do, you should not experience any errors other than when you enter the old name from muscle memory.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


package.json
```diff
@@ -12,38 +12,43 @@
   "author": "sebworks",
   "private": true,
   "scripts": {
-    "dev": "pnpm install && pnpm i18n:create-locales-list && pnpm i18n:update-locales && nuxt --hostname 0.0.0.0",
+    "prepare": "husky install",
+    "ver": "func() { pnpm ls --depth=0 | grep -e \"^${1}\\s\" | awk '{print $2}'; }; func",
```
zackkrida, line 16:
```suggestion
    "version": "func() { pnpm ls --depth=0 | grep -e \"^${1}\\s\" | awk '{print $2}'; }; func",
```

I think I'd prefer `version` personally, it isn't that much longer.

dhruvkb:
This [is the reason](https://pnpm.io/next/cli/run#differences-with-npm-run) why this is a draft PR.

package.json
dhruvkb: For some reason, when I run `pnpm version` it runs `npm version` automatically. It might be some incorrect setup on my end or a feature/bug of `pnpm` but given that behaviour, `ver` seemed like the safer/non-conflicting choice. 

image

dhruvkb:
Turns out I do need reviews and repros so gotta undraft it.

package.json
zackkrida: Oh, good call. Apologies for not considering that!

sarayourfriend:
@dhruvkb pre and post scripts are executed because of this `.npmrc` line: https://github.com/WordPress/openverse-frontend/blob/79c74e98842aa557fd809299fdd7dee7f4631a90/.npmrc#L2

It's a necessary override of the default `pnpm` behavior to get some of our packages to work which assume pre/post script existence.

README.md
```diff
@@ -101,18 +96,18 @@ The standalone mode which has a large header with logo and a footer, can be enab
...
 
 ```bash
-pnpm test:e2e
+pnpm test:playwright
 ```
 
 You might have to run `npx playwright install` to get the browsers installed if e2e tests fail.
 
-When writing e2e tests, you can also use `pnpm generate-e2e-tests` to generate tests and test selectors.
+When writing e2e tests, you can also use `pnpm test:playwright:gen` to generate tests and test selectors.
```
sarayourfriend, line 110: Let's delete this section altogether and link to the TESTING_GUIDELINES document so we don't duplicate this information.

Some of it is out of date (for example, the `npm playwright install` advice is generally going to be unnecessary for most people who will just be running things in docker anyway).

dhruvkb:
Ahhh, thanks for pointing that out @sarayourfriend. That discrepancy between the docs and the actual behaviour was very weird!

README.md
```diff
@@ -133,7 +128,7 @@ ngrok http 8443 -host-header="localhost:8443"
 
 ## Formatting and Linting
 
-The code in this repository is formatted using `prettier`. If you have prettier setup in your code editor it should work out of the box; otherwise you can use the `pnpm lintfix` script to format and fix lint errors in your code. Checks are run to lint your code and validate the formatting on git precommit using [husky](https://github.com/typicode/husky).
+The code in this repository is formatted using `prettier`. If you have prettier setup in your code editor it should work out of the box; otherwise you can use the `pnpm lint:fix` and `pnpm format:fix` scripts to format and fix lint errors in your code. Checks are run to lint your code and validate the formatting on git precommit using [husky](https://github.com/typicode/husky).
```
sarayourfriend, line 131: I wonder why we split `format` and `lint` in this way? Could we just have a single `lint:fix` script that runs the formatting fix?

package.json
```diff
@@ -12,38 +12,44 @@
...
```
sarayourfriend, line 16: This pains me _deeply_ to read. Does this need to be generic, is it used elsewhere, could it be a YAGNI to complicate things in this way? If so, could we move it to a proper shell script in `bin/get-package-version.sh` or something?

sarayourfriend (commented):
Love this clean up, thank you! Just a few questions before approving.

package.json
dhruvkb: It is a complicated expression, yes, but it's just a simple generalisation of the `playwright-version` script that existed before it:
```bash
pnpm ls --dev --depth=0 | grep playwright | awk '{print $2}'
```
It's only used in one place (https://github.com/WordPress/openverse-frontend/pull/1207/files#diff-ad76caf254756b7bb67a44577cdd651afc57e3b77289360b59868361b3a62a12R7) and that too seems more informational that functional. We could get rid of it altogether. Having a separate shell script for a one-liner would be overkill.

README.md
dhruvkb: Great advice! Implemented in 11ba387.

README.md
dhruvkb: Mentally I associated `lint` with ESLint and `format` with Prettier. While they do overlap in the formatting department, aren't their concerns slightly different?

package.json
sarayourfriend: It's not just informational, the environment variable is used in the docker-compose.yml to determine what version of the playwright image to pull.

We could just add the version getting function into the `playwright.sh` script instead of leaving it to pollute the package.json scripts.

I would prefer that a lot actually, it doesn't need to be in package.json at all (I think it was only in there originally because it was being used directly by another package.json script but that's no longer the case).

README.md
sarayourfriend: They are technically different, but we don't make the distinction in our Python code (it's all just one big pre-commit run with everything, formatting and linting included) and I think I do commonly see prettier grouped under linting :shrug: 

Is there ever a situation in which you would want to run one and not the other? It seems to me that both should always run whenever one or the other is run (which to me is the most compelling reason to make them a single script, despite there being a technical difference between the two).

package.json
dhruvkb: Seems that we agree, moved in 8b60e2b.

README.md
dhruvkb: That makes sense, one less command to remember. Combined in b4a001c.

sarayourfriend: Nice! That is much nicer to work with :tada:

.github/workflows/ci.yml
```diff
@@ -89,7 +86,7 @@ jobs:
       - uses: ./.github/actions/setup-node-env
 
       - name: Run Storybook smoke-test
-        run: pnpm storybook -- --ci --smoke-test
+        run: pnpm storybook:smoke
```
sarayourfriend, line 89: Another thing occurring to me, sort of as a consequence of our discussion about the playwright version script.

I wonder if things that aren't used (or meant to be used) generally, like this smoke test script, should actually be in the package.json at all. It makes the list of scripts longer just to slightly shorten the already not-too-complex one-liner in the CI scripts.

sarayourfriend (approved):
Great! I had one more philosophical question but feel free to merge this either way :grin: Thanks again for the awesome clean up :tada:

dhruvkb:
While we await the second approval, it'll be nice to discuss whether we should and how we can trim the list of scripts down to the bare minimum.

.github/workflows/ci.yml
dhruvkb: That does make sense, a lot of scripts could be eliminated if we only keep those in the `package.json` file that are used frequently and/or mostly invoked manually.

krysal (approved):
Nice clean-up 💯 Thanks, @dhruvkb!

PR #1210, opened 2022-03-31
More TS-ification and refactoring

dhruvkb:
## Fixes

No issue closed but this belongs to the [TypeScriptification milestone](https://github.com/WordPress/openverse-frontend/milestone/10).

## Description

- src
	- **assets/licenses** | renamed to match CC's [press kit](https://mirrors.creativecommons.org/presskit/icons/) & license elements
	- components
 		- License
     	 	- **VLicense.vue** | updated to use the `getElements` util added to `utils/license.ts`
      		- **VLicenseElements.vue** | updated to use the `getElements` util added to `utils/license.ts`
			- meta
       	 		- **VLicenseElements.stories.mdx** | added stories, test in Storybook
    	- VContentReport
    		- **VContentReportForm.vue** | updated to match flat constants structure
    		- **VReportDescForm.vue** | updated to match flat constants structure
    	- VErrorSection
    	    - **VErrorSection.vue** | function renamed, no change
		- VFilters
			- **VLicenseExplanation.vue** | import path change, can ignore
		- **VLink.vue** | minor type change, can ignore
  		- VMediaInfo
  			- **VCopyLicense.vue** | hollow out component using new attribution util `getAttribution` 😎 
  			- **VMediaLicense.vue** | import path change, can ignore
  			- **VMediaReuse.vue** | removed unused props passed to component that doesn't accept them
  	- constants
		- **license.ts** | added constants and types for license elements
	- models
		- **media.ts** | replaced `string` with types `License` and `LicenseVersion`
	- pages/audio
		- **_id.vue** | removed unused props passed to component that doesn't accept them
	- utils
		- **attribution-html.ts** | _converged all markup generation logic into one function_ 💪
		- **license.ts** | wrote a util to split a license into it's constituent elements
- tests/unit/specs/utils | added/updated tests for the above changes

The main code change to review in this PR is `attribution-html.ts`. I tried my best to keep it modular and readable but I'm happy to provide a walkthrough, perform further refactors if the code is too messy.

## Testing Instructions

Since this is a refactor PR, play around with with the site (touching upon the pages with the affected components) and ensure everything works like before. Ideally nothing should be broken.

Also check out the Storybook for some added/fixed stories.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb:
On further thought, gonna refactor some more.

src/components/VContentReport/VReportDescForm.vue
```diff
@@ -44,7 +44,7 @@ export default defineComponent({
      */
     reason: {
       type: String,
-      validator: (val) => Object.values(reasons).includes(val),
+      validator: (val) => reasons.includes(val),
```
obulat, line 47: It would be nice to extract this kind of validation to a separate function to ensure it works well with TypeScript. The problem here is that `reasons` is a `const` array, and using `includes` with values that are not in that array throws an error: https://stackoverflow.com/questions/56565528/typescript-const-assertions-how-to-use-array-prototype-includes

src/constants/content-report.ts
```diff
@@ -0,0 +1,14 @@
...
+export const MATURE = 'mature'
+export const OTHER = 'other'
+
+export const reasons = [DMCA, MATURE, OTHER] as const
+
+export const SENT = 'sent'
+export const FAILED = 'failed'
+export const WIP = 'wip'
+
+export const statuses = [SENT, FAILED, WIP] as const
```
obulat, line 11: Can we either:
a. add exported types for `reason` and `status` here, similar to the `ErrorCode` type in `errors.ts` below, and use them for prop validation
or 
b. create a util function that can take a string value, and validate it against a `const` array so that TS doesn't throw an error. Now, if you try to use `reasons.includes(somestringvalue)` when validating with TS, you would get an error.

src/models/media.ts
```diff
@@ -18,8 +19,8 @@ export interface Media {
   url: string
   foreign_landing_url: string
 
-  license: string
-  license_version: string
+  license: License
```
obulat, line 22: Nice update!

src/utils/attribution-html.ts
```diff
@@ -1,70 +1,191 @@
 import type { Media } from '~/models/media'
```
obulat, line 1: Very nice cleanup of the attribution creation function!
Do you think we should also open an issue for internationalizing it? It would be out of scope for this PR.

test/unit/specs/utils/attribution-html.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  creator_url: 'https://creator/url',
+  license: 'pdm',
+  license_version: '1.0',
+  license_url: 'https://license/url',
+}
+
+describe('getAttribution', () => {
+  it.each`
+    sendI18n | attributionText
+    ${true}  | ${'"Title" by Creator is marked with Public Domain Mark 1.0 .'}
```
obulat, line 26: Can we remove the space before the final period in the attribution text?

obulat (approved):
I really like the cleanup, and additional tests!
I've added some comments inline, but they are non-blockers.

src/constants/content-report.ts
dhruvkb: Good call, updated in 4b0b9ee.

src/components/VContentReport/VReportDescForm.vue
dhruvkb: Good call, updated in 4b0b9ee.

test/unit/specs/utils/attribution-html.spec.ts
dhruvkb: This is not an actual space, we're comparing the textual content of the HTML with this string (there are actually `` tags with license icons there). 

I've added an extra test case in 9a66217 to confirm that there are no stray spaces in the output.

src/utils/attribution-html.ts
```diff
@@ -1,70 +1,191 @@
...
+  /* Attribution */
 
-  return `${title} by ${creator} is licensed under ${license}.`
-}
+  const i18nBase = 'media-details.reuse.credit'
+  const isPd = isPublicDomain(mediaItem.license)
+
+  let attribution: string
+  if (i18n) {
+    let fillers: object = {
```
sarayourfriend, line 146: Why the `object` type annotation? Unfortunately this type is mostly non-useful for 99% of applications.

Did you want `Record<string, any>` maybe?

sarayourfriend (commented):
I appreciate a lot of the changes in this PR but it is very big. Even the list of changes (which is helpful) is long. Would you mind splitting this PR into some smaller PRs? It would spread the review burden around better and make it easier to verify the changes to the application and generally easier to review.

test/unit/specs/utils/attribution-html.spec.ts
```diff
@@ -0,0 +1,70 @@
+import Vuei18n from 'vue-i18n'
```
sarayourfriend, line 1: Thank you for adding tests! The changes to the attribution HTML generation are really nice!

dhruvkb:
@sarayourfriend I've moved as much out of this PR as I safely could before breaking stuff. I hope this is slightly more reviewable now?

src/utils/attribution-html.ts
dhruvkb: Yes, updated in d7b9bde.

src/components/License/VLicenseElements.vue
```diff
@@ -23,21 +23,26 @@
...
-import samplingPlus from '~/assets/licenses/sampling-plus.svg'
+import samplingPlus from '~/assets/licenses/sampling.plus.svg'
 
 export default defineComponent({
   name: 'VLicenseElements',
   components: { VIcon },
   props: {
+    /**
+     * the slug of the license
+     * @values
```
sarayourfriend, line 44: Is this jsdoc incomplete?

src/utils/attribution-html.ts
```diff
@@ -1,70 +1,197 @@
...
+  attrs: Record<string, string>,
+  children: string[] | null = null
+): string => {
+  const map = Object.entries(attrs)
+    .map(([key, value]) => `${key}="${value}"`)
+    .join(' ')
+  const opening = map ? `<${name} ${map}>` : `<${name}>`
+  const closing = `</${name}>`
+  return `${opening}${(children ?? []).join('\n')}${closing}`
+}
```
sarayourfriend, line 32: Love this! Really helps make the markup generation more readable :rocket:

test/unit/specs/utils/license.spec.ts
```diff
@@ -39,3 +39,24 @@ describe('getFullLicenseName', () => {
...
+    ${'by'}           | ${['cc', 'by']}
+    ${'by-sa'}        | ${['cc', 'by', 'sa']}
+    ${'by-nd'}        | ${['cc', 'by', 'nd']}
+    ${'by-nc'}        | ${['cc', 'by', 'nc']}
+    ${'by-nc-sa'}     | ${['cc', 'by', 'nc', 'sa']}
+    ${'by-nc-nd'}     | ${['cc', 'by', 'nc', 'nd']}
+    ${'nc-sampling+'} | ${['cc', 'nc', 'sampling-plus']}
+    ${'sampling+'}    | ${['cc', 'sampling-plus']}
+    ${'pdm'}          | ${['pd']}
+    ${'cc0'}          | ${['cc', 'zero']}
```
sarayourfriend, line 55: If this is such a 1:1 mapping I wonder if we shouldn't just create a `Record<License, LicensePart[]>` object. We're basically already hand-writing that mapping for the tests anyway (which we could safely remove if it was just a frozen configuration object :wink: )

Up to you though!

sarayourfriend (commented):
Thanks for paring down this PR! I'll do a functionality review tomorrow, but for now I've just taken a look over the code. It looks good! I anticipate approving tomorrow :slightly_smiling_face:

test/unit/specs/utils/license.spec.ts
dhruvkb: That's a valid point, but for now I want to keep the code lean and tests verbose. Can revisit this in the future though, especially if we add licenses that don't follow this pattern.

src/components/License/VLicenseElements.vue
dhruvkb: Yeah, had to specify the options after `@values`.

dhruvkb:
Resolving conflicts introduced by #1233.

test/unit/specs/utils/license.spec.ts
sarayourfriend: I'd probably describe an object mapping one to the other more lean than the current code which has a few branches of conditional logic.

But totally up to you if you prefer the imperative approach to a declarative one for this :+1:

src/components/VMediaInfo/VCopyLicense.vue
```diff
@@ -34,91 +34,26 @@
...
-                }}
-              </template>
-              <template v-if="licenseUrl" #URL>
-                {{ licenseUrl }}
-              </template>
-            </i18n>
-          </template>
-        </i18n>
+
+        <div v-if="tab === 'plain'">
```
sarayourfriend, line 54: This is missing the `id` attribute used by the copy button for grabbing the copyable text. The "rich text" one is missing it as well. This breaks the copy button for this PR.

When you fix this, would you mind adding an e2e test for this behavior, just to check that copying the text works for each of the tabs?

sarayourfriend (changes requested):
Noticed a regression in the attribution copying, but otherwise looks good!

src/components/VMediaInfo/VCopyLicense.vue
dhruvkb: Good catch, I'll fix it right away and add the tests.

sarayourfriend (approved):
LGTM! I do personally prefer for test files to have at least one root describe block to help give context in the test output but I don't know if that preference is shared among the team, so feel free to ignore.

Looks great though!

PR #1233, opened 2022-04-05
Add types to basic components

obulat:
## Fixes

We don't have issues for component typescriptification yet.

## Description

This PR TS-ifies base components: `VButton`, `VIcon`, `VLicenseElement`, `VLicense`. 
It also extracts some commonly used data such as license elements and icon mappings to a separate file for easier reuse.
Some button and icon props are also extracted to `Types.ts` file, but I'm not sure that's the best placement. Should it be in the constants?

## Testing Instructions

Icons, buttons and licenses should work as usual in the running app.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/Button.Types.ts
```diff
@@ -0,0 +1,14 @@
...
+  'action-menu',
+  'action-menu-secondary',
+  'action-menu-muted',
+  'plain',
+  'plain-dangerous',
+  'full',
+] as const
+export type ButtonVariant = typeof buttonVariants[number]
+export const buttonTypes = ['button', 'submit', 'reset'] as const
+export type ButtonType = typeof buttonTypes[number]
```
sarayourfriend, line 14: I think I'd prefer these to be defined in the component SFC itself, otherwise it has the effect you pointed out a while ago (at least I think it was you, might have been @krysal maybe?) that it splits the prop type information from the place where you'd need it most, in the component code itself.

Though I guess with TS you can find it more easily than before, either in the auto-complete type information or with "go to definition" in an editor.

If you decide to leave it in separate files (which is fine with me in the end) can we exclude `*.types.ts` from the title case naming ESLint rule so that we don't have the weird `*.Types.ts` situation :sweat_smile: 

It should just be a matter of updating [this line](https://github.com/WordPress/openverse-frontend/blob/3922733043c1616944934f3997f15ca14f5f28b4/src/components/.eslintrc.js#L9) to include TypeScript files.

src/components/Icon.Types.ts
```diff
@@ -0,0 +1,8 @@
+export type IconPath = object | string
```
sarayourfriend, line 1: TypeScript's `object` type has some very weird properties.

```ts
export type IconPath = object | string
const iconPath: IconPath = new Number(1)
```

Is valid, for example, because `new Number` produces an object :scream: [Here's a good StackOverflow clarifying the difference between them](https://stackoverflow.com/questions/49464634/difference-between-object-and-object-in-typescript) (which I have to reference constantly because I cannot keep them straight in my head!)

In this case we need to indicate that it can be the value of an `svg` import, right?

https://github.com/WordPress/openverse-frontend/blob/4552c37e4f8f0f2f17eb46b3d79073b39c3844bb/typings/openverse/index.d.ts#L1-L9

All `svg` imports are typed as `string` so I think just the string type should be okay. If the type definition there is wrong though, and there's something more specific, then we can change the definition to be more accurate, but I don't see much value in it as we're likely not going to, and probably shouldn't, work on those values directly.

src/components/License/VLicense.vue
```diff
@@ -72,37 +75,30 @@ export default {
     const isCcLicense = computed(() => isCc(props.license))
 
     const icons = computed(() => {
-      let iconList = props.license.split(/[-\s]/)
+      let iconList: (LicenseElement | 'ccLogo')[] = props.license.split(
+        /[-\s]/
+      ) as LicenseElement[]
       if (isCcLicense.value) iconList = ['ccLogo', ...iconList]
       return iconList
```
sarayourfriend, line 82: This mutable array reference pains me! I think we can get rid of some of the painful type juggling here by treating `iconList` as a `const`, if you're interested:

```
      const iconList = props.license.split(/[-\s]/) as LicenseElement[]
      if (isCcLicense.value) return ['ccLogo', ...iconList] as const
      return iconList
```

Take it or leave it though :slightly_smiling_face:

src/components/VIconButton/VIconButton.vue
```diff
@@ -15,13 +15,33 @@
...
+<script lang="ts">
+import { computed, defineComponent, PropType } from '@nuxtjs/composition-api'
 
 import VIcon from '~/components/VIcon/VIcon.vue'
 import VButton from '~/components/VButton.vue'
+import type { ButtonType } from '~/components/Button.Types'
+import type { IconProps } from '~/components/Icon.Types'
 
-export default {
+const SIZE_MAP = {
```
sarayourfriend, line 26: nit: Could be good to deep freeze this.

sarayourfriend (approved):
LGTM! Thanks for carrying this process along :tada: And good work avoiding runtime changes too :slightly_smiling_face: It can make it so much easier and more confident reviewing TSificiation PRs.

src/components/Button.Types.ts
obulat: Yes, it was me complaining about not understanding what props a component gets when the props are in a different file :) For `buttonTypes`, I had to extract it because the types are also used by `VIconButton`. I will check the `buttonVariants`, and move them back to SFC if they are not used anywhere else.

Thank you so much for linking the line in ESLint config, I tried to find it this morning, but couldn't :)

sarayourfriend: Is it possible to import types from a vue SFC? Given that `VIconButton` already imports from `VButton`, if we export the `types` type from the button component itself it'd be convenient (and help localize the types) to get it from the same place as the component.

obulat: Done! This is much better!

src/components/Icon.Types.ts
obulat: This was done to mimic the check for `isTest` in the Vue component props validation to make sure that tests (that for some reason receive `iconPath` as an object, not as a string) don't break. Now I see that if we declare the type in the interface as a `string`, and keep the `isTest ? object : string` check in the prop validator, it all works as it should! So, I simply left the TS type as `string`.
Thank you for the explanation! I am slowly learning about module typings as well.

PR #1239, opened 2022-04-05
Fix command for code formatting 

krysal:
Follow up of #1207 by @dhruvkb 

## Description

The `pnpm lint:fix` command was not applying format corrections, it was just showing the errors and warnings as in `pnpm lint`. Here I added new commands to keep the style of the cleanup 🙂 Let me know if it isn't too much, maybe we can omit `lint:prettier`.

## Testing Instructions

1. In `main` make any code style modifications, it could be something simple like adding several new lines in a component template, then run `pnpm lint:fix` and observe the code doesn´t change.
2. Check out this branch and repeat, the code should be formatted now.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


package.json
```diff
@@ -38,8 +38,10 @@
     "test:playwright:update-tapes": "UPDATE_TAPES=true pnpm test:playwright",
     "test:playwright:gen": "playwright codegen localhost:8443/",
     "types": "vue-tsc --noEmit",
-    "lint": "eslint --ext .js,.vue,.ts --ignore-path .gitignore --ignore-path .eslintignore . && prettier --check .",
-    "lint:fix": "pnpm lint -- --fix && prettier --write --loglevel=warn .",
```
dhruvkb, line 42: Maybe replacing `&&` with `;` would have sufficed? That would allow both commands to run regardless if one failed.

krysal: It certainly continues even if it fails but that way does not apply eslint fixes and shows error as output that prettier fixed, so it's a bit confusing/annoying because one might think there are still errors when is not the case.

obulat (approved):
Thank you for fixing it! I got a bit confused by `unknown flag --fix received` error when running `lint:fix` previously.

package.json
sarayourfriend: Would reversing the order and running prettier first _then_ eslint also fix this problem without adding new scripts?

In part of Dhruv's PR we discussed trying to pair down the scripts in package.json to ones that are truly useful (mostly reusable), otherwise it adds a lot of noise. Individual eslint and prettier scripts were something that ended up being avoided because you never want to run one without the other.

krysal: @sarayourfriend that works! Done in [`35b8a19`](https://github.com/WordPress/openverse-frontend/pull/1239/commits/35b8a199e723956fc3d5e5ff370e68b333e81173). I don't mind much two extra lines of scripts and this solution execute prettier check two times but it's not a big burden either 🤷🏻‍♀️

krysal: @sarayourfriend that works! Done in [`35b8a19`](https://github.com/WordPress/openverse-frontend/pull/1239/commits/35b8a199e723956fc3d5e5ff370e68b333e81173). I don't mind much two extra lines of scripts and this solution execute prettier check two times but it's not a big burden either 🤷🏻‍♀️

package.json
```diff
@@ -38,8 +38,8 @@
...
+    "lint": "prettier --check .; eslint --ext .js,.vue,.ts --ignore-path .gitignore --ignore-path .eslintignore .",
+    "lint:fix": "prettier --write --loglevel=warn . && pnpm lint -- --fix",
```
dhruvkb, line 42: We should also change this to a `;`, just to make sure both parts run indepedently. Other than that, this was much needed.
```suggestion
    "lint:fix": "prettier --write --loglevel=warn .; pnpm lint -- --fix",
```

sarayourfriend (approved):
LGTM! Running prettier twice in the `fix` script is annoying but not that bad overall! I can live with :slightly_smiling_face:

PR #1241, opened 2022-04-06
Fix unruly URLs in source page table for small viewports

### LINKED ISSUE #1223: Broken Table styling on small device
## Description
Broken Table styling on iPad and mobile viewport size

## Reproduction
1. Visit this following Link :https://wordpress.org/openverse/sources
2. check this responsive device
3. See issue.

## Screenshots
https://prnt.sc/X-n2gCU1wUHE

## Environment
 - Device: Laptop, iPad, mobile
 - Browser: google Chrome

## Resolution

sarayourfriend:
## Fixes

Fixes #1223 by @umeshggupta

## Description

At first I just truncated the source URLs. But on narrow screens that means that all you would see usually was the protocol and the first couple letters of the URL.

So I added some logic to make the URLs more presentable, stripping protocol, leading www and trailing slashes. This does _not_ update the actual linked URL, just the text representing it.

I also added snapshot tests for this, so it should make it easier to review.

## Testing Instructions

Checkout the branch locally and run the app with `pnpm dev`, then visit `/sources` and verify it looks good.

Alternatively review the new snapshots which essentially do the same thing :slightly_smiling_face: 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


bin/playwright.sh
```diff
@@ -10,4 +10,4 @@ export PLAYWRIGHT_VERSION=$(version)
 
 echo Running Playwright v$PLAYWRIGHT_VERSION as $USER_ID with Playwright arguments $PLAYWRIGHT_ARGS
 
-docker-compose -f docker-compose.playwright.yml up --abort-on-container-exit --exit-code-from playwright
+docker-compose -f docker-compose.playwright.yml up --build --force-recreate --abort-on-container-exit --exit-code-from playwright
```
krysal, line 13: I understand we need to rebuild the container because of the bump in playwright but won't this build the container every time the tests are run? 🚨 It is _not_ a small image.

test/playwright/utils/page.ts
```diff
@@ -0,0 +1,13 @@
+import type { Page } from '@playwright/test'
+
+export const removeHiddenOverflow = async (page: Page) => {
```
krysal, line 3: What is the reason for this?

krysal (commented):
Looks great on mobile! It was surprisingly an easy fix with the `truncate` Tailwind utility but I like how you improve it ✨ Nice! I'm just a little worried about test setup changes.

bin/playwright.sh
sarayourfriend: It will just cause the `Dockerfile.playwright` build to re-run but all the layers are still cached, so if the build results in a no-op then it's only a few milliseconds slower than not re-building.

The reason to do this is so that when the playwright version changes our local machines will actually use the new version. Docker-compose `up` doesn't rebuild the image when arguments to the Dockerfile change (like the PLAYWRIGHT_VERSION one).

If you try it out locally you'll find that the first time, yes it will be slow because we're getting a new playwright version, but subsequent runs are just as fast as they were before because the build is cached.

test/playwright/utils/page.ts
sarayourfriend: If you don't remove this you can't take full-page screenshots :disappointed: There's something about the overflow-hidden stuff that causes this. I don't understand why we need it, it seems related to the sticky header, but I'm not sure. @zackkrida said he might have a solution for removing it soon?

You can see the problem here: https://try.playwright.tech/?l=javascript&s=zbv3d2d

Note that even though the screenshot is for a "full page" it doesn't actually take the full page (because the overflow is hidden :upside_down_face:)

sarayourfriend:
For some reason the talkback tapes aren't being found for some of the requests on CI, but I can't reproduce this locally :cry: 

```
playwright_1  | [WebServer] 2022-04-06T15:03:21.877Z [Openverse e2e proxy] [INFO] Tape for /v1/images/?q=cat not found and recording is disabled (fallbackMode: NOT_FOUND)
playwright_1  | 2022-04-06T15:03:21.877Z [Openverse e2e proxy] [INFO] {"url":"/v1/images/?q=cat","headers":{"host":"localhost:49153","connection":"keep-alive","pragma":"no-cache","cache-control":"no-cache","sec-ch-ua":"","accept":"application/json, text/plain, */*","accept-language":"en-US","sec-ch-ua-mobile":"?0","user-agent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/101.0.4929.0 Safari/537.36","sec-ch-ua-platform":"","origin":"http://localhost:8443","sec-fetch-site":"same-site","sec-fetch-mode":"cors","sec-fetch-dest":"empty","referer":"http://localhost:8443/","accept-encoding":"gzip, deflate, br"}}
playwright_1  | [WebServer] 2022-04-06T15:03:21.880Z [Openverse e2e proxy] [INFO] Tape for /v1/audio/?q=cat not found and recording is disabled (fallbackMode: NOT_FOUND)
playwright_1  | 2022-04-06T15:03:21.880Z [Openverse e2e proxy] [INFO] {"url":"/v1/audio/?q=cat","headers":{"host":"localhost:49153","connection":"keep-alive","pragma":"no-cache","cache-control":"no-cache","sec-ch-ua":"","accept":"application/json, text/plain, */*","accept-language":"en-US","sec-ch-ua-mobile":"?0","user-agent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/101.0.4929.0 Safari/537.36","sec-ch-ua-platform":"","origin":"http://localhost:8443","sec-fetch-site":"same-site","sec-fetch-mode":"cors","sec-fetch-dest":"empty","referer":"http://localhost:8443/","accept-encoding":"gzip, deflate, br"}}
```

I have no idea why this is happening :confused: The tapes are definitely there for both requests, maybe there's some kind of header mismatch or something? But why isn't it happening locally! So confusing.

sarayourfriend:
Setting `allowHeaders: ['connection']` seems to have fixed the issue with the talkback tapes.

@krysal and @obulat please review when y'all have time :pray:

obulat (commented):
It seems that the inputs don't get focused anymore: there's no cursor line in the searchBox in the screenshots that change in this PR. Can it be related to the overlay removal?

sarayourfriend:
@obulat I think that could be a timing thing, because the cursor flashes.

We could use [caret-color](https://developer.mozilla.org/en-US/docs/Web/CSS/caret-color) to hide it completely to remove it as a variable from the e2e tests.

I'll add that to this PR, if that sounds good to you?

sarayourfriend:
Oh BTW, the fields are still focused, you can see from the pink outline on the field. I'll add the `caret-color: transparent` to the page before taking the screenshot.

obulat (approved):

src/pages/sources.vue
```diff
@@ -183,6 +183,14 @@ const SourcePage = {
...
       this.sort = { direction, field }
     },
+    cleanSourceUrlForPresentation(url) {
+      const stripProtocol = (s) => s.replace(/https?:\/\//, '')
+      const stripLeadingWww = (s) =>
+        s.startsWith('www.') ? s.replace('www.', '') : s
+      const stripTrailingSlash = (s) => (s.endsWith('/') ? s.slice(0, -1) : s)
+
+      return stripTrailingSlash(stripLeadingWww(stripProtocol(url)))
+    },
```
zackkrida, line 193: I see you planting the seeds of some FP in the codebase 😉. We'll refactor this to `pipe(stripTrailingSlash, stripLeadingWww, stripProtocol)(url)` someday!

test/playwright/utils/page.ts
zackkrida: Yes, I'll make an issue for the necessary layout refactoring today!

src/pages/sources.vue
sarayourfriend: One day!

PR #1243, opened 2022-04-06
Fix audio results not being keyboard accessible in all results view

### LINKED ISSUE #623: Audio single result page keyboard navigation is broken
## Description

It is not possible to tab out of the waveform component on the audio result page, nor does the waveform respond to keyboard events anymore for seeking or play/pause.

## Reproduction

1. Run `main` and go to http://localhost:8443/audio/632d0076-fdbd-4919-8b8f-b498ef246ee8
2. Use `tab` to reach the waveform
3. Try to "tab out" of the waveform by hitting tab again.
4. Try to seek the audio using your arrow keys.
5. Try to toggle the audio playback using the spacebar.

Note that none of 3-5 will be successful.

## Expectation

All of 3-5 should work. You should be able to tab out of the audio waveform, toggle the playback state, and seek, all from the keyboard.

## Resolution

### LINKED ISSUE #1242: Cannot open audio results with keyboard navigation on all results grid
## Description

It is not possible to open the audio results with keyboard on all results grid.

## Reproduction

1.  Go to https://search.openverse.engineering/search?q=birds
2.  Tab to the first audio result
3.  Hit enter

## Expected Results
The link should open the audio results

## Actual result
The audio plays

## Resolution

sarayourfriend:
## Fixes

Fixes #1242   by @sarayourfriend 

## Description

Makes the boxed audio result a VLink wrapped element so that when focused, it is keyboard accessibly openable and also reports the actual URL of the page that will be opened, unlike the existing solution which mimicked link behavior using JavaScript.

We will later continue to work on the audio result link by transforming it into a composite element (probably `role="grid"`) that has custom keyboard navigation to enter the link to be able to play and pause the audio.

## Testing Instructions

Check the new e2e tests pass. Also test it in your own browser and in particular check the focus styles.

I could add visual regression tests for the focus styles actually, I will do that tomorrow!

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -398,8 +398,10 @@ export default defineComponent({
     })
     const ariaLabel = computed(() =>
       isBoxed.value
-        ? i18n.t('audio-track.aria-label-interactive')
-        : i18n.t('audio-track.aria-label')
+        ? i18n.t('audio-track.aria-label-interactive', {
+            title: props.audio.title,
+          })
+        : i18n.t('audio-track.aria-label', { title: props.audio.title })
     )
     /**
      * @param {KeyboardEvent} event
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -1,6 +1,6 @@
 <template>
   <Component
-    :is="isBoxed ? 'VLink' : 'div'"
+    :is="isBoxed ? 'VLink' : 'VWarningSuppressor'"
     class="audio-track group"
     :aria-label="ariaLabel"
     role="region"
@@ -65,6 +65,7 @@ import VRowLayout from '~/components/VAudioTrack/layouts/VRowLayout.vue'
 import VBoxLayout from '~/components/VAudioTrack/layouts/VBoxLayout.vue'
 import VGlobalLayout from '~/components/VAudioTrack/layouts/VGlobalLayout.vue'
 import VLink from '~/components/VLink.vue'
+import VWarningSuppressor from '~/components/VWarningSuppressor.vue'
 
 const propTypes = {
   /**
@@ -116,6 +117,7 @@ export default defineComponent({
     VPlayPause,
     VWaveform,
     VLink,
+    VWarningSuppressor,
 
     // Layouts
     VFullLayout,
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -1,5 +1,6 @@
 <template>
-  <div
+  <Component
+    :is="isBoxed ? 'VLink' : 'div'"
     class="audio-track group"
     :aria-label="$t('audio-track.aria-label')"
     role="region"
@@ -34,7 +35,7 @@
         />
       </template>
     </Component>
-  </div>
+  </Component>
 </template>
 
 <script>
@@ -61,6 +62,7 @@ import VFullLayout from '~/components/VAudioTrack/layouts/VFullLayout.vue'
 import VRowLayout from '~/components/VAudioTrack/layouts/VRowLayout.vue'
 import VBoxLayout from '~/components/VAudioTrack/layouts/VBoxLayout.vue'
 import VGlobalLayout from '~/components/VAudioTrack/layouts/VGlobalLayout.vue'
+import VLink from '~/components/VLink.vue'
 
 const propTypes = {
   /**
@@ -111,6 +113,7 @@ export default defineComponent({
   components: {
     VPlayPause,
     VWaveform,
+    VLink,
 
     // Layouts
     VFullLayout,
@@ -385,7 +388,7 @@ export default defineComponent({
     const layoutBasedProps = computed(() => {
       if (!isBoxed.value) return {}
       return {
-        tabindex: isBoxed.value ? 0 : -1,
+        href: `/audio/${props.audio.id}`,
         class:
           'block focus:bg-white focus:border-tx focus:ring-[3px] focus:ring-pink focus:ring-offset-[3px] focus:outline-none rounded-sm overflow-hidden cursor-pointer',
       }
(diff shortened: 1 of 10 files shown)


krysal (approved):
Nice! Links work properly hitting the enter key here, there is just a little visual regression: the category and license icons show in pink. Would be nice to fix that before merging.

CleanShot 2022-04-06 at 12 55 42@2x

zackkrida (commented):
This does what you want it to, and looks good. Some notes:

- The `layoutBasedListeners` click event, which emits a `boxedAudioClick`, can be deleted
- The `layoutBasedListeners` for `keydown` events no longer works, so the spacebar no longer play/pauses the audio track

The reason that JavaScript was originally used to navigate on click and enter (which _was_ working at one point in time) was, if I remember correctly, related to NuxtLink not executing event handlers, or something like that. I felt that navigation via click+enter and play/pause via space was the best tradeoff at the time of writing this.

sarayourfriend:
> The reason that JavaScript was originally used to navigate on click and enter (which was working at one point in time) was, if I remember correctly, related to NuxtLink not executing event handlers, or something like that. I felt that navigation via click+enter and play/pause via space was the best tradeoff at the time of writing this.

That makes sense. I think prioritizing natural navigation over interactivity on the all results page achieves a "graceful degradation" of the page interaction. I don't think we should sacrifice using the native anchor element, which has accessible features that we do not want to have to recreate with JavaScript, for non-essential interactions. To preview the audio you can navigate to the result page and play the preview from there, after all.

I'll see if I can resolve the issue with NuxtLink not emitting native events, I think @obulat was able to work around this in the past using the `.native` modifier on the event listener, so I will try that.

sarayourfriend:
@zackkrida I was able to use the `.native` modifier to make it work.

sarayourfriend:
@obulat could we use a regular anchor instead of a NuxtLink in those cases?

obulat:
> @obulat could we use a regular anchor instead of a NuxtLink in those cases?

I think it would be simpler to set `prefetch=false` than to use a regular anchor. This way, the order of function calls would be as we expect. It might make the loading of the single page JS slower, but only for the first time. And because we get the data from the search results, it would probably still make the overall page load faster.

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -1,10 +1,11 @@
...
-  <div
+  <Component
+    :is="isBoxed ? 'VLink' : 'div'"
     class="audio-track group"
-    :aria-label="$t('audio-track.aria-label')"
+    :aria-label="ariaLabel"
     role="region"
     v-bind="layoutBasedProps"
-    v-on="layoutBasedListeners"
+    @keydown.native="handleKeydown"
```
obulat, line 8: We probably cannot use an event handler like this on a `div`. Besides, we would probably need to add a `tabindex` if a div has an keydown handler. Or is it a requirement only for a click handler? This is a warning I get  on a single audio page:
```
[Vue warn]: The .native modifier for v-on is only valid on components but it was used on <div>.
found in
---> <VAudioTrack> at src/components/VAudioTrack/VAudioTrack.vue
```

Maybe we can add a keydown handler with v-bind, in addition to the `layoutBasedProps`?

obulat (changes requested):

src/components/VAudioTrack/VAudioTrack.vue
sarayourfriend: I couldn't figure out how to add the `.native` modifier when using `v-on`, the event handler never got applied to the element when I tried this for example:

```js
const eventHandlers = computed(() => {
  if (!isBoxed.value) return {}

  return {
    'keydown.native': ...
  }
})
```

Is there a different way to do it? As far as I can tell modifiers aren't available for the `v-on` object syntax: https://github.com/vuejs/vue/issues/7846

sarayourfriend: Regarding the div problem, the div won't be focusable ever, there's no tabindex, so the key handler will just be ignored, right? And in those cases, indeed it _should_ be ignored as the event handler itself will just do an early return anyway.

Is there a way to suppress that vue warning? In this case I think it is safe to ignore considering that it doesn't appear to be possible to conditionally bind a `.native` modified event handler.

zackkrida:
Apologies, I pushed a suggestion accidentally, but it was a simple change to fix the text color:

CleanShot 2022-04-11 at 15 06 47@2x

zackkrida (approved):
LGTM! So glad to have this working with a real link.

obulat (changes requested):
The changes currently don't work for me 😞

When I click on the audio box, I get redirected to the page, but the audio does not start playing. When I click on the play button inside the box, the app breaks. There are errors related to resizeObserver: `TypeError: ResizeObserver.observe: Argument 1 is not an object.`.

What do you think about extracting a `VBoxedAudioTrack` component to make the click handlers work correctly?

zackkrida:
@obulat I realize that I didn't test _clicking_ the play/pause button, thank you for catching that. This should be fixed pretty straightforwardly by preventing bubbling on the play/pause click event, with the `.stop` modifier on the `src/components/VAudioTrack/VPlayPause.vue` click event, @sarayourfriend .

sarayourfriend:
Play/pause is fixed and e2e tests added for both ways of toggling audio, via keyboard and via mouse.

sarayourfriend:
@obulat can you re-review when you get a chance? Thanks :pray:

obulat (approved):
It works really well.
I initially misunderstood the functionality, and thought that when you click on the boxed audio, it should open the link and _also start playing the audio._ Now I see that it wasn't intended to do that.
One concern I still have is with the warnings. Can we do something using the Vue config: https://v2.vuejs.org/v2/api/#warnHandler ?

sarayourfriend:
Sure, let me try to filter it using the warnHandler.

sarayourfriend:
Warn handler doesn't work for it because it doesn't let you suppress warnings as far as I can tell, just do "extra stuff" with the, if you want.

sarayourfriend:
I found a fix for it though @obulat, let me know what you think.

obulat (approved):
🤣 It's a really funny solution :) Thank you for adding it!

PR #1255, opened 2022-04-11
Fix VIconButton default variant

### LINKED ISSUE #1206: Incorrect styles on reporting form close button
## Description

The content reporting form shows the incorrect button styles, with a pink background that should not be present.

## Reproduction

1. Submit a mature content report.
2. Observe the incorrect button styles

## Screenshots

CleanShot 2022-03-30 at 09 39 53@2x

sarayourfriend:
## Fixes

Fixes #1206  by @zackkrida 

## Description

Fixes the problem in the linked issue by fixing the root cause in the VIconButton component. Also does some TypeScript stuff to make the VButton easier to use with TypeScript and able to type things like VIconButton that "spread props" to VButton.

## Testing Instructions

Check out the VIconButton implementations in the app and ensure they look good:

* PlayPause button
* Content report form close button
* License explanation tooltip in the filter list
* Global audio player close button

I could add visual-regression tests for each of these but in the interest of expediency I only added it for the particular button in question.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VButton.vue
```diff
@@ -95,7 +101,7 @@ const VButton = defineComponent({
     as: {
       type: String as PropType<ButtonForm>,
       default: 'button',
-      validate: (val: ButtonForm) => buttonForms.includes(val),
+      validate: (val: unknown) => buttonForms.includes(val as ButtonForm),
```
obulat, line 104: TIL!

sarayourfriend: I only changed this because I was trying to get the `ExtractPropTypes` to work correctly with props that have default values. I'll set it back, there's no functional or operational difference either way.

src/components/VAudioTrack/VPlayPause.vue
```diff
@@ -36,22 +36,20 @@ export default defineComponent({
      * the current play status of the audio
      */
     status: {
-      type: String,
-      validator: (val) => ['playing', 'paused', 'played'].includes(val),
+      type: String as PropType<'playing' | 'paused' | 'played'>,
```
obulat, line 39: We could use the type from active media store here: 
https://github.com/WordPress/openverse-frontend/blob/bccbd1333fd0deb51fdd90a391d65e4666caf554/src/stores/active-media.ts#L7

src/components/VButton.vue
sarayourfriend: (Though the `unknown` parameter type is _technically_ more correct).

sarayourfriend: It just doesn't matter because we don't call `validate` directly so the type checking in this function is basically a "whatever" kind of thing.

src/components/VAudioTrack/VPlayPause.vue
sarayourfriend: It's a different status, this one has "played" instead of "ejected". IIRC there _is_ some distinction, perhaps one for when the player is closed ("ejected") and the other for when the audio has simply finished playing to the end (in which case we show the replay button).

obulat: Oh, thank you for explaining :)

obulat (approved):
The buttons' colors are fixed. There are some size inconsistencies with the mockups:

The close button on the License Popover is the correct size (40x40px), but it probably should be right at the top of the popover (without any padding/margin).
The close button on the content report form is too big (48x48 instead of 40x40)

Do you think we should open a new issue for that?

sarayourfriend:
@obulat I can fix those here.

For the button position, I moved it because it looked awkward and wrong for it to not be vertically aligned with header text of the popover. cc @panchovm on that one.

panchovm:
In both the license Popover and the content report form, I see `48px` as size for the `button` element. Is that for a reason? I am using a `24px` size in the mockups to match other similar buttons. 

Regarding the close button position in popovers, I have pending reviewing the popover style. Some cases have more content than others, and the vertical alignment can blend with the popover's title area. Here is an example.

CleanShot 2022-04-14 at 14 19 05@2x

For now, I would keep it top-right aligned with a top and right margin of `8px`. 

Also, notice that the resting style of the close button is `dark-charcoal` instead of `dark-charcoal-70`. I can create a ticket to add the state styles that interactive components have.

obulat:
I think that the close buttons should be at least 40x40px, because on the mobile phones it's very difficult to click on buttons that are smaller than 40px. This is also an accessibility/SEO requirement that is checked by Google Lighthouse.
If you have a margin of 8px, the 24x24px button looks like it's a 40x40px button. We can have a button that is 40x40px right in the corner of the popover, that looks exactly as the designs, don't you think, @panchovm ?

src/components/VButton.vue
```diff
@@ -63,6 +65,10 @@ const buttonTypes = ['button', 'submit', 'reset'] as const
 
 export type ButtonType = typeof buttonTypes[number]
 
+export type ButtonProps = ProperlyExtractPropTypes<
+  NonNullable<typeof VButton['props']>
+>
```
krysal, line 70: Shouldn't this be after the declaration of `VButton`?

krysal (approved):
LGTM!

panchovm:
The [target size criterion](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html) requires a `48px` size, but that is for AAA level. For AA it seems not necessary. Nonetheless, you are right about the size being too small. We already have a `close` [component](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=851%3A6222) that has a `32px` for the smallest size. We might go with that.

What do you think?

Issue #1212, opened 2022-03-31
Pages menu is cut off in mobile landscape view


Asadpalasara-multidots:
## Description
Some of Menu items not visible.

## Reproduction
1. Visit: https://wordpress.org/openverse/
2. Search anything using images search option in search box.
3. Change the screen viewport to the mobile devices.
4. Click on Button(Dots) https://prnt.sc/HutwN50nGtmO
5. Menu items showing till "API"  then rest of items not visible.

## Screenshots
https://prnt.sc/aUuprxKjkp2m

## Environment
 - Device: iPhone 12 Pro Landscape mode
 - Browser: Chrome

krysal:
Thanks for reporting this issue with all details @Asadpalasara-multidots. We'll work on general improvements for the whole header. CC @panchovm.

panchovm:
I can not reproduce the bug. Once reaching step 3 to see mobile views, the header changes and shows secondary menu's items inside the content switcher modal. See [this gif](https://cln.sh/6fD1Vk).

krysal:
@panchovm Try in an actual phone, not the browser emulator, that gives you the limited height that prevents seeing the whole dropdown menu, as @Asadpalasara-multidots shown in the screenshot.

sarayourfriend:
Changing the horizontal breakpoint probably doesn't fix the problem. If you have a wide but short device (like a horizontal smartphone, or a browser window aligned to the top half of your screen) this still happens.

We could make the inside of the menu scrollable?

panchovm:
Most devices' aspect ratios display the content in a proper layout, so the challenge seems to be picking a breakpoint where the navigation adapts correctly, as both navigational elements and search results, whether it is landscape or portrait. On desktop, or where you can control the browser size, there will always be cases of cutting the UI due to window size.

The scrollable menu sounds reasonable for this case, but I can not envision a smooth interaction in other situations unless you are thinking of special conditions for this behavior.

zackkrida:
The core issue, to my mind, is having a sticky header combined with flyout/dropdown menus. This means that the height of the dropdowns can never exceed the user's screen height, which is a difficult thing to always guarantee. We'd have to find a way to use javascript to make sure the dropdowns are scrollable and never extend past the bottom edge of the viewport.

sarayourfriend:
[...]

PR #1257, opened 2022-04-11
Fix app layout issues

### LINKED ISSUE #845: Header on search page not displayed on IPad initially
## Description

When searching on an IPad, the header is not displayed at first. You have to scroll up twice to get to see it:
https://user-images.githubusercontent.com/15233243/153350152-3baea109-02f0-41cd-a698-5f5563949728.mp4

## Reproduction

1. Search for something from the homepage.
2. See error: There appears to be no header on the page. If you scroll, the header doesn't appear, and therefore it's not fixed. 

## Expectation

The header should be seen on the search page at all times

## Environment

 - Device: IPad 
 - OS: 
 - Browser: Chrome
 - Version: 
 - Other info: landscape mode (it's fine in portrait)

## Resolution

### LINKED ISSUE #1043: Header does not fit on phones
## Description

The end-side of the header is cut off on smaller widths such as phones.

## Reproduction

1. Visit Openverse from a smaller width device.
2. See error.

## Expectation

The Search bar could possibly shrink to accommodate the content.

## Screenshots

IMG_5935

## Environment

 - Device: iPhone Xr 
 - OS: iOS 15.3 
 - Browser: Safari 
 - Version: 15.3 
 - Other info: viewport width = 414px 

## Resolution

### LINKED ISSUE #1060: Mobile scrolling problem
## Description

Mobile web sites commonly use scroll detection on the `<body>` element to hide/toggle ui elements. For example, mobile Firefox on Android uses this to hide/show the address bar in the footer. 

Because we're _not_ scrolling the body, and instead scrolling an internal element for the search results area, the browser never detects user scroll and never hides the the address bar. 

## Reproduction

1. Go to search.openverse.engineering on an android device in Firefox
2. Make a search and scroll the page
3. Observe that the browser bottom addressbar never hides itself
4. Bonus: Play an audio file and observe the global audio player is partially obscured behind the address bar!

The fix for this would be to reassess our global layout. Some potential solutions:

1. On mobile only, enable body overflow scroll and make the header sticky with a top position of zero
2. Revise the entire layout with a truly sticky header, and sticky or fixed position for the filters sidebar. The main content area will need a hardcoded padding or margin end to prevent the sidebar from covering the page contents.

### LINKED ISSUE #1214: Filters Icon Cutting Off
## Description
Once Page scroll and header is sticky, i can See Filters Icon Cutting Off.

## Reproduction
1. Visit: https://wordpress.org/openverse/
2. Search anything using images search option in search box.
3. Change the screen viewport to the mobile devices.
4. You will notice that the filters icon cutting Off.

## Screenshots
https://prnt.sc/Y6JECzM8GokE

## Environment
 - Device: iPhone 12 Pro
 - Browser: Chrome

zackkrida:
## Fixes

Fixes #1214
Fixes #1060
Fixes #1043 
Fixes #845

## Description

Switches Openverse's frontend to use a truly-sticky header, an actual fixed sidebar, and removes code to hide overflow in parent containers. 

Also fixes the mobile header issues by adding `w-full` to the search input, which overrides the automatic `size="20"` applied to unsized inputs!

## Testing Instructions

Visit the site and observe the filter sidebar, mobile filter popup, and mobile header. If you're on MacOS please try with different scrollbar settings:

CleanShot 2022-04-14 at 17 16 23@2x

You can search 'scrollbar' in system preferences to find those controls. Additionally, _please_ test on mobile devices, the smallest you can find, to make sure that all header elements fit within the viewport without any icons cutting off.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
@WordPress/openverse-frontend please review this when you can. The one thing I haven't fixed here, that I would love someone to look into, is fixing the header scroll detection code which is used to change the header layout on mobile.

src/components/VFilters/VSearchGridFilter.vue
```diff
@@ -1,8 +1,5 @@
 <template>
-  <div
-    class="filters py-8 px-10 min-h-full md:bg-dark-charcoal-06"
-    data-testid="filters-list"
-  >
+  <div class="filters py-8 px-10" data-testid="filters-list">
```
zackkrida, line 2: Styles specific to the search filter _sidebar_ have been moved up in the component tree.

src/layouts/default.vue
```diff
@@ -1,17 +1,16 @@
...
     </div>
-    <main
-      class="main embedded overflow-x-hidden"
-      :class="{ 'has-sidebar': isSidebarVisible }"
-    >
+    <main class="main embedded" :class="{ 'has-sidebar': isSidebarVisible }">
       <Nuxt ref="mainContentRef" class="min-w-0 main-page" />
-      <VSidebarTarget class="sidebar" />
+      <VSidebarTarget
+        class="sidebar fixed pb-20 right-0 [calc(100%-81px)] bg-dark-charcoal-06 border-s border-dark-charcoal-20 overflow-y-scroll"
```
zackkrida, line 12: I don't _love_ that we need to know the height of the desktop header here, but it's the only way to do it.

src/layouts/default.vue
```diff
@@ -1,17 +1,16 @@
...
     </div>
-    <main
-      class="main embedded overflow-x-hidden"
-      :class="{ 'has-sidebar': isSidebarVisible }"
-    >
+    <main class="main embedded" :class="{ 'has-sidebar': isSidebarVisible }">
       <Nuxt ref="mainContentRef" class="min-w-0 main-page" />
-      <VSidebarTarget class="sidebar" />
+      <VSidebarTarget
+        class="sidebar fixed pb-20 right-0 h-[calc(100%-81px)] bg-dark-charcoal-06 border-s border-dark-charcoal-20 overflow-y-scroll"
```
zackkrida, line 12: I hate needing to know the header height here (81px), but it's the only way to do this.

obulat: What about the banners? They currently break the calculations

sarayourfriend: CSS variables and calc outside of storybook would probably be a better fit than this kind of thing.

src/components/VHeader/VSidebarContent.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <VTeleport v-if="visible" :to="to"
-    ><VSearchGridFilter @close="$emit('close')"
-  /></VTeleport>
+  <VTeleport v-if="visible" :to="to">
```
obulat, line 2: 🔥

src/layouts/default.vue
obulat: Did you mean outside of _tailwind_, not storybook, @sarayourfriend ? 
If so, I agree with you :)

sarayourfriend: Sorry, yes tailwind, not storybook! :sweat_smile:

sarayourfriend:
Something here has broken the body scroll lock for the filters modal on mobile (but not the search type switcher :thinking:)

sarayourfriend (commented):
This is exciting!

Couple of issues I'm noticing:

1. Filter sidebar has a scrollbar on desktop widths when the translation banner is visible
2. When both banners are open the header is _so big_ it obscures almost half the screen on mobile devices
Screen Shot 2022-04-15 at 06 59 22
3. There is horizontal scroll on smaller viewports in the filter and search type modals
4. On smaller viewports the header wraps, I'm not sure what the right solution there is, but maybe the search type switcher button isn't truncating as expected? :thinking: 
Screen Shot 2022-04-15 at 07 01 04
5. The mobile global audio player is behaving quite strangely. It's not actually meant to show on the single result page anyway. Some of this may be caused by mysterious horizontal scrolling on the audio single results page but other aspects of it (like the global audio player not sticking to the bottom of the screen) are unique to this branch.
Screen Shot 2022-04-15 at 07 02 38

zackkrida:
Great review @sarayourfriend, thank you! I think I can fix this stuff 😄

obulat:
We can probably fix # 4 in Sara's review if we update the `xs` breakpoint in `SCREEN_SIZES` to 360[px], because the header wraps when it's between 340 and 360 px.

obulat (changes requested):
Could you also add 
```
      window.scrollTo({
        top: 0,
        left: 0,
        behavior: 'smooth',
      })
```
(instead of `this.$el.parentElement.scrollTo`) in the `VScrollButton` ?

src/layouts/default.vue
```diff
@@ -1,17 +1,16 @@
...
-    <main
-      class="main embedded overflow-x-hidden"
-      :class="{ 'has-sidebar': isSidebarVisible }"
-    >
-      <Nuxt ref="mainContentRef" class="min-w-0 main-page" />
-      <VSidebarTarget class="sidebar" />
+    <main class="main embedded" :class="{ 'has-sidebar': isSidebarVisible }">
+      <Nuxt class="min-w-0 main-page" />
+      <VSidebarTarget
+        class="sidebar fixed pb-20 end-0 bg-dark-charcoal-06 border-s border-dark-charcoal-20 overflow-y-auto"
```
obulat, line 12: The border color should be conditional here, something like:
`:class={ 'border-dark-charcoal-20': isSidebarVisible }`. Otherwise, you can see a gray border even when the sidebar is closed.

src/layouts/default.vue
```diff
@@ -92,14 +87,20 @@ const embeddedPage = {
...
 }
 export default embeddedPage
 </script>
+
 <style scoped>
+.sidebar {
+  height: calc(100vh - var(--header-height, 81px));
+}
+.has-sidebar .sidebar {
+  width: 325px;
```
obulat, line 101: The sidebar width in the mockups is 316px. It keeps changing in the code, for some reason :)

src/layouts/default.vue
```diff
@@ -92,14 +87,20 @@ const embeddedPage = {
...
-      mainContentRef,
-      mainRef,
     }
   },
 }
 export default embeddedPage
 </script>
+
 <style scoped>
+.sidebar {
```
obulat, line 97: Can we add the CSS variables for banner height here?

obulat (changes requested):
Great changes! I've left some comments inline.

src/layouts/default.vue
zackkrida: Done!

src/layouts/default.vue
zackkrida: I've made this a global css var so it is consistent everywhere!

src/layouts/default.vue
zackkrida: Using CSS variables instead of inline tailwind.

src/layouts/default.vue
zackkrida: It actually shouldn't need those to work correctly anymore, but if you notice that we need them in testing please let me know.

zackkrida:
@WordPress/openverse-frontend I think this is ready for re-review! I've merged in @krysal's additions. I've reduced and made consistent the size, line height, and font size of the two banners on mobile. I am not seeing any unnecessary horizontal or vertical scrollbars, if you _are_ please let me know in which browser.

CleanShot 2022-04-24 at 10 54 55@2x

Side note: I'm not sure in userland if folks will ever see the two banners active at once. Any search.creativecommons.org url will always redirect to the english site at https://wordpress.org/openverse.

src/components/VNotificationBanner.vue
```diff
@@ -84,5 +84,6 @@ export default defineComponent({
 
 .announcement {
   @apply bg-trans-blue text-white;
+  border-left: 0.25rem solid transparent;
```
zackkrida, line 87: this keeps the text aligned when both banners are present.

sarayourfriend:
> Side note: I'm not sure in userland if folks will ever see the two banners active at once. Any search.creativecommons.org url will always redirect to the english site at https://wordpress.org/openverse.

Good to know! That had not occurred to me at all.

sarayourfriend (commented):
This is looking so good. I've found a few more bugs, I suspect they won't be too bad to solve.

1. The header z-index clashes with the "back to search results" z-index.
image
2. The filter sidebar jumps a bit and has a small amount of whitespace after it on desktop
https://user-images.githubusercontent.com/24264157/164990127-d44eaffb-2b1a-428d-8869-6571fc64133d.mp4
3. The modal scroll lock doesn't appear to work anymore, visible on small width viewports:
https://user-images.githubusercontent.com/24264157/164990143-06519521-55df-4136-afdf-3c8bbfba51ed.mp4
4. The header changes height when scrolling on the single results page:
https://user-images.githubusercontent.com/24264157/164990269-51aa59c0-2931-4c3d-b234-ce1b3b532f40.mp4

obulat:
> 4\. The header changes height when scrolling on the single results page:
>         https://user-images.githubusercontent.com/24264157/164990269-51aa59c0-2931-4c3d-b234-ce1b3b532f40.mp4

The heights of scrolled/ non-scrolled headers are different because they have different paddings in the mockups ( I have forgot to add that detail when talking about the CSS variable for header height!).

The resting header has 16px paddings: https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=352%3A443
And the scrolled header has only 12px paddings: https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=352%3A544

Would this mean we need to set the CSS variables in JS?

src/layouts/default.vue
```diff
@@ -86,14 +78,20 @@ const embeddedPage = {
...
+  height: calc(100vh - var(--header-height, 81px));
```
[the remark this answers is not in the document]
obulat: Considering that the header height can vary based on the scrolled/non-scrolled status (which I remembered very late in the process of reviewing, sorry!), I think we should define `header-height` in `setup`.

src/styles/tailwind.css
```diff
@@ -3,6 +3,10 @@
 @tailwind utilities;
 
 @layer base {
+  :root {
```
obulat, line 6: Nice!

obulat (changes requested):
I've added the POT file and removed z-index from back to search results.
The only blocker I can see here is the header height: it needs to be smaller (72px + 1px border) when the header is scrolled.

sarayourfriend:
> Would this mean we need to set the CSS variables in JS?

Yes, I think that's right. If it could be done fully in CSS somehow that would be excellent because back-forward navigation does preserve scroll position, but it doesn't seem that there's a way to detect scroll position in CSS.

src/layouts/default.vue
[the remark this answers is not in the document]
sarayourfriend: It'd be good to define in CSS and update it in JS (otherwise it'll cause more jumpiness :slightly_frowning_face:)

obulat:
> > Would this mean we need to set the CSS variables in JS?
> 
> Yes, I think that's right. If it could be done fully in CSS somehow that would be excellent because back-forward navigation does preserve scroll position, but it doesn't seem that there's a way to detect scroll position in CSS.

We could have a class or a data attribute added by JS to an outer element when the header is scrolled, and have CSS use that...

sarayourfriend:
We could, but that still more or less just relies on JS as much as the other solution :sweat_smile:

Let's fix it in the basic sense first and then we can see if the back/forwards is actually an issue.

zackkrida:
@obulat is this ready for review, or are their still issues? I'm happy to work on it, in my testing everything appears to be working correctly so I'm not sure what's needed yet.

sarayourfriend:
The gap at the bottom of the filter sidebar is fixed, but the header changes sizes on scroll for seemingly no reason on desktop :thinking: It looks like the only thing that changes is the border on the search type switcher?

sarayourfriend:
Mobile search results scrolling and header also has some pretty strange interactions... :thinking:

sarayourfriend (commented):
I'm confused why those `xs` screenshots of the attribution tabs changed width :thinking:

zackkrida:
I'm seeing this branch quite broken on mobile, and am investigating today.

zackkrida:
I believe modal interactions are the only thing left to fix here, I'm investigating today. 

I have made the header the same height, regardless of scrolling. It's a much better experience with less erratic movement and only a 10px or so difference. The only consequence of this is that the sticky header is a bit taller (cc @panchovm).

zackkrida:
@sarayourfriend I think named layers are a good idea. We could create an eslint rule that disallows `z-{1-10}` classes. If someone is using a `z-[]` JIT class I think that's enough of a code smell that reviewers would catch it, but we could also write a lint rule for those if needed.

Please go ahead and create an issue. I'd like to do that work but don't want to add even more to this already critical, deployment blocking PR. The z indices are currently pretty 'healthy', we're using z-10,z-20,z-30,z-40 and z-50 all quite reasonably. No hacks or anything yet. It would still be nice to get the named values in soon in case this changes!

zackkrida:
I discovered that body scroll locking was not working for the mobile filter menu, but we didn't notice in production because the body isn't actually scrollable in the production site, rather the direct children of the `main` element were. Because this PR makes the body truly scrollable, it revealed the existing bug.

obulat:
Looking through the failing snapshots in the `test-results` artifact:
Screen Shot 2022-05-10 at 6 37 59 AM

- Several `scrolled` snapshots expectedly fail because everything is pushed down by the amount of extra padding when scrolled.
- RTL media reuse snapshots seem to not capture all of the height, I'll investigate it. I'll also 'dismiss' the RTL translation banner for them.

I'll try fixing these problems, update the snapshots, and try running the tests again :)

obulat:
> I'm confused why those `xs` screenshots of the attribution tabs changed width 🤔

I think that it was caused by the `main` being wider than the screen width, and it was overflowing horizontally. At least, when I set main to use `class="w-screen"`, all of the component started fitting into the `xs` screen, and it seems nothing else broke.

src/layouts/default.vue
```diff
@@ -80,20 +76,36 @@ const embeddedPage = {
         isSearchRoute.value && !isHeaderScrolled.value && !isMinScreenMd.value
     )
     provide('headerHasTwoRows', headerHasTwoRows)
+
```
obulat, line 79: We should probably remove this CSS variable if we removed the scrolled/non-scrolled padding change.

obulat:
The changes in the latest PRs:
- the SearchBar search button was set not to shrink. This stops the search button from being too narrow on small screens.
- `main` width is set to 100vh (Tailwind `w-screen` class). This stops the app from being too wide for the `xs` screens. This fixes the media reuse visual regression tests, and also makes sure that everything fits into the screen on `xs` breakpoint.
-  Mobile header vertical padding is set to the smaller `py-3` value to match the mockups.

I think this PR is ready to be merged 🚀

obulat (approved):

zackkrida:
@obulat 

> main width is set to 100vh (Tailwind w-screen class). This stops the app from being too wide for the xs screens. This fixes the media reuse visual regression tests, and also makes sure that everything fits into the screen on xs breakpoint.

`w-screen` does not account for the width of a visible scrollbar, so it adds horizontal scrolling when scrollbars are set to visible, like on Windows.

Trying `w-full` instead, I believe the tests still pass.

sarayourfriend:
> As far as I know there are no mobile browsers which display vertical scrollbars. There's likely a child misbehaving that's causing the expansion on small sizes that I'll also investigate if time permits.

Note that that's just a stopgap. I regularly use Openverse under the `md` breakpoint on my laptop (with visible scrollbars).

zackkrida:
@sarayourfriend a full, non stop-gap fix for this is applied, and the Playwright tests are passing

krysal (approved):
Such a huge improvement here. I love that the header doesn't make two lines when is scrolled on mobile anymore! :)

#1212 is still an unresolved issue, but that could be addressed apart. Great work here!

zackkrida:
Ah you're right @krysal #1212 isn't related to this PR. Thanks for the review!

PR #1270, opened 2022-04-14
Create a proof-of-concept for feature flags

dhruvkb:
## Fixes

Implements WordPress/openverse#186

## Description

This PR creates a PoC for working with feature flags inside the Openverse frontend. This code adds
- feature flag constants and types for TS
- a feature flag store for storing flag statuses and computing feature states
- middleware to init the store from cookies
- a makeshift preference page to update the cookies based on preferences and to demo the functionality (will be deleted before this is merged and replaced with a solid framework)

## Testing Instructions

1. Visit /preferences and see how the preferences are stored.
2. Play with the feature toggles.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


feat/feature-flags.json
```diff
@@ -0,0 +1,17 @@
+{
```
dhruvkb, line 1: 🤔 How should this file be managed across multiple environments?

src/locales/en.json
```diff
@@ -625,5 +625,14 @@
     "see-image": "See all images",
     "see-audio": "See all audio"
   },
-  "skip-to-content": "Skip to content"
+  "skip-to-content": "Skip to content",
+  "pref-page": {
```
dhruvkb, line 629: 🤔 How will we preview "switchable" features to our audience?
🎨 This might need to be an actual page in our designs.

src/pages/preferences.vue
```diff
@@ -0,0 +1,116 @@
...
+    <h2>{{ $t('pref-page.store-state') }}</h2>
+    <pre><code>{{ flags }}</code></pre>
+
+    <h2>{{ $t('pref-page.content-filtering') }}</h2>
+    <template
+      v-for="featName in ['feat_enabled', 'feat_disabled', 'feat_switchable']"
+    >
+      <template v-for="featState in [ON, OFF]">
+        <i18n
+          v-if="featureState(featName) === featState"
```
dhruvkb, line 42: `v-if` seems to be the best solution for SSR compatibility. Nuxt does not run custom directives like the proposed `v-feat` in SSR, so that HTML is sent to the client and then deleted on client side (which is pretty wasteful).

dhruvkb (commented):
`// TODO`

feat/feature-flags.json
sarayourfriend: Would we name the file after the environment?

`local.json`, `staging.json`, `production.json`?

src/types/feature-flag.ts
```diff
@@ -0,0 +1,20 @@
+import type { FeatureState, FlagStatus } from '~/constants/feature-flag'
+
+interface BaseFeatureFlag {
+  status: FlagStatus
+  description?: string
+  data?: object
```
sarayourfriend, line 6: `object` is kind of a weird type in TypeScript and generally is safer to avoid it: https://stackoverflow.com/questions/49464634/difference-between-object-and-object-in-typescript

This is probably better typed as `unknown` to force users of a flag to have to cast the data to the type they think it should be for the given flag name.

src/pages/preferences.vue
```diff
@@ -0,0 +1,116 @@
...
+import { computed, defineComponent, useContext } from '@nuxtjs/composition-api'
+
+import { useFeatureFlagStore } from '~/stores/feature-flag'
+import { SWITCHABLE, ON, OFF } from '~/constants/feature-flag'
+
+import VContentPage from '~/components/VContentPage.vue'
+import VButton from '~/components/VButton.vue'
+import VCheckbox from '~/components/VCheckbox/VCheckbox.vue'
+
+export default defineComponent({
```
sarayourfriend, line 69: Should we hide this page in production? It doesn't seem like it should be exposed in prod and it'd be nice to _not_ expose it so that we can ignore things like UI/UX and the expectation that it's anything other than an "internal" development/testing tool.

src/pages/preferences.vue
```diff
@@ -0,0 +1,116 @@
...
+    <VButton variant="primary" @click="handleSave">{{
+      $t('pref-page.save')
+    }}</VButton>
+
+    <h2>{{ $t('pref-page.store-state') }}</h2>
+    <pre><code>{{ flags }}</code></pre>
+
+    <h2>{{ $t('pref-page.content-filtering') }}</h2>
+    <template
+      v-for="featName in ['feat_enabled', 'feat_disabled', 'feat_switchable']"
```
sarayourfriend, line 38: In actual implementation this would just iterate over the flags in the store right?

nuxt.config.ts
```diff
@@ -173,6 +173,7 @@ const config: NuxtConfig = {
     '@nuxtjs/redirect-module',
     '@nuxtjs/sentry',
     '@nuxtjs/sitemap',
+    'cookie-universal-nuxt',
```
sarayourfriend, line 176: It's a shame we're not on Nuxt3 and can't use `useCookie` yet :cry:

sarayourfriend (commented):
Looks great so far! The preferences page looks excellent as it is imo, as long as it's meant to be internal and disabled in production.

One improvement that could be made to it is feedback on the save button, right now it feels like the button is a noop when you click it because it doesn't give any feedback that it successfully saved the preferences.

nuxt.config.ts
dhruvkb: +💯

src/pages/preferences.vue
dhruvkb: I'm more inclined towards refining the UI and making it a prod-ready page. That way we might be able to do beta-releases of new features and UI changes.

feat/feature-flags.json
dhruvkb: We could do that, but then the JSON import will need to be dynamic. Not sure how that'll affect TS infererence.

src/stores/feature-flag.ts
```diff
@@ -0,0 +1,75 @@
...
+        if (flag.status === SWITCHABLE) flag.preferredState = cookies[name]
+      })
+    },
+    /**
+     * Toggle the feature flag of the given name to the given preferred state.
+     *
+     * @param name - the name of the flag to toggle
+     * @param targetState - the desired state of the feature flag
+     */
+    toggleFeature(name: string, targetState: FeatureState) {
```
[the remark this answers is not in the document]
sarayourfriend: I see :thinking: Why is that? The store already knows how to initialize itself from cookies. Is there a usecase where we'd want to call `toggleFeature` and it _wouldn't_ update the cookie?

src/pages/preferences.vue
sarayourfriend: > That way we might be able to do beta-releases of new features and UI changes.

How would this work?

src/types/feature-flag.ts
dhruvkb: Resolved in 90b2a37.

src/pages/preferences.vue
dhruvkb: Beta releases as in features going to production, while still being iterated or improved? If a feature is switchable in production with a [BETA] badge on it, that's essentially possible if we enable this page in prod.

src/stores/feature-flag.ts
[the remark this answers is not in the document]
dhruvkb: The store doesn't read cookie directly it only reads the JSON (which is read from cookies in middleware). The store also doesn't write cookies, it just returns an object (which is written to the cookie in a Vue component).

src/pages/preferences.vue
dhruvkb: We should be doing the same here too. Fixed in 9bb1283.

src/pages/preferences.vue
sarayourfriend: Ah, understood!

src/stores/feature-flag.ts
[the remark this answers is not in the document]
sarayourfriend: Gotcha!

I wonder of toggle feature is a confusing function expose then? Couldn't the "save" function on the preferences page just pass the updated cookie JSON to the `init` function? What's the use-case for this function outside of the preferences page? In the current implementation you could have some confusing behavior because the preferences page actually does toggle the feature in the store without updating the cookie, meaning if you navigate away from the preferences page without reloading then the rest of your session will be feature flags that are out of sync with your cookies.

To me it feels like splitting the store from the cookie introduces two sources of truth that can very easily drift (and indeed do in the currently implementation).

src/stores/feature-flag.ts
```diff
@@ -0,0 +1,75 @@
...
+  state: () => ({
+    flags: featureData.features as Record<string, FeatureFlag>,
+  }),
+  getters: {
+    /**
+     * Get the state of the named feature, based on config and cookie.
+     */
+    featureState:
+      (state: FeatureFlagState) =>
+      (name: string): FeatureState => {
```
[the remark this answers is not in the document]
dhruvkb: Updated in 735ac97.

src/stores/feature-flag.ts
```diff
@@ -0,0 +1,75 @@
...
+import featureData from '~~/feat/feature-flags.json'
+
+import { warn } from '~/utils/console'
+
+import type { FeatureFlag } from '~/types/feature-flag'
+import type { FeatureState } from '~/constants/feature-flag'
+import { ENABLED, SWITCHABLE, ON, OFF } from '~/constants/feature-flag'
+
+export interface FeatureFlagState {
+  flags: Record<string, FeatureFlag>
```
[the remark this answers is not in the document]
dhruvkb: Updated in 735ac97.

src/stores/feature-flag.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+const FEATURE_FLAG = 'feature_flag'
+
+export const useFeatureFlagStore = defineStore(FEATURE_FLAG, {
+  state: () => ({
+    flags: featureData.features as Record<string, FeatureFlag>,
```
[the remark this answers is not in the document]
dhruvkb: Updated in 735ac97.

src/stores/feature-flag.ts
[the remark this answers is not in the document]
dhruvkb: Technically the source of truth is always the store, 100% of the time. It's the only place from which the feature state is determined. The cookie is only meant for persistence of a very small subset of the info held in the store, i.e. mapping of switchable flags to their preferred states.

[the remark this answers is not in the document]
sarayourfriend: Yup! Sounds good to me. As long as can make sure the cookies and the store state don't drift (so that the unexpected "reload" with suddenly different features than you expected based on the previous page's state doesn't happen) then it's good to me!

src/stores/feature-flag.ts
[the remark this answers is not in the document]
dhruvkb: @sarayourfriend this is making it much harder to write tests when the JSON file is mocked.

src/stores/feature-flag.ts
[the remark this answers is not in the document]
dhruvkb: @sarayourfriend this is making it much harder to write tests when the JSON file is mocked.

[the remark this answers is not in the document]
dhruvkb: Converted tests to plain JS in bd00d5d.

src/components/VContentSwitcher/VSearchTypes.vue
```diff
@@ -84,7 +86,7 @@ export default defineComponent({
      * For example having all content types under `content.types` and making them filterable,
      * like `const additional = [...content.types.filter(i => i.status === ADDITIONAL)]`.
      */
-    if (isDev) {
+    if (featureFlagStore.featureState('models_3d') === ON) {
```
obulat, line 89: In #1285, I moved the enabling of 3d models to the constants module (by adding or not adding it to the list of the media types). Do you think we should maybe get that merged first?
I wonder if we should have both video and 3d models switchable? Or should they have different status?

sarayourfriend: > In https://github.com/WordPress/openverse-frontend/pull/1285, I moved the enabling of 3d models to the constants module (by adding or not adding it to the list of the media types). Do you think we should maybe get that merged first?

FWIW this makes more sense to me! Push the flag check as high up in the app as possible and the rest can just operate as usual, just iterating through the list of supported media types as it does today.

>  I wonder if we should have both video and 3d models switchable? Or should they have different status?

Does video require more than a single PR? What is the current goal for it, I wasn't even aware that it was something we were working on :sweat_smile: If it's multiple PRs of work then adding a flag would make sense (it doesn't cost anything to put it behind a flag, other than the amount of time spent cleaning the flag up after the feature is stable).

dhruvkb: We can break them up into two feature flags with different statuses.

feat/feature-flags.json
sarayourfriend: That stinks! I wonder if we could update the `defaultState` and `status` keys to be objects that define it per environment, defaulting to being the same for all environments if they're not an object?

Something like:

```
"favorite_feature": {
  "status": {
    "local": "enabled",
    "staging": "switchable",
    "production": "disabled",
  }
}
```

We'd have to define a cascading behavior though, like if a lower environment isn't defined then it uses the definition for the next highest environment? I.e., if local isn't there, then use staging, if not then production (at which point it shouldn't be the object version anyway). And if we're on a higher environment than is defined, default to `disabled`.

Do you think that could work Dhruv? It makes things a bit more complicated but it would give us a lot more flexibility.

src/locales/en.json
sarayourfriend: We could add a note that says that the disabled checkboxes are not switchable for the current environment, or something like that?

feat/feature-flags.json
dhruvkb: I'll explore this and see if we can work with that.

src/locales/en.json
dhruvkb: Updated the text. The UI should be much clearer now.

src/constants/node-env.ts
```diff
@@ -0,0 +1,19 @@
+export const LOCAL = 'local'
+export const DEVELOPMENT = 'development'
```
sarayourfriend, line 2: This is interesting, because `NODE_ENV` is usually set to `development` when doing local development. I wonder what the best way to update scripts is for this?

This also gets into something that @zackkrida mentioned about the difference between NODE_ENV and the deployment stage. Like in our staging environment, `NODE_ENV` will be `production`, but maybe `STAGE_ENV` or some other custom variable we set, would be `staging`.

Should we rely on that instead maybe?

These are complicated!

I'm also wondering if we should encourage the use of `CI`, and `TEST` at all. In test presumably you would be testing specific flag combinations rather than some pre-set _single_ combination of flags (or at least we should be) and CI sort of conflicts with test in my head, I don't know when CI would be true but not test, at least when it comes to feature flags :thinking: 

Do you know if `NODE_ENV` is set to the string `'ci'` ever?

dhruvkb: I was also hesitant to use `NODE_ENV` for this purpose. It's very unclear when it's set to what and what it's possible values are. I'm not also sure about what else we could use.

sarayourfriend: I think we can just create a new variable for it. `STAGE` is what our infrastructure uses to denote the "stage" of deployment, but I find it confusing with the word "staging" as one of those "stages" of deployment.

Could we use `DEPLOYMENT_ENV` and default it to `local`? Then we'd just need to update our staging and prod variables in our Terraform to define `DEPLOYMENT_ENV`. We also need this for Sentry as well.

sarayourfriend (approved):
LGTM!

Should we add the `preferences` link to the pages menu... _behind a feature flag_??? :scream:

sarayourfriend:
I just noticed I have a bunch of warnings in my console:

```
 WARN  Invalid feature flag accessed: feat_nonexistent                                             10:48:54


 WARN  [vue-i18n] Value of key 'feat_nonexistent' is not a string or function !                    10:48:54


 WARN  [vue-i18n] Cannot translate the value of keypath 'feat_nonexistent'. Use the value of keypath as default.


 WARN  [vue-i18n] Value of key 'on' is not a string or function !                                  10:48:54


 WARN  [vue-i18n] Cannot translate the value of keypath 'on'. Use the value of keypath as default. 10:48:54
```

src/constants/deploy-env.ts
```diff
@@ -0,0 +1,10 @@
+export const LOCAL = 'local'
```
obulat, line 1: What does `local` mean here? How is it different from `dev`?

src/stores/feature-flag.ts
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Get the state of the named feature, based on config and cookie.
+     */
+    featureState:
+      (state: FeatureFlagState) =>
+      (name: keyof typeof featureData['features']): FeatureState => {
+        if (name in state.flags) return getFeatureState(state.flags[name])
+        else {
+          warn(`Invalid feature flag accessed: ${name}`)
+          return ON
```
obulat, line 72: Wouldn't it be safer to return OFF for an invalid feature?

src/pages/preferences.vue
```diff
@@ -0,0 +1,125 @@
...
+     */
+    const handleChange = ({
+      name,
+      checked,
+    }: {
+      name: string
+      checked: boolean
+    }) => {
+      featureFlagStore.toggleFeature(name, checked ? ON : OFF)
+      app.$cookies.set('features', featureFlagStore.flagStateMap)
```
obulat, line 109: Is there a scenario where we would want to toggle a feature but not set the cookie?
If not, can we move setting the cookie to the store so that components don't have to know about this implementation detail?

obulat (commented):
Amazing work! The store code is really clean, and figuring out the json typing was really tricky.
I have left a question about the setting of the cookie inside the store, other than that this LGTM!

src/pages/preferences.vue
sarayourfriend: Ideally components aren't toggle feature flags, only the preferences page is doing so.

Dhruv and I discussed the same separation of concerns question in one of the comments on [this review](https://github.com/WordPress/openverse-frontend/pull/1270#pullrequestreview-941984689) (I can't find the exact comment right now because they won't load for me at the moment).

I tend to agree with you that I'd like the store to encapsulate all the behavior including reading and writing the cookies, but Dhruv's reasons for not doing so also made sense at time, IIRC trying to separate the concerns :slightly_smiling_face:

obulat: Oh, I missed the fact that a feature flag can ONLY be updated by the preferences page.
I found [the discussion](https://github.com/WordPress/openverse-frontend/pull/1270#discussion_r850293918) you've had:

> The store doesn't read cookie directly it only reads the JSON (which is read from cookies in middleware). The store also doesn't write cookies, it just returns an object (which is written to the cookie in a Vue component).

And still wasn't convinced that not updating the cookie in the store is the best course of action. But since it's only the preferences page, it makes sense.

obulat (approved):

src/constants/deploy-env.ts
dhruvkb: I'm not sure to be honest, but this is my interpretation:
- local: the instance on your work computer/laptop
- dev: an instance in the cloud that runs the up-to-date code from master
- staging: an instance in the cloud that runs the latest release for QA
- production: an instance in the cloud that runs the latest after it's passed QA in staging

src/stores/feature-flag.ts
dhruvkb: That's a good point, but returning `ON` by default allows us to drop the flags from the file and ensure a rolled-out feature does not get rolled back.

src/pages/preferences.vue
dhruvkb: The other issue was that `app.$cookies` would not be available inside the store. I'd defintitely move it to the store (the reasons in favour of doing so are rather strong) if there is a good way to get `app.$cookies` inside the store.

src/constants/deploy-env.ts
sarayourfriend: That's my understanding too.

We currently use `dev` and `staging` roughly interchangeably in our current infrastructure setup, but that's something that's liable (hopefully!) to change. We could eventually collapse dev and staging into a single environment once we get close to true continuous delivery.

dhruvkb:
I'm going to make a new PR that drops these demo flags and adds some real ones for 3D models. That should resolve https://github.com/WordPress/openverse-frontend/pull/1270/#issuecomment-1115920438.

PR #1274, opened 2022-04-14
Convert nav store to an options store

### LINKED ISSUE #1254: Correctly hydrating the server state in Pinia setup stores
## Problem

We are almost done converting Vuex stores to Pinia ([milestone](https://github.com/WordPress/openverse-frontend/milestone/5)). However, we are incorrectly re-initializing the store on the client side instead of reusing the server-side state, and this is causing a lot of errors. 

## Description

All of the stores are now converted to Pinia setup stores.

Final conversion surfaced a very important flaw in this conversion process. During the state hydration on the client, the Pinia store setup function is ran again as is. This means that instead of reusing the state passed from the server, we are re-creating the state with initial values. This causes numerous errors in the app logic.
For example, when you click on 'Load more' button on a server-rendered search page, the media fetch state page value is set to undefined, so we are re-fetching the same page that has already been fetched. Other bugs such as the audio search page with fewer than 20 results being unresponsive, and no change when you select an extension filter and decelect it, are also caused by the incorrect state on the client.

The solution I propose (and try out in the Provider store PR) is to add a custom function for initializing the state by setting it to the initial value on the server, and adding our own hydration solution on the client. Nuxt context is not injected inside the setup store setup function. However, it is available as part of the `$nuxt` property of the global `window` object.
```
const initialState = { searchType: ALL_MEDIA }

const initializeState = () => {
    // add `isClient` checks to make sure window is not undefined 
    const serverState = window.$nuxt.context.pinia.state.search
    if (serverState) {
        // Making sure that the state is a simple object, not a proxy or a ref.
        return JSON.parse(JSON.stringify(serverState))
    } else {
        return initialState
    }
}

const searchStore = defineStore('search', () => {
    const state = reactive(initializeState())

    return { state }
})
```

This can only work if we can serialize and deserialize the state correctly. Here, the problem in the `media` store is with the current implementation of the `fetchState`, which is a composable that returns both simple properties, but also functions to change the `fetchState`. The functions cannot be stringified, and need to be filtered out when hydrating the state. We also cannot use a computed that would only return the inner simple `FetchState` object from this computed because Pinia treats computed refs as getters, and it only passes the state (i.e. the `ref`s and `reactive`s) from the server to client as part of the `window.$nuxt.context` value.
Instead, we can create the `_fetchState` as an inner composable that would handle state updates, and a fetchState `ref` that would be updated by a `_fetchState` watcher, and would be serialized by the server, and deserialized by the client to initialize client state correctly.

This conversion will be easier if the current large media store is split into separate stores: audio, image, allMedia, and singleMedia (for the image/audio detail pages), but that would make for a huge PR that would be difficult to review :(

## Alternatives

Another to go back and refactor the stores to use options API. The options store uses a separate function for creating the store, and because of that does correctly handle hydration on the client side. It calls the `state` function on the server, but hydrates the server state values and makes them reactive on the client. There is even a special `hydrate` option that is only available for the options stores.
With this solution, we won't have to _"fight the tool"_, and will be using the options it provides instead.

## Additional context

The solution for store hydration is implemented in the Provider store PR: #1158.
I have also opened a [discussion in the Pinia repository](https://github.com/vuejs/pinia/discussions/1213) about the composition store state hydration, but it's still unanswered.

## Implementation

obulat:
## Fixes

Fixes #1254 by @obulat

## Description

This is the last store that needed to be converted from a Pinia setup store to an options store. 
`active-media` store can also be converted, but because it's not used in server-rendered pages, it's not necessary.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/stores/nav.ts
+++ b/src/stores/nav.ts
@@ -1,39 +1,26 @@
 import { defineStore } from 'pinia'
-import { reactive, readonly, toRefs } from '@nuxtjs/composition-api'
 
 export interface NavState {
   isEmbedded: boolean
   isReferredFromCc: boolean
 }
 
-const NAV = 'nav'
-
 /**
  * Store information about navigation.
  */
-export const useNavStore = defineStore(NAV, () => {
-  /* State */
-
-  const state: NavState = reactive({
+export const useNavStore = defineStore('nav', {
+  state: (): NavState => ({
     isEmbedded: true,
     isReferredFromCc: false,
-  })
-  const { isEmbedded, isReferredFromCc } = toRefs(state)
+  }),
 
   /* Actions */
-
-  function setIsEmbedded(isEmbedded = true) {
-    state.isEmbedded = isEmbedded
-  }
-  function setIsReferredFromCc(isReferredFromCc = true) {
-    state.isReferredFromCc = isReferredFromCc
-  }
-
-  return {
-    isEmbedded: readonly(isEmbedded),
-    isReferredFromCc: readonly(isReferredFromCc),
-
-    setIsEmbedded,
-    setIsReferredFromCc,
-  }
+  actions: {
+    setIsEmbedded(isEmbedded = true) {
+      this.isEmbedded = isEmbedded
+    },
+    setIsReferredFromCc(isReferredFromCc = true) {
+      this.isReferredFromCc = isReferredFromCc
+    },
+  },
 })
--- a/src/layouts/blank.vue
+++ b/src/layouts/blank.vue
@@ -9,7 +9,7 @@
 <script>
 import { computed, defineComponent } from '@nuxtjs/composition-api'
 
-import { useNavStore } from '~/stores/nav'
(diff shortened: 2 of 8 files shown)


src/stores/nav.ts
```diff
@@ -1,39 +1,26 @@
...
-const NAV = 'nav'
-
 /**
  * Store information about navigation.
  */
-export const useNavStore = defineStore(NAV, () => {
-  /* State */
-
-  const state: NavState = reactive({
+export const useNavStore = defineStore('nav', {
```
krysal, line 11:
```suggestion
export const useNavigationStore = defineStore('navigation', {
```
🙏

krysal (approved):
I will appreciate if we use the full name of this module as the rest. Otherwise, LGTM.

sarayourfriend (approved):
LGTM!

PR #1275, opened 2022-04-14
Remove `mob` breakpoint and simplify VFilterButton

sarayourfriend:
## Fixes

Actually there are no issues for this. The files I renamed to TS were already type checked.

The removal of the `mob` breakpoint happened in a private discussion between @dhruvkb and myself about why it existed that was prompted by me trying to refine the types for the media query composable. It turns out the `mob` breakpoint was redundant with the `xs` so we removed it.

This also prompted some clean up with breakpoint modifiers in the templates which then led me to removing a ton of JavaScript from the VFilterButton and replacing it with breakpoint-based conditional CSS :tada: This makes the SSR rendering & hydration of the filter button much more pleasant.

Also because the button no longer only uses JavaScript to determine how to present itself, I moved the tests to visual regression tests. It still requires some programatic stuff for some things, so I added a utility `gotoWithArgs` in the storybook-specific playwright utils that makes it easier to write tests against a story with specific arguments passed to it.

This also prompted me to clean up the filter button story, it had some confusing stuff in it that didn't appear to be doing anything (probably just left over from copy/pasting another story I'd imagine).

## Description

* Renames several files to TS and moves the types from JSDoc to TypeScript
* Removes the `mob` breakpoint in favor of just using xs as the lowest (which was the lowest in Tailwind anyway).
* Fixes usages of the `xs` breakpoint modifier in the templates. These were basically noop modifications because `xs` is the lowest breakpoint and is always met (unless a device goes under 340 but there are no known devices used today that do that we would need to target specifically anyway)
* Removes a ton of JavaScript-based conditional style rendering from the VFilterButton to improve the code and the rendering in SSR matching the client without a jump after the page hydration. As much of this kind of JavaScript as we can remove the better. Some of it isn't possible to remove (like the button variant one, unfortunately, though I keep trying to think of a way to make that work) but any of it that can be removed should be.

## Testing Instructions

Checkout the branch and compare it to staging, visually. It should be nearly identical with the only differences showing in the filter button in the SSR render before hydration (there should be marginally less jumpiness now).

Be sure to test applying filters and scrolling or not scrolling, etc at various page widths. I've written snapshot tests for these cases, if I missed one let me know and I will add it.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (commented):
I love code cleanup, but I think it un-did the fixes we had for the very small screens (below 370px). Here's a link to the discussion of the problem: https://github.com/WordPress/openverse-frontend/pull/718#issuecomment-1023164665

sarayourfriend:
@obulat good catch, thanks! You added visual regression tests in another PR that I think will cover this right?

obulat:
> @obulat good catch, thanks! You added visual regression tests in another PR that I think will cover this right?

I only added the searchbar there, but I was thinking that we would eventually add all of the header. Do you think it's better to split the visual header tests by component (and have different states there), or use all of the header with various states of different components?

sarayourfriend:
> I only added the searchbar there, but I was thinking that we would eventually add all of the header. Do you think it's better to split the visual header tests by component (and have different states there), or use all of the header with various states of different components?

I think a single suite of tests for the full header under certain conditions is probably easier to manage rather than a ton of smaller individual component tests? I'm not sure. @dhruvkb what do you think, given your previous experience with visual regression tests?

sarayourfriend:
On second thought, bigger full-header snapshots, while nice for checking the layout, would also introduce more frustrating merge conflicts if there are multiple PRs that update the header. But I think that might just be something we have to accept :shrug:

sarayourfriend:
@obulat @krysal this is ready for another review :grin:

src/components/VHeader/VFilterButton.vue
```diff
@@ -3,27 +3,32 @@
...
-      'w-10 h-10 px-0': isIconButton,
-    }"
+    :class="
+      filtersAreApplied ? 'px-3' : 'w-10 md:w-auto h-10 md:h-auto px-0 md:px-3'
+    "
     :pressed="pressed"
     aria-controls="filter-sidebar"
-    :aria-label="label"
-    @click="toggleFilters"
+    @click="$emit('toggle')"
```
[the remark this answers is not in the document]
sarayourfriend: Hmm, I'm not sure why I removed the label! The label should definitely be more verbose, preferably the full version of the label that we use on desktop when filters are selected.

Thanks for catching this.

test/playwright/e2e/filters.spec.ts
```diff
@@ -152,9 +152,9 @@ test('filters are updated when media type changes', async ({ page }) => {
 
   // Only CC0 checkbox is checked, and the filter button label is '1 Filter'
   await assertCheckboxStatus(page, 'cc0')
-  const filterButtonSelector =
-    '[aria-controls="filter-sidebar"], [aria-controls="filter-modal"]'
-  await expect(page.locator(filterButtonSelector)).toHaveText('1 Filter')
+  await expect(
```
obulat, line 155: `aria-controls` doesn't seem to work anymore: I couldn't find any elements with `filter-sidebar` or `filter-modal` ids. If the sidebar and the modal are not present in the DOM at the same time, we could just use a single `filters` id, and simplify this locator to use `[aria-controls="filters"].

tsconfig.json
```diff
@@ -57,13 +58,9 @@
...
     "src/composables/types.js",
     "src/composables/use-event-listener-outside.js",
-    "src/composables/use-active-audio.js",
-    "src/composables/use-get-locale-formatted-number.js",
-    "src/composables/use-browser-detection.js",
-    "src/composables/use-media-query.js",
-    "src/composables/window.js",
-    "src/constants/screens.js",
-    "src/utils/deep-freeze.js"
+
```
obulat, line 61: 🎉

obulat (changes requested):
So nice to simplify the logic and reduce the number of files listed in `tsconfig`! :) 
I will approve after the `aria-controls` logic is fixed (I think it was working and then during the changes got accidentally deleted?)

test/playwright/e2e/filters.spec.ts
sarayourfriend: The `filter-modal` aria-controls doesn't appear to exist anywhere in the codebase on `main`, that's why I removed it from this selector.

Not sure why the `filter-sidebar` one isn't working anymore though!

sarayourfriend: Actually, it seems this must be a regression from a previous PR as I can't find anything with a `filter-sidebar` ID in `main` either.

obulat: It seems to be a regression from a long time ago, because I couldn't find neither of those ids in the code. But it would surely be nice to fix :)

sarayourfriend: Where are those id's meant to go? On the wrapper elements of the sidebar and the modal?

obulat (approved):
👍

sarayourfriend:
@dhruvkb or @krysal (when you're back from resting) this would be a high value PR to get in as it unblocks work on a critical issue.

dhruvkb (approved):
Sorry I forgot to approve this. It's a massive improvement!

PR #1281, opened 2022-04-15
Fix narrow width audio results

### LINKED ISSUE #1278: Mobile audio results using incorrect audio track layout
## Description

The audio results isn't switching to the correct track layout on small displays.

## Reproduction

1. View https://wordpress.org/openverse/search/audio?q=dog on a mobile device
2. Observe incorrect audio layout

## Screenshots

CleanShot 2022-04-14 at 20 42 17@2x

## Environment

 - Device: 
 - OS: 
 - Browser: 
 - Version: 
 - Other info: 

## Resolution

sarayourfriend:
## Fixes

Fixes #1278  by @zackkrida 

## Description

The fix in this PR is actually just two lines, flipping one boolean and switching an and to an or.

The rest is all tests and test support.

## Testing Instructions

Checkout the snapshots and also run the app locally by checking out the branch and running `pnpm dev`. Visit the audio results page, both via client side navigation and via SSR with and without a mobile UA and verify that the correct layout and size is used for every situation. The snapshots should help with that as well.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (approved):
This is wonderful! Some of the screenshots show some xs screen horizontal cropping but we can look at that outside of this PR.

obulat (commented):
On a narrow desktop view, the SSR-ed page is still displaying incorrect layout: there is a large play-pause button on the right side, and the little play-pause button is lower than it should be (it's actually outside of the thumbnail). If you open the same page on the client, it looks correct.
Sometimes, when you use a computed on SSR, it doesn't get re-evaluted on the server. Using conditional `hidden` class on the server to hide something can also not be updated on the server, and the element can stayl visible. I'm not sure why or when it occurs, but I had a lot of trouble debugging a conditional class not being applied on the server render when working on the filters button label.

I tried replacing `audioTrackSize` computed with a `ref`, `watch` and `onMounted`:
```
    // On SSR, we set the size to small if the User Agent is mobile, otherwise we set the size to medium.
    const isMobile = ref(useBrowserIsMobile())
    const updateAudioTrackSize = () => !isMinScreenMd.value || isMobile ? 's' : props.isFilterVisible ? 'l' : 'm'

    const audioTrackSize = ref(updateAudioTrackSize())

    onMounted(() => {
      audioTrackSize.value = updateAudioTrackSize()
    })

    watch(
      [isMinScreenMd, isMobile, props],
      ([isMinScreenMd, isMobile, isFilterVisible]) => {
        updateAudioTrackSize()
      }
    )
```

It seems to fix the layout on narrow computer screen, at least after the component is mounted.

sarayourfriend:
@obulat Interesting, I didn't notice that.

Using `onMounted` in the audio track doesn't seem to change anything, indeed, if you check the props being passed it is passing the correct size to the layout component (`s` in this case). It appears to be some kind of issue where the layout isn't re-rendering even though the prop value changes.

This commit (https://github.com/WordPress/openverse-frontend/pull/1281/commits/d5a965632a56c2bb974da8054b4336ae6dca0edd) fixed it :shrug:

krysal (approved):
LGTM! I see the audio player layout changing and no regressions 👍

Issue #1124, opened 2022-03-15
Additional type page


panchovm:
## Problem
The additional type integration needs a standalone page to list the related sources. In the last release assets ([Audio release](https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=133%3A6869)) there is no page for this case.

## Description

|Desktop|Mobile|
|:---|:---|
|![image](https://user-images.githubusercontent.com/895819/158461259-ae13b04e-5f41-4134-b6cf-79dabc53a407.png)|![image](https://user-images.githubusercontent.com/895819/158461287-e2b1aed7-7175-4fc4-ade9-f0dbccd413af.png)|

The page proposed above is based on **No results** and **Timeout** pages, and the content is center-aligned in 6 columns. The header has disabled the filter button as some sources do not allow specific filters. If we agree with the idea, I need to add the button to the Design Library and create a ticket for its integration.

I am not fully sure about the h1 text. The description below is from the meta-search sources list visible under the load more button.

## Additional context
- Related ticket: #739

sarayourfriend:
@panchovm should this ticket have needs design tag on it?

sarayourfriend:
Thank you, Francisco!

PR #1285, opened 2022-04-16
Support for video page and additional content types

### LINKED ISSUE #631: Content switcher with Audio as beta and Video as Meta search
## Problem
As discussed on the public channel, the audio releases will be in Beta as we are still improving the integration. In parallel, openverse displays a list of additional sources for video content under an open license. None of the current mockups include these two situations.

## Description
Here is a new version of the content switcher showing both actions: Audio in Beta, and Video as Additional type. Here is a screen recording showing the interaction.

### Desktop
https://user-images.githubusercontent.com/895819/149994168-cb7bb81e-aec1-400c-bc8b-2037312bdef7.mp4

### Mobile
https://user-images.githubusercontent.com/895819/149994715-a9822e9e-555d-4345-89f9-b584574cc464.mp4

Note: The video above misses a step. After selecting the video option, the menu closes and video results page is visible. The video item selected is only visible once opening the menu on video results page.

## Additional context
The Beta pill is a new component requested on issue #630

obulat:
## Fixes

Fixes #631 by @panchovm 

## Description

This PR adds support for the video page with Additional sources form. It also refactors the way search types are handled to enable using a non-supported search type, too.

I am setting this as medium priority, but it may be 'high' because the `model_3d` work depends on these changes.

I have a question about what to do when a user tries to search from a non-supported type search page (i.e., video). When the user enters a search term on the Video page and clicks 'Search', we can:
1. automatically search for All Content - redirect to the All Content page
2. update the path and set the search term in the store, but do nothing else. This way, the user can change the search type using the content switcher, and be redirected to the same query for other search types? Here, we can also use some 'popup'/ 'toast' indication saying that search for video is not available.
3. do nothing (maybe with some kind of indication like a toast saying that search is not available for 'video')

I think I would expect the second option if I was the user. What do you think, @panchovm ?

## Testing Instructions

Open the video page (http://localhost:8443/search/video or http://localhost:8443/search/video/?q=cat). It should not be broken, and should show the Additional sources buttons that redirect to video search on the provider sites.
Changing the search type in the content switcher should also work.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
This looks great! I think #2 would be best, so that the 'additional source' urls update when a new search is submitted. 

@panchovm we can discuss this in a separate PR, but I experimented with reducing the vertical padding of the content switcher. I'm just concerned about the height of this element as we add more types. Here are some examples of reduced padding:

CleanShot 2022-04-16 at 11 19 18@2x
CleanShot 2022-04-16 at 11 18 39@2x

zackkrida:
Oh @obulat one more thing, should/could we add the Models 3D page in this PR as well? Seems like a fairly easy addition given the work you've put in!

obulat:
Do we want to use `/model_3d/` as a slug? If so, then the file name for the page (`model_3d.vue`) becomes not "kebab-cased", so we'll need to overrule the ESLint rule for this file.

zackkrida:
That's a good question regarding what the best slug is. A hyphen would be much more conventional–maybe we should jut change to `model-3d` everywhere

panchovm:
I like solution 2 too. I think it is the most expected behavior.

Noting that this implementation is related to ticket #1124 and the mockups are available in [this Figma file](https://www.figma.com/file/IFIqt7Q3MEKMLSnBedzHBO/Next-Release-%5BTBD%5D?node-id=2%3A48).

Regarding @zackkrida's exploration, I like the outcome. I would reduce the sides' paddings to have the same white space effect on all sides. Did you check [the component's last version](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=558%3A2283)? It looks very similar, and I think there is a current issue requesting the spacing update.

zackkrida:
@obulat have you seen @panchovm's design for these pages, in this PR? https://github.com/WordPress/openverse-frontend/issues/1124

It includes disabling the filter bar on these pages.

sarayourfriend:
We never got to discuss the language around them.

I'll reshare my thoughts about the names here:

* "Additional" implies that they are in Openverse somehow, like "audio" in addition to "image" search. It doesn't follow from "additional" that the searches are _external_, which to me is a clearer term, "External search engines" or even more simply "Other search engines".
* The word "source" probably doesn't mean much to a new Openverse user. Even provider could be vague, as we don't educate users clearly about where the results are coming from until they get to the single result page and start seeing that such and such result is from Flickr or Freesound or wherever. Design earlier in the search funnel could help this (and we know that some of our providers feel too hidden by our current approach)? It's hard to know if this is actually a problem though, given we can't user test :cry:  I'm just guessing based on what I think the average new user will know about Openverse ("a search engine for openly licensed media", I wonder if the average non-technical person thinks of "search engine" as an aggregator like technical folks probably understand it to mean).

panchovm:
I like the word External. It is more explicit about the browser experience where Openverse showcase other search engines.

Regarding sources, I agree that is difficult to understand. It is not just the term but also the experience we are providing. This needs an improvement on the single result view to stand out the source/provider name, and probably some Homepage tweak showing the variety of sources living in the search results.

Fortunately, we have discussed the latter, so I would keep the discussion once I can work on a proposal for this scenario.

zackkrida:
I like the simplicity of 'Other search engines'; the fact that we describe openverse as a 'search engine' should mean they're familiar with the term. I'm also comfortable with 'other sources', since we have a 'source' filter and a 'sources' page.

obulat:
So, looking at https://github.com/WordPress/openverse-frontend/issues/1124, the new copy for the video and 3d model pages will be:

> _Video_ search is still under development. Try one of the _<???>_ below.

I like the "other search engines" for the <???> part:

> _Video_ search is still under development. Try one of the _other search engines_ below.

obulat:
I am going to close this PR because the implementation of Video pages after the introduction of Feature flags changes a bit. #1430 makes it so that going to the `/search/video` URL doesn't crash the app, and #1410 puts the additional search types behind a feature flag.

obulat:
I am going to close this PR because the implementation of Video pages after the introduction of Feature flags changes a bit. #1430 makes it so that going to the `/search/video` URL doesn't crash the app, and #1410 puts the additional search types behind a feature flag.

PR #1289, opened 2022-04-17
Open results in modal

sarayourfriend:
## Fixes

Fixes #601  by @panchovm 

## Description

Still very much a work in progress. More or less works in the happy path for images (all that's been set up yet, haven't touched audio at all but once images are sorted out it shouldn't be too bad to get audio, or any media type, working).

What works:

Open an image result from the all content results into a modal. Closing the modal works and even using the back button to navigate back to the results (i.e., close the modal with the back button). You can even _reopen_ the modal using the forward button (but then the back button stops working to go back to search).

Navigating related results works until you try to use the browser forward button, then they stop working. Otherwise you can visit several related results and use the back button to go all the way back to the search results. However, if you hit "back" again after closing the modal from a related result it will open the modal but with the first result you selected instead of the last result that was viewed.

What does not work:

As I mentioned above, if you open the modal then go "back - forward - back" in the browser it should close then reopen the modal, then on the second back, close the modal again. The closing the modal the second time does not work for some reason.

Basically a whole bunch of edge cases with the browser back/forward navigation don't work. It feels like history state is supposed to be useful for this but for some reason `popstate`'s Event.state property just has a `key: number` property in it no matter what I pass for the state in `pushState`. From what I read in the docs that's not the expected behavior. My only guess is that this is something to do with vue-router also writing to state (potentially using `replaceState` in a sneaky way?) and overwriting the state that I pass in myself.

I'm considering the idea that maybe a custom history stack needs to be maintained that can actually be inspected, as `history` doesn't allow you to do that. The idea here would be that we'd monitor `popstate` and check for back/forwards navigation and maintain our own separate routing path stack in a ref or something. Of course that would be difficult to do actually because the popstate event doesn't actually tell you if back or forwards was pressed, so we'd have to be guessing by maintaining a separate ref to know the previous location and compare the current, then try to guess which direction in the stack we're moving. The problem with that, of course, is that the if you have a stack that goes `A B A B`, if you're on the second A and then navigate to B, it's impossible to know whether it's a forward or back action. I don't think we actually have the possibility for that in the current application but it'd be a nasty bug to run into down the line if we did. Overall re-implementing history stack sounds like a real source for headaches.

I'm glad I've been able to get what's working so far actually working.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/SketchFabViewer.vue
+++ b/src/components/SketchFabViewer.vue
@@ -28,6 +28,7 @@ export default {
       required: true,
     },
   },
+  emits: ['failure'],
   data() {
     return {
       client: null,
--- a/src/components/VAllResultsGrid/VImageCellSquare.vue
+++ b/src/components/VAllResultsGrid/VImageCellSquare.vue
@@ -4,6 +4,7 @@
     :title="image.title"
     :href="'/image/' + image.id"
     class="group block focus:ring-[3px] focus:ring-pink focus:ring-offset-[3px] focus:outline-none rounded-sm"
+    :data-resultid="image.id"
   >
     <figure
       itemprop="image"
--- a/src/components/VImageGrid/VImageCell.vue
+++ b/src/components/VImageGrid/VImageCell.vue
@@ -6,6 +6,7 @@
     :style="`width: ${containerAspect * widthBasis}px;flex-grow: ${
       containerAspect * widthBasis
     }`"
+    :data-resultid="image.id"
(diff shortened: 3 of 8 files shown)


sarayourfriend:
I'm going to close this PR. The code can be picked up my someone else as a starting point when this work actually gets prioritized.

PR #1316, opened 2022-04-23
Add a Tabs component

### LINKED ISSUE #749: Accessible "Tabs" component
## Problem

Currently we're creating inaccessible, ad-hoc experiences with our "tabs" components on the image and audio results page.

There is an established accessible pattern for these kinds of interfaces called the ["tabpanel"](https://www.w3.org/TR/wai-aria-practices/#tabpanel).

## Description

We need to create a component to encapsulate this interaction and make it accessible. Ideally the API would look something like this:

```vue
<template>
  <VTabs>
    <template #tabs="{ activeTab }">
      <VTab id="one" :active="activeTab === 'one'">Tab one</VTab>
      <VTab id="two" :active="activeTab === 'two'">Tab two</VTab>
      <VTab id="three" :active="activeTab === 'three'">Tab three</VTab>
    </template>

    <VTabPage for="one">Tab one content</VTabPage>
    <VTabPage for="two">Tab one content</VTabPage>
    <VTabPage for="three">Tab one content</VTabPage>
  </VTabs>
</template>
```

It would be nice if we could arbitrarily order the tabs and tab pages so that you could do something like this:

```vue
<template>
  <VTabs>
    <VTab id="one">Tab one</VTab>
    <VTabPage for="one">Tab one content</VTabPage>

    <VTab id="two">Tab two</VTab>
    <VTabPage for="two">Tab one content</VTabPage>

    <VTab id="three">Tab three</VTab>
    <VTabPage for="three">Tab one content</VTabPage>
  </VTabs>
</template>
```

We could try using `provide/inject` to create an API that each tab and page registers itself against and the `VTabs` component can manage rendering them in the correct place and order.

This may not be workable though, in which case the first example should work and we would just use DOM apis to hide/show the appropriate page when a given tab is selected.

## Implementation

### LINKED ISSUE #1000: Feedback tabs are broken
Come custom (read: bad, inaccessible) styles were used for the tabs on the feedback page, and those are now broken. The styles must have been removed at some point.

CleanShot 2022-02-23 at 18 22 38@2x

We should instead do something simple, like add two buttons that are jump links to each form. Maybe we could show them next to each other on desktop and stacked on mobile.

obulat:
## Fixes

Fixes #749 by @sarayourfriend 
Fixes #1000 by @zackkrida 

## Description

This PR adds the accessible Tabs component and uses it for **_License attribution copy_** component and the **_Feedback_** page. This implementation is based on the Tailwind Labs [Headless UI ](https://headlessui.dev/) Tabs component implementation (MIT licensed).
The `focus-management` file added probably duplicates a lot of functionality that we are already using from AriaKit.

## Testing Instructions

The number of changed files might seem intimidating, but it's mostly snapshot and tape files.

Open a single result page (audio and image) and check that the Copy attribution tabs work correctly, and you can move through the tabs using arrow keys. Arrow keys move in the correct direction in RTL pages, too! When you <kbd>Tab</kbd> from a selected tab, you focus the tabpanel.

Open the feedback page and see that the tab panels are working correctly. To open a tab using the keyboard you need to first use your arrow keys to focus on the tab, and then click `Enter` or `Space` to actually select it.

Run the Storybook and check that the two tab stories work correctly.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/utils/focus-management.ts
```diff
@@ -0,0 +1,183 @@
...
+
+import type { Ref } from '@nuxtjs/composition-api'
+
+function getOwnerDocument<T extends Element | Ref<Element | null>>(
+  element: T | null | undefined
+) {
+  if (typeof window === 'undefined') return null
+  if (element instanceof Node) return element.ownerDocument
+  // eslint-disable-next-line no-prototype-builtins
+  if (element?.hasOwnProperty('value')) {
```
sarayourfriend, line 12: We could follow the [advice from the rule](https://eslint.org/docs/rules/no-prototype-builtins) instead of disabling it, if you want:

```suggestion
  if (element && Object.prototype.hasOwnProperty.call(element, 'value')) {
```

src/utils/focus-management.ts
```diff
@@ -0,0 +1,183 @@
...
+  Error,
+  Overflow,
+  Success,
+  Underflow,
+}
+
+enum Direction {
+  Previous = -1,
+  Next = 1,
+}
```
sarayourfriend, line 71: Let's avoid using the `enum` feature from TypeScript. It violates TS's promise of only adding types to JS but also adding new language features that JS doesn't have natively (same with the TS decorator implementation).

In these cases, we can just as easily use an object for the first one and a union for the other two:

```suggestion
export const Focus = Object.freeze({
  /** Focus the first non-disabled element */
  First: 1 << 0,

  /** Focus the previous non-disabled element */
  Previous: 1 << 1,

  /** Focus the next non-disabled element */
  Next: 1 << 2,

  /** Focus the last non-disabled element */
  Last: 1 << 3,

  /** Wrap tab around */
  WrapAround: 1 << 4,
}

export type FocusResult =
  | 'Error'
  | 'Overflow'
  | 'Success'
  | 'Underflow'

// This could be an object too if we want to keep the nice names.
type Direction =
  | -1
  | 1
```

test/playwright/utils/breakpoints.ts
```diff
@@ -133,13 +133,15 @@ const describeEachBreakpoint =
   }
 
 const describeEvery = describeEachBreakpoint(Array.from(SCREEN_SIZES.keys()))
+const describeDesktopMobile = describeEachBreakpoint(['lg', 'sm'])
```
sarayourfriend, line 136: This name is kind of confusing :sweat_smile: I'd have a hard time understanding the difference between this an `describeEvery` without looking at the implementation.

Is this a pattern you think we'd use in lots of different e2e tests, only testing `lg` and `sm` specifically? If not, the tests themselves can call `breakpoints.describeEachBreakpoint`.

What's the goal behind testing only `lg` and `sm` instead of using `describeEvery`?

sarayourfriend (commented):
I LOVE this so far!

src/components/VTabs/VTab.vue
```diff
@@ -0,0 +1,169 @@
...
+  <Component
+    :is="as"
+    :id="`tab-${id}`"
+    ref="internalTabRef"
+    role="tab"
+    :tabindex="selected ? 0 : -1"
+    size="disabled"
+    variant="plain-dangerous"
+    v-bind="tabProps"
+    class="py-3 md:py-4 px-4 md:px-6 border-t border-x border-tx rounded-t-sm bg-white text-base font-semibold focus:shadow-[0_0_0_1.5px_#c52b9b_inset]"
```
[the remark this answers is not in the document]
obulat: I'm not sure who added this particular piece, I just copied it from the previous implementation :)

src/components/VTabs/VTab.vue
```diff
@@ -0,0 +1,169 @@
...
+import { dom } from '~/utils/dom'
+import { Focus, focusIn } from '~/utils/focus-management'
+
+export default defineComponent({
+  name: 'VTab',
+  props: {
+    as: {
+      type: String,
+      default: 'VButton',
+    },
```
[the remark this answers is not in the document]
obulat: I just removed it locally, before looking at notifications :)

test/playwright/utils/breakpoints.ts
obulat: The goal was to save some space :) And yes, I couldn't find a better name.
I removed it in the end, and used `describeEvery` instead.

src/components/VTabs/VTab.vue
```diff
@@ -0,0 +1,169 @@
...
+    id: {
+      type: String,
+      required: true,
+    },
+  },
+  setup(props) {
+    const tabContext = inject(tabsContextKey)
+    if (!tabContext) {
+      throw new Error(`Could not resolve tabContext in VTab`)
+    }
```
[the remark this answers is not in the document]
obulat: I didn't realize that hoisting also works in the transpiled Vue code. Or, actually, that the arrow functions and normal `function()`s are transpiled differently. Replacing all the `function`s with arrow functions fixed the problem.

src/components/VTabs/VTabs.vue
```diff
@@ -0,0 +1,108 @@
...
+        if (idx !== -1) panels.value.splice(idx, 1)
+      },
+    }
+    provide(tabsContextKey, tabGroupContext)
+
+    const accessibleLabel = computed(() =>
+      props.label.startsWith('#')
+        ? { 'aria-labelledby': props.label.slice(1) }
+        : { 'aria-label': props.label }
+    )
```
sarayourfriend, line 102: Clever!

src/components/VTabs/meta/VTabs.stories.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+`tab`: navigate to the tablist and then to the selected tabpanel. The tabs use 'rowing index' to only allow to tab to the currently selected tab using keyboard.
+
+`home`, `end`: set the first or the last tab as focused (or selected).
+
+## Manual plain tabs
+
+To activate the tabs using keyboard, you need to focus on the tab you want by using the `Tab` key, and then click `Space` or `Enter`.
+
+These tabs also use the second available variant, `plain`.
```
sarayourfriend, line 65: I wonder about this "manual" mode. From an accessibility standpoint, is it better or worse? Also, is there anything to indicate to the user that they need to make an additional action to activate the tab, versus the typical arrow key navigation?

src/utils/dom.ts
```diff
@@ -0,0 +1,12 @@
+// Credit: https://github.com/tailwindlabs/headlessui/
+import type { ComponentPublicInstance } from '@vue/runtime-dom'
```
sarayourfriend, line 2: TIL about this type! I've been searching for it for a while :grin:

src/utils/focus-management.ts
```diff
@@ -0,0 +1,179 @@
...
+  /** Wrap tab around */
+  WrapAround: 1 << 4,
+})
+
+export const FocusResult = Object.freeze({
+  Error: 1,
+  Overflow: 2,
+  Success: 4,
+  Underflow: 8,
+})
```
sarayourfriend, line 65: Using this mapping to numbers has the strange effect of making the `focusIn` function's return type a `number`, which, at least to me, is pretty unclear. Switching this to be strings instead of numbers would make it clearer to a user of the function what the return type is meant to indicate, I think:

```suggestion
export const FocusResult = Object.freeze({
  Error: 'Error' as const,
  Overflow: 'Overflow' as const,
  Success: 'Success' as const,
  Underflow: 'Underflow' as const,
})
```

Then the return type of `focusIn` becomes `'Error' | 'Overflow' | 'Success' | 'Underflow'` which is much clearer than `number` (though we could still probably due to document what overflow and underflow mean as it is not immediately clear to me without reading the `focusIn` function's implementation carefully :slightly_smiling_face:)

src/utils/license.ts
```diff
@@ -54,8 +54,8 @@ export const getFullLicenseName = (
  * @param license - the license slug to check
  * @returns `false` if `license` is 'cc0' or 'pdm', `true` otherwise
  */
-export const isLicense = (license: License): boolean =>
-  !(PUBLIC_DOMAIN_MARKS as ReadonlyArray<License>).includes(license)
+export const isLicense = (license: string): boolean =>
+  !(PUBLIC_DOMAIN_MARKS as ReadonlyArray<string>).includes(license)
```
sarayourfriend, line 58: I wonder if the `ReadonlyArray` annotation here is necessary? It seems like noise to me...

```suggestion
  !(PUBLIC_DOMAIN_MARKS as string[]).includes(license)
```

sarayourfriend (approved):
This also fixes #1245 by virtue of the wonderful tests you added :tada: 

This also LGTM, I tested it locally on desktop and my phone as well and all appears to be perfect. Really nice work!

I think there is some duplication with the focus management code, but honestly the code here is a bit cleaner than the stuff I implemented for the VItemGroup. It'd be nice if we could generalize the keyboard navigation bit somehow and also use it for VItemGroup.

src/utils/dom.ts
obulat: I've been searching for it, too! headlessui/vue used it.

src/components/VTabs/meta/VTabs.stories.mdx
obulat: The W3 guidelines recommend using "manual" mode if you make a network request to display the tabpanel content: https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_selection_follows_focus

When I look at the network tab, I see that both requests for the feedback forms are made almost simultaneously, and each tab is  ~15kb. I wonder if it's worth the inconsistency of tab selection experience?

src/components/VCopyButton.vue
```diff
@@ -16,49 +16,64 @@
...
+  defineComponent,
+  onBeforeUnmount,
+  onMounted,
+  ref,
+} from '@nuxtjs/composition-api'
+
 import VButton from '~/components/VButton.vue'
 
-export default {
+import type ClipboardJS from 'clipboard'
```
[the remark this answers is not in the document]
obulat: Thank you! I tried joining both imports, but since they both appear to be default exports, decided to leave as is. Didn't realize that `Clipboard` could be used as type, too.

src/components/VTabs/meta/VTabs.stories.mdx
sarayourfriend: Ah, okay! Thank you for sharing the document :slightly_smiling_face: It had not occurred to me that opening a tab could incur latency!

It's probably fine if it's part of the spec, I'd say! I was surprised that there wasn't an additional aria attribute to indicate the change in behavior to the user.

src/utils/license.ts
obulat: Without ReadonlyArray this causes TS error:
`TS2352: Conversion of type 'readonly ["pdm", "cc0"]' to type 'string[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.   The type 'readonly ["pdm", "cc0"]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.`

src/components/VTabs/meta/VTabs.stories.mdx
obulat: Yes, that difference in behavior did make me think it's kind of inconsistent. But I didn't do anything about it :)

src/components/VTabs/VTab.vue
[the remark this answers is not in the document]
zackkrida: WOW

zackkrida (approved):
This is excellent. I noticed one thing that I am not sure about. I see that there's a group with an aria-label/aria-labelledby around the _entire tab component_ (the div with `v-bind="accessibleLabel"`). It seems fine to me, but most implementations I've seen seem to have a group only around the tab_list_. Here's an example:

https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

CleanShot 2022-05-02 at 21 34 36@2x

The keyboard controls appear to work perfectly, and the experience of using this in Voiceover was quite nice. 

Regarding the 'rich text' tab, I wonder if there's an aria label we could add, or make the rendered html sr-only, to make the user experience better. Right now it reads one small piece of the attribution at a time, instead of simply reading the entire sentence, which is very unpleasant.

obulat:
>  see that there's a group with an aria-label/aria-labelledby around the entire tab component (the div with v-bind="accessibleLabel"). It seems fine to me, but most implementations I've seen seem to have a group only around the tab_list_.

Thank you for catching this, I didn't notice that the label was applied for the tablist!

Issue #1324, opened 2022-04-25
`timeFmt` called with invalid date in VWaveform


sarayourfriend:
## Description

https://sentry.io/share/issue/ea8316be37c94b7296c867797cc0dd41/

Not sure why this happens, according to the data Sentry has scraped, we are passing an actual number of seconds for the current time: `6.011812837` (this isn't visible in the shareable Sentry link, but if you have access to the full Sentry page then you can scroll down to the "vue" section and see the `propsData` that was collected)

Going to the search the error is for, `/search/audio?q=lofi` doesn't cause the error for me locally. It is happening frequently though, so I'm prioritizing as high.

## Reproduction

Not sure how to reproduce.

## Environment
According to Sentry this appears to happen almost exclusively to Macs and iPhones with Safari.

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

sarayourfriend:
@WordPress/openverse-frontend This issue requires a Mac and Safari to work on it. Could someone with the appropriate hardware/software setup take a look at this soon?

obulat:
I managed to reproduce it on Firefox, too, but can't reproduce it the second time . I searched for 'ears ringing', as in the linked issue, and on 'All content' search result page clicked on the first audio, and got a 404 page.
Also, sometimes when you click on short audios from Freesound, you get `Unexpected value NaN parsing width attribute.` warnings in the console

obulat:
`progressBarWidth` can sometimes be NaN...
https://github.com/WordPress/openverse-frontend/blob/7eac75b019f3b3e65462b26c27c82f899d9a7bcf/src/components/VAudioTrack/VWaveform.vue#L370-L373

sarayourfriend:
Aha! Good find Olga! I wonder which of those is undefined :thinking:

obulat:
There is something wrong with the way the `defaultRef` sets the duration value, @sarayourfriend:
https://github.com/WordPress/openverse-frontend/blob/7eac75b019f3b3e65462b26c27c82f899d9a7bcf/src/components/VAudioTrack/VAudioTrack.vue#L223-L233

I tried debugging, but I can't understand what and how to fix it. For some reason, by the time `localAudio` is available, the `defaultRef`'s value is already set to `NaN`.

obulat:
[...]

PR #1328, opened 2022-04-26
Stop setting audio duration when it's NaN

obulat:
## Fixes

Related to #1324 by @sarayourfriend (I'm not sure it fixes it)

## Description

`VGlobalAudioTrack` was setting audio duration when it was unavailable, so the duration was set to `NaN`.

The duration is already updated on `durationchange`. When I check the component data using Vue devtools, I see that duration for `VGlobalAudioTrack` is set correctly.

## Testing Instructions

Try opening audio search results, playing audio from "All Content" and "Audio" search result pages, from single result and from related audio. There should be no `Unexpected value NaN parsing width attribute.` warnings in the console.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VAudioTrack/VGlobalAudioTrack.vue
```diff
@@ -154,7 +154,6 @@ export default defineComponent({
         audio.addEventListener('timeupdate', setTimeWhenPaused)
         audio.addEventListener('durationchange', setDuration)
         currentTime.value = audio.currentTime
-        duration.value = audio.duration
```
sarayourfriend, line 157: I guess it would be covered by `durationchange` and `setDuration` once the file is finished loading?

I wonder if this has to do with some providers not sending the content length for the preview audio, so then the audio object only has a coherent duration once the file is actually loaded and the calculation has been made :thinking:

sarayourfriend (approved):
Seems safe to try to me!

I wonder if instead of outright skipping setting the duration from the audio object there, we could check if it is defined and only skip it if it would cause problems?

src/components/VAudioTrack/VGlobalAudioTrack.vue
obulat: Oh, that might be the reason!

obulat: I tried different audios from Jamendo and Freesound, and all of them have the duration of NaN here initially. If you click on the same audio for the second time (play one audio - then play a different audio - then go back to the first one and click play), the duration is correct.

src/components/VAudioTrack/VWaveform.vue
```diff
@@ -258,7 +258,7 @@ export default defineComponent({
      */
     const timeFmt = (seconds: number): string => {
       const date = new Date(0)
-      date.setSeconds(seconds)
+      date.setSeconds(isNaN(seconds) ? 0 : seconds)
```
sarayourfriend, line 261: Given that the way `isNaN` works can be confusing, (even the `Number.isNaN` isn't straightforward usecase for this) it might make more sense to use `Number.isFinite` and reverse the order:

```suggestion
      date.setSeconds(Number.isFinite(seconds) ? seconds : 0)
```

@sirreal and I were chatting offline the other day about potentially trying to find eslint rules for the projects we work on to prevent using the legacy `window.isNaN` and `window.isFinite` functions, neither of them works entirely as expected :cold_sweat:

dhruvkb (approved):
This is a good patch. TIL about [`window.isFinite` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite) vs [`Number.isFinite` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite).

PR #1332, opened 2022-04-26
Add bundle size reporting to PRs

### LINKED ISSUE #567: Add bundle size reporting to PRs
## Problem

Currently if we add or update a dependency, we don't have any automatic insight into how it effects our bundle sizes. Especially when adding a new dependency, this can be helpful in narrowing down options or debugging sudden unexpected increases in bundle sizes.

## Description

Let's add a bundle size analysis step to the CI similar to how Gutenberg has it. [Here is an example comment from a PR where this feedback was helpful for avoiding a sudden unexpected increase in bundle sizes](https://github.com/WordPress/gutenberg/pull/27184#issuecomment-731874775).

We can copy the strategy used by [Gutenberg in this workflow action](https://github.com/WordPress/gutenberg/blob/f06eaf26811a8a8c01602ca8091bca1c302fcb86/.github/workflows/bundle-size.yml#L1). 

The [preactjs@compressed-size](preactjs/compressed-size-action) action should still work fine for us even though we're not using preact or React. It just does a Webpack bundle analysis to get the data.

If for whatever it doesn't work for us then we'll need to explore a bespoke strategy. I tried searching around for `nuxt` specific solutions but there don't appear to be any but we could build off of `nuxt build --analyze --spa --no-generate` and use the generated json stats output in the `./.nuxt/stats` folder.

## Implementation

sarayourfriend:
## Fixes

Fixes #567 by @sarayourfriend 

## Description

Uses the example from Gutenberg and uses the preact action used there as well. Hopefully it works :grin: 

Update: it works!

Note that the list of files in this PR is going to be super long because I've also changed how the bundle names are generated.

Maybe we could actually only change the way the bundle names are generated during this specific build by passing an environment variable or something. What do y'all think about that? Then we wouldn't have to worry about hashes changing at all, we'd just see names 1:1? I'll try it out and see if it works.

## Testing Instructions

Check out the new workflow and confirm that it is working :tada: 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -225,6 +225,24 @@ const config: NuxtConfig = {
         dst: 'index.js',
       },
     ],
+    filenames: {
+      app: ({ isDev, isModern }) =>
+        isDev
+          ? `[name]${isModern ? '.modern' : ''}.js`
+          : `[name].[contenthash:7]${isModern ? '.modern' : ''}.js`,
+      chunk: ({ isDev, isModern }) =>
+        isDev
+          ? `[name]${isModern ? '.modern' : ''}.js`
+          : `[name].[contenthash:7]${isModern ? '.modern' : ''}.js`,
+      css: ({ isDev }) =>
+        isDev ? '[name].css' : 'css/[name].[contenthash:7].css',
+      img: ({ isDev }) =>
+        isDev ? '[path][name].[ext]' : 'img/[name].[contenthash:7].[ext]',
+      font: ({ isDev }) =>
+        isDev ? '[path][name].[ext]' : 'fonts/[name].[contenthash:7].[ext]',
+      video: ({ isDev }) =>
+        isDev ? '[path][name].[ext]' : 'videos/[name].[contenthash:7].[ext]',
+    },
     friendlyErrors: false,
     postcss: {
       plugins: {
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -108,6 +108,26 @@ const head = {
   ],
 }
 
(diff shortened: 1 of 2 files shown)


sarayourfriend:
> Total Size: 0 B

Hmmm, that doesn't seem right :frowning_face:

sarayourfriend:
I think it looks good with the file names for each bundle instead of the hashes. What do y'all think? Theoretically this shouldn't change the build output at all!

nuxt.config.ts
```diff
@@ -108,6 +108,26 @@ const head = {
   ],
 }
 
+const baseProdName = process.env.CI ? '[name]' : '[contenthash:7]'
+
+const filenames: NonNullable<NuxtConfig['build']>['filenames'] = {
```
obulat, line 113: Awesome! It is definitely much clearer this way!

obulat (approved):
Thank you for adding this, it's really important!

sarayourfriend:
@dhruvkb do you have time to review this this week? It'd be a nice improvement to devex if we could start tracking this, especially when adding or removing dependencies.

dhruvkb (approved):
I didn't realise it was ready for a review, the PR description mentioned some things you were planning to try. Nonetheless, this is great and will be very helpful to track changes to our bundle sizes.

sarayourfriend:
Ah sorry, subsequent comments added details about that and I never updated the PR description.

PR #1349, opened 2022-04-28
Await fetching related media in the single result store

### LINKED ISSUE #1339: Related results never load for SSR
## Description

In some cases, the "related results" loader spins continuously and never loads results.

## Reproduction

1. Visit https://search-staging.openverse.engineering/ru/audio/a56ff303-9983-4c21-a7f4-e1e454153ebe
2. Observe that related results never load

## Environment

 - Device: Mobile & Desktop
 - OS: Android & Ubuntu
 - Browser: Firefox

## Resolution

obulat:
## Fixes

Fixes #1339 by @AetherUnbound 

## Description

This PR adds `await` to ensure that related media are fetched on SSR as well as on CSR.
The related fetch call is moved after the main single result is fetched successfully.

## Testing Instructions

Try loading a single result page on SSR, or opening a single result page and refreshing. The related media should be there.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal (commented):
I was wondering why the call to `useRelatedMediaStore().fetchMedia()` can't be async so inspecting Pinia in the single results page I found that the `single-result` store does **not persist or it is not initialized**, it's not shown in the list on Vue Devtools.

CleanShot 2022-04-29 at 13 56 14@2x

So making the call from the `setup` function instead of the `single-result` store seems to allow it to work asynchronously.

obulat:
Thank you for catching this, @krysal ! 
Actually, I guess you cannot see the single-result store because the call from it is done inside the `asyncData`, not in the `setup` function. Even though you couldn't see the single-result store on the first load, it **_was_** fetching the media.
 However, I found a problem with the first implementation here: I moved the `fetchRelatedMedia` call to after the `fetchMediaItem`. This meant that if we were reusing the media object (when you navigate from the search results or open an image from the related media), `fetchRelatedMedia` wasn't being called at all! 
 Trying to fix it, I actually converted the detail pages to use `defineComponent`, composition API and TypeScript. I did not add them to `tsconfig` because they depend on the Reakit composables, which currently cause errors because they are using React types.

sarayourfriend (approved):
LGTM! :tada:

src/stores/media/single-result.ts
```diff
@@ -49,32 +59,48 @@ export const useSingleResultStore = defineStore('single-result', {
       return mediaItem
     },
 
-    async fetchMediaItem(type: SupportedMediaType, id: string) {
+    /**
+     * If the result is available in the media store (from search results or related media),
+     * we can re-use it. Otherwise, we fetch it from the API.
+     * The media objects from the search results currently don't have filetype or filesize properties.
```
krysal, line 65: Why reuse it? 🤔 I'm not sure if we gain much from this optimization as the search endpoint doesn't return all the fields we need in the single result page.

krysal (approved):
💯+ for more conversion to Composition API. It would be nice to add some tests but this effectively fixes the problem ☑️

sarayourfriend:
Did fixes #1377 :grin:

dhruvkb:
@sarayourfriend does it? I can't reproduce #1377 on `main` either.

src/stores/media/single-result.ts
obulat: I didn't know that the search endpoint doesn't return all the fields when I implemented this. I think there are only `filetype` and `filesize` fields that are not returned currently. I think re-using still improves the performance of single result loading.

obulat:
> Yes, that's my suspicion, that the Pinia instance from the `asyncData` call does not persist, since if we don't keep the `await` for the related media then the call will not complete before `asyncData` ends and the data will be lost. That is why I still think we can avoid waiting for the fetch of the related media if it's done from the `setup()` function in each `_id.vue` page, but this also works.

The state data from the server is saved to `window.$nuxt` and is re-used when the Pinia store is initialized on the client (for the options stores, not the setup ones :) ). Getting the order of the calls of `asyncData` and `fetch` is kind of tricky. I was really surprised to know that `asyncData` for the single result pages is run before you click on the search link because it's prefetched. I'm not sure this is relevant here, though.

Let's see how it goes now, and, if necessary, we can refactor later.

PR #1372, opened 2022-05-02
WIP - Add Vue Ariakit implementation

sarayourfriend:
## Description

This is a big big big WIP that probably won't get merged for a long time, but I wanted to share what I'd done so far. Maybe this is something we can do iteratively as a team now that there's a rough pattern established for how to convert Ariakit stuff from React hooks to Vue composition API.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/ariakit/README.md
+++ b/src/ariakit/README.md
@@ -1,0 +1,7 @@
+# Ariakit vendorization
+
+Due to [Ariakit's](https://github.com/ariakit/ariakit) current React focus, the published code pulls in the `@types/react-dom` package, which clashes with the `@vue/runtime-dom` types for "JSX" (templates). Because of this, we cannot use the published ariakit/reakit packages directly and have had to vendorize them here.
+
+The process of adding these files was literally a copy paste job. Thanks to Ariakit's MIT license, this is safe to do. This README and the title of the directory serves as attribution for the original code.
+
+Additionally, there are several composables in this directory which are adapted from ariakit React hooks. These are _not_ vendorized from ariakit directly but because they are essentially copied from ariakit, we've included them in this directory to consistently indicate where the original code is from.
--- a/src/ariakit/dialog/use-dialog-heading.ts
+++ b/src/ariakit/dialog/use-dialog-heading.ts
@@ -1,0 +1,46 @@
+import {
+  Ref,
+  ref,
+  computed,
+  watch,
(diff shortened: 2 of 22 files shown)


krysal:
Does it need to be changed all at once? Is there a way we can do it incrementally?

sarayourfriend:
We can definitely do it incrementally. The only reason this one got so big was because I chose a place to start that had lots of internal dependencies. I think this was a good way to start because it forced me to reckon with different patterns that exist in Ariakit that we would have to solve with Vue analogues. In that sense this PR is probably good to keep around but we can start at the smaller parts of this PR (which we probably wouldn't be using directly).

It _may_ be helpful for me or someone else to check out this PR and explicitly document the patterns I'm talking about here. Some things like how I decided to convert Ariakit's "state" objects into Vue reactives, the prop type shapes, other concepts like the `listeners`, `attributes` and `classes` standard pattern for the hook return values that allows them to compose.

One problem that still needs to be solved, however, is multiple ordered dispatch of events through hooks that all define the same listener. For example, if you are composing multiple hooks and more than one of them define a `click` handler, we'll need to recreate the behavior `stopPropagation` in an order dependent way. This could be as easy as looping over the list of handlers and breaking the loop once `stopPropagation` has been called on the event. Having a standard `compose` type function that composes a list of hook return values (i.e., it knows how to combine `attributes`, `listeners` and `classes` in a reactive way) would be a good way to do that.

What do you think @krysal? I've also chatted briefly with @diegohaz, the maintainer of Ariakit sometime earlier this year about contributing a Vue implementation. Working in Ariakit to make the baseline changes necessary to make Ariakit able to also house a Vue implementation (this mostly comes down to removing React as a base dependency from the base Ariakit packages) could be another good path to follow for this. It would just benefit us less immediately because it would take longer to be able to use that (much more up-front work). Having a standalone Vue implementation might be fine for now though... however Ariakit has undergone some big refactors. Maybe it'd be better to wait until we know Ariakit's API is more stable, even though there will inherently be differences between the React and Vue (and any other framework) version.

sarayourfriend:
Closing this PR. We can and should continue this work but in smaller discrete chunks as Krystle suggested.

PR #1389, opened 2022-05-04
Replace `lodash.sortby` and fix sources table ordering by url

### LINKED ISSUE #1108: Replace `lodash.sortby` with `Array.prototype.sort`
## Problem

We are looking to remove dependency on lodash altogether (see milestone description).

## Description

Replace all usages of `lodash.sortby` with [`Array.prototype.sort`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort).

Please note that `Array.prototype.sort` is not a pure function and mutates the array. Be sure to copy the array before sorting it with native sort.

## Implementation

### LINKED ISSUE #1388: Domain sorting in sources table doesn't take the clean URL
## Description

When sorting the table by the **Domain** field it doesn't appear like it's sorted because the `lodash.sortby` function uses the raw URL instead of the cleaned version, which is the one shown.

## Reproduction

1. Go to the /sources page (in wp.org/sources, [staging](https://search-staging.openverse.engineering/sources) or local)
2. Click on the "**Domain**" header cell of the table
3. See "clevelandart.org " after "postalmuseum.si.edu" among other misplaced URLs.

## Resolution

krysal:
## Fixes

Fixes #1108 by @sarayourfriend 
Fixes #1388 by @krysal

## Description

This PR completes the **Remove lodash** milestone by replacing `lodash.sortby` with `Array.prototype.sort` as Sara suggested. There is still a `lodash.isempty` dependency that I'm not sure if it can be safely removed, doesn't seem to be used anywhere.
 
Also fixes the sorting of the Sources table, something I noticed while working in the first issue :)

## Testing Instructions

Reproduce steps in #1388 and see URLs correctly sorted.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,6 @@
     "focus-visible": "^5.2.0",
     "glob": "^7.1.6",
     "lodash.isempty": "^4.4.0",
-    "lodash.sortby": "^4.7.0",
     "node-html-parser": "^5.3.3",
     "nuxt": "^2.15.4",
     "pinia": "^2.0.13",
--- a/package.json
+++ b/package.json
@@ -83,7 +83,6 @@
     "focus-trap-vue": "1.1.1",
     "focus-visible": "^5.2.0",
     "glob": "^7.1.6",
-    "lodash.isempty": "^4.4.0",
     "node-html-parser": "^5.3.3",
     "nuxt": "^2.15.4",
     "pinia": "^2.0.13",
--- a/src/components/VSourcesTable.vue
+++ b/src/components/VSourcesTable.vue
@@ -57,7 +57,6 @@
 </template>
 
 <script lang="ts">
-import sortBy from 'lodash.sortby'
 import {
   computed,
   defineComponent,
@@ -69,6 +68,7 @@ import { useProviderStore } from '~/stores/provider'
 import { useGetLocaleFormattedNumber } from '~/composables/use-get-locale-formatted-number'
 
 import type { SupportedMediaType } from '~/constants/media'
+import type { MediaProvider } from '~/models/media-provider'
 
 import TableSortIcon from '~/components/TableSortIcon.vue'
 import VLink from '~/components/VLink.vue'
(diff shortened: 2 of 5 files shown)


src/components/VSourcesTable.vue
```diff
@@ -110,15 +111,28 @@ export default defineComponent({
...
       return removeAfterSlash(stripLeadingWww(stripProtocol(url)))
     }
+
     const getLocaleFormattedNumber = useGetLocaleFormattedNumber()
     const providerStore = useProviderStore()
 
+    function compareProviders(prov1: MediaProvider, prov2: MediaProvider) {
+      let field1 = prov1[sorting.field]
+      let field2 = prov2[sorting.field]
+      if (sorting.field == 'source_url') {
```
zackkrida, line 121:
```suggestion
      if (sorting.field === 'source_url') {
```

src/components/VSourcesTable.vue
```diff
@@ -90,10 +90,10 @@ export default defineComponent({
   setup(props) {
     const sorting = reactive({
       direction: 'asc',
-      field: 'display_name',
+      field: 'display_name' as keyof Omit<MediaProvider, 'logo_url'>,
```
obulat, line 93: I wonder if we should remove the provider `logo_url` parameter. Do we use it anywhere?

obulat (commented):
The bundle size change is 🔥
I will try to approve this tomorrow, after running and making sure it works.

package.json
```diff
@@ -84,7 +84,6 @@
     "focus-visible": "^5.2.0",
     "glob": "^7.1.6",
     "lodash.isempty": "^4.4.0",
```
sarayourfriend, line 86: It should be safe to remove this! It's not explicitly imported by anything. If it's needed by another dependency it'll still get included in the lock file. It's not hurting our bundle for it to be here, but someone might accidentally use it :stuck_out_tongue:

sarayourfriend (approved):
LGTM! Works perfect :tada: 

Once this is merged we can close the "remove lodash" milestone! Congrats @krysal!

zackkrida (approved):
LGTM!

obulat (approved):

src/components/VSourcesTable.vue
krysal: No, I don't think we are showing the sources logo anywhere, though it was the intention in CC to show them on the homepage if I recall correctly? we can probably remove them!

PR #1391, opened 2022-05-05
Break long searched words if necessary

### LINKED ISSUE #1293: Long search term breaks the layout of the search page
## Description

Try to search for a very long word on 'All Content' view. If the word is longer than the page width, a horizontal scroll bar appears. 

## Reproduction

1. Try to search for a very long search term on 'All Content' view, such as ' %22Claude%20Monet%22 '
2. If the word is longer than the page width, it doesn't fit and a horizontal scroll bar appears. 

## Screenshots

Screen Shot 2022-04-20 at 8 45 02 AM

## Additional context

Found it when trying to fix the links on the Search Guide page. The specific links should also probably be url-unencoded. The screenshot also shows a broken search, of course, because the search result and the search term do not match 😆 

## Resolution

rafaelytakei:
## Fixes

Fixes #1293 by @obulat 

## Description
If the user tried searching for a long word, for example `dog -dskaodkadokaosdadadasdadadasdsadadsad`, the long word would generate a horizontal overflow. The same thing happened on the 'no results' page.

## Testing Instructions
Try searching for something with a long word in it. I recommend using something like `dog -dskaodkadokaosdadadasdadadasdsadadsad` because it's easier to write whatever as a long excluding word, and check if there is a horizontal overflow on the page. (check issue for sample screenshot)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VErrorSection/VNoResults.vue
+++ b/src/components/VErrorSection/VNoResults.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="no-results text-center md:text-left">
-    <h5 class="text-5xl">
+    <h5 class="text-5xl break-words">
       {{ $t('no-results.heading', { query: query.q }) }}
     </h5>
     <h6 class="text-3xl font-normal md:font-bold mt-10 md:mt-16">
--- a/src/components/VSearchResultsTitle.vue
+++ b/src/components/VSearchResultsTitle.vue
@@ -1,6 +1,6 @@
 <template>
   <h1
-    class="leading-tight sr-only md:not-sr-only"
+    class="leading-tight sr-only md:not-sr-only break-words"
     :class="[size === 'large' ? 'text-[90px]' : 'text-6xl']"
   >
     <slot default />


sarayourfriend:
image

Great, simple fix, thank you again @rafaelytakei :slightly_smiling_face: The screenshot above could be fixed as well though, if you get the "no results found" page, it seems like the picture of the sad pianist should have a `min-width` set on it. Potentially we'd want to truncate the line-broken long query to 3 lines or something so that it doesn't push the rest of the text on the page far down.

krysal (commented):
Thanks for jumping into this issue @rafaelytakei!

I've thought this could be truncated in desktop view as it seems to take too much space; see the screenshots below. I would not recommend that for mobile (tablet size and down) since the whole text will not be visible in the title or search box, but I wonder what @panchovm would recommend here.

### Desktop truncating

search-desktop-truncate

### Desktop breaking words

search-desktop-break-words

### Tablet-like truncating

search-tablet-truncate


### Tablet-like breaking words

search-tablet-break-words


For now, @rafaelytakei would you mind rebasing this to update with `main`?  Just run:

```
git rebase main
```

panchovm:
Thanks for bringing this up.

In any case, I would not truncate the text as it is crucial to see what was the search term about. Regarding the heading style, I like having it big and bold as it strengthens the brand, especially on the All content results where the brand yellow color also stands out.

Time ago, I was thinking of having an adaptable size for this heading based on its container. Something like [this example](https://css-tricks.com/examples/ViewportTypography/) but where we define a minimum size for short sentences like `Cat`, but as it gets bigger, the font size becomes smaller.

What do you think of that?

sarayourfriend:
> Time ago, I was thinking of having an adaptable size for this heading based on its container. Something like [this example](https://css-tricks.com/examples/ViewportTypography/) but where we define a minimum size for short sentences like Cat, but as it gets bigger, the font size becomes smaller.

Is this possible with pure CSS? :thinking: I'd be worried that if it requires JavaScript to adjust the font-size the SSR rendering of the page could look _very_ different.

zackkrida:
@panchovm what do you think of the following:

1. Automatically break very long words
2. Reduce the line height of the title

That produces a result like this:

CleanShot 2022-05-30 at 12 10 00@2x

zackkrida (approved):
This PR simply adds `break-words` to the page title, and therefore I think it can be merged as-is even if we need to add more changes later. Could another @WordPress/openverse-frontend developer give this a review this week?

PR #1394, opened 2022-05-06
Tab order

### LINKED ISSUE #989: Opening the filter sidebar should focus the filter sidebar
## Description

Activating the filter sidebar should move the user's focus to the filter sidebar.

## Reproduction

1. Enable the filter sidebar
2. Observe that the page focus has _not_ shifted to the filter sidebar, and that one must tab through all search results to finally access the filters.

## Expectation

After pressing the filter button, the users focus should be moved to the filters sidebar. Likely the first element within the sidebar, or perhaps some kind of aria region, but this will require some research.

## Resolution

obulat:
## Fixes

Fixes #989 by @zackkrida

## Description

This PR automatically focuses the filters sidebar when the user presses <kbd>Tab</kbd> key when the Filters button is pressed.
~~I will also add automatic focusing of the sidebar when the button is clicked.~~ Do we really need this? I think to add it, we would need to do something like `Vue.nextTick`, because when you click the filters button, the sidebar is not yet rendered.
When the user presses <kbd>Tab</kbd> key when on the last focusable item on the sidebar, the focus moves to the first focusable element in the `main` element. When the user pressed <kbd>Shift</kbd><kbd>Tab</kbd> on the first focusable element in the sidebar (whether it's a checkbox or the 'Clear filters' button), the focus goes to the Filters button.

Also, when the user presses <kbd>Tab</kbd> on the last focusable element in the main search page (one of the Meta Search buttons, or the scroll to top button), the focus moves to the first focusable element in the layout instead of moving to the filters sidebar again.

## Testing Instructions

Open the app on desktop. Try Tabbing from the Filters button, and tabbing through the filters sidebar to the main element. Try All content, Image and Audio views.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VFilters/VSearchGridFilter.vue
```diff
@@ -53,11 +52,17 @@ import { kebab } from 'case'
...
+import VButton from '~/components/VButton.vue'
 
-export default {
+export default defineComponent({
   name: 'VSearchGridFilter',
   components: {
+    VButton,
     VFilterChecklist,
   },
   setup() {
```
sarayourfriend, line 68: Can you add the missing `emits` as well, given this is being converted to TS?

src/components/VFilters/VSearchGridFilter.vue
```diff
@@ -98,16 +105,47 @@ export default {
...
+     */
+    const lastFocusableElement = ref<HTMLElement>(null)
+    const updateLastFocusable = () => {
+      const focusable = getFocusableElements(filtersFormRef.value)
+      lastFocusableElement.value = focusable[focusable.length - 1]
+    }
+    onMounted(() => {
+      updateLastFocusable()
+    })
+    watch(filterTypes, () => updateLastFocusable())
```
sarayourfriend, line 121: Couldn't this be a computed?

src/components/VFilters/VSearchGridFilter.vue
```diff
@@ -98,16 +105,47 @@ export default {
...
+    /**
+     * When the user presses 'Tab' on the last focusable element, we need to
+     * move focus to the first focusable element in main.
+     * @param event
+     */
```
sarayourfriend, line 126: So in addition to the last focusable element... we also need the _first_ focusable element to shift focus to the filter button when `shift + tab` is pressed.

Currently if you tab into the filter sidebar from the filter button (works great, amazing!) but then press <kbd>Shift</kbd> + <kbd>Tab</kbd> you end up on the 'back to top button'

sarayourfriend (commented):
This is working great already! Nice work, glad to get this a11y issue resolved :tada:

obulat:
> There's one last quirk but I'm not sure how to address it or if it needs to be addressed. IF you have the filter sidebar open and tab over to the results, then shift tab back up, you never reach the filter sidebar, instead you jump straight to the header and have to tab to the filter button and then tab into the filter sidebar.
> 
> I'm not sure whether this needs to be changed, but it _might_ be more contextually confusing if you remembered that the order went "filter button -> filter options -> results" but then going backwards the filter options are skipped. In fact, it might lead you to assume the filter options were _closed_ and then have a moment of confusion when you land on the filter button again and find that it's still reporting itself as pressed.
> 
> But that's just a guess. I'm not sure if that's what users would expect or not.

I was afraid that I missed one of the scenarios. Thank you for describing it!

zackkrida (approved):
This is working great! At first I was surprised with the functionality, expecting that pressing enter on the filter button would then shift focus to the first filter, and that tabbing from the button would continue through the regular document flow into the results themselves, but the behavior here is consistient with google search and Pexels, both which have fairly similar filter UIs.

PR #1403, opened 2022-05-10
Fix incorrect text styles

### LINKED ISSUE #1365: Incorrect text styles on image single result page on mobile
## Description

The single image result page has different text styles.

This issue is blocked until @panchovm confirms the necessary changes.

@panchovm , Could you please update this issue, correcting any wrong todos here? I copied the styles with their names from the Figma mockup, but the styles with the same name have different values 🤷
- Update only for mobile (below `md` screen)?
- Update the  _Go to image's website_ button text style to **Caption bold**:
```
font-family: Inter;
font-size: 12px;
font-weight: 600;
line-height: 16px;
```
- Update the image title text style to **Description bold**:
```
font-family: Inter;
font-size: 14px;
font-weight: 600;
line-height: 18px;
```
- Update the image author text style to **Description bold**:
```
font-family: Inter;
font-size: 14px;
font-weight: 600;
line-height: 18px;
```

## Reproduction

1. Go to the single result view of an image (eg. https://search-staging.openverse.engineering/image/f9384235-b72e-4f1e-9b05-e1b116262a29).
2.  The text styles of CTA (_Go to image's website_ button), image title, and image author all look different. This feels odd and inconsistent.

## Additional context

Figma mockup of the single image result page: https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Audio-Release?node-id=958%3A19663

## Resolution

rutterjt:
Fix incorrect text styles on image single result page to match figma design.

## Fixes

Fixes  #1365 by @obulat

## Description

Updates image author text to have `font-size: 14px` and `line-height: 18.2px`, to be in line with figma design.

## Checklist

- N/A I added or updated tests for the changes I made (if applicable).
- N/A I added or updated documentation (if applicable).

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat:
Thank you for your contribution!

You are right that the desktop (normal) version has correct text styles. The problem is in the mobile styles, and the correct values for mobile are listed in the issue: https://github.com/WordPress/openverse-frontend/issues/1365#issue-1222649573

So, you will need to prefix the current text style values with `md:`, and add the new (mobile) ones before them.

src/pages/image/_id.vue
```diff
@@ -45,7 +45,7 @@
           v-if="image.creator"
           path="image-details.creator"
           tag="span"
-          class="font-semibold leading-[130%]"
+          class="font-semibold leading-[130%] text-base"
```
obulat, line 48:
```suggestion
          class="leading-snug"
```
Could you add `font-semibold text-base` classes to the outer span (line 40), so that we don't have to repeat it for all children? We can then remove `text-base font-semibold` from the `h1` with image title. It would also be nice to replace the outer span (line 40) with a `div` because you are not supposed to have block element inside a `span`.

obulat (changes requested):

rutterjt:
Hi @obulat, I'm sorry for taking so long to get back to you. I was on vacation and just got back last night. I just made the changes now!

Just to verify, I believe from @panchovm's comment in #1365, some of the text styles were previously fixed in a commit since the issue was filed, so I think all the text styles should be correct now. But if there are any problems, let me know and I'll be happy to fix them!

sarayourfriend:
I believe these problems also exist on the audio single result page. For example, the "Audio information" header is the incorrect size. Could we port these fixes over there as well?

rutterjt:
> I believe these problems also exist on the audio single result page. For example, the "Audio information" header is the incorrect size. Could we port these fixes over there as well?

Sure thing! I’ll take a look!

sarayourfriend:
@rutterjt while you're at it, if you notice any opportunities for consolidating the two single results pages, please feel free to share those here or create additional issues. As we add additional media types it'll be critical to not have to duplicate efforts like this in the future. No pressure if you're low on time though :+1:

rutterjt:
> @rutterjt while you're at it, if you notice any opportunities for consolidating the two single results pages, please feel free to share those here or create additional issues. As we add additional media types it'll be critical to not have to duplicate efforts like this in the future. No pressure if you're low on time though +1

I compared the figma design with the audio single results page, and I think I caught all the inconsistencies!

I would help, but I don't think I feel comfortable enough with Vue/Nuxt yet to take a stab at that. I'm mostly a React developer, but I use Tailwind all the time, so I'm comfortable making styling changes and smaller fixes, but less sure about making more substantial changes to the Vue/Nuxt codebase. But in the future when I'm more familiar with Vue, I'll be happy to help more with other things! :grin:

obulat (approved):
Thank you for your contributions, @rutterjt! 🚀

zackkrida:
@panchovm could you review this PR since it is focused on making text sizing changes to better match the Figma mockups?

panchovm (changes requested):
`Caption bold` in CTA and `description bold` in title are correct, but `description bold` in author's name has a slightly bigger `line-height`.

I also noticed the line-height value in CTA and author's name have no unit, while title uses `%`. I am concerned that this might impact consistency as a future change will not update all text elements.

rutterjt:
> `Caption bold` in CTA and `description bold` in title are correct, but `description bold` in author's name has a slightly bigger `line-height`.

Thanks for the feedback! I made the changes and checked the computed styles in the browser, and I believe everything should be fixed now.

> I also noticed the line-height value in CTA and author's name have no unit, while title uses `%`. I am concerned that this might impact consistency as a future change will not update all text elements.

That makes sense! I oped to convert all of the percentages to unit-less values, since according to MDN unit-less values are the preferred way of setting line-height, and using percentages can cause [unexpected inheritance issues](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values). But let me know if you'd like me to go the other way, or if there are any other issues that need to be addressed!

sarayourfriend:
> That makes sense! I oped to convert all of the percentages to unit-less values, since according to MDN unit-less values are the preferred way of setting line-height, and using percentages can cause [unexpected inheritance issues](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values). But let me know if you'd like me to go the other way, or if there are any other issues that need to be addressed!

FWIW I think this makes sense from an engineering perspective. @panchovm when we go to update text sizes again in the future, we can make sure to check the line-heights as well.

Ideally this wouldn't be an issue at some point in the near future it'd be nice to have a `VHeading` and `VText` primitive components that just take `size` or `variant` props and are used across the board so that it's trivial for us to update all headings of a particular style and size at the same time.

Opened an issue to create these component primitives based on the Figma typography stuff here: https://github.com/WordPress/openverse-frontend/issues/1489

panchovm:
Oh, the unit-less approach is interesting. I did not know about its impact. My main concern was consistency and a11y, but it should be fine as the post shared shows. Thank you both.

src/pages/image/_id.vue
```diff
@@ -33,20 +33,17 @@
...
+        class="btn-main flex-initial w-full md:w-max mb-4 md:mb-0 leading-[1.3]"
         size="large"
         >{{ $t('image-details.weblink') }}</VButton
       >
-      <span class="flex-1 flex flex-col justify-center">
-        <h1 class="text-base md:text-3xl font-semibold leading-[130%]">
+      <div
+        class="flex-1 flex flex-col justify-center font-semibold text-base leading-[1.3]"
+      >
+        <h1 class="md:text-3xl">
```
zackkrida, line 43:
```suggestion
        <h1 class="md:text-3xl font-semibold">
```

I'm noticing that the `h1` isn't inheriting the semibold `600` weight from the parent div, so it it needs this class specified.

panchovm (approved):
It looks great ✨

sarayourfriend:
I'm working on updating the snapshots for this.

PR #1410, opened 2022-05-11
Put external sources behind feature flag

sarayourfriend:
## Fixes

Also fixes a minor bug on the preferences page where switchable features using per-environment configuration were not being displayed as such due to the page checking the flag status directly instead of using `featureState` which handles the state-object configuration.

## Description

Fixes the bug mentioned above.

Adds `isOn` getter to simplify the majority of use cases for the feature flag store.

Removes the `isDev` check for the external/additional sources in the search types header dropdown.

## Testing Instructions

Checkout the branch and run `pnpm dev`. Visit the site locally and confirm that the external sources are not present in the header dropdown. Navigate to `/preferences` and toggle the `external_sources` flag on. Go back to the search results page and confirm that the external sources in the header dropdown are now present.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VContentSwitcher/VSearchTypes.vue
+++ b/src/components/VContentSwitcher/VSearchTypes.vue
@@ -40,7 +40,6 @@ import { computed, defineComponent } from '@nuxtjs/composition-api'
 
 import { supportedSearchTypes } from '~/constants/media'
 import useSearchType from '~/composables/use-search-type'
-import { useFeatureFlagStore } from '~/stores/feature-flag'
 
 import VItemGroup from '~/components/VItemGroup/VItemGroup.vue'
 import VSearchTypeItem from '~/components/VContentSwitcher/VSearchTypeItem.vue'
@@ -71,22 +70,23 @@ export default defineComponent({
   setup(props, { emit }) {
     const content = useSearchType()
 
-    const contentTypeGroups = [
-      {
-        heading: 'heading',
-        items: content.types,
-      },
-    ]
+    const contentTypeGroups = computed(() => {
+      const base = [
+        {
+          heading: 'heading',
+          items: content.types,
+        },
+      ]
 
-    const featureFlagStore = useFeatureFlagStore()
+      if (content.additionalTypes.value.length) {
+        base.push({
+          heading: 'additional',
+          items: content.additionalTypes.value,
+        })
+      }
 
-    if (featureFlagStore.isOn('external_sources')) {
-      content.additionalTypes = ['model_3d']
-      contentTypeGroups.push({
-        heading: 'additional',
-        items: content.additionalTypes,
-      })
-    }
+      return base
+    })
 
     const bordered = computed(() => props.size === 'small')
     const handleClick = (item) => {
--- a/src/components/VContentSwitcher/VSearchTypes.vue
+++ b/src/components/VContentSwitcher/VSearchTypes.vue
@@ -38,10 +38,9 @@
 <script>
 import { computed, defineComponent } from '@nuxtjs/composition-api'
 
-import { isDev } from '~/utils/node-env'
-
 import { supportedSearchTypes } from '~/constants/media'
 import useSearchType from '~/composables/use-search-type'
+import { useFeatureFlagStore } from '~/stores/feature-flag'
 
 import VItemGroup from '~/components/VItemGroup/VItemGroup.vue'
 import VSearchTypeItem from '~/components/VContentSwitcher/VSearchTypeItem.vue'
@@ -79,12 +78,9 @@ export default defineComponent({
       },
     ]
 
-    /**
-     * @todo This is for testing purposes only! We may want a different abstraction here;
-     * For example having all content types under `content.types` and making them filterable,
-     * like `const additional = [...content.types.filter(i => i.status === ADDITIONAL)]`.
-     */
-    if (isDev) {
+    const featureFlagStore = useFeatureFlagStore()
+
+    if (featureFlagStore.isOn('external_sources')) {
       content.additionalTypes = ['model_3d']
       contentTypeGroups.push({
         heading: 'additional',
--- a/src/composables/use-search-type.ts
+++ b/src/composables/use-search-type.ts
@@ -32,9 +32,7 @@ export default function useSearchType() {
     featureFlagStore.isOn('external_sources') ? ([MODEL_3D] as const) : []
   )
   const searchTypes = [...supportedSearchTypes]
-  const setActiveType = (
-    searchType: SupportedSearchType | typeof additionalTypes['value'][number]
-  ) => {
+  const setActiveType = (searchType: SupportedSearchType) => {
     if (previousSearchType.value === searchType) return
     useSearchStore().setSearchType(searchType as SupportedSearchType)
     previousSearchType.value = searchType as SupportedSearchType
--- a/src/composables/use-search-type.ts
+++ b/src/composables/use-search-type.ts
@@ -2,6 +2,7 @@ import { computed, ref } from '@nuxtjs/composition-api'
 
 import {
   supportedSearchTypes,
+  SupportedSearchType,
   ALL_MEDIA,
   AUDIO,
   IMAGE,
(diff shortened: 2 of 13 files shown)


obulat (changes requested):
Can we move the search types declaration to `constants/media.ts` as in this PR: https://github.com/WordPress/openverse-frontend/pull/1285

That PR was almost ready, but I got stuck on disabling the Filters button and hiding the Sidebar, so I need more time getting it to work. But the logic of types, I think, is better done there.

sarayourfriend:
@obulat I like the fixes you made to the approach, but the problem is that the conditional is at the top-level of a module and therefore outside the Nuxt context. It's also static, meaning it can't have a different value per request.

I'm imagining we might need a store to hold the media types rather than having them be as constant, so that we can also check the feature flag. That will complicate a lot more code though.

src/components/VContentSwitcher/VSearchTypes.vue
```diff
@@ -79,12 +78,9 @@ export default defineComponent({
...
-    /**
-     * @todo This is for testing purposes only! We may want a different abstraction here;
-     * For example having all content types under `content.types` and making them filterable,
-     * like `const additional = [...content.types.filter(i => i.status === ADDITIONAL)]`.
-     */
-    if (isDev) {
+    const featureFlagStore = useFeatureFlagStore()
+
+    if (featureFlagStore.isOn('external_sources')) {
       content.additionalTypes = ['model_3d']
```
obulat, line 84: `content` here refers to the return from the composable:

```
const searchTypes = [...supportedSearchTypes]


export default function useSearchType() {
  const activeType = computed(() => useSearchStore().searchType)
  const previousSearchType = ref(activeType.value)
  const setActiveType = (searchType) => {
    if (previousSearchType.value === searchType) return
    useSearchStore().setSearchType(searchType)
    previousSearchType.value = searchType
  }
  return {
    setActiveType,
    activeType,
    types: searchTypes,
    icons,
  }
}
```
And I can't see how we can add an `additionalTypes` property to a composable return object. Will it be available to other places where it's used?

sarayourfriend:
You're right, it won't. Will my suggestion here work? https://github.com/WordPress/openverse-frontend/pull/1410#discussion_r870499856

src/components/VContentSwitcher/VSearchTypes.vue
obulat: That's a good idea!
It's this part that we need to change:
`const searchTypes = [...supportedSearchTypes]`
It's currently set outside of the composable to keep the global state.

sarayourfriend: We'll need to update all the code that uses that I guess to call the composable instead?

sarayourfriend:
@obulat I pushed the change to move the check into the store. We still need to decide how to handle the _rest_ of the code though that imports `supportedSearchTypes` directly :thinking: Should all of those be refactored to use the composable? :thinking:

src/composables/use-search-type.ts
```diff
@@ -21,20 +23,28 @@ const icons = {
...
+    featureFlagStore.isOn('external_sources') ? ([MODEL_3D] as const) : []
+  )
+  const searchTypes = [...supportedSearchTypes]
+  const setActiveType = (
+    searchType: SupportedSearchType | typeof additionalTypes['value'][number]
+  ) => {
     if (previousSearchType.value === searchType) return
-    useSearchStore().setSearchType(searchType)
-    previousSearchType.value = searchType
+    useSearchStore().setSearchType(searchType as SupportedSearchType)
```
obulat, line 39: I think we need to add a check for whether the search type is in `supportedSearchTypes` here, otherwise if for some reason we pass an additional type to the `useSearchStore.setSearchType`, it will throw an error.

src/composables/use-search-type.ts
```diff
@@ -21,20 +23,28 @@ const icons = {
...
 }
-const searchTypes = [...supportedSearchTypes]
 
 export default function useSearchType() {
   const activeType = computed(() => useSearchStore().searchType)
   const previousSearchType = ref(activeType.value)
-  const setActiveType = (searchType) => {
+  const featureFlagStore = useFeatureFlagStore()
+  const additionalTypes = computed(() =>
+    featureFlagStore.isOn('external_sources') ? ([MODEL_3D] as const) : []
```
obulat, line 32: How about `video`? I think we should add it to the array.

obulat:
> It seems like some can't or shouldn't... Like the frontend search type switcher probably doesn't need to include the additional types, so why would it care.
> 
> Really I think the question is where needs to be updated to also consider _additional_/_external_ types? I'm supposing we don't want to update the `supportedSearchTypes` list to include anything unless it actually is "supported" (as in has an Openverse search)?

I agree that `supportedSearchType` list does not need to be changed because it actually refers to the non-external types. However, to be able to have additional/external type _pages_, we will need to update the type of the `search` store `searchType`:
https://github.com/WordPress/openverse-frontend/blob/2315ecd45df18dcb36fcaac6709a88f91e6ae750/src/stores/search.ts#L30
It should be set to `SupportedSearchType | AdditionalSearchType`, with maybe some additional checks in that store to make sure that we only set the `searchType` to an `AdditionalSearchType` if the flag is On.

We would also need to add page handling for external types. What I mean is that we would need to update:

https://github.com/WordPress/openverse-frontend/blob/2315ecd45df18dcb36fcaac6709a88f91e6ae750/src/stores/search.ts#L253

`queryStringToSearchType` does not currently support video or `model_3d`, so if you go to `/search/video?q=cat`, you would get a 404.
And then, there is no blank page for `model_3d`, so if you _do_ add `model_3d` to the `queryStringToSearchType`, and go to `/search/model_3d/?q=cat`, you will get an error because we don't currently have `/page/model_3d.vue`. These might need to be handled in separate PRs, though.

src/composables/use-search-type.ts
sarayourfriend: Ah, okay, so this function really should only accept SupportedSearchType? The additional types are just links and don't modify the store, is that right :thinking:

sarayourfriend:
> queryStringToSearchType does not currently support video or model_3d, so if you go to /search/video?q=cat, you would get a 404.
> And then, there is no blank page for model_3d, so if you do add model_3d to the queryStringToSearchType, and go to /search/model_3d/?q=cat, you will get an error because we don't currently have /page/model_3d.vue. These might need to be handled in separate PRs, though.

This PR isn't _too_ unweildy so far... but it might be best to put it into a different PR afterall, you're right. It's behind a feature flag anyway :grin:

dhruvkb (changes requested):

obulat:
I think #1430 and this PR can together accomplish two things:

- make the `/search/video` and `/search/model-3d` pages open the "Other search engines" page with buttons (we already have the objects that create those links for 3D models and video), if the user opens the pages directly, by typing the URL in the address bar or by redirecting from CC Search. - this feature would be available in prod, without the feature flag.
- add the additional media types (currently video and model-3d) to the Content switcher and thus available in the UI if the feature flag is On.

obulat (approved):

src/composables/use-search-type.ts
obulat: We can actually use the `additionalSearchTypes` from `~/constants/media`

src/stores/search.ts
```diff
@@ -115,7 +123,17 @@ export const useSearchStore = defineStore('search', {
     },
   },
   actions: {
-    setSearchType(type: SupportedSearchType) {
+    setSearchType(type: SearchType) {
+      const featureFlagStore = useFeatureFlagStore()
+      if (
+        featureFlagStore.isOn('external_sources') &&
```
obulat, line 129:
```suggestion
        !featureFlagStore.isOn('external_sources') &&
```

sarayourfriend:
Closing in favor of @obulat's more comprehensive PR.

PR #1385, opened 2022-05-04
Fix pnpm + docker

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse-frontend/runs/6283206618

## Description

This PR does the following:

* Pins `pnpm` to a specific version and updates instructions and installations of `pnpm` to indicate how to use the "pinned" version. We need a custom solution because volta does not yet support pinning `pnpm` and we don't use volta to install it in the docker builds anyway.
* Removes the local docker-compose stack for the frontend. It was broken ever since the playwright dockerization and no one noticed, so I'm assuming it was not being used.
* Updates documentation regarding the build
* Removes the multi-stage docker build. Given that the development image is gone now, the multi-stage build gave us nothing. In fact, even before, it wasn't doing anything but making build times longer if you weren't using BuildKit, so good riddance! It's good to delete code.
* Switches to using `pnpm fetch` to populate the store in the docker build from `pnpm-lock.yaml` instead of `pnpm install`. Details on how this works available [on the documentation page for it here](https://pnpm.io/6.x/cli/fetch). I've cross referenced the API and usage with pnpm 7 and there are no differences from pnpm 6, so the "experimental" warning at the top of that documentation page can be safely ignored.
* Updates the production image to run the app using the `node` user rather than root. This is a basic docker security best practice that we've been ignoring for a while. It's a simple fix here, especially without the multi-stage build, so why not.

## Testing Instructions

Checkout this branch and run `docker build . --build-arg PNPM_VERSION=$(cat .pnpm-version) -t openverse-frontend:latest` and confirm that it builds a working image that you can run using `docker run -it -p 127.0.0.1:8443:8443/tcp openverse-frontend:latest` at `localhost:8443`.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [N/A] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

[...]

PR #1420, opened 2022-05-12
Remove need to pin pnpm version

sarayourfriend:
## Description

Previously in #1385 we pinned the pnpm version to the latest 6.x.x. We did this because version 7 introduced an installation failure when peer dependencies were not being met.

Rather than continue to fight to pin pnpm, it seemed to make sense to actually try to fix the failures it was complaining about. Previously I'd avoided this because I wasn't sure what to do about react being a peer dependency of storybook and use clearly not wanting to declare a dependency on React (it makes no sense). I'm not sure why Storybook needs it considering it works fine without it. Maybe something else in our toolchain is installing it directly :shrug: 

For this PR I went through our pnpm-lock.yaml and identified the packages that were already in there anyway with version that were already being pulled into our node_modules. I added those to our package.json. Finally, I added some ignore rules and some overrides to let pnpm know that we know what we're doing and that the versions we have will be okay (at least, they have been okay so far).

Finally, this PR also pins our pnpm version in `engines` at >=7.0.1 and properly restricts our node version to >=16 <17. Node 17 continues not to play nice with webpack 4 or Nuxt, it seems (I guess the other way around maybe).

All of this has the added benefit of simplifying everywhere that got more complicated by #1385. We can remove the `.pnpm-version` file and all the juggling we had to do to get its values into Dockerfiles and other places it didn't want to be on its own.

## Testing Instructions

Make sure all of CI passes including the docker build etc.

Checkout the branch, delete `node_modules`, `volta install pnpm@7` and run `pnpm install`. Make sure it's successful and doesn't complain about any unmet peer dependencies.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


package.json
```diff
@@ -147,11 +153,26 @@
...
+      "ignoreMissing": [
+        "react",
+        "react-dom",
+        "consola"
+      ],
+      "allowedVersions": {
+        "babel-core": "7.0.0-bridge.0",
+        "postcss": "5.2.18"
+      }
+    }
```
krysal, line 171: I was looking for something like this but I didn't know what to do with these "peerDependencies", nice!

krysal (approved):
Awesome that there are no more errors when installing or removing packages! And the simplified setup of course :) 🌟 Thank you!

PR #1430, opened 2022-05-13
Stop opening `/search/video` from crashing the app

### LINKED ISSUE #1428: "/search/video" URL crashes app
## Description

Navigating to `/search/video` causes the app to crash.

## Reproduction

- https://search.openverse.engineering/search/video/
- https://search-staging.openverse.engineering/search/video

## Additional context

https://sentry.io/share/issue/0168474ee4fd4a62b3f891089a2f21d9/

## Resolution

obulat:
## Fixes

Fixes #1428 by @AetherUnbound

## Description

This PR makes it possible to open a `/search/video` page without crashing the app (mostly for users that get redirected from CC Search). It doesn't add `Video` to the Content Switcher.
To make sure everything works, I had to make the types more flexible, and make some methods accept `SearchType` (i.e., All Content and all media types including `video` and `media_3d`) instead of `SupportedSearchType`.

This PR is probably a good precursor for #1410, which will add the ability to add more types to the Content Switcher, making it possible to open the `model-3d` and `video` pages through the UI and not only by using the URL with the type in path.

With this PR, you will not be able to open `model-3d` page because we still haven't added a page component (`pages/model-3d`). I think we should add it, no matter what the `AdditionalTypes` flag value is, so that opening a `/search/model-3d` page doesn't crash the app, and the `AdditionalTypes` flag should control whether we have the ability to change the search type via UI (the Content switcher in the header or homepage searchbar).

## Testing Instructions

Try opening `/search/video` or `search/video?q=cat` and make sure it doesn't crash anymore.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VContentSwitcher/VSearchTypePopover.vue
+++ b/src/components/VContentSwitcher/VSearchTypePopover.vue
@@ -2,7 +2,7 @@
   <VPopover
     ref="contentMenuPopover"
     class="flex items-stretch"
-    :label="$t('search-type.label')"
+    :label="$t('search-type.label').toString()"
     placement="bottom-end"
   >
     <template #trigger="{ a11yProps }">
@@ -28,10 +28,18 @@
   </VPopover>
 </template>
 
-<script>
-import { computed, defineComponent, ref } from '@nuxtjs/composition-api'
+<script lang="ts">
+import {
+  ComponentInstance,
+  computed,
+  defineComponent,
+  PropType,
+  ref,
+} from '@nuxtjs/composition-api'
 
 import useSearchType from '~/composables/use-search-type'
+import type { SearchType } from '~/constants/media'
+import { defineEvent } from '~/types/emits'
 
 import VPopover from '~/components/VPopover/VPopover.vue'
 import VSearchTypeButton from '~/components/VContentSwitcher/VSearchTypeButton.vue'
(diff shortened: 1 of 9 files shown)


src/components/VContentSwitcher/VSearchTypePopover.vue
```diff
@@ -2,7 +2,7 @@
   <VPopover
     ref="contentMenuPopover"
     class="flex items-stretch"
-    :label="$t('search-type.label')"
+    :label="$t('search-type.label').toString()"
```
krysal, line 5: why does localized text need to be converted to string in templates? 🤔

src/stores/search.ts
```diff
@@ -259,6 +270,8 @@ export const useSearchStore = defineStore('search', {
         this.setSearchTerm(urlQuery.q.trim())
       }
       this.searchType = queryStringToSearchType(path)
+      console.log('search type set to', this.searchType)
```
krysal, line 273:
```suggestion
```

krysal (commented):
I think we should add `video` to the Content Switcher as the metasearch is available and it was a part of the old CC Search. Currently is showing 3D models, which doesn't crash the app but returns the 404 error page, which seems correct to me since we don't support it even in the metasearch(?), so to me, we should remove 3D models from there until is "semi-supported" at least.

src/components/VContentSwitcher/VSearchTypePopover.vue
obulat: Because for TypeScript, i18n does not return string, it returns `TranslationResult`, and it's getting flagged in the templates. It should be fixed in newer versions of i18n library, and hopefully we'll remove all `toString()`s when we update.

obulat:
> I think we should add `video` to the Content Switcher as the metasearch is available and it was a part of the old CC Search. Currently is showing 3D models, which doesn't crash the app but returns the 404 error page, which seems correct to me since we don't support it even in the metasearch(?), so to me, we should remove 3D models from there until is "semi-supported" at least.

We actually do support 3D models in metasearch, and if you add the page component, it will open the page with buttons with links to several providers.
We could define the "support" for additional types under the flag as the addition of the types to the Content switcher. Otherwise, if both page and the Content switcher are under the flag, we would have video treated separately because we need the support for `/search/video` for redirects.

krysal:
@dhruvkb Could you take a look at this today or tomorrow? Thank you!

zackkrida (approved):
This looks like a good approach to fix the crash, and the extra types in places are a nice touch. It will be very nice to just enable video and 3d model meta search soon as well, but this is a great solution until then.

PR #1438, opened 2022-05-13
Workflow improvements to fix CI pipline

krysal:
## Description

This PR should fix the pipeline in `main` by doing the following:
- Skip jobs trying to comment instructions on the `main` branch
- Removes checking steps from the `playwright-test-failure-comment` job as they were interfering when we really want it to run
- Add dependency on `build` in the `nuxt-playwright` job, tests would fail if nuxt can't be built anyway

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,20 +129,7 @@ jobs:
       - storybook-playwright
 
     steps:
-      - uses: actions/download-artifact@v2
-        with:
-          path: test-results
-
-      - name: Check if test results exist
-        id: check-results
-        run: |
-          has_results=$(stat test-results/nuxt-test-results || stat test-results/storybook-test-results) || false
-          if [[ $has_results ]]; then
-            echo "::set-output name=has-failure::true"
-          fi
-
       - uses: peter-evans/create-or-update-comment@v2
-        if: steps.check-results.outputs.has-failure
         with:
           issue-number: ${{ github.event.pull_request.number }}
           body: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,8 @@ jobs:
   nuxt-playwright:
     name: Nuxt Playwright tests
     runs-on: ubuntu-latest
+    needs:
+      - build
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -84,6 +85,8 @@ jobs:
   nuxt-playwright:
     name: Nuxt Playwright tests
     runs-on: ubuntu-latest
+    needs:
+      - build
```
sarayourfriend, line 89: I wish it was possible to _use_ the build from this step for the Nuxt playwright. I thought about doing this before but hesitated because it means that the _start_ of the Nuxt playwright tests in CI is now _two_ build times instead of just one :confused: But I agree that this makes sense, especially to prevent the comment from showing up when it's just a build failure anyway.

sarayourfriend (approved):
Love this clean up!

zackkrida (approved):
oh, this is a nice surprise. lgtm!

.github/workflows/ci.yml
krysal: > I wish it was possible to use the build from this step for the Nuxt playwright.

Me too! I was thinking about that but doesn't seem like something trivial for now...

.github/workflows/ci.yml
```diff
@@ -121,26 +124,13 @@ jobs:
...
-          path: test-results
-
-      - name: Check if test results exist
-        id: check-results
-        run: |
-          has_results=$(stat test-results/nuxt-test-results || stat test-results/storybook-test-results) || false
-          if [[ $has_results ]]; then
-            echo "::set-output name=has-failure::true"
-          fi
-
```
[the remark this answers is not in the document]
krysal: It's interesting to see these workarounds too 🙂

.github/workflows/ci.yml
sarayourfriend: We could push the build as an artifact or something :thinking: And then devise some way of skipping the build step in CI... though we need to wait talkback to be available or you run into a race condition, I think this is documented in the playwright configuration:

https://github.com/WordPress/openverse-frontend/blob/f56af03179a68b090067fe6f517bd4bafb6a888b/test/playwright/playwright.config.ts#L16-L21

This package might be a thing we could use with npm-run-all: https://www.npmjs.com/package/wait-on

Something like `npm-run-all -p talkback "wait-on localhost:3000" -s start`? There's some way of getting the right parallel and serial tasks into `npm-run-all` to make it work I think: https://www.npmjs.com/package/npm-run-all/v/1.8.0#run-a-mix-of-sequential-and-parallel-tasks

PR #1443, opened 2022-05-15
Add metrics middleware

### LINKED ISSUE #1256: Add metrics endpoint for prometheus
## Problem

We need to be able to gather metrics from Nuxt for prometheus. To do this, we'll add a `metrics` endpoint that reports the prometheus node.js client library's gathered data.

## Description

We can base ours off of https://github.com/franckaragao/nuxt-prometheus-module but that one is fairly basic and doesn't directly allow customization.

Ideally we can create a separate repository (or package if we've got a JS monorepo any time soon) `@{wordpress,openverse}/nuxt-prometheus-module` for example, and share our solution with the world. Ideally it would be more easy to customize and add custom metrics while also covering a wider array of default metrics like the ones from django-prometheus (i.e., view timing, status code counting, etc).

## Alternatives

Just use the basic existing nuxt prometheus module and call it a day. This would be faster, but wouldn't serve us as well in the long run.

## Additional context

https://github.com/WordPress/openverse/pull/193

## Implementation

sarayourfriend:
## Fixes

Fixes #1256 by @sarayourfriend

## Description

'
Adding this to the frontend because it's the easiest to deploy (to search-staging) and it'll be nice to have something live to point the Grafana Agent at while I work on setting that up this week.

## Testing Instructions

Checkout the branch and run `pnpm dev`. Ensure that you cannot reach any `/metrics` endpoint. Comment out [this line](https://github.com/WordPress/openverse-frontend/blob/50f9eb6a9b98f3251c9d836a7f400abdd3c645aa/nuxt.config.ts#L244) and then repeat and confirm that you can reach the metrics endpoint. Visit some pages and confirm that the metrics endpoint records them appropriately. In particular visit some single result pages and make sure that the mapping works to unify all the individual single result pages into a single entry in the metrics listing (per search type).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (approved):
Tested it locally and it works well.
I don't quite understand the outputs, though. 
This is a part of metrics I get for `/favicon.ico`:
```
http_request_duration_seconds_bucket{le="0.003",status_code="200",method="GET",path="/favicon.ico"} 2
http_request_duration_seconds_bucket{le="0.03",status_code="200",method="GET",path="/favicon.ico"} 4
http_request_duration_seconds_bucket{le="0.1",status_code="200",method="GET",path="/favicon.ico"} 4
http_request_duration_seconds_bucket{le="0.3",status_code="200",method="GET",path="/favicon.ico"} 4
http_request_duration_seconds_bucket{le="1.5",status_code="200",method="GET",path="/favicon.ico"} 4
http_request_duration_seconds_bucket{le="10",status_code="200",method="GET",path="/favicon.ico"} 4
http_request_duration_seconds_bucket{le="+Inf",status_code="200",method="GET",path="/favicon.ico"} 4
```
Are there 7 buckets of response durations (up to 0.003, 0.003-0.03, 0.03-0.1, 0.1 - 0.3, 0.3-1.5, 1.5 - 10 and above 10 - seconds?). And 4 out of the requests for favicon took longer than 10 (seconds?)

sarayourfriend:
> Are there 7 buckets of response durations (up to 0.003, 0.003-0.03, 0.03-0.1, 0.1 - 0.3, 0.3-1.5, 1.5 - 10 and above 10 - seconds?). And 4 out of the requests for favicon took longer than 10 (seconds?)

They're actually all "le" which is "lower than or equal". So rather than buckets you're seeing upper bounds. All 4 were less than 0.03 seconds (and each upper bound above that), but only 2 were less than 0.003 seconds.

krysal (approved):
Works as indicated. I was wondering why it wasn't getting metrics for some pages until I realized that it only works for SSR pages, is this expected? Should be documented somewhere?

sarayourfriend:
> Works as indicated. I was wondering why it wasn't getting metrics for some pages until I realized that it only works for SSR pages, is this expected? Should be documented somewhere?

I'm not sure where to document this, but it is indeed a limitation and it makes sense once you consider how an SSR SPA works (SSR the initial page load, then the rest is client side). The client doesn't have access to the prom client (nor should it). When we want to start measuring individual client-side performance of things we'll need to look for a RUM library rather than a system monitoring library. I think given that client-side measurements are really outside of the scope of Prometheus (once you think about and understand it, which took me a second to do as well) then it makes sense.

I guess if we had some vague documentation about the difference between SSR and CSR page navigations etc then we could add a note there, but I don't think we do have any information about that outside of our testing guidelines which seems like a weird place to mention anything about prometheus.

krysal:
I believe that putting it with the middleware declaration will suffice.

sarayourfriend:
Added in the most recent commit @krysal

PR #1457, opened 2022-05-19
Use local provider data on the first load

### LINKED ISSUE #1456: Sources filter choices don't display on client-side render
## Description

When searching the sources filter in the filter sidebar is empty, if navigating to the page from the client side, as most users entering the site from the homepage are.

## Reproduction

1. Visit wordpress.org/openverse
2. Make a search
3. Switch to the audio or image views
4. Observe that no source filters appear in the filter sidebar.

obulat:
## Fixes

Fixes #1456 by @zackkrida

## Description

This PR loads the provider data from local data at first and requests the data from the server to update providers. This means that the user will see provider filters instantly, and will see the updated version after the request is finished.

## Testing Instructions

Search for something from the homepage, and then go to 'Images' view and make sure that the `Source` filter in the sidebar is not empty.
Try reloading the same page.
Check that sources page is not empty.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/data/sources.json
```diff
@@ -0,0 +1,349 @@
+{
```
sarayourfriend, line 1: Should we generate this file at build time, you think? :thinking: At least then it would only ever be minimally out of date (as opposed to the situation where we forget to update this file manually :sweat_smile:)

sarayourfriend (approved):
LGTM! It'd be nice not to have to manage the file manually, if we can download the data and build the file from the stat's endpoints I think that'd be pretty nice :slightly_smiling_face:

src/data/sources.json
obulat: Agree! I'll add a script similar to the locale scripts.

scripts/providers.mjs
```diff
@@ -0,0 +1,48 @@
...
+try {
+  if (shouldUpdate) {
+    updateProviderData().then((providerData) => {
+      fs.writeFileSync(providerFilePath, JSON.stringify(providerData, null, 2))
+      console.log(`Wrote the updated provider data to ${providerFilePath}`)
+    })
+  }
+} catch (error) {
+  console.log(`The providers data is outdated, but there was an error during update: ${error}`)
+}
```
sarayourfriend, line 48: Can we make the exit code non-0 whenever `NODE_ENV` isn't `development`? It'd be nice for this to fail during the docker build and cause the build to stop, for example, given the generated resource is critical for running the application at this point.

obulat:
> This works well and thanks for writing it! I am curious about what benefit this approach has compared to the 'old' approach back when the filter was working.
> 
> My concern is that this adds another layer of 'cache' to the providers list, and means that every time we update the providers we need to rebuild the app.
> 
> The alternative approaches I remember thinking about in the past are:
> 
>     * We load the providers from the API on the initial server render, and put them in the store for subsequent page visits
> 
>     * We load the providers from the API on every single '/search' page load, and separately on the '/sources' page
> 
> 
> Those approaches seem like they keep the provider list more up-to-date than this one, without adding more than a few milliseconds of delay.
> 
> I'm still open to the approach in the PR; would you be able to share how it differs from those other approaches?

Thank you for the comment, it made me re-think the approach to solving this problem.

My main motivation for adding the local file was the issues with the timeout of the `/stats/` endpoint I was sometimes seeing locally during the development. But you're right that having so many cache layers is not efficient, and the need for re-building the app when the providers change would add an unnecessary complication.

I realized that there was another problem in the load logic: updating the filters through the provider store on each request did not work, and adding the check for provider stats in the store on every single '/search' page load solved the empty source filters problem in #1460. And because we check for time passed since the last update, we don't actually re-fetch the providers too much.

I think the #1460 implementation is better, so I'm going to close this PR.

By the way, I'm not entirely sure when the source filters got broken on the client. I know they were working with Vuex, but after the Pinia update, I'm not so sure. Do you remember them working correctly after the Pinia update? If so, I'm curious to what changed this.

PR #1476, opened 2022-05-29
Fix waveform misalignment for short audios

### LINKED ISSUE #1304: Very short audio tracks have misaligned progress bar upon track completion
## Description

Example: https://search-staging.openverse.engineering/audio/d4c5806f-1e86-4f41-835f-8fcda6c62cb7

The issue is very slight but it is there.

Originally reported by @zackkrida 

## Reproduction

1.  Visit https://search-staging.openverse.engineering/audio/d4c5806f-1e86-4f41-835f-8fcda6c62cb7 and play the track through to the end, note the slight misalignment of the progress bar at the end.

## Screenshots

image

## Additional context

We could potentially solve this by setting the progress bar width to exactly 100% of the waveform width once the track completes. This seems like it's probably a rounding error.

## Resolution

obulat:
## Fixes

Fixes #1304 by @sarayourfriend 

## Description

This PR fixes the waveform misalignment at the end of the waveform. It was especially noticeable for very short audios.   We were only updating the `currentTime` value when the audio was in `playing` status, so the last update when the audio went from `playing` to `played` status was never made, and the audio waveform `progressBarWidth` was not updated for the last time. 
There was also an error when we used a computed without calling its `.value`.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (approved):
LGTM! I thought this would be an unsolvable problem related to web api rounding or some other issue. So glad there was a fix.

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -188,10 +188,13 @@ export default defineComponent({
...
-        currentTime.value = localAudio.currentTime
-        window.requestAnimationFrame(updateTimeLoop)
-      }
+      if (localAudio)
+        if (status.value === 'playing') {
+          currentTime.value = localAudio.currentTime
+          window.requestAnimationFrame(updateTimeLoop)
+        } else if (status.value === 'played') {
+          currentTime.value = localAudio.currentTime
+        }
```
sarayourfriend, line 197: I'm going to look into why later but I wanted to note it here so I don't forget. It looks like the final update to put the audio current time happens slightly later and causes a jump in the waveform progression at the end for short audio files :thinking: I want to try to find a fix for that too if you don't mind @obulat?

sarayourfriend (approved):
Approving this for now. I haven't had time to check out the issue I mentioned. @obulat if you don't get around to checking on this now can you open an issue to address it? I think it's very low priority, but this fix is better to have than to block on it for the minor jumpiness.

zackkrida:
@sarayourfriend does this video look like what you're seeing? I haven't been able to reproduce the jump you mentioned.

https://user-images.githubusercontent.com/6351754/172632555-af3422e0-c56a-4ef7-b8c9-5558f7becc8b.mp4

I have, however, @obulat, noticed interesting behavior with audio around 1 second long. Here's the 'end point' of a bunch of >1s audios:


CleanShot 2022-06-08 at 09 46 21@2x

They all finish in a different position! I wonder if the position is deterministic and related to their exact length. Perhaps for these we'd need to round up the progress bar to 1 second.

I also wonder: should we alter the timestamp to show milliseconds for audio tracks less than one second long? We can make a separate issue for that if needed.

sarayourfriend:
@zackkrida it does not look like your video. Unfortunately I'm completely unable to play any audio files locally at the moment (no idea why, it's very strange) so I can't take a video to show what I was seeing.

obulat:
> They all finish in a different position! I wonder if the position is deterministic and related to their exact length. Perhaps for these, we'd need to round up the progress bar to 1 second.

This is a really interesting bug! Sometimes I get the correct end position every time, sometimes it's incorrect on the first time, sometimes - on following replays. I'm going to look into it.

src/components/VAudioTrack/VAudioTrack.vue
obulat: I wonder if it was caused by the audio sometimes ending with a `paused` status, and sometimes with `played`. I still don't understand why it happens, but I removed the check for `status` being played to update the `currentTime` on line 196 here.

obulat:
> I'm wondering if instead of relying on the animation loop we need to just set `currentTime = duration` when [the audio is finished playing (via the `ended` event)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/ended_event).

It seems to be working now, after removing the check for played status. Do you think this solution is good enough, @sarayourfriend ?

zackkrida:
Looks better to me, nice 👍

obulat:
> It makes me wonder if we really should only show milliseconds for sub-second files, mostly to minimize the visibility of that frustrating data issue 😬

This was an easy thing to fix :)

PR #1405, opened 2022-05-11
Use `defineComponent` and TS in all components

obulat:
## Fixes

Related to Typescriptification milestone

## Description

This PR converts almost all components to `defineComponent`, and adds `lang="ts"` to most of the components. I did not add many of them to `tsconfig` because until we convert ariakit utils, we type checking will fail for all components that use Popover or Modal. But we can still use TS for writing them!

I feel like it's better to get this change over with and prevent cognitive overload when you have to add TS or defineComponent to other PRs that are not related to TS-ification.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #1480, opened 2022-06-02
Use `defineComponent`, add prop types and emit types

obulat:
## Fixes

Related to Typescriptification milestone

## Description

This PR is extracted from #1405 to make it easier to review. 
It adds `defineComponent`, sets the language to `ts`, documents `emits`, and replaces JSDoc comments with TS annotations for more components. It only changes the files that can be added to `tsconfig` (i.e., don't have popup ariakit dependencies)

## Testing Instructions

Passing CI should be enough.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VFilters/VLicenseExplanation.vue
```diff
@@ -36,7 +36,9 @@
 </template>
 
 <script>
```
sarayourfriend, line 38: Should this have lang=ts on it as well?

sarayourfriend (commented):
Is this PR meant to also have updates to tsconfig.json for these?

obulat:
> Is this PR meant to also have updates to tsconfig.json for these?

I've added the files that pass the types check.

sarayourfriend (approved):
Lovely!

src/components/VLogoLoader/VLogoLoader.vue
```diff
@@ -39,16 +39,24 @@
...
-import { propTypes } from './VLogoLoader.types'
-
 export default defineComponent({
   name: 'VLogoLoader',
-  props: propTypes,
+  props: {
+    status: {
+      type: String as PropType<'loading' | 'idle'>,
+      default: 'idle',
+      required: false,
```
dhruvkb, line 53: `required: false` is the [default behavior](https://vuejs.org/guide/components/props.html#prop-validation).

dhruvkb (approved):
I really wish there was a better way than listing each component individually in the `tsconfig.json` file. This is a good PR. I only have one minor nit and a question:

- When we remove the validator from the props, there would be no runtime warnings for invalid values because all the TS will be transpiled to JS, so shouldn't we keep them?

src/components/VLogoLoader/VLogoLoader.vue
obulat, line 53:
```suggestion
      default: 'idle',
```

obulat:
> I really wish there was a better way than listing each component individually in the `tsconfig.json` file. This is a good PR. I only have one minor nit and a question:
> 
>     * When we remove the validator from the props, there would be no runtime warnings for invalid values because all the TS will be transpiled to JS, so shouldn't we keep them?

We have discussed with @sarayourfriend before that we want to remove the validators. I don't quite remember what the decision was regarding the runtime warnings and why we might not need them. 
I am going to merge this PR as is, @dhruvkb , and open an issue to discuss whether we want to keep the validators, and document the reasons for the decision.

Issue #914, opened 2022-02-20
Mock media requests for the e2e tests


obulat:
## Problem

When running e2e tests, we make network requests for media files. This can make the tests non-deterministic and slower than necessary.
Currently, in e2e tests we block the requests for images based on their extensions, and all the requests with to `jamendo`:
https://github.com/WordPress/openverse-frontend/blob/99a803542c46b20b7cac36bc7d421ffcaa8398af/test/e2e/utils.js#L64-L66
Audio requests for `jamendo` files don't have an extension that we could use, so we block all the requests with `jamendo` in the domain:
https://github.com/WordPress/openverse-frontend/blob/99a803542c46b20b7cac36bc7d421ffcaa8398af/test/e2e/search-query-server.spec.js#L19
We are probably missing some network requests that also need to be blocked.

## Description

We need to audit what requests are being made and block them. This does not include requests to the Openverse API, which are handled by a proxy that already stops network requests and uses the pre-recorded responses in the repository in `test/tapes` folder.

## Alternatives

We could also set up proxies for all providers, but I suspect that it would be untenable.

## Additional context

It would be nice to have a way to monitor the network requests made during the e2e tests in the CI, and fail the PRs that add new un-handled network requests.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

PR #1493, opened 2022-06-08
Add a fallback for broken provider image

obulat:
## Fixes

Related to #914 by @obulat 

## Description

We are currently requesting provider media during the Playwright tests to render single image detail and audio detail pages. This creates flaky and slow tests. Until we create a solution to proxy those requests, we can block or mock the provider media requests entirely. 

Blocking the provider media requests would result in a broken image on the image detail page during Playwright tests. To prevent this, this PR adds the logic to replace the provider image with the thumbnail on load error. Since the thumbnails have the same dimensions as the provider image they allow us to test that the different dimensions are treated correctly). 

## Testing Instructions

Search for 'birds', and open the first image result. Block the requests to 'Flickr', and reload the image.
Screen Shot 2022-06-08 at 11 17 21 AM

 Visually, the page should be the same as before, but it should be using the 'thumbnail' from our API as the main image.
Also, try blocking the main thumbnail image. You should see the error image placeholder.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (commented):
I wonder if we could accomplish this using just CSS: https://stackoverflow.com/a/33959259

It may even have the extra benefit of preventing people from trying to download the image from Openverse instead of visiting the provider directly for it (though I can't decide if that's a dark pattern for us to implement).

I'm fine with the JS solution you've taken here but if we can do it without JavaScript that would be great :slightly_smiling_face:

sarayourfriend (approved):
LGTM!

dhruvkb (approved):
This is really good. Hopefully it will drastically reduce flakiness.

Issue #1244, opened 2022-04-06
Placeholder for critical path documentation that will lead to e2e tests


sarayourfriend:
## Assigned reviewers

- [] @obulat 
- [] @krysal 

## Description
I (Sara) will fill this in tomorrow morning :slightly_smiling_face:

sarayourfriend:
Note that parts of the flow are already covered (we have tests from homepage to results page for example). We just want to augment the existing tests with end-to-end user flow tests.

sarayourfriend:
@WordPress/openverse-maintainers please review this RFC/proposal :pray: It totally fell off my radar.

sarayourfriend:
I guess they'd be doing both. I'm not totally sure if it makes sense to separate them for these "long flow" type tests because we'd want to confirm they work in the contexts of a user actually going through the core flows.

On the other hand, we would need to have some way to disable them when testing against staging otherwise the visual regression tests will end up needing to somehow ignore differences between the tape data and production data. That's an issue we don't need to solve now though.

Do you think there's a good reason to separate these particular snapshots out into their own tests (that still preserves their ability to test the visual results of full user flows)? Right now our visual regression tests run at the same time as the other e2e tests. I'm not totally sure the distinction is _that_ important. What do you think?

PR #1517, opened 2022-06-10
Remove secrets and fix shell step in task definition redeploy

sarayourfriend:
## Description

I'm modifying the composite action here and then will backport the fixes to the template in the infrastructure repository where it is generated once it's fixed.

## Testing Instructions

See if it works on this branch, remove the testing code, then merge and hope it also works on main.

Update: it worked! https://github.com/WordPress/openverse-frontend/runs/6828684979?check_suite_focus=true

Confirmed in AWS that the service did indeed redeploy with the new task definition and reached a steady state :tada: 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/actions/frontend-deploy/action.yml
+++ b/.github/actions/frontend-deploy/action.yml
@@ -12,6 +12,10 @@ inputs:
   tag:
     required: true
     description: Image tag to deploy.
+  aws_access_key_id:
+    required: true
+  aws_secret_access_key:
+    required: true
 
 runs:
   using: "composite"
@@ -20,12 +24,12 @@ runs:
     - name: Configure AWS credentials
       uses: aws-actions/configure-aws-credentials@v1
       with:
-        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        aws-access-key-id: ${{ inputs.aws_access_key_id }}
+        aws-secret-access-key: ${{ inputs.aws_secret_access_key }}
         aws-region: us-east-2
 
     - name: Download task definition
-      with: shell
+      shell: bash
       run: |
         aws ecs describe-task-definition \
           --task-definition staging-frontend \
--- a/.github/workflows/frontend_deploy.yml
+++ b/.github/workflows/frontend_deploy.yml
@@ -24,3 +24,5 @@ jobs:
       - uses: ./.github/actions/frontend-deploy
         with:
           tag: ${{ (github.event_name == 'workflow_dispatch' && github.event.inputs.sha) || (github.event_name == 'push' && 'main') || github.event.release.tag_name }}
+          aws_access_key_id: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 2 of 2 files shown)


sarayourfriend:
Note: the current configuration of this action doesn't really make much sense because it's deploying to the same place on releases, pushes to main, and workflow dispatches. Once the ECS infrastructure is finalized and we have two environments for it (staging and production) then the pushes to main will only deploy staging and releases (and workflow dispatches) will deploy to the production service. The intention here is to mirror our current production deployment cadence where we deploy to production through a "release" cycle rather than on every push to main. The workflow dispatch is included as a way to rollback to a previous version.

In the future (like maybe by the end of this year/early next year) I'd like to get us to a point where every push to main implies a production deployment through some kind of pattern that looks like this:

1. Merge to main
2. Staging deploys
3. Verify changes in staging
4. Press a button somewhere that triggers the production deployment (maybe a Workflow to dispatch)

We'll need to think about safe guards so someone doesn't accidentally merge to main and then ignore the rest of the process (because then it will fall on subsequent deployers to test the previous merge in staging before they're able to push their new revision to production). We'll also want to try doing things like running e2e tests against the actual deployed staging environment and block the push to production until the e2e tests pass.

In short: there's a _lot_ of work for us to get to the quick "pushing to main implies a production deployment" setting, but we can get there eventually. The work I'm doing now sets us up for that.

cc @WordPress/openverse-maintainers please check out this comment as I believe it's good context for how I'm approaching this work currently and what the goals are right now vs the far future goals we've discussed to varying levels of depth.

obulat:
> get us to a point where every push to main implies a production deployment 

This sounds really dangerous, even if there's a button that you need to press after checking that staging is fine. I wonder if the _checking that staging is fine_ after every PR merge instead of _checking for release every [x] days_ would add more work... Or maybe we would need to change our perceptions of what deploy means :)

sarayourfriend:
> This sounds really dangerous, even if there's a button that you need to press after checking that staging is fine. I wonder if the checking that staging is fine after every PR merge instead of checking for release every [x] days would add more work... Or maybe we would need to change our perceptions of what deploy means :)

There are lots of big projects that follow this pattern (wp-calypso for example as well as other projects I've worked on in the past) and it's pretty safe. There are a bunch of details I'm leaving out like:

1. Rollbacks have to be easy and fast to do
2. Deploying to Production from Staging needs to _block_ on anything that doesn't pass (like e2e tests)
3. Complementing that second point, we need to have robust e2e tests that confirm the entire core user flows are healthy (implementing #1244 which is _well_ past it's due date, I've just realized, is a good step in that direction)
4. We have to trust contributors to do their due diligence when testing their changes in staging so that it doesn't break production

It also helps to have the perspective that if we're deploying every commit to main to staging and then to production after staging is verified, then our production deployment units are _much_ smaller than they are today. You're not worried about "am I going to break the entire app" because 99% of the time the changes you merged only change a small piece of functionality. Even in long-running feature development, we'll leverage feature flags to avoid having to do massive significant deployments.

zackkrida (approved):
lgtm!

PR #1522, opened 2022-06-13
Make search.vue media fetching non-blocking

### LINKED ISSUE #1516: Pages are loading slowly
## Description

We have recently moved to using `asyncData` for search pages to fix the API being called twice for each page. This, however, had the unintended consequence of slowing the page load. Now, when you navigate to the search page from the homepage, it loads only after all the API calls have been completed, instead of loading the page skeleton first and showing the images/audio as soon as they are available.

## Reproduction

1. Open the homepage at https://search-dev.openverse.engineering
2. Search for something, and then open one of the results.
3. If you are on a not extremely fast network, you should see the loading indicator (black line) at the top of the pages instead of the more smooth page transitions, just like in the screenshot below.

## Screenshots

transition

## Resolution

obulat:
## Fixes

Fixes #1516 by @obulat

## Description

This PR moves the call to fetch media from `asyncData` to `fetch`. If we fetch data in `asyncData`, the page transitions only after the media are loaded from the API, and this makes the transition feel slower. With this PR, the page is being loaded concurrently with the media items, the search page opens as soon as the user clicks on the homepage 'Search' button, so it seems more responsive.
The Nuxt docs describe the difference between the `asyncData` and `fetch` hooks here: https://nuxtjs.org/docs/features/data-fetching#async-data

Since the `search` page fetches media on each route change, we only need to set the route in the header, and the media fetching should be handled by the `search` page. So, I removed the calls to `fetchMedia` from the header.

## Testing Instructions

Search for something from the homepage, from the search page, and from content pages. The search page should not show the black progress bar at the top when loading. The navigation to the Search page should also feel faster than on main currently.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VHeader/VHeader.vue
+++ b/src/components/VHeader/VHeader.vue
@@ -166,12 +166,14 @@ export default defineComponent({
      * - search term hasn't changed:
      *   - on a search route, do nothing.
      *   - on other routes: set searchType to 'All content', reset the media,
-     *     change the path to `/search/` (All content) and fetch media.
+     *     change the path to `/search/` (All content).
      * - search term changed:
      *   - on a search route: Update the store searchTerm value, update query `q` param, reset media,
      *     fetch new media.
      *   - on other routes: Update the store searchTerm value, set searchType to 'All content', reset media,
-     *     update query `q` param, fetch new media.
+     *     update query `q` param.
+     * Updating the path causes the `search.vue` page's route watcher
+     * to run and fetch new media.
      */
     const handleSearch = async () => {
       window.scrollTo({ top: 0, left: 0, behavior: 'auto' })
@@ -198,7 +200,6 @@ export default defineComponent({
           query: searchStore.searchQueryParams,
         })
         router.push(newPath)
-        await mediaStore.fetchMedia()
       }
     }
     const areFiltersDisabled = computed(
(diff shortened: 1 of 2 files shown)


src/pages/search.vue
```diff
@@ -95,14 +97,24 @@ export default defineComponent({
...
       path: route.path,
       urlQuery: route.query,
     })
+  },
+  /**
+   * Fetch media, if necessary, in a non-blocking way.
+   */
+  async fetch() {
+    const mediaStore = useMediaStore()
+    const searchStore = useSearchStore()
```
krysal, line 117: What do you think about moving this fetching to the `setup` block? I believe it also works, we could save duplicated lines of declarations and keep the code in a more standard style instead of using custom Nuxt hooks.

src/components/VHeader/VHeader.vue
```diff
@@ -198,7 +200,6 @@ export default defineComponent({
           query: searchStore.searchQueryParams,
         })
         router.push(newPath)
-        await mediaStore.fetchMedia()
```
krysal, line 201: Nice to see the responsibilities of this component reduced 💯

krysal (commented):
Nice improvement! Thanks for taking care of the loading time, the extra waiting before changing the page is slightly annoying before this indeed.
I left some comments for consideration but it's already working great.

src/pages/search.vue
obulat: As far as I understand, the `setup` block is blocking the rendering of the page, so it would mean the black loading indicator on the top of the page while loading the media items :(
I hate duplication, too. I might actually be able to return the necessary store methods from the `setup`, and use _them_ in fetch...

PR #1526, opened 2022-06-14
Remove extension from title if it's the same as filetype

### LINKED ISSUE #1303: Strip extensions from audio results where the title appears to be a file name with extension
## Description

Example: https://search-staging.openverse.engineering/audio/b9487d22-4927-4d09-a3f0-465e3ec4ceb6

Originally reported by @dhruvkb 

This would probably require compiling a list of common audio file extensions. mp3, wav, ogg, flac are probably the biggest ones (except for flac I've noticed each of those with this problem before).

## Resolution

obulat:
## Fixes

Fixes #1303 by @sarayourfriend 

## Description

This PR removes the extension from the audio file title if it matches the audio filetype. We can assume that the extension was not added to the title intentionally. You can see the changes in the old and updated visual regression snapshots.

#### Before (the 2nd, 3rd and 5th audio items have `.mp3` extensions in the title)

Snapshot showing the redundant extensions in the title

#### After (the 2nd, 3rd and 5th audio items do not have redundant `.mp3` extensions in the title):

Snapshot without the redundant extensions in the title

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/utils/decode-media-data.ts
```diff
@@ -1,14 +1,84 @@
...
+ * @param filetype - the type of the file.
+ */
+const isFiletypeMatching = (extension: string, filetype?: string) => {
+  if (filetype === extension) {
+    return true
+  }
+  return (
+    filetype &&
+    ['jpg', 'jpeg'].includes(filetype) &&
+    ['jpg', 'jpeg'].includes(extension)
```
dhruvkb, line 38: Since there is the possibility to have many cases of an extensions having multiple forms, would a regex based elimination be more suited at this?

A simple regex to match extensions would look something like this: `.*\.jpe?g$`. This covers both considerations
- extension being the last part after the dot
- jpg or jpeg both being captured identically

src/components/VErrorSection/VErrorImage.vue
```diff
@@ -51,6 +51,7 @@ export default defineComponent({
         let image = errorItem.image
         const errorImage: ErrorImage = {
           ...image,
+          originalTitle: image.title,
```
dhruvkb, line 54: I would've advocated for us to keep the original title as sent from the API in `title` (same as the field in the API) and store the transformed value in a different field. But I can see how that could have caused ripple effects across the codebase.

dhruvkb (approved):
This is a great improvement in terms of UI. I have a few thoughts but they can be safely ignored.

src/components/VErrorSection/VErrorImage.vue
obulat: That was my original thought, too, @dhruvkb !  But as you say, it would require a lot of changes in the codebase.

PR #1529, opened 2022-06-16
Share audio loaded state through pinia

### LINKED ISSUE #1528: Audio loading icon bug after navigating to a single audio result while the audio is already playing
## Description

The audio loading icon displays on single audio results if you navigate to the single result page after pressing 'play' on the audio track in the search result.

## Reproduction

1. Go to https://search-staging.openverse.engineering and make a search with audio results
2. Start playing any track
3. Click on any track's play button
4. Click on the track to navigate to the single result
5. Observe the loading spinner is active

## Screenshots

https://user-images.githubusercontent.com/6351754/173934535-2cfd5497-84c4-4385-b5ca-07195f01a22c.mp4

sarayourfriend:
## Fixes

Fixes #1528  by @zackkrida 

## Description

Based on my debugging it seemed like the issue was that the `loaded` event was never firing for the audio after it had already been loaded on the search results page, so when the audio track component is initialized on the single result page, `hasLoaded` is set to false and never updated to `true` because the `loaded` event doesn't fire again after the first time. This made the status get stuck on `loading`.
This PR now follows @obulat's suggestion to manage the audio state within the audio object itself so that it's easily shared throughout the app without the need for an additional store. This treats the audio object itself as the source of truth. It works great I think! And is maybe a little less unexpected. Thanks for the suggestion @obulat!

## Testing Instructions

Follow the repro instructions in the linked issue and confirm it's no longer reproducible.

I'm not sure whether to add an e2e test for this as I don't know whether we can reliably mock audio responses at the moment? It's probably worth looking into though.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (commented):
Great catch!
I was also thinking that we could update the duration in the Pinia store as soon as it is updated from the provider. This would make the audio replays more consistent.

I wonder if it makes more sense to make the Audio list items editable, and add the `isLoaded` property there? This way we could avoid creating a new Pinia store. I do have reservation about making the audio item not a `const`, as that might cause other problems if it is updated incorrectly, but it seems like a more flexible approach if we need to update some properties after the list was loaded from the API.

The same approach could be used to update the filetype of the images, otherwise if we open an image without a filetype twice, we would make 2 `head` requests.

sarayourfriend:
That's a good idea, Olga. I will try to implement that here and see if it's viable for a future issue to do with the duration (and probably status too) as you suggested.

src/stores/media/index.ts
```diff
@@ -376,5 +376,16 @@ export const useMediaStore = defineStore('media', {
         throw new Error(errorMessage)
       }
     },
+
+    setMediaProperties(
```
obulat, line 380: Here, if the audio item was loaded on SSR, it's not in the `results` array. It's in the single audio store. I think this is the reason for the crash you get if you load an audio on SSR, and then try clicking on a related audio:
audio

obulat (commented):
Any media item can be stored in three stores:
- the search results store (`media/index.ts`)
- the related items store (`media/related-media.ts`)
- the single result store (`media/single-result.ts`)

When getting an item by `id` in the `media` store(`/media/index.ts`), for instance, we try both the search results and the related media (I'm not sure why we don't look in the `single-result` 😬):
```
    getItemById: (state) => {
      return (mediaType: SupportedMediaType, id: string): Media | undefined => {
        const itemFromSearchResults = state.results[mediaType].items[id]
        if (itemFromSearchResults) return itemFromSearchResults
        return useRelatedMediaStore().getItemById(id)
      }
    },
```

obulat:
Another bug I've noticed is if you click on an play for some audio in the 'All results` view, and then click on a different audio link to open the single result, the first audio keeps playing, and there's no way to stop it

src/stores/media/index.ts
obulat: I think we could have a helper that would return which media store this item is in: `main`, `related-media` or `media-result`. And `setMediaProperties` could then use this helper internally to update the correct store.

sarayourfriend: Interesting.

I wonder if we should actually change the media store to be an "every piece of media ever returned by the API during this session" store and then have a separate "results" store? :thinking:

src/stores/media/index.ts
```diff
@@ -391,5 +392,20 @@ export const useMediaStore = defineStore('media', {
...
+      type: SupportedMediaType,
+      id: string,
+      properties: Partial<DetailFromMediaType<typeof type>>
+    ) {
+      const item = this.getItemById(type, id)
+      if (item) {
+        Object.assign(item, properties)
```
obulat, line 403: I didn't realize that you could update the store objects like this!

sarayourfriend: I didn't either! I was pleasantly surprised that it worked. But it makes sense, I suppose, as Object.assign mutates the existing object which is a Vue reactive (I think?) or at least part of a deep-reactive tree...

obulat (approved):
Thank you for fixing this issue and adding the base for any future work adding properties to loaded audio! Everything works well, and I don't think we need any more tests.

src/stores/media/index.ts
dhruvkb: The fact that it works is even more bizarre because [the Vue docs](https://v2.vuejs.org/v2/guide/reactivity.html#For-Objects) state that properties added using `Object.assign` will not be reactive. Maybe that's changed in the composition API 🤷?

dhruvkb (approved):
Looks good, apart from one concern regarding discrepancy with the docs.

src/stores/media/index.ts
sarayourfriend: Strange. I'm doing something wrong in this code sandbox because nothing is working as expected, but when I try `Object.assign(obj, { [count]: count ])` in the `doMutation` function it doesn't cause the computed to update. When I do `obj[count] = count` however, it also doesn't cause the computed to update...

https://codesandbox.io/s/relaxed-black-nqhos7?file=/src/components/HelloWorld.vue

Any ideas? I could definitely see Object.assign bypassing the Proxy that Vue uses to track reactive objects...

sarayourfriend: Okay, I realized that what I was doing in that previous example was slightly different. I've updated it with something more representative (watching a specific property of the obj) and it works.

Maybe the composition API changed this after all.

PR #1530, opened 2022-06-16
Add Playwright utilities

### LINKED ISSUE #797: Move common e2e test functionality into reusable utilities
## Problem

Our e2e tests mostly all block image requests for example, but duplicate the code to do this across each of the test suites that require it.

## Description

We can move this into some test environment setup utility that could get called from the e2e suites instead. This would make it easier to add new e2e tests without further duplicating this code.

## Alternatives

Continue duplicating code in the test suites. Tests don't necessarily need to adhere to DRY principles. But I think common test environment set up tasks should.

## Implementation

obulat:
## Fixes

Fixes #797 by @sarayourfriend

## Description

This PR creates some reusable functions for the Playwright tests. The functions are based on the descriptions in https://github.com/WordPress/openverse-frontend/issues/1244:
- `goToSearchTerm` - has an option of using `SSR` page (i.e., going directly) or `CSR` page (i.e., by searching through the homepage). We dismiss the translation banner for RTL pages here.
- `dismissTranslationBanner`. Here, I've edited the banner `id` to remove the language value from it. This way, it's easier to select the banner through the universal `id`, instead of having to add the language value to get the correct banner `id`.
- `scrollToBottom`/`scrollToTop`. These also have some timeouts built-in to ensure that the interface has changed (for instance, the header needs some time to update the buttons/border to the new scrolled state).
- `scrollToBottomAndUp` - this is necessary to load all images when the page has lazy-loaded images.
- `openFirstResult` - opening the first result/related result from the search page or from the detail page Related Media section.

## Testing Instructions

All the Playwright tests should pass.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VTranslationStatusBanner.vue
```diff
@@ -1,8 +1,9 @@
 <template>
   <VNotificationBanner
-    :id="bannerKey"
+    id="translation"
```
sarayourfriend, line 3: This prop needs to be unique between all instances as `VNotificationBanner` uses it to track which banners have been dismissed. With this change all translation banners have the same id here and if you dismiss one, all the others will also be treated as dismissed. You can try this by clearing your local storage, visiting `/pt` and dismissing the banner then visiting `/ru` and see that the banner will already be dismissed.

Is there any reason we can't use `data-testid` instead?

test/playwright/utils/navigation.ts
```diff
@@ -1,10 +1,41 @@
...
 const buttonSelectors = {
   filter: '[aria-controls="filters"]',
   contentSwitcher: '[aria-controls="content-switcher-modal"]',
 }
 
+export function sleep(ms: number) {
+  return new Promise<void>((resolve) =>
+    setTimeout(async () => {
+      resolve()
+    }, ms)
```
sarayourfriend, line 23: Totally a nit, but this can be cut down (and `async` declaration isn't necessary either way here):

```suggestion
    setTimeout(resolve, ms)
```

test/playwright/utils/navigation.ts
```diff
@@ -90,7 +121,100 @@ export const currentContentType = async (page: Page) => {
...
+  searchType: SearchType = ALL_MEDIA,
+  options: {
+    mode?: RenderMode
+    dir?: 'ltr' | 'rtl'
+    query?: string // Only for SSR mode
+  } = {}
+) => {
+  const dir = options.dir || 'ltr'
+  const mode = options.mode ?? 'SSR'
+  const query = options.query ? `&${options.query}` : ''
```
sarayourfriend, line 139: It seems like maybe `searchType` should also go in `options` to make it so someone could easily do a CSR of the all content search. Otherwise they either have to pass ALL_MEDIA or undefined to coerce the default out.

```suggestion
  options: {
    searchType?: SearchType
    mode?: RenderMode
    dir?: 'ltr' | 'rtl'
    query?: string // Only for SSR mode
  } = {}
) => {
  const searchType = options.searchType ?? ALL_MEDIA
  const dir = options.dir || 'ltr'
  const mode = options.mode ?? 'SSR'
  const query = options.query ? `&${options.query}` : ''
```

test/playwright/utils/navigation.ts
```diff
@@ -90,7 +121,100 @@ export const currentContentType = async (page: Page) => {
...
+  })
+  await sleep(300)
+}
+
+export const scrollToTop = async (page: Page) => {
+  await page.evaluate(() => {
+    window.scrollTo(0, 0)
+  })
+  await sleep(200)
+}
```
sarayourfriend, line 212: Rather than these semi-arbitrary sleep times, could we poll the `clientY` on the window until it matches the expected value?

```suggestion
const waitForScrollPosition = async (toY: number) => new Promise((resolve) => {
  const interval = setInterval(() => {
    page.evaluate(() => document.clientY === toY).then((isScrolledToY) => {
      if (isScrolledToY) {
        resolve()
        clearInterval(interval)
      }
    })
  }, 50)
})

export const scrollToBottom = async (page: Page) => {
  const bottomY = await page.evaluate(() => document.body.scrollHeight)
  await page.evaluate((y) => {
    window.scrollTo(0, y)
  }, [bottomY])
  await waitForScrollPosition(bottomY)
}

export const scrollToTop = async (page: Page) => {
  await page.evaluate(() => {
    window.scrollTo(0, 0)
  })
  await waitForScrollPosition(0)
}
```

I'm writing the `page.evaluate` stuff from memory so I don't think this exact code will work but it's the general idea I wanted to propose more than anything. We'd also want to implement the playwright timeout for maximum wait times as well :wink:

sarayourfriend (changes requested):

sarayourfriend:
Love this refactor :star_struck: Especially the search helpers make the test code a lot clearer in intention right off the bat for a reader.

test/playwright/utils/navigation.ts
obulat: Thank you! This will make calling the helper function much easier :)

test/playwright/utils/navigation.ts
obulat: I have written some tests following the paths you've outlined in https://github.com/WordPress/openverse-frontend/issues/1244, @sarayourfriend , and I had a lot of problems with the page being scrolled to top, but still using the styles as if it wasn't. Namely, the header buttons would not have borders, or the logo Openverse text would not display. 

I wonder if we should keep the arbitrary wait times for now until we can test if this approach works well or not? I don't want to add the tests themselves here because this PR is already too big...

src/components/VTranslationStatusBanner.vue
obulat: I saw why I wasn't using `data-testid`: it's because the `VNotificationBanner` uses the `id` prop to create its own `data-testid` that is set as `banner-${id}`, so it becomes something like `data-testid="banner-translation-ar"` :(

obulat: Buuut, I can override the `VNotificationBanner`'s `data-testid` by providing my own `data-testid` from `VTranslationStatusBanner`, so it works well!

test/playwright/utils/navigation.ts
sarayourfriend: Sure, we can. Maybe let's open another issue to remove them. I'm concerned that they'll be (a) a source of flakiness and (b) unnecessarily slow down tests waiting for things that have already settled (even if only a little bit).

test/playwright/utils/navigation.ts
```diff
@@ -89,8 +116,135 @@ export const currentContentType = async (page: Page) => {
...
+
+export const renderDirs = ['ltr', 'rtl'] as const
+
+export const renderModes = [
+  ['SSR', 'ltr'],
+  ['SSR', 'rtl'],
+  ['CSR', 'ltr'],
+  ['CSR', 'rtl'],
+] as const
```
sarayourfriend, line 242: Kind of a nit but maybe not because render is a verb and a noun so there is some potential for confusion here: what is render meant to indicate here? Could `renderDirs` just be `languageDirs` or `pageDirections`? I'd prefer `directions` in this case (and in the `pageWithDir` function name) because `dir` is also easily confused with "directory" (in fact it took me a second to understand what `dir` meant in the `pageWithDir` function based on the signature and name of the function).

`renderModes` might be clearer as `renderingContexts`.

What do you think?

test/playwright/utils/navigation.ts
```diff
@@ -89,8 +116,135 @@ export const currentContentType = async (page: Page) => {
...
+
+/**
+ * Adds '/ar' prefix to a rtl route. The path should start with '/'
+ */
+export const pathWithDir = (rawPath: string, dir: string) => {
+  const path = rawPath.startsWith('/') ? rawPath : `/${rawPath}`
+  return dir === 'rtl' ? `/ar${path}` : path
```
sarayourfriend, line 249: We can have clearer types here (and remove the `startsWith` check using a template string type :slightly_smiling_face:)

```suggestion
export const pathWithDir = (path: `/${string}`, dir: 'ltr' | 'rtl') => {
  return dir === 'rtl' ? `/ar${path}` : path
```

test/playwright/visual-regression/components/header.spec.ts
```diff
@@ -2,65 +2,43 @@ import { test } from '@playwright/test'
...
-        test('searchbar hovered', async ({ page }) => {
-          await page.hover('input')
-          await hideInputCursors(page)
-          await expectSnapshot(
-            'searchbar-hovered-ltr',
-            page.locator(headerSelector)
-          )
-        })
+  for (const dir of renderDirs) {
+    test.describe(dir, () => {
```
sarayourfriend, line 12: Given iterations over page directions should always be followed by `test.describe(dir, ...)`, should we just go ahead and make a function like the breakpoints have? `describeDirections((dir) => ...)`?

obulat: Makes sense 👍

obulat: Actually, I don't know how to implement this. The problem is that the direction is set by the URL, and not the Playwright settings as in `breakpoints`.

test/playwright/utils/navigation.ts
obulat: I am always grateful for naming suggestions! I wonder if it should be `languageDirections` to be very clear?
And `renderingContexts` is perfect!

sarayourfriend: `languageDirections` is perfect!

test/playwright/visual-regression/components/header.spec.ts
sarayourfriend: Ahh, interesting. The `describeDirections` callback could be `(getPath: (path: `/${string}`) => `${'/ar' | ''}${path}`, dir: 'ltr' | 'rtl') => void`. Then we eliminate having to import the path with direction function too:

```ts
describeDirections((getPath, dir) => {
  test('should render whatever', async ({ page }) => {
    await page.goto(getPath('/'))
    expect(page.screenshot()).toMatchSnapshot({ name: `whatever-${dir}` })
  })
})
```

What do you think?

obulat: I've tried implementing this, but I really don't understand how :( Can we create a separate issue for `describeDirections`, too, @sarayourfriend?

sarayourfriend: Sure, no problem.

dhruvkb (approved):
A much welcome refactor that makes the tests read like a series of natural-language instructions!

sarayourfriend (approved):
LGTM! Great refactor. Excited to see how this leads into more user-story focused e2e tests :tada:

PR #1535, opened 2022-06-17
Move PO file into special branch

### LINKED ISSUE #1051: PO file is not regenerated when directories are renamed
## Description

The `pre-commit` action configured by `husky` only generates the PO file when the `en.json` file is changed. But the PO file also changes when directories are named in the code, as that affects the usage of the translations.

This means that there could be events like this:
- Developer changes usage of string from `en.json`.
- Developer lints the code which does not generate the PO file.
- Developer commits this to Git.
- CI check on GitHub flags this as a lint failure.

## Expectation

The PO file generation should be able to identify other cases where it can be changed. At least it should be able to catch the same changes as the CI job does, to prevent commits from reaching GitHub and failing there.

## Additional context

If the PO generation is fast, we could just do it for every commit.

## Resolution

sarayourfriend:
## Fixes

Fixes #1051 by @dhruvkb

## Description

Fixes the problem described by @dhruvkb in the linked issue by completely changing how we manage the POT file altogether :slightly_smiling_face: 

You can check out the new `translations` branch here: https://github.com/WordPress/openverse-frontend/tree/translations

The new `generate_pot.yml` workflow will push a new version of the `openverse.pot` file every time we merge to main.

We'll need to update the GlotPress configuration to pull `openverse.pot` from the translations branch specifically once this is merged. cc @zackkrida if you can document how that is done that would be great!

Because the PO file is just used to hydrate GlotPress and then we generate the json locale files from the GlotPress API responses, we don't need to store the PO file in the main branch at all. Everything is derived.

## Testing Instructions

Verify the approach is sane and will work. For the workflow, check that the code makes sense. Then we'll find out if it works when we merge it.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb:
I like the idea of having data files in a separate branch (with their own separate history). It might be interesting to see if we can do the same for visual regression images.

Does this also mean that cloning a repo with `--single-branch` will be faster because of fewer stuff to download?

sarayourfriend:
> It might be interesting to see if we can do the same for visual regression images.

I keep wracking my brain for solutions to the visual regression images since WCEU but the problem always comes down to how to review updates to them during the course of a PR. It's critical to the workflow for visual regression images to be able to explicitly review and approve updates to them. The only solution I can think of is to use Percy or some other SaaS that stores the images off site for you and allows you to review them in their UI, with a blocking requirement for approval there before PRs can be merged.

> Does this also mean that cloning a repo with --single-branch will be faster because of fewer stuff to download?

I suppose so. However, `openverse.pot` is only 54KB, so the gain here is marginal at best under most conditions.

sarayourfriend:
Once the approach is reviewed and approved (if it gets there) then I'll rebase this PR to fix the conflict with the openverse.pot file... otherwise the conflict will just re-arise every time someone else pushes to main because I'm deleting the file entirely here :slightly_smiling_face:

obulat (commented):
I love this approach!

How are we using the `pot` file?.. We are uploading it to GlotPress, but how?

sarayourfriend:
I think actually GlotPress downloads it, IIRC. I vaguely remembering there being something where we had to configure the location of the PO file in GlotPress so it knew where to find it. Hoping that @zackkrida can clarify this as I believe he was the one to figure out that configuration.

zackkrida:
Ah, missed this ping, but yes @dd32 set this up for us the first time around. All I did was give Dion the url to the .pot file:

https://github.com/WordPress/openverse-frontend/blob/main/src/locales/po-files/openverse.pot

And as I recall GlotPress reads this file in some cron job.

zackkrida (approved):
This looks awesome, I think this approach will be perfect.

zackkrida:
The workflow didn't work. I suspect it's because the `checkout` action only checks out the current commit, and not all branches of a repository:

https://github.com/actions/checkout

We might be able to fix it by adding:

```
  with:
    fetch-depth: 0
```

to fetch _all_ branches and commits? that seems slow, so there might be a better way.

sarayourfriend:
Maybe we can use the checkout action to checkout the specific translations ref? It accepts a `ref` input: https://github.com/actions/checkout#Checkout-a-different-branch

PR #1547, opened 2022-06-28
Add audio duration filter clarifications as strings

### LINKED ISSUE #1538: Add clarification for length filters
## Problem

It can be difficult to understand what the length filter names mean.

## Description

We should add clarification. Something like:
- [ ] shortest under 30 seconds
- [ ] short 30 seconds - 2 minutes
- [ ] medium 2 - 10 minutes
- [ ] long above 10 minutes
Before implementing this, we need a design for it. This is what the current filters look like:
Screen Shot 2022-06-19 at 7 11 18 AM

## Implementation

obulat:
## Fixes

Fixes #1538 by @obulat

## Description

I've added the clarification as text so this PR does not change anything in the code. But if we want to style them differently, I would need to separate the names from the clarification texts. Please, let me think what you think.

Audio length filters with clarification

## Testing Instructions

Run the app and open the Audio search page filters.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/locales/en.json
```diff
@@ -333,10 +333,10 @@
     },
     "lengths": {
       "title": "Durations",
-      "shortest": "Shortest",
-      "short": "Short",
-      "medium": "Medium",
-      "long": "Long"
+      "shortest": "Shortest (< 30s)",
```
zackkrida, line 336:
```suggestion
      "shortest": "Shortest (<30s)",
```

For consistency with the 'long' label

src/locales/en.json
```diff
@@ -333,10 +333,10 @@
...
+      "short": "Short (30s - 2min)",
+      "medium": "Medium (2 - 10min)",
```
zackkrida, line 338:
```suggestion
      "short": "Short (30s-2min)",
      "medium": "Medium (2-10min)",
```

I believe with these kinds of ranges there isn't generally a space.

panchovm:
It looks simple and works great.

My only suggestion is to eliminate the classification (shortest, short, medium, and long) and keep the time length, as SoundCloud does.

CleanShot 2022-06-29 at 09 53 51@2x

On the other hand, I wonder if it is fine using abbreviations such as `min` for meeting i18n requirements. If so, then I suggest using `sec` for seconds and adding a whitespace character between the number and the word. The outcome would be:

```
< 30 sec
30 sec-2 min
2-10 min
> 10 min
```

zackkrida:
I decided to ask about this in the #polyglots channel. We can see if they have any suggestions:

https://wordpress.slack.com/archives/C02RP50LK/p1656514587496599

obulat:
> My only suggestion is to eliminate the classification (shortest, short, medium, and long) and keep the time length, as SoundCloud does.

That is actually much simpler and easier to understand ! :)

Thank you for asking in the polyglot channel, @zackkrida ! I think that these translations should be simple because they are hard-coded, and the values for seconds or minutes in strings like '30 sec - 1 min' do not change. If they did, we would have to go the pluralization route, which usually requires more effort. I'm going to go look at the channel :)

obulat:
Updated screenshot:

Screen Shot 2022-06-30 at 5 31 54 PM

src/locales/en.json
```diff
@@ -333,10 +333,10 @@
     },
     "lengths": {
       "title": "Durations",
```
panchovm, line 335: For consistency, I think we should use `Duration`, without the `s`. All the other categories are written in a singular form.

obulat: Great catch, @panchovm !

obulat, line 335:
```suggestion
      "title": "Duration",
```

zackkrida (approved):

panchovm (approved):
It looks great ✨

zackkrida (approved):

PR #1556, opened 2022-07-08
Deploy production frontend on release & correctly time staging deployment

### LINKED ISSUE #1557: Generate a version.json file based on the ref of the current build
## Problem

We'd like to have an Nginx sidecar container to go along with our Nuxt container in ECS. This nginx container at first will just be responsible for serving the https://search.openverse.engineering/version JSON file.

## Description

Add a second stage to the Dockerfile for the repository that builds an nginx image. Add building and publishing this image to the GHCR workflow. Push it to `openverse-frontend-nginx`. Generate a `version.json` file in the Nginx container based on the ref of the current build. For staging, it'd be nice to list the actual commit SHA of the `main` branch that is deployed rather than just `main`. It might be nice to update the staging build to do this as well anyway, instead of always just using the `main` tag.

## Alternatives

Generate and push the `version.json` file into the `static` directory inside the docker image so that Nuxt can serve it. This might be easier and could put off having to mess with and deal with potentially complicated nginx configurations. I think I prefer this approach but want to hear from @WordPress/openverse-maintainers before committing to it.

## Implementation

sarayourfriend:
## Fixes
Fixes #1557 by @sarayourfriend 

## Description

Deploys https://search-production.openverse.engineering whenever a release is cut in this repository. At least hopefully! We'll need to see if the GHCR configuration actually works for releases and if it will successfully trigger the "deploy-production" workflow to completion with the release tag. Unfortunately I'm not aware of any way to test this functionality other than by merging it and see if it works with a test release.

I also noticed that the staging deployment and the GHCR push to the `main` tag have a race condition. I've fixed that in this PR by only running the staging deployment after the GHCR image is uploaded after a push to the `main` branch.

Also: I've removed the old push_prod and push_staging workflows that were used to push images to ECR. We no longer use ECR, so those workflows will fail until they're removed.

Also: this adds the `version.json` requested by the Nginx sidecar issue. I think we can safely close that issue for now and in the future look into whether we want to use Nginx to serve static files. We don't currently do that, and the version.json endpoint isn't worth the extra complexity for now.

## Testing Instructions

I think just merge and see how it goes. Note that the existing GHCR push should still work for this branch. You can see I did a bunch of test commits to get the action changes verified as working, in particular the tag extraction for the RELEASE build arg.

To test the build arg, checkout this branch and run:

```
docker build . --build-arg RELEASE=super-sweet-release-sha-1234abcd -t openverse-frontend:latest
```

Run the generated image and verify that you can reach the `/version.json` endpoint at `http://localhost:8443/version.json`:

```
docker run -p 8443:8443/tcp openverse-frontend:latest
```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (approved):
Let's try it! Regarding rollbacks, how were we going to handle those? Was it manual runs of the GitHub workflow with a `tag` input field?

sarayourfriend:
> Let's try it! Regarding rollbacks, how were we going to handle those? Was it manual runs of the GitHub workflow with a tag input field?

Right! I forgot about that. I'll create a new "rollback production" workflow that can be dispatched with a `ref` input. It'll just call  through to the `production-frontend-deploy` action with the input ref.

.github/workflows/ghcr.yml
```diff
@@ -34,11 +52,63 @@ jobs:
...
           tags: ${{ steps.metadata.outputs.tags }}
           labels: ${{ steps.metadata.outputs.labels }}
           push: true
+          build-args: |
+            RELEASE=${{ steps.get-version.outputs.version }}
+
+  deploy-staging:
+    name: Deploy staging frontend
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
```
krysal, line 85: Will this deploy to staging on each push in a pull request (to main)?

krysal (approved):
This is exciting and fun automation! 🚀 I left some questions but nothing blocking.

.github/workflows/ghcr.yml
```diff
@@ -34,11 +52,63 @@ jobs:
           tags: |
             type=sha
             type=ref,event=tag
-            ${{ github.event_name == 'push' && 'type=raw,value=main' || '' }}
+            type=raw,value=main,enabled=${{ github.event_name == 'push' }}
+            type=raw,value=main
```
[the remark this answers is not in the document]
sarayourfriend: Oops, yes, I meant to remove this second one, it was just here for testing. Thanks for catching it!

.github/workflows/ghcr.yml
sarayourfriend: No because the push event is scoped to the main branch.

You can confirm this is true by seeing that this workflow step was skipped in this PR: https://github.com/WordPress/openverse-frontend/runs/7283335174?check_suite_focus=true

krysal: I see, makes sense! Though then is the `on: -> pull_request: ...` scope doing something in this workflow? 🤔
I remember we used to publish an image for each PR but seems that has been lost at some point.

zackkrida:
I wanted to test the version route locally after building the dockerfile with something like `docker build . -t openverse-frontend:latest --build-arg RELEASE=whatever` but the Dockerfile doesn't build on an M1 mac.

sarayourfriend:
> I wanted to test the version route locally after building the dockerfile with something like docker build . -t openverse-frontend:latest --build-arg RELEASE=whatever but the Dockerfile doesn't build on an M1 mac.

Ah, this is a known issue that we decided to accept. It has to do with some permissions issue with the COPY commands, IIRC.

I've pushed an update to change the `chmod` to `chown`. @zackkrida can you try again?

sarayourfriend:
I'm going to merge this as it's working in CI and for me locally, the worst case scenario is that it still doesn't build on an M1, but I think it should fix it. If not we can open another issue if we still wish to address this.

Issue #1237, opened 2022-04-05
Warnings in dev command: `WARN  Cannot stringify a function bound`


zackkrida:
## Description

Run `pnpm dev` and observe multiple instances of the error `WARN  Cannot stringify a function bound`.

I would _suspect_ that this is a matter of a server-rendered setup function that returns a function which is stringified.

sarayourfriend:
This is related to `@nuxt/devalue` which marshals state to and from the serialized version used for hydrating the client. Whomever looks into this might want to throw some debugging statements in devalue to see if they can tell _which_ function it is complaining about... It'd be nice if it would just call `toString` on the function so you could at least see the implementation and know where to debug in your app :shrug:

sarayourfriend:
Appears this could be related to the media store to Pinia changes, I think that's when about I noticed this started to happen.

https://github.com/nuxt-community/composition-api/issues/386

The `search.vue` component uses both `fetch` and `setup`. Apparently this warning is unavoidable. Should we just ignore it? It's pretty annoying.

Alternatively, we can convert `search.vue` to not use `setup` anymore OR use `useFetch` from `@nuxtjs/composition-api` instead of `fetch` (which will be deprecated, but either way we'll need to re-write this component when upgrading to Bridge/Nuxt 3 it seems).

sarayourfriend:
Assigning medium priority because this is pretty annoying for DX currently.

obulat:
I think #1260 solves this issue. At least I couldn't see any warnings when I tried searching, viewing single media items and loading more items.

sarayourfriend:
@obulat looks to me like it is fixed on that branch as well! Yay! :rocket:

PR #1569, opened 2022-07-15
Frontend tokens

### LINKED ISSUE #1558: Allow configuring an API key to use for server-side Openverse API requests
## Problem

Currently the production frontend exempts itself from the API throttling by auto-magically adding its IP address to the list of exempted IPs in our production Redis. This has several bad problems:

1. It requires managing bespoke throttling classes (see https://github.com/WordPress/openverse-api/issues/585) to enable this behavior
2. It requires bad and leaky permissions to be given to the frontend EC2 instances for it to be able to access and write to the Redis cache that it otherwise does not need access to. This creates a new potential vector for someone to attack our API.

We don't want to replicate the second part in our ECS deployment (indeed, it would be a pain in the butt!) so let's just remove this anti-pattern altogether and use API keys which are expressly designed for this purpose.

## Description

Add a new environment variable to the Nuxt app, `OPENVERSE_API_KEY` that is used for server-side API requests. Ensure that it is _only_ used for server side requests and that the key is under no circumstances leaked to the client.

## Implementation

dhruvkb:
## Fixes

Fixes #1558 by @sarayourfriend 

## Description

This PR adds a mechanism to send authenticated requests when an API key has been provided (and unauthenticated ones) when not.

This means that API calls during SSR will be authenticated, and not throttled, but API calls from the client (subsequent calls) will be throttled as an authenticated user because they won't have any API key set.

## Testing Instructions

0. Check out the repo.
1. Follow instructions in `TESTING_GUIDELINES.md` for setting up your local environment with API authentication tokens after registering an app on your local dev server or on staging.
    * Be sure to use the version of the document from this branch as I've introduced new documentation for how to do this!
2. Visit the application as normal and confirm in the logs you can see the API token flow. Especially confirm that the _first_ request does wait for the token to come back before continuing.

Furthermore, test using the docker image:

```
docker build . --build-arg RELEASE=v0-test --build-arg API_URL=https://api-dev.openverse.engineering/ -t openverse-frontend:latest
docker run --env-file .env -p 8443:8443/tcp openverse-frontend:latest
```

> **Note:** Earlier this testing instructions read to use a locally running API with the docker container. This doesn't work due to CORS issues and having different URLs for the client and server to hit the API (due to the different network contexts they're running in). To test docker, please use a live API instance. To test against a local API, run the Nuxt app outside of docker using `pnpm prod` and an `.env` file set up to point to your local API.

Make sure to visit the docker-running service at `localhost:8443` and _not_ the IP address that nuxt spits out. That IP address is internal to the docker network. For some reason it still lets some requests through but most of them will fail (including static assets). If you stick to `localhost:8443` it works fine though.

When making requests to the docker image, you will see more than one API token being generated now in the logs, depending on the number of CPUs docker has access to. That is because pm2 runs a worker per CPU it finds. If you're using docker desktop and it is not configured with multiple CPUs then you will not see this behavior. Please enable at least 2 CPUs so that you can test this!

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
If we put the client key and secret into the `privateRuntimeConfig`, could we put a function into context instead of the actual API key? Otherwise, we could add a server-only middleware that checks a module-level variable that stores the API key and expiration time.

```ts
// ~/plugins/api-authentication.server.ts

import type { Plugin, Context } from '@nuxt/types'

interface State {
  apiKey: string
  expiration: Date
}

const local: { state: State | null } = {
  state: null,
}

const getApiKey = async (ctx: Context): State => {
  /* ... */
}

const expiresSoon = (s: State): boolean => {
  /* ... */
}

const apiAuthentication: Plugin = async (ctx, inject) => {
  if (!local.state || expiresSoon(local.state)) {
    local.state = await getApiKey(ctx)
  }

  inject('openverseApiKey', local.state.apiKey)
}
```

sarayourfriend:
It just sucks to have to have the `apiService` wired up and dependent on Nuxt state. It's suck a hack, like you said, and it infects _everything_.

sarayourfriend:
With the code above I suggested I think `getApiKey` would need an async lock on it as well to prevent multiple requests per worker to get a token :thinking:

dhruvkb:
@sarayourfriend I tried this with a server-side plugin and it was refreshing the access token on every request. It seems to me that plugins are reinitialised on every request and so their state always starts as a blank one.

sarayourfriend:
> it was refreshing the access token on every request

This is odd. I can reproduce it in the dev build but in the production build it works as expected (like a normal node module resolution).

I'll have some changes up soon.

sarayourfriend:
Changes pushed. They work if you do the following:

1. Create a `.env` file with `API_URL` pointing to your local API
2. Add `API_CLIENT_ID` and `API_CLIENT_SECRET` with client credentials for a local API exempted token
3. Build and run the server: `env $(cat .env) pnpm build && env $(cat .env) pnpm start`

You can try the `/secret-test` page but also just check out the console as I've added additional logging to make it easy to track the request through the plugin.

sarayourfriend:
Unit tests are currently failing due to missing API token so we'll have to sort out a way around that. Pushing the token into context makes the component code that relies on it a lot simpler. No need to do conditional checks or handle asynchrony.

sarayourfriend:
This PR is ready for review now and the description has been updated with up-to-date testing instructions. You can also visit the `/secret-test` page, but I will remove that before merging the PR.

src/data/api-service.ts
```diff
@@ -34,14 +35,61 @@ Please check the url: ${config.baseURL}${config.url}`
...
+  /** to make authenticated requests to the API */
+  accessToken?: string
+  /** whether to use the `'v1/'` prefix after the base URL */
+  isVersioned?: boolean
+}
+
+/**
+ * the schema of the API service
+ */
+export interface ApiService {
```
obulat, line 52: Nice!

TESTING_GUIDELINES.md
```diff
@@ -12,10 +12,40 @@ Staging is redeployed everytime we merge to the `main` branch, so if you're look
...
+For the first, [run the Openverse API locally](https://github.com/WordPress/openverse-api). Then create a `.env` file with the following:
+
+```shell
+API_URL="http://localhost:8000/"
+```
+
+The run the dev server with environment loading:
+
+```shell
+pnpm dev:env
```
krysal, line 35: Is this necessary? I was using and `.env` file before this PR, and variables were getting in fine.

sarayourfriend: Oh really? I didn't realize but I guess Nuxt has it built in! Thanks for pointing it out. I will revise the instructions.

sarayourfriend: Fixed in https://github.com/WordPress/openverse-frontend/pull/1569/commits/0976ebd6bfa10b7aa937c67307ca300755ed76cf

Thanks again!

dhruvkb (commented):
I ran this locally and the "happy flow" seems to be working pretty well. I'm currently trying to understand how we could test these events:
- access tokens expiring
- token fetch failing (perhaps invalid credentials)
- requests being blocked till the token is refreshed

src/plugins/api-token.server.ts
```diff
@@ -0,0 +1,140 @@
...
+    log(`Next token expiry for worker expiry=${res.data.expires_in}`)
+  } catch (e) {
+    /**
+     * If an error occurs, serve the current request (and any pending)
+     * anonymously and hope it works. By setting the expiry to 0 we queue
+     * up another token fetch attempt for the next request.
+     */
+    error('Unable to retrieve API token, clearing existing token', e)
+    process.tokenData.accessToken = ''
+    process.tokenData.accessTokenExpiry = 0
```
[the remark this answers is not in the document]
sarayourfriend: It would. Ideally we'd have some way to monitor this. We could dispatch a Sentry error? Alternatively CloudWatch log insights could be used to build an alarm... but given we already have a workflow around Sentry I'll try to add that!

sarayourfriend:
I think we could test all of those using the axios mock adapter. Thanks for pointing them out. I will write those tests.

krysal (approved):
This works well on my end, even the docker build on mac with some extra packages. I have some questions, but nothing worth blocking, I believe 👏

Perhaps when the token expires we can force a hard refresh at the next request, so the request is done server side, and then the token gets refreshed? It's not ideal but could be a solution in the interim.

src/pages/search.vue
```diff
@@ -145,10 +136,23 @@ export default defineComponent({
...
+    handleTab(event: KeyboardEvent, element: string) {
+      if (this.showScrollButton.value && element !== 'scroll-button') {
+        return
+      }
+      focusIn(document.getElementById('__layout'), Focus.First)
+    },
+    fetchMedia(...args: unknown[]) {
+      const mediaStore = useMediaStore(this.$pinia)
+      return mediaStore.fetchMedia(...args)
+    },
```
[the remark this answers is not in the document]
sarayourfriend: Because you can't return functions in `setup` and also use Options API for some things or else Nuxt will raise the "cannot stringify unbound function" error in the console. See https://github.com/WordPress/openverse-frontend/issues/1237 for where we've fixed this before.

nuxt.config.ts
```diff
@@ -163,14 +163,19 @@ const config: NuxtConfig = {
...
+    '~/plugins/url-change.ts',
+    '~/plugins/migration-notice.ts',
+    '~/plugins/ua-parse.ts',
+    '~/plugins/focus-visible.client.ts',
+    '~/plugins/api-token.server.ts',
   ],
   css: ['~/styles/tailwind.css', '~/assets/fonts.css', '~/styles/accent.css'],
   head,
-  env,
+  env, // TODO: Replace with `publicRuntimeConfig`
```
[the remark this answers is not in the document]
sarayourfriend: I don't think this is a valid todo. All of the variables in `env` are required by the client and are also safe to send to the client.

sarayourfriend:
> Perhaps when the token expires we can force a hard refresh at the next request, so the request is done server side, and then the token gets refreshed? It's not ideal but could be a solution in the interim.

I don't I understand the suggestion or what problem it is trying to solve. The token requests always happen on the server side and the token is never sent to the client (this would be unsafe). When the token is about to expire, we pre-emptively request a new one and all other requests that come in will wait until the new token is retrieved.

@krysal I've also dismissed your review because I made some non-trivial changes to the logic and I want to make sure the additional documentation in the code is reviewed and comprehensible. Can you re-review with an eye towards that?

I've also added extensive unit tests, some of which are rather complicated due to juggling asynchrony in the tests themselves. @dhruvkb please review the tests and let me know if they are sufficient.

If anyone can think of additional tests to write, please let me know.

src/data/api-service.ts
```diff
@@ -34,14 +35,61 @@ Please check the url: ${config.baseURL}${config.url}`
...
+  ): Promise<AxiosResponse<T>>
+}
+
+export const createApiService = ({
+  baseUrl = process.env.apiUrl,
+  accessToken = undefined,
+  isVersioned = true,
+}: ApiServiceConfig = {}): ApiService => {
+  const axiosParams: AxiosRequestConfig = {
+    baseURL: isVersioned ? `${baseUrl}v1/` : baseUrl,
```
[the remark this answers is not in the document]
sarayourfriend: I think `isVersioned` here assumes that the api service being created will only be used for a particular set of resources or requests rather than all requests. Because we create multiple axios clients for different api service usages, we pass `isVersioned` when we want the URLs to automatically include the v1 string.

However, I also disagree with this change. @dhruvkb can you explain the benefit of this implementation over explicitly including the version string at the request site? If we introduce new versions of existing endpoints or have api services used for mixed version requests in the same cycle then this implementation will either have to be reverted back to the explicit version or be significantly increase the complexity of this code.

I prefer explicit request URLs personally, they are easier to search for in the app and make zero assumptions about the future or current versions of API endpoints that will or are being used.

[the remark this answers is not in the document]
sarayourfriend: Well, I think undoing this refactor is actually more trouble than it is worth and will complicate this PR. The existing "versioned API" approach we took was effectively the same and had the same bad assumptions (that `v1` is the only version and that all requests performed by a given API service instance will all reference the same version of endpoint).

I personally find the current approach with the `getResourceSlug` function over-complicated, sneaky, and bakes assumptions into the base `api-service.ts` module that are not necessarily true.

Either the API service instances are only for making requests to the media endpoints (and we should use a plain axios instance for anything else) or we have to reconsider how the API service is used in the first place.

I think it's potentially an anti-pattern, the way it's currently implemented, and we might want to consider stuffing it into Nuxt context, either by hand-spinning our own module or using the existing nuxt/axios module.

This would have some significant benefits:

1. We could insert the plugin after the API token request plugin so that it can automatically include the access token on the server. This allows downstream requesters to completely ignore the api token detail. Currently all api service instantiators have to consider the api token. This duplicates the work and complicates the downstream code.
2. We can eliminate the complexity in the `api-service` module completely. The only issue that remains is the mapping of the media type to the endpoint names, but this can easily be put into a function or a static mapping for reference by requesters to construct API urls.

I'll open an issue for this. For now I think this should remain as is with @dhruvkb's refactors because it keeps the existing assumptions, but makes them ever so slightly more flexible to accommodate non-versioned API requests.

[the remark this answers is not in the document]
sarayourfriend: https://github.com/WordPress/openverse-frontend/issues/1591

src/data/api-service.ts
```diff
@@ -13,7 +13,8 @@ const DEFAULT_REQUEST_TIMEOUT = 30000
  */
 export const getResourceSlug = (resource: string): string => {
   const slug = { [AUDIO]: 'audio', [IMAGE]: 'images' }[resource] ?? resource
-  return `${slug}/`
+  const noTrailingSlash = ['rate_limit'] // TODO: Ensure all API endpoints have trailing slashes
```
obulat, line 16: Is this something that needs to be done in the API? If so, maybe we can add a link to the API issue here to follow up?

src/pages/secret-test.vue
```diff
@@ -0,0 +1,67 @@
...
+        apiClientSecret,
+      },
+      apiRes: {
+        message: { error: 'Something is not working properly.' },
+        isSuccess: false,
+      },
+    }
+    try {
+      const apiService = createApiService({ accessToken: $openverseApiToken })
+      data.apiRes.message = (await apiService.query('rate_limit/', {}))
```
obulat, line 53:
```suggestion
      data.apiRes.message = (await apiService.query('rate_limit', {}))
```

obulat: Currently, this gives an error because `noTrailingSlash` array contains `rate_limit`, not `rate_limit/` with trailing slash, so `getResourceSlug` not only does not _not add_ a slash, but adds one more.

obulat:
I cannot build docker using the commands in the testing instructions because python cannot be found in the path.

<details>

<summary>Error message: installing deasync requires Python</summary>

```shell
#9 26.19 .../node_modules/@sentry/cli install: Done
#9 26.27 .../deasync@0.1.24/node_modules/deasync install$ node ./build.js
#9 26.37 .../deasync@0.1.24/node_modules/deasync install: gyp info it worked if it ends with ok
#9 26.37 .../deasync@0.1.24/node_modules/deasync install: gyp info using node-gyp@8.4.1
#9 26.37 .../deasync@0.1.24/node_modules/deasync install: gyp info using node@16.16.0 | linux | arm64
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python 
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python Python is not set from command line or npm configuration
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python Python is not set from environment variable PYTHON
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python checking if "python3" can be used
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python - "python3" is not in PATH or produced an error
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python checking if "python" can be used
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python - "python" is not in PATH or produced an error
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python 
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python **********************************************************
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python You need to install the latest version of Python.
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python Node-gyp should be able to find and use Python. If not,
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python you can try one of the following options:
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python - Use the switch --python="/path/to/pythonexecutable"
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python   (accepted by both node-gyp and npm)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python - Set the environment variable PYTHON
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python - Set the npm configuration variable python:
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python   npm config set python "/path/to/pythonexecutable"
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python For more information consult the documentation at:
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python https://github.com/nodejs/node-gyp#installation
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python **********************************************************
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! find Python 
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! configure error 
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack Error: Could not find any Python installation to use
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack     at PythonFinder.fail (/usr/local/lib/node_modules/pnpm/dist/node_modules/node-gyp/lib/find-python.js:330:47)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack     at PythonFinder.runChecks (/usr/local/lib/node_modules/pnpm/dist/node_modules/node-gyp/lib/find-python.js:159:21)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack     at PythonFinder.<anonymous> (/usr/local/lib/node_modules/pnpm/dist/node_modules/node-gyp/lib/find-python.js:202:16)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack     at PythonFinder.execFileCallback (/usr/local/lib/node_modules/pnpm/dist/node_modules/node-gyp/lib/find-python.js:294:16)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack     at exithandler (node:child_process:406:5)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack     at ChildProcess.errorhandler (node:child_process:418:5)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack     at ChildProcess.emit (node:events:527:28)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack     at Process.ChildProcess._handle.onexit (node:internal/child_process:289:12)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack     at onErrorNT (node:internal/child_process:478:16)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! stack     at processTicksAndRejections (node:internal/process/task_queues:83:21)
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! System Linux 5.10.104-linuxkit
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! command "/usr/local/bin/node" "/usr/local/lib/node_modules/pnpm/dist/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! cwd /home/node/app/node_modules/.pnpm/deasync@0.1.24/node_modules/deasync
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! node -v v16.16.0
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! node-gyp -v v8.4.1
#9 26.41 .../deasync@0.1.24/node_modules/deasync install: gyp ERR! not ok 
#9 26.42 .../deasync@0.1.24/node_modules/deasync install: Build failed
```

</details>

src/data/api-service.ts
sarayourfriend: It's already fixed in the API: https://github.com/WordPress/openverse-api/pull/801

This condition needs to stay, however, as the API isn't deployed with that fix yet.

.env.template
```diff
@@ -0,0 +1,3 @@
+#API_URL=http://127.17.0.1:8000/
+#API_CLIENT_ID=""
+#API_CLIENT_SECRET=""
```
zackkrida, line 3:
```suggestion
API_URL=http://127.17.0.1:8000/
API_CLIENT_ID=""
API_CLIENT_SECRET=""
``` 

Nit, maybe we uncomment these so there's one fewer step to edit the values once the user copies `.env.template` to `.env`?

TESTING_GUIDELINES.md
```diff
@@ -12,10 +12,46 @@ Staging is redeployed everytime we merge to the `main` branch, so if you're look
...
+1. You can change your local Nuxt server to point to a local API server that does not have throttling enabled
+2. You can introduce the environment variables necessary for authenticating with an API
+
+For the first, [run the Openverse API locally](https://github.com/WordPress/openverse-api). Then create a `.env` file by copying the `.env.template` file and update it with the following:
+
+```shell
+API_URL="http://localhost:8000/"
+```
+
+Be sure to remove the leading `#` to uncomment the variable from the copied template.
```
zackkrida, line 32:
```suggestion
```

If we decide to remove the comments entirely (if not disregard).

TESTING_GUIDELINES.md
```diff
@@ -12,10 +12,46 @@ Staging is redeployed everytime we merge to the `main` branch, so if you're look
...
+
+The run the dev server as usual:
+
+```shell
+pnpm dev
+```
+
+For the second, you'll need to follow [the instructions to set up an OAuth application](https://api.openverse.engineering/v1/#section/Register-and-Authenticate) and then fill in the `API_CLIENT_ID` and `API_CLIENT_SECRET` in the `.env` file copied from `.env.template`. Be sure to remove the leading `#` to uncomment these variables from the copied template.
```
zackkrida, line 40:
```suggestion
For the second, you'll need to follow [the instructions to set up an OAuth application](https://api.openverse.engineering/v1/#section/Register-and-Authenticate) and then fill in the `API_CLIENT_ID` and `API_CLIENT_SECRET` in the `.env` file copied from `.env.template`.
```

If we decide to remove the comments entirely (if not disregard).

TESTING_GUIDELINES.md
```diff
@@ -12,10 +12,46 @@ Staging is redeployed everytime we merge to the `main` branch, so if you're look
...
+You can also access it from other devices in your same network (like a mobile phone) for additional testing. See the [finding your local IP address](./README.md#finding-your-local-ip-address) section of the README for how to identify the local IP adress Nuxt is served on. Once you have identified your local IP address, you can access the website running on your computer by visiting `https://<local IP>:8443` replacing `<local IP>` (including the brackets) with the value you found using the instructions above in your mobile device's browser.
 
 Testing from multiple different devices as often as possible is a great way to contribute to Openverse's frontend development.
 
+### API Authentication
```
zackkrida, line 19: This new section is great, very straightforward 👍

test/unit/specs/plugins/api-token.server.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+       * to check the mutex and await on the fetching promise.
+       *
+       * This relies on the behavior of the Node event loop where
+       * several async functions called synchronously in succession will exectue
+       * up until the first blocking `await` and then return the promise. This allows
+       * us to effectively get all three of the async api token plugin function
+       * calls up to the first blocking await which will either be the the call to
+       * `refreshApiAccessToken` which makes the axios call (blocked by the adapter
+       * mock in this test) _or_ awaiting the promise shared by the entire process.
+       */
```
zackkrida, line 162: phenomenal documentation here, thanks

test/unit/test-utils/api-service-mock.ts
```diff
@@ -0,0 +1,80 @@
...
+ *     await createApiService().get('hello')
+ *   ).toEqual(', Openverse!')
+ * })
+ * ```
+ *
+ * Note in the example that despite `createApiService` being called
+ * twice, the mock adapter is not re-applied and the second registered
+ * response still works as expected.
+ *
+ * With the particual way this mock creation function works,
```
zackkrida, line 40:
```suggestion
 * With the particular way this mock creation function works,
```

.env.template
zackkrida: Eh I'm not sold on this, and it's not important.

src/plugins/api-token.server.ts
```diff
@@ -0,0 +1,193 @@
...
+const getApiAccessToken = async (
+  context: Context
+): Promise<string | undefined> => {
+  const { apiClientId, apiClientSecret } = context.$config
+  if (!(apiClientId || apiClientSecret)) {
+    log('No API client secrets found.')
+    return undefined
+  }
+
+  log('We have client information')
```
zackkrida, line 133:
```suggestion
  log('API client secrets found.')
```

For parity with the other log

src/plugins/api-token.server.ts
```diff
@@ -0,0 +1,193 @@
...
+
+const apiToken: Plugin = async (context, inject) => {
+  let openverseApiToken: string | undefined
+  try {
+    openverseApiToken = await getApiAccessToken(context)
+
+    if (openverseApiToken) {
+      log('injecting openverseApiToken into request context')
+    } else {
+      log('using empty openverseApiToken')
```
zackkrida, line 183:
```suggestion
      log('injecting empty openverseApiToken into request context')
```

test/unit/.eslintrc.js
```diff
@@ -3,5 +3,13 @@ module.exports = {
...
   rules: {
     'import/no-named-as-default-member': ['off'],
+    'no-restricted-imports': [
+      'error',
+      {
+        name: 'pinia',
+        message:
+          'Please import pinia test utils from `~~/test/unit/test-utils/pinia`. The test-utils version ensures proper setup of universally necessary Nuxt context mocks.',
+      },
+    ],
```
zackkrida, line 13: Nice!

test/unit/test-utils/api-service-mock.ts
```diff
@@ -0,0 +1,80 @@
...
+const originalApiService = jest.requireActual('~/data/api-service')
+
+const createApiService = apiService.createApiService as jest.Mock<
+  ReturnType<typeof apiService.createApiService>,
+  Parameters<typeof apiService.createApiService>
+>
+
+/**
+ * Creates an API Serivce that get's reused throughout the lifetime of
+ * the current context in which it is called.
```
zackkrida, line 15:
```suggestion
 * Creates an API Service that gets reused throughout the lifetime of
 * the current context in which it is called.
```

zackkrida (approved):
I just lost a long review comment I was writing to an accidental pagerefresh, but I spent a lot of time testing this and don't have nearly as much feedback as I'd hoped to be able to give. I applied @krysal's docker changes from #1609 and verified the correct behavior there, made sure that `pnpm dev` worked locally with API credentials that were invalid, empty, valid, and with no `.env` file. 

Awesome work.

obulat (commented):
I couldn't run the Docker setup anyways, even with the changes from @krysal's PR (on that PR, Docker ran fine. Here, with the copy of those changes, it starts running and then stops for some reason. I would have to investigate more into why)
I have a really general question:
>This means that API calls during SSR will be authenticated, and not throttled, but API calls from the client (subsequent calls) will be throttled as an authenticated user because they won't have any API key set.
With the changes in this PR, will the client calls be throttled as _unauthenticated_ user? So, you would only be allowed _100 requests per day and 5 requests per hour rate limit_? 

I think I'm misunderstanding this PR, because if we only have 5 requests per hour, the frontend will be unusable. A single 'All content' search makes 2 requests (images & audio). So, 3 "All content" searches (or clicking on "Load more" or one or two filters selected) would overuse the limit...

sarayourfriend:
> I think I'm misunderstanding this PR, because if we only have 5 requests per hour, the frontend will be unusable. A single 'All content' search makes 2 requests (images & audio). So, 3 "All content" searches (or clicking on "Load more" or one or two filters selected) would overuse the limit...

This is already the case, clients do not make authenticated or exempted requests. I do not know why throttling doesn't appear to be a problem. I tried to trigger the throttling against the production API recently and wasn't able to either. Keep in mind that clients individually making anonymous requests to the API do not add to the same throttle count as the server individually making requests to the API to fulfill SSR requests.

This PR doesn't change the client story at all, it's only changing the way the _server_ exempts itself from throttling.

obulat:
Thank you for the explanation, @sarayourfriend !

> I do not know why throttling doesn't appear to be a problem. I tried to trigger the throttling against the production API recently and wasn't able to either.

This is weird...

obulat:
I tried running this code (from Docker) against the newly-deployed dev api (https://api-dev.openverse.engineering), and it got an error response 401, 'www-authenticate': 'Bearer error="invalid_client"' when posting to '/v1/auth_tokens/token/' from SSR.

obulat:
Here's what I'm doing to run front end in Docker with the local API:
1. Add `host.docker.internal` to the `allowed_hosts` in Django settings: 
```
ALLOWED_HOSTS = config("ALLOWED_HOSTS").split(",") + [
    gethostname(),
    gethostbyname(gethostname()),
    "host.docker.internal"
]
```
3. Create an `.env` file at the root of the front end repository and add `API_URL="http://localhost:8000/"` there, together with API_CLIENT_ID and API_CLIENT_SECRET.
4. docker build . --build-arg RELEASE=v0-test --build-arg API_URL=http://host.docker.internal:8000/ -t openverse-frontend:latest  (from the testing instructions,replacing the API_URL)
5. docker run --env-file .env -p 8443:8443/tcp openverse-frontend:latest

I don't get any results due to CORS error: `Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://host.docker.internal:8000/v1/images/?q=cat. (Reason: CORS request did not succeed). Status code: (null).` :(

src/data/api-service.ts
obulat: @sarayourfriend , I think the API was deployed yesterday?

sarayourfriend: Yup! Thanks :+1:

sarayourfriend:
> I don't get any results due to CORS error: Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://host.docker.internal:8000/v1/images/?q=cat. (Reason: CORS request did not succeed). Status code: (null). :(

Is this on the client side or the server side? I've run into that problem on the client side a bunch when trying to run the frontend against a local API with everything in Docker. The problem is having to use a different URL on the client side than is used on the server side to hit the API, _I think_.

It's really far outside the scope of this change though. Please test this with `api-dev` as it's a much more realistic "true to prod" version. Everything, client and server, will be using the same URL to access the API.

sarayourfriend:
I've just re-tested and confirmed that it works fine if I use `api-dev` for the API url rather than trying to do it locally.

If you need to test it locally then please don't test it with the docker container, do it by running `pnpm prod` with the `.env` file set up with local API url and client information.

sarayourfriend:
Okay, so for me the 401 error ended up being because I missed the final character of my client ID in the env file. It was `xxxxxxxx` and it should have been `xxxxxxxxi`. This was just a bad copy paste job. Can you double check @obulat that your client information is correct in the .env file?

zackkrida:
I can also confirm; not seeing any errors with the API_URL set to `https://api-dev.openverse.engineering/` and using my client and client secret that I registered and verified.

obulat:
Woohoo 🎉 I can confirm that with `api-dev` this works well, and I see no errors!

obulat (approved):

Issue #1287, opened 2022-04-17
Audio without category shows 'filters.audio-categories.null' in the search result box 


obulat:
## Description

Audio without category shows 'filters.audio-categories.null' in the search result box.

## Reproduction

1.  Go to https://wordpress.org/openverse/search/?q=coffee
2.  On the 4th row (on desktop) you will see a box with the `audio-categories.null` message.

## Screenshots

<img width="444" alt="Screen Shot 2022-04-17 at 6 37 23 AM" src="https://user-images.githubusercontent.com/15233243/163699382-706ca639-f301-49de-bcac-31ec29c35416.png">

## Additional context

It's in prod, too.

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

PR #1575, opened 2022-07-20
Use saved locale files required for CI from test/locales

### LINKED ISSUE #1570: Remove need to make external requests for i18n in CI
## Problem

We get rate limited in CI when making requests to get translation files.

https://github.com/WordPress/openverse-frontend/runs/7388883247?check_suite_focus=true

## Description

It shouldn't be necessary for the CI to have these translation files pulled from external sources, at least not for _every_ step.

We could try some of the following options:
1. Request the translations in one step that uploads an artifact with them and then add it as a dependency of all the other steps in `depends-on`, who can then download them as artifacts.
2. Remove the need to get these translation files entirely by using auto-generated [fake "pirate" translations for testing](https://make.wordpress.org/polyglots/2017/01/05/hi-4/).

## Implementation

obulat:
## Fixes

Fixes #1570 by @sarayourfriend

## Description

This is an attempt to fix the linked issue by manually saving the locale files to the `test/locales` folder and copying them to `src/locales` during the `setup-node-env` step.

To run the Openverse Nuxt app, we need a list of locale objects inside the `src/locales/scripts/valid-locales.json`, and a corresponding set of translation files inside `src/locales` (e.g. `en.json`, `ar.json` and `es.json`). 
`valid-locales.json` is a list of locales available on `wordpress.org`. Most properties of the locales are updated very infrequently (probably at most once every WordPress release?). The only locale object property frequently updated is `translated`, the percentage of Openverse strings (from `en.json`) that have been translated on `translate.wordpress.org/projects/openverse`.  We display the translation banner for the specific locale if it is below a certain threshold. To test the translation banner, we could manually set this property, or use an existing locale that has low translated percentage.

Since updates to `valid-locales.json` are not significant for CI, and the file size is small, we can simply use the saved file for CI by copying it from `test/locales` to `src/locales` during `setup-node-env` step.

This PR saves 4 locale-related files:
- "valid-locales.json" with 3 locales (`es`, `ru`, `ar`)
- "es" locale for testing redirects.
- "ru" locale with its current `translated` value of ~40% for testing translation banner.
- "ar" locale for testing RTL. I used Google translate to manually translate all the strings from `en.json` into Arabic, as discussed in the Public Slack meeting [[start of the discussion]](https://wordpress.slack.com/archives/C02012JB00N/p1658245462022729).

We will still make one request per CI run to `translate.wordpress.org` because the `Check Nuxt build` task runs the `package.json` `prebuild` script: `"prebuild": "pnpm install && pnpm i18n". This is necessary to ensure that nothing breaks in the build process.

## Testing Instructions

Check the CI. All the tests should pass, including the RTL Playwright tests (e2e and visual regression)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


package.json
```diff
@@ -27,7 +27,7 @@
     "prod": "pnpm build:only && pnpm start",
     "storybook": "nuxt storybook --port=54000",
     "storybook:build": "nuxt storybook build",
-    "storybook:smoke": "pnpm storybook --ci --smoke-test",
+    "storybook:smoke": "pnpm i18n && pnpm storybook --ci --smoke-test",
```
sarayourfriend, line 30: Why does storybook smoke test need to fetch translations?

test/playwright/utils/navigation.ts
```diff
@@ -184,16 +184,17 @@ export const goToSearchTerm = async (
     await searchInput.type(term)
     // Click search button
     // Wait for navigation
+    const searchLabel = dir === 'ltr' ? 'Search' : 'يبحث'
```
sarayourfriend, line 187: It'd be great if we could pull these from the actual JSON files instead :thinking:

sarayourfriend (commented):
> We will still make one request per CI run to translate.wordpress.org because the Check Nuxt build task runs the package.json prebuild script: `"prebuild": "pnpm install && pnpm i18n". This is necessary to ensure that nothing breaks in the build process.

The Docker image build also runs `pnpm i18n`, FWIW :+1: Along with the one in storybook:smoke I think it would be three requests to translate.wordpress.org?

This works great, however, do we need _some_ process to update the "100%" files with new strings any time `en.json` is updated?

test/playwright/utils/navigation.ts
obulat: Do you think finding elements by text is still the best way, even with i18n complications? It kind of made me wish for the simplicity of data-testid when editing these tests...

package.json
obulat: It's a leftover from my previous attempt at implementing this. Thank you for catching it!

test/playwright/utils/navigation.ts
obulat: I've updated the commented line, and the search type names object to use the actual JSON files.
Do you know if we can get the page's `dir` property from `Playwright` `page` object? Otherwise, we would have to pass `direction` to all the utility functions that locate page elements :(

obulat:
> This works great, however, do we need some process to update the "100%" files with new strings any time en.json is updated?

Ideally, yes. In practice, I'm not sure how best to proceed with this.

We could have a script that would run weekly (monthly) and analyze which keys that were added/removed from the `en.json` are missing in the translation files, and add them in, possibly also getting some machine translation for them. 
I think this is a low priority for two reasons:
- Vue i18n automatically falls back on the English text if it doesn't find a string in the translation file. So, CI should still pass as it would be testing the existing strings, not the new ones.
- The translation banner would appear only after the `translated` percentage falls below 90%, and I don't anticipate that we will change that much in the near future.

test/playwright/utils/navigation.ts
sarayourfriend: It's interesting. I think from an a11y perspective it's best to use the label strings because those most closely reflect how an actual user will identify the elements. If our e2e tests can't work with those, then it would indicate truly that a user may not be able to use it either.

On the other hand, we run into this issue where we have to copy the strings out of the translations... I'm really not sure :slightly_smiling_face: Both seem sub-optimal but one has a11y implications and the other is just annoying from a maintenance perspective, so I'm inclined to choose the one that has better a11y guarantees (which you've used here) even if it's more annoying to maintain.

sarayourfriend:
> The translation banner would appear only after the translated percentage falls below 90%, and I don't anticipate that we will change that much in the near future.

This was the main thing I was thinking would need to be avoided somehow, but you're right, it's very low priority. We are not moving that quickly at adding new features. Can we cut an issue for it though? I like the idea of just doing it in a cron workflow as you suggested.

obulat:
> > The translation banner would appear only after the translated percentage falls below 90%, and I don't anticipate that we will change that much in the near future.
> 
> This was the main thing I was thinking would need to be avoided somehow, but you're right, it's very low priority. We are not moving that quickly at adding new features. Can we cut an issue for it though? I like the idea of just doing it in a cron workflow as you suggested.

Issue #1582 opened :)

@sarayourfriend, I removed the `Check translation files` step from the CI because it does not really suggest actionable steps and has some false positives for when the missing key is dynamic (something like `$t(\`search.${media-type}\`))`. But I guess this step is set as required in Terraform, so it shows up as `Waiting for status to be reported`.

sarayourfriend:
That is really unfortunate that the check doesn't work with dynamic labels. It's a very nice check to have and not one I'm super comfortable removing without some good forethought.

Could we follow any of the suggestions in this issue for dealing with dynamic keys? https://github.com/Spittal/vue-i18n-extract/issues/46

Potentially we could also get rid of dynamic keys altogether. It makes the code slightly more verbose but I'm sure we can abstract it in other ways to reduce the burden of it.

> But I guess this step is set as required in Terraform, so it shows up as Waiting for status to be reported

You can update it manually in the repository settings and open a PR in the infrastructure repository with the removed check. Alternatively, make the Terraform change with the PR and block this PR on that one. Given that I think we should move more carefully with removing the string extraction check the latter is probably better as it does not effect existing PRs.

obulat:
> That is really unfortunate that the check doesn't work with dynamic labels. It's a very nice check to have and not one I'm super comfortable removing without some good forethought.

It would really be a nice check if it worked correctly for us. Currently, it checks for **missing** and **unused** keys, prints out a report and always exits with a 0 status. So, unless you check the CI logs, you will not see any errors, and CI **will not** fail even if the check fails.

In our locale files, we only have the keys that are translated, so for many languages, there are lots of missing keys. Which is OK because Vue-i18n falls back on the English strings. However, the `i18n:check` was reporting the missing keys for all locales. I've changed that to only check the `en.json` locale. This helped me find and fix a missing key in the Search guide (the OR description was missing 😱). 

I tried adding the `--ci` key to make the CI fail if the `check` script fails. However, it also fails because of all the keys it thinks are unused. There is no option to only check for `missing` keys without checking the `unused` keys, so we cannot currently use the `--ci` flag :( 

> Could we follow any of the suggestions in this issue for dealing with dynamic keys? [Spittal/vue-i18n-extract#46](https://github.com/Spittal/vue-i18n-extract/issues/46)

Unfortunately, the changes to dynamic keys handling described in this comment were reverted by a later PR: https://github.com/Spittal/vue-i18n-extract/pull/124

> Potentially we could also get rid of dynamic keys altogether. It makes the code slightly more verbose but I'm sure we can abstract it in other ways to reduce the burden of it.

We could try that, too. That will require a lot of changes.

> > But I guess this step is set as required in Terraform, so it shows up as Waiting for status to be reported
> 
> You can update it manually in the repository settings and open a PR in the infrastructure repository with the removed check. Alternatively, make the Terraform change with the PR and block this PR on that one. Given that I think we should move more carefully with removing the string extraction check the latter is probably better as it does not effect existing PRs.

Considering all the problems with the check step, do you think it is still useful to have a CI step that only logs the failures _without failing_ as required for CI?

src/plugins/vue-i18n.ts
```diff
@@ -4,7 +4,7 @@ export default () => {
     silentFallbackWarn: true,
     pluralizationRules: {
       /**
-       * @param choice - a choice index given by the input to $tc: `$tc('path.to.rule', choiceIndex)`
+       * @param choice - a choice index given by the input to $tc: $tc('header.filter-button.with-count', choiceIndex)
```
obulat, line 7: `vue-i18n-extract` was reporting `path.to.rule` key as missing even though it's in the comments. So, I replaced it with a key that actually exists.

sarayourfriend:
> It would really be a nice check if it worked correctly for us. Currently, it checks for missing and unused keys, prints out a report and always exits with a 0 status. So, unless you check the CI logs, you will not see any errors, and CI will not fail even if the check fails.

Given this, I guess it is fine to disable it for now. But let's open an issue to track the need for this. Even if it's a lot of work to stop using dynamic keys, this check catches easy to make and sometimes subtle mistakes (like the `.null` key we accidentally had on the audio track component recently, #1287).

sarayourfriend (approved):
LGTM!

test/playwright/utils/navigation.ts
```diff
@@ -11,7 +11,57 @@ import {
...
+ * @param path - The label to translate.
+ * @param dir - The language direction.
+ */
+export const t = (path: string, dir: LanguageDirection = 'ltr'): string => {
+  let value = ''
+  if (dir === 'rtl') {
+    value = getNestedProperty(messages.rtl, path)
+  }
+  return value === '' ? getNestedProperty(messages.ltr, path) : value
+}
```
sarayourfriend, line 52: I love this!!!

I actually started to write something like this when I was working on the flaky audio test fix but realized I didn't need it then, I'm so glad we're getting it for real!

Question: should the case where the string doesn't match raise an error, by any chance? :thinking: I guess it could interrupt the `ar` one if we're not consistently updating it with the new strings?

obulat: > Question: should the case where the string doesn't match raise an error, by any chance?

Sorry, I don't fully understand this question. What should the string match? I don't think we can compare to previous versions here, because we don't have the previous versions of the locales...
I mimicked the behavior of Vue-i18n here so that if the locale doesn't have the key, it falls back to `en.json` value.

sarayourfriend: Oh I see! I misread the code. Looks great then, never mind :slightly_smiling_face:

test/locales/README.md
```diff
@@ -0,0 +1,8 @@
+# Locale files for Playwright tests
+
+Normally when you build Openverse, you download the updated information about locales and translation files to `src/locales` folder. To prevent too many requests during CI, we save the locale files that are necessary to run the app during Playwright tests.
+
+1. `scripts/valid-locales.json` contains a list of valid locales with properties including their translation status (the percentage of strings that are translated).
+2. `es.json` is the `es` locale file used for search navigation tests (`../playwright/e2e/search-navigation.spec.ts`).
+3. `ru.json` is the `ru` locale file used for testing the translation banner because the current translated percentage is ~40%, which is below the banner threshold of 90%.
+4. `ar.json` is the machine-translated `ar` locale file used for RTL testing.
```
sarayourfriend, line 8: For completion's sake, should we also pull in a partially machine translated Urdu that is below the banner threshold so we have an RTL one as well?

obulat: Actually, we can simulate untranslated Urdu by adding it to `valid-locales` with `translated` set to 0, and adding an empty `ur.json` file (or setting `translated` to a low percentage and adding several strings to `ur.json`). Then, there should be a translation banner, but the strings will be mostly in English.
Should we open a separate issue for that?

test/playwright/utils/navigation.ts
```diff
@@ -11,7 +11,57 @@ import {
...
+        return (acc as Record<string, string | Record<string, unknown>>)[part]
+      }
+      return ''
+    }, obj)
+  return typeof value === 'string' ? value : JSON.stringify(value)
+}
+
+/**
+ * Simplified i18n t function that returns English messages for `ltr` and Arabic for `rtl`.
+ * It can also handle nested labels ('header.title').
```
zackkrida, line 42:
```suggestion
 * It can also handle nested labels using dot notation ('header.title').
```

test/playwright/utils/navigation.ts
```diff
@@ -11,7 +11,57 @@ import {
...
+ * @param path - The label to translate.
+ * @param dir - The language direction.
+ */
+export const t = (path: string, dir: LanguageDirection = 'ltr'): string => {
+  let value = ''
+  if (dir === 'rtl') {
+    value = getNestedProperty(messages.rtl, path)
+  }
+  return value === '' ? getNestedProperty(messages.ltr, path) : value
```
zackkrida, line 51:
```suggestion
return getNestedProperty(messages[dir], path) ?? ''
```

what about something like this instead?

zackkrida (approved):
LGTM!

test/playwright/utils/navigation.ts
obulat: We need to fall back to `ltr` (i.e., English) locale if we can't find the value in other locales. I could change `getNestedProperty` to return `null` if it doesn't find the key, and then replace the lines above with something like this:

```
let value = dir === 'rtl' ? getNestedProperty(messages.rtl, path) : null;
return value ?? getNestedProperty(messages.ltr, path) || '';
```
Would it be clearer?

zackkrida: Oh I see, I was missing that context that we need to fallback to `ltr` if `rtl` isn't present. Of course! I think it's fine as-is, then.

Issue #1561, opened 2022-07-11
Snackbar component for displaying messages


panchovm:
## Description
We need to include a snack bar component to display messages for different contexts. The component is based on the [Gutenberg one](https://wordpress.github.io/gutenberg/?path=/story/components-snackbar--default).

<img src="https://user-images.githubusercontent.com/895819/178352421-22ab8630-cf4d-4f0e-a765-c1e629579628.png" width=300>

## References
- Audio track navigation issue #1297  
- Audio track navigation [presentation](https://www.figma.com/proto/GeG5nSTdPI887CH39jzjzo/Audio-track-navigation?page-id=26%3A13657&node-id=37%3A28663&scaling=min-zoom&starting-point-node-id=37%3A28663)
- Figma component in the [Design Library](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=2557%3A4436)

## Implementation
- [ ] 🙋 I would be interested in implementing this component.

Issue #1549, opened 2022-06-29
Audio track navigation


fcoveram:
## Problem
The current audio component needs to allow the composite navigation to meet a11y requirements.

### Issues related
- #1297

### Desktop

https://user-images.githubusercontent.com/895819/176517211-ffbad05a-5a8f-489e-bb87-c8126188f434.mp4

|Snackbar. safe area|Overlapping snackbar|
|:---|:---|
|![image](https://user-images.githubusercontent.com/895819/176517276-7e3bf21d-cabb-40f6-8fac-7b2c8736e5c8.png)|![image](https://user-images.githubusercontent.com/895819/176517424-670f2ba7-0766-4574-bb1c-0d612f3a7c43.png)|

### Mobile

<img src="https://user-images.githubusercontent.com/895819/176517546-a86e4a41-e8e2-47fd-8da8-df32ab16bcee.png" width="300">

## Description
Show the focus style of the composite area and display a snackbar element indicating how to interact with the audio track. The snackbar can disappear after a few seconds or have a closing action that reminds the action and no longer appear during the visitor’s session.

Regarding the interaction, I wonder if the screen can scroll down once the navigation reaches the last visible element. Otherwise, the snackbar element will overlap the audio track; like in the example shared above.

I would like your thoughts on the idea.

## Design file
Figma: [Audio track navigation](https://www.figma.com/file/GeG5nSTdPI887CH39jzjzo/Audio-track-navigation?node-id=0%3A1)

sarayourfriend:
> As for closing the notification, [...] it could close after the user's first interaction, like the next time they press a key or click on something.

I think this is the best way. For signed keyboard users who do not use a screen reader, we cannot assume that the length of time it takes for it to be read by the computer is sufficient for someone reading.

panchovm:
I like @zackkrida's idea. Yet I would like to see how others have solved the problem. Meanwhile, the `sm` breakpoint flow should look like the prototype attached below.

Since visitors can control the wave area with the arrow keys on bigger breakpoints, we should do the same on `sm` and the wave area in the global player react to the time change. And to close the player, I like the idea of doing it by pressing `escape` or a similar action that does not depend on focusing on an element.

https://user-images.githubusercontent.com/895819/176781574-83f96355-d0d6-4c0c-a31b-37d29973554b.mp4

> I wonder if the screen can scroll down once the navigation reaches the last visible element. Otherwise, the snackbar element will overlap the audio track; like in the example shared above.

[...]

PR #1593, opened 2022-07-29
Enable audio interactions on the audio track component

### LINKED ISSUE #1297: Transform audio track component on search results into composite element for quicker navigation
## Problem

Each audio result has three tabbable elements on the single audio view. Takes a very long time to tab through the audio results.

## Description

The audio track component used for search results should be turned into a composite element with special navigation for entering and exiting the composite element.

See [WAI-ARIA grid](https://www.w3.org/TR/wai-aria-1.1/#grid) specification for more details about how to implement this accessibly.

## Additional context

This can be ignored for the all-content (boxed) audio track as the problem is already solved there.

## Design solution
- Design introduction: [Presentation](https://www.figma.com/proto/GeG5nSTdPI887CH39jzjzo/Audio-track-navigation?page-id=26%3A13657&node-id=37%3A28663&scaling=min-zoom&starting-point-node-id=37%3A28663)
- Figma file: [Audio track navigation](https://www.figma.com/file/GeG5nSTdPI887CH39jzjzo/Audio-track-navigation?node-id=0%3A1)
- Design issue #1549

## Implementation

dhruvkb:
## Fixes

Fixes #1297 by @sarayourfriend 

## Description

This PR builds upon work done by @sarayourfriend. It enables audio track controls using the keyboard when the audio track is focused and makes the subcomponents non-focusable to provide accessible navigation in the search results.

## Testing Instructions

1. Search for any given term and switch to the audio-only results.
2. Navigate using the <kbd>Tab</kbd> key. The focus should move from one result track to the next and not go inside the component.
3. Manipulate playback by seeking (with arrows <kbd>&larr;</kbd> and <kbd>&rarr;</kbd>) and toggling (with the <kbd>Space</kbd>) key.
4. Try to access the same functionality with the mouse by clicking the buttons and dragging the seekbar etc. Existing functionality should continue to work. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/utils/time-fmt.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Format the time as hh:mm:ss, dropping the hour part if it is zero.
+ * @param seconds - the number of seconds in the duration
+ * @returns the duration in a human-friendly format
+ */
+export const timeFmt = (seconds: number): string => {
+  const date = new Date(0)
+  date.setSeconds(Number.isFinite(seconds) ? seconds : 0)
+  return date.toISOString().substr(11, 8).replace(/^00:/, '')
```
sarayourfriend, line 9: Just noting that this will fail for audio files longer than 24 hours (as it'll just wrap back around to 0 at that point).

It'd be nice if seconds and minutes weren't base _whatever they are_.

Here's a solution on SO that is to the point and doesn't have any problems with very, very long audio:

https://stackoverflow.com/a/11486026

dhruvkb: Fixed in 1ae858d with a slightly tweaked implementation of the SO answer.

src/locales/en.json
```diff
@@ -577,6 +577,7 @@
   "audio-track": {
     "aria-label": "{title} - Audio Player",
     "aria-label-interactive": "{title} - audio player - press the spacebar to play and pause a preview of the audio",
+    "aria-label-interactive-seekable": "{title} - audio player - press the spacebar to play and pause a preview of the audio; use the left and right arrow keys to seek through the waveform",
```
dhruvkb, line 580: I think the word 'track' is more appropriate here as compared to 'waveform'.

```suggestion
    "aria-label-interactive-seekable": "{title} - audio player - press the spacebar to play and pause a preview of the audio; use the left and right arrow keys to seek through the track",
```

obulat: Do you think we should also add a note that this is also a link, and if you press Enter, you'll see the details of the audio, for user navigating with VoiceOver? Currently, it might seem like a list of playable audios without links.

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -1,12 +1,12 @@
...
+  <VLink
+    v-bind="containerAttributes"
     class="audio-track group"
     :aria-label="ariaLabel"
     role="region"
-    v-bind="layoutBasedProps"
+    :href="`/audio/${audio.id}`"
     @keydown.native.shift.tab.exact="$emit('shift-tab', $event)"
-    @keydown.native.space="handleSpace"
+    @keydown.native.exact="handleKeydown"
```
obulat, line 9: What is the purpose of `exact` modifier here? I thought it only works if you add the key modifiers (just like the `.shift.tab.exact` above (there, it's only called when both keys were pressed together).

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -385,7 +386,7 @@ export default defineComponent({
      * Sets default size if not provided.
      */
     const _size = computed(() => {
-      if (isBoxed && !props.size) {
+      if (isBoxed.value && !props.size) {
```
obulat, line 389: Aargh! I wonder if there's a linter to check for `.value` not being added.

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -406,38 +407,44 @@ export default defineComponent({
...
-      if (!isBoxed.value) return {}
-      return {
-        href: `/audio/${props.audio.id}`,
-        class:
-          'block focus:bg-white focus:border-tx focus:ring-[3px] focus:ring-pink focus:ring-offset-[3px] focus:outline-none rounded-sm overflow-hidden cursor-pointer',
-      }
-    })
+    const layoutBasedProps = computed(() => ({
+      href: `/audio/${props.audio.id}`,
+      class:
```
obulat, line 412: Should the `class` be a constant?

zackkrida:
This looks amazing! I'll do a full review later today but I noticed some quick things:

- The timestamps have turned pink
- focusing a track shows the little round indicators at the top and bottom of the 'playhead'. I think it ends up looking a bit like a visual error and should only be visible once a user presses one of the arrow keys, and 'hidden' again on subsequent presses of tab, spacebar, etc.

Quick video to show both issues as I see them:

https://user-images.githubusercontent.com/6351754/182158486-27eaee91-d149-452c-ad76-2ee6d04b3e7a.mp4

src/components/VAudioTrack/VAudioTrack.vue
dhruvkb: Yeah, this `.exact` was required by the linter. Apparently the presence of `.exact` for the `@keydown` event above means all other `@keydown` listeners must have `.exact` specified.

src/components/VAudioTrack/VAudioTrack.vue
dhruvkb: I'm looking forward to the [reactivity transform](https://vuejs.org/guide/extras/reactivity-transform.html) proposal that enables the compiler to automatically handle this for us.

src/locales/en.json
dhruvkb: Ideally the nature of the element being an anchor tag would be pointed out by the screen-reader.

src/components/VAudioTrack/VAudioTrack.vue
sarayourfriend: The TypeScript ESLint plugin has type-aware linting that is able to catch things like this as well.

dhruvkb:
Drafting to address feedback.

src/components/VAudioTrack/VAudioTrack.vue
dhruvkb: True, moved out of `script` to the `class` attribute in the `template` in 03a2ae6.

src/locales/en.json
dhruvkb: Changed in 8508e4f.

zackkrida:
Lots of progress! Some more notes:

CleanShot 2022-08-04 at 11 01 10@2x

- the single result title is now pink
- the single result title and author are highlighted with an underline when any part of the waveform is hovered
- in general, the grouped controls look a bit odd on the single result page, instead of being able to highlight the individual waveform, play/pause, title, and author, but i think it would be *okay* once the above issues are cleaned up. Ideally the changes your making would only apply to the 'track' version of the audio component but I can imagine that's pretty complex and might be okay to skip as a result.

zackkrida:
Related audio looks great already! I do notice that I'm still seeing the 'seek' bar when tabbing through the audio tracks, however.

CleanShot 2022-08-04 at 11 05 33@2x

zackkrida:
This looks really good now! One thing I notice when using VoiceOver in Safari is that I'm still able to keyboard navigate through the various sub-components of the audio track. So it works the same as the current behavior with some differences:

1. I can press the arrow keys to navigate the the entire grouped audio track. In this state spacerbar plays/pauses the audio while option+arrow key tracks through the waveform. This seems really good!
2. Subsequent right/left arrow presses navigate _into_ the grouped audio tracks. That doesn't seem ideal, but does work.
3. Since I can navigate into the components, when I arrow-key over to the waveform it now has the incorrect aria label. Instead of the formatted timestamps it reads the 'raw' number with a long decimal sequence. 

I don't know how to feel about those aria issues. If the label was fixed on the waveform it doesn't _really_ feel like a regression from the current behavior, but I'm also not sure if tabbing into the component should be blocked for keyboard+screen reader users in the first place.

panchovm:
I noticed the following:
1. I can not move through the time with `⇧` + `← / →` to jump more seconds.
2. The padding is too tight compared to [the design](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=2235%3A5579).
3. In hover, the text info shows a pink underline. It should only be dark for the audio title, like the current implementation.

The design ticket (#1549) has all the assets.

dhruvkb:
Addressing feedback

https://github.com/WordPress/openverse-frontend/pull/1593#issuecomment-1205381921
- the single result page text is no longer pink
- the text is not underlined on hover
- it's possible to individually focus the waveform and the play/pause toggle

https://github.com/WordPress/openverse-frontend/pull/1593#issuecomment-1205385256
- the seek bar does not appear when the user is not seeking

https://github.com/WordPress/openverse-frontend/pull/1593#issuecomment-1209748517
- Shift + Arrow and Meta + Arrow behaviour has been re-enabled
- the row layout has been updated to match the designs (including the hover background) 
- underline color has been reverted to black and it only under the track title

dhruvkb:
@panchovm I've left a few comments on Figma regarding what I perceive to be a discrepancy in focus-outline widths. Other than that, I appreciate you calling out the the differences between the UI in the PR and the mockups. If you could take another look, that'll be very helpful. Thanks!

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -401,52 +407,72 @@ export default defineComponent({
...
+                'focus:ring-offset-0 focus:ring-[1.5px]':
+                  props.layout === 'row',
+              },
+            ],
+          }
+        : {}
+    )
     const ariaLabel = computed(() =>
-      isBoxed.value
+      isComposite
```
krysal, line 430: If we're not allowing go inside the composite audio component then we should add the author, license and category to the label, otherwise, it's not announced and screen reader users will not receive the same information.

krysal (commented):
This is a great improvement! I didn't know I needed this so much until now 😄 Love it!
I'm just leaving a small comment, not sure if we will allow to go inside the composite element.

src/composables/use-seekable.ts
```diff
@@ -0,0 +1,100 @@
...
+  duration,
+  currentTime,
+  isReady,
+  onSeek,
+  onTogglePlayback,
+}: UseSeekableOptions) => {
+  const i18n = useI18n()
+
+  const attributes = computed(() => ({
+    'aria-valuemax': duration.value,
```
sarayourfriend, line 26: I think we may need to add [`'role': 'slider'`](https://www.w3.org/WAI/ARIA/apg/example-index/slider/slider-seek.html) :thinking:

src/components/VAudioTrack/VWaveform.vue
```diff
@@ -1,23 +1,19 @@
 <template>
   <div
+    v-bind="seekAttributes"
```
obulat, line 3: The seekable attributes should only be binded if the waveform is interactive and has a role of slider. Currently, when it's not a slider (on the audio search page), Lighthouse shows aria attributes mismatch.

src/composables/use-seekable.ts
obulat: `'role': 'slider'` is currently only added when the Waveform is interactive. When it's not (on the search page), then we either need to remove these aria-attributes, or to move the `'role': 'slider'` here so that it's always applied. Can we have slider role inside of an aria-disabled element, though? I'm confused

obulat (changes requested):
The duration computation needs to be changed, I haven't noticed it for a very long time. This audio should be something like 3 minutes:
Screen Shot 2022-08-12 at 7 32 33 AM

src/composables/use-seekable.ts
sarayourfriend: > Can we have slider role inside of an aria-disabled element, though

I think so, but it should not be hard to test. Can you check if VO ignores the role of elements inside an aria-disabled container, for example?

panchovm:
It looks great ✨ 

Since we are adding `16px` on all sides, the gap between audio tracks is `40px` (`2.50 rem` as I see in the inspector). Can we change it to `16px`? In that way, the results layout will show more results on the screen.

krysal:
> The duration computation needs to be changed, I haven't noticed it for a very long time. This audio should be something like 3 minutes:

@obulat Is that a bug introduced here or do we need to create an issue for that?

obulat:
> > The duration computation needs to be changed, I haven't noticed it for a very long time. This audio should be something like 3 minutes:
> 
> @obulat Is that a bug introduced here or do we need to create an issue for that?

This bug is introduced here. What I meant by "not noticing it for a very long time" is the numbers were off, but I didn't see them even though I was reviewing the PR and looking at the AudioTrack :) I just wasn't concentrated on the numbers.

zackkrida:
This is really looking awesome, especially once the vertical spacing is tightened up:

CleanShot 2022-08-12 at 11 33 53@2x

Such a huge improvement to the audio browsing experience!

dhruvkb:
^ These commits tighen the spacing between results in the search results and the related audios. They also fix the incorrect timestamps introduced the `timeFmt` function was extracted into a util and used uniformly everywhere.

sarayourfriend:
Can folks confirm they've tested this change VO and either JAWS or NVDA?

dhruvkb:
I can reproduce what @zackkrida said about being able to still navigate into the track using <kbd>&larr;</kbd> and <kbd>&rarr;</kbd> arrow keys in VoiceOver. I don't know how we'd prevent that but to me it feels like a good thing because it allows you to listen to more information about the track like the license and duration that would be clunky if always announced as @krysal noted. Pressing <kbd>Tab</kbd> moves to the next track as expected, even with VoiceOver on.

zackkrida:
@dhruvkb and @sarayourfriend I'll make videos today walking through the changes in this PR and the old version of the component for everyone's evaluation. Perhaps there's some a11y folks who we could share the videos with as well.

panchovm:
I tested it using VO and it works correctly. I can play, pause, jump over time, and move among audio tracks with `tab`. My only question is that VO every time reads the audio interaction instructions and wondering if that could annoy recurrent visitors.

src/components/VAudioTrack/layouts/VRowLayout.vue
```diff
@@ -75,11 +76,13 @@
           name="play-pause"
           :size="isLarge ? 'medium' : 'large'"
           :layout="'row'"
+          :is-tabbable="false"
```
obulat, line 79: In RowLayout on line 54 we also need to add `true` to `timeFmt`.

obulat (dismissed):

sarayourfriend:
> I can reproduce what @zackkrida said about being able to still navigate into the track using arrow keys in VoiceOver. I don't know how we'd prevent that but to me it feels like a good thing because it allows you to listen to more information about the track like the license and duration that would be clunky if always announced as @krysal noted. Pressing Tab moves to the next track as expected, even with VoiceOver on.

This is the correct behaviour for a composite component anyway :slightly_smiling_face:

obulat:
I tried using the Lighthouse A11y check, and it still shows `[aria-*] attributes do not match their roles` error for each `a.audio-track` element. I'm not sure how significant it is. Here's the explanation they give:
_Each ARIA `role` supports a specific subset of `aria-*` attributes. Mismatching these invalidates the `aria-*` attributes. [Learn more](https://web.dev/aria-allowed-attr/?utm_source=lighthouse&utm_medium=devtools)._

sarayourfriend:
> I tested it using VO and it works correctly. I can play, pause, jump over time, and move among audio tracks with tab. My only question is that VO every time reads the audio interaction instructions and wondering if that could annoy recurrent visitors.

This is definitely not ideal. I guess it's because the aria-label on the group always includes the instructions. We might want to push the instructions into an alert rather than putting it into the aria-label (something we can probably do for other things as well).

Gutenberg's `a11y` module may be instructive here (though would likely require some re-implementation into a Vue compatible approach: https://github.com/WordPress/gutenberg/blob/2cc282ad3968699d61c83a34704b064c739c860e/packages/a11y/src/index.js#L60

zackkrida:
Will the "audio interaction instructions" problem be fixed once #1561 lands and we integrate it here? I think I'll bump the priority of that up and we may consider marking this 'blocked' until that's finished.

obulat:
I tried using the VO with a sample slider: https://www.w3.org/WAI/ARIA/apg/example-index/slider/slider-temperature.html , and I've never heard the same things with the Audio Track. I don't hear the current time or the full duration, all I hear is "You are currently in a group of 5 items" or something similar. I wonder if I'm not using the VoiceOver correctly, or if the role of "link" overrides the slider aria-attributes.

sarayourfriend:
Explicitly setting an `aria-role` will override the implicit roles and attributes on an element. Links _could_ be different, I'm not sure.

krysal:
@zackkrida:
> Will the "audio interaction instructions" problem be fixed once https://github.com/WordPress/openverse-frontend/issues/1561 lands and we integrate it here?

That PR only builds the look of the component, I wasn't sure if I needed to add some interaction there but I undrafted it so people can review it.

---

@dhruvkb:
> I can reproduce what @zackkrida said about being able to still navigate into the track using arrow keys in VoiceOver.

It would be nice to have a step with the explicit key combination to enter the composite element in the **Testing Instructions**, as I'm not able to reproduce that behavior 🙂

dhruvkb:
Drafting to prevent review notifications while I incorporate the snackbar.

src/components/VAudioTrack/layouts/VRowLayout.vue
dhruvkb: Fixed in commit e3dd1eb.

krysal (commented):
So I see it behaves differently in different browsers.

In Safari:
- With Voice Over (VO), here is where I can enter the composite element with simple arrow keys
- However, I cannot navigate the results with <kbd>Tab</kbd>, it takes me to the "Load more results" button, and is difficult to reach the audio results again. I needed to get the "Skip to content" button first, which is skipped in a cycle of <kbd>Tab</kbd> navigation through the whole page. Once you reach the search input, you press <kbd>shift</kbd> + <kbd>Tab</kbd> and only then the button appears.

In Chrome & Firefox:
- I cannot enter the composite element and therefore get nothing more than the title of the track from VO
- I can navigate the different results with <kbd>Tab</kbd>

The waveform seeking continues to work as instructed in the snackbar in the three browsers.

dhruvkb:
@krysal I did not experience the <kbd>Tab</kbd> key jumping to the "Load more results" button. It seems that what you're seeing could be the result of a configuration difference between our systems. Can you see if the instructions in [this post from the a11y project](https://www.a11yproject.com/posts/macos-browser-keyboard-navigation/) fix the problem?

krysal:
Right! I can navigate each element with <kbd>Tab</kbd> in Safari after changing the config (or use the default <kbd>Option</kbd> + <kbd>Tab</kbd> combination, which is annoying that is different). So the issue there is fixed, great.

But does it work the same for you in Chrome and Firefox?

dhruvkb:
Yes @krysal, I was able to navigate with the keyboard in Firefox, Safari and Vivaldi (Chromium) but for each browser I had to tweak some settings first. Now they all behave identically and as expected.

sarayourfriend:
Just a note too that VoiceOver is known to not work well with any browsers other than Safari.

If someone can confirm that they will test with JAWS or NVDA, that would be great. VoiceOver is essential but its market share is tiny compared to JAWS especially.

zackkrida (changes requested):
Part of me _really_ wants to approve this as-is, because visually, behaviorally, and from an a11y perspective it seems much better than the current implementation. It seems much more accessible than our previous component, gives me much clearer instructions than the old version did, and allows for seeking/playing/pausing/navigating to the single items very easily from one element. 

I tried testing SoundCloud with voiceover and I can't seek, I the labels are unclear, and overall it's a very unpleasant experience. We're doing a much better job. I do have some small notes:

- [ ] Playing audio with the mouse should _also_ dismiss the snackbar. The criteria for removing it should be 'an audio track has been seeked,played,or paused' with no regard for what input type.
- [ ] Can the snackbar visibility go in the global audio store? Right now if you go to a single result and then back the audio page, it alerts the snackbar again.

I'd like to work on finding us some real a11y experts and screen reader users, because ultimately I feel like any testing we do here is symbolic with our limited expertise and a 'best guess'.

zackkrida:
Because this is better than our current implementation, I also feel comfortable getting it merged after my requested changes are made so we can share the page easily with some screen-reader users.

panchovm:
The snackbar looks great. Although I noticed the following:

1. The snackbar should appear once focusing on the audio component rather than landing in the audio-only results. This is to work and be consistent with audio components on "all content" results.
2. On smaller breakpoints, the snackbar does not show up.
3. On smaller breakpoints, the audio component [has a bigger padding](https://share.cleanshot.com/A9Dq5U) of the focus area than the design.
4. `Spacebar` and arrows (`←`, `→`) in the snackbar have a text style not defined in the component. We should address this new text style in a separate issue as it needs to consider other similar interface elements and style changes for other breakpoints.

The [design presentation](https://www.figma.com/proto/GeG5nSTdPI887CH39jzjzo/Audio-track-navigation?page-id=26%3A13657&node-id=37%3A28663&scaling=min-zoom&starting-point-node-id=37%3A28663) describes the points mentioned above.

On the other hand, I wonder what we can do with the following

## Hidden element
While navigating with the keyboard, the focused element lies hidden under the global player once reaching the end of the viewport. In the image below you can see how the audio track is covered by the global player and part of the focus ring is visible.

CleanShot 2022-08-29 at 10 20 27@2x

## Navigating and interacting
On smaller breakpoints, listening to an audio track and jumping among the results does not allow you to control it by pausing and seeking a new time. You need to navigate back to the audio playing to control it. This consistency with desktop makes sense, but I wonder how we can improve this interaction. Here below I recorded a video of what I am pointing. We talked about pressing `escape` to close the global player, but this use case is new.

https://user-images.githubusercontent.com/895819/187230906-3cbc9735-6cdf-450b-926b-e67d8fd1d392.mp4

zackkrida (approved):
My core concerns were resolved, so going to approve this. I see that @panchovm still has feedback and am aware that @obulat plans to do some testing in Windows screen-readers, but that we also want to get this deployed to staging, _at least_, so we can try to get some actual screen readers to test the url for us. I can make a post on our make blog once that's available.

Excellent, excellent, complicated work, @dhruvkb!

obulat (changes requested):
I tested the new interactions in NVDA on Windows.

NVDA is different from VoiceOver in that it has several navigation modes. The default mode is "Browse mode". In it, the keypresses are handled by NVDA, and are not sent to the application. This means that when you press "Spacebar" on the audio track, you get the default event handling - you open the link. The AudioTrack component does not receive the keypress event at all. This is the screen recording of the default behavior now:

https://user-images.githubusercontent.com/15233243/187631823-9258b660-3c32-4f9b-8c53-23a4f6784256.mp4


To fix this, we can add `role="application"`[^1] to the AudioTrack component. This prevents NVDA from hijacking the keypress events and makes the component behave just as we want: Play on pressing spacebar, Seek on pressing arrow keys and Navigate to the link on pressing Enter.

So, my requested change is to add `role="application"` to the VAudioTrack component. After that, I think the best way forward is to create new issues with other feedback in the review here, and merge this PR.

panchovm (changes requested):
The interaction feels great and smooth. Excellent 🌟 

Although I am still seeing the following:
* The snackbar has a left and right padding of `32px` instead of `24px`. 
* I am still seeing the "JetBrains Mono" font for the keyboard action words.
* On smaller breakpoints
   * Snackbar message is not visible.
   * The audio component's focus ring is bigger than the mockups due to padding values.

dhruvkb:
The last Playwright failure is from a regression on `main`, so I will not be attempting to fix that in this PR.

obulat:
> The last Playwright failure (2 failing tests) is from a regression on `main`, so I will not be attempting to fix that in this PR.

I've disabled them by adding `.skip` in #1646, and opened an issue to fix and re-enable them.
The lock file also needs updating.

src/components/VAllResultsGrid/VAllResultsGrid.vue
```diff
@@ -49,15 +58,19 @@ import { Focus } from '~/utils/focus-management'
...
 import { useI18n } from '~/composables/use-i18n'
 
+import VSnackbar from '~/components/VSnackbar.vue'
 import VImageCellSquare from '~/components/VAllResultsGrid/VImageCellSquare.vue'
 import VAudioCell from '~/components/VAllResultsGrid/VAudioCell.vue'
 import VLoadMore from '~/components/VLoadMore.vue'
 import VContentLink from '~/components/VContentLink/VContentLink.vue'
 import VGridSkeleton from '~/components/VSkeleton/VGridSkeleton.vue'
 
+import { useUiStore } from '../../stores/ui'
```
obulat, line 68:
```suggestion
import { useUiStore } from '~/stores/ui'
```

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -360,6 +366,7 @@ export default defineComponent({
           pause()
           break
       }
+      emit('interacted')
```
obulat, line 369: Could you add this to `emits` property on this component?

obulat (approved):

PR #1629, opened 2022-08-08
Check paginated versions during Rollback version check

### LINKED ISSUE #1623: Rollback script only looks at most recent 10 versions of the frontend image
## Description

If you try to rollback to a valid version that is not on the first page of results the rollback workflow will fail: https://github.com/WordPress/openverse-frontend/actions/runs/2803805784

## Reproduction

1.  Dispatch the rollback workflow to a valid version that is not on the first page of results from the package version endpoint
2.  See the failure from the link in the description.

## Resolution

sarayourfriend:
## Fixes

Fixes #1623 by @sarayourfriend 

## Description

Paginates through the image versions instead of relying just on the first page of 30 results.

## Testing Instructions

Merge and try to re-deploy production to v3.4.2 (which is not previously on the first page).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/rollback.yml
+++ b/.github/workflows/rollback.yml
@@ -41,16 +41,26 @@ jobs:
         uses: actions/github-script@v6
         with:
           script: |
-            const { data: versions } = await github.rest.packages.getAllPackageVersionsForPackageOwnedByOrg({
-              package_type: 'container',
-              package_name: 'openverse-frontend',
-              org: 'WordPress',
-            })
+            let exists = undefined,
+              versions = [],
+              page = 0;
 
-            const exists = versions.some(v => v.metadata.container.tags.includes("${{ inputs.tag }}"))
+            do {
+              page += 1;
+              ({ data: versions } = await github.rest.packages.getAllPackageVersionsForPackageOwnedByOrg({
+                package_type: 'container',
+                package_name: 'openverse-frontend',
+                org: 'WordPress',
+                page,
+                // max of `per_page`
+                per_page: 100,
+              }));
+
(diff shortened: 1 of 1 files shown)


zackkrida:
I believe the loop runs infinitely in the case where the tag doesn't exist and it runs out of pages. 

BTW I was able to test this locally with Octokit. Just need a personal access token with `read:package` permissions.

```js
const { Octokit } = require("@octokit/rest");

const github = new Octokit({
  auth: "your_key_here",
});

async function go() {
  let exists = undefined,
    versions = [],
    page = 0;

  do {
    page += 1;
    ({ data: versions } = await github.rest.packages.getAllPackageVersionsForPackageOwnedByOrg({
      package_type: "container",
      package_name: "openverse-frontend",
      org: "WordPress",
      page,
      // max of `per_page`
      per_page: 100,
    }));

    exists = versions.some((v) => v.metadata.container.tags.includes("v3.4.2"));
  } while (!exists || versions.length === 0);

  if (!exists) {
    throw new Error(
      "`${{ inputs.tag }}` does not appear to be a valid tag for the ghcr.io/wordpress/openverse-frontend image."
    );
  }
}

go().then(() => console.log("done!"));

```

sarayourfriend:
Ah, thanks for the tip! I will try it locally. I think the while condition needs to be an `&&` :slightly_smiling_face:

sarayourfriend:
It ended up being clearer to just use an explicit break instead of overloading the condition.

AetherUnbound (approved):
Looks good!

Issue #1634, opened 2022-08-09
Use `v-deep` correctly


obulat:
## Problem

`v-deep` was previously used as a combinator selector (i.e., it was written before another CSS selector. Now, it was converted to a [pseudo-class](https://vuejs.org/api/sfc-css-features.html#scoped-css).

## Description

In `VErrorImage`, replace the `::v-deep some-selector` declarations with `::v-deep(some-selector)` style declarations.

## Implementation

- [x] 🙋 I would be interested in implementing this feature.

PR #1635, opened 2022-08-09
Update Vue to v2.7 

obulat:
## Description

This PR updates Vue to version 2.7 which backports the composition API to Vue 2. This should (hopefully) make the future conversion to Nuxt 3 easier.

The main updated packages are Vue (to version 2.7) and `@nuxtjs/composition-api` which supports Vue 2.7 (0.33.1).

Other changes:
- `vue-loader` added to devDependencies to force the ^15.10 version as described in Vue 2.7 update post.

## Problems that I encountered:
- I had to pin the Playwright version because the tests can't start with a version higher than 1.20. See #1646 (Wonder if this issue is related: https://github.com/npm/cli/issues/4308)

## Testing Instructions

The CI tests should pass. The app should behave as normal everywhere.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


package.json
```diff
@@ -137,15 +140,14 @@
...
-    "eslint-plugin-vue": "^8.6.0",
+    "eslint-plugin-vue": "^9.0.0",
     "eslint-plugin-vuejs-accessibility": "^1.1.1",
     "husky": "^7.0.1",
     "jest": "^26.6.3",
     "jest-transform-stub": "^2.0.0",
     "lint-staged": "^12.3.7",
     "module-alias": "^2.2.2",
     "npm-run-all": "^4.1.5",
-    "pinia": "^2.0.11",
```
obulat, line 148: We apparently had `pinia` both in dependencies and devDependencies, with somewhat different versions specified.

test/playwright/utils/breakpoints.ts
```diff
@@ -56,42 +56,39 @@ const makeBreakpointDescribe =
...
-        const expectSnapshot = async <T extends ScreenshotAble>(
-          name: string,
-          screenshotAble: T,
-          options?: Parameters<T['screenshot']>[0]
-        ) => {
-          const { name: snapshotName } = getConfigValues(name)
-          return expect(
-            await screenshotAble.screenshot(options)
-          ).toMatchSnapshot({
+    test.describe(`screen at breakpoint ${breakpoint} with width ${screenWidth}`, () => {
```
obulat, line 59: For some reason, this was re-formatted by the linter. The main change in this file is the type of ExpectSnapshot, which was throwing an error.

src/components/VErrorSection/VErrorImage.vue
```diff
@@ -70,13 +70,13 @@ export default defineComponent({
 </script>
 
 <style scoped>
-::v-deep .attribution {
+::v-deep(.attribution) {
```
obulat, line 73: The warning in #1634 starts appearing with the update to Vue 2.7, so I just fixed it here.

sarayourfriend:
I managed to fix the playwright issue in https://github.com/WordPress/openverse-frontend/pull/1646#issuecomment-1226016354. One line fix to be able to update Playwright and avoid having to pin it.

src/composables/use-focus-on-show.js
```diff
@@ -35,26 +36,27 @@ export const useFocusOnShow = ({
...
-
-      const tabbable = getFirstTabbableIn(dialog, true)
-
-      if (tabbable) {
-        ensureFocus(tabbable, { preventScroll: true, isActive })
-      } else {
-        ensureFocus(dialog, { preventScroll: true, isActive })
-        if (dialog.tabIndex === undefined || dialog.tabIndex < 0) {
-          warn(noFocusableElementWarning)
+      Vue.nextTick(() => {
```
sarayourfriend, line 39: `nextTick` is available from the composition API. What is the reason to pull it off of `Vue` directly rather than importing it from `@nuxtjs/composition-api`?

src/stores/feature-flag.ts
```diff
@@ -106,10 +106,12 @@ export const useFeatureFlagStore = defineStore(FEATURE_FLAG, {
...
     initFromCookies(cookies: Record<string, FeatureState>) {
-      Object.entries(this.flags).forEach(([name, flag]) => {
-        if (getFlagStatus(flag) === SWITCHABLE)
+      for (const flagEntry of Object.entries(this.flags)) {
+        const [name, flag] = flagEntry as [string, FeatureFlag]
+        if (getFlagStatus(flag) === SWITCHABLE) {
           flag.preferredState = cookies[name]
-      })
+        }
+      }
```
sarayourfriend, line 114: Is this a change to appease TypeScript?

I don't remember if this is still the case, but at one point using a `for...of` loop actually incurred a non-trivial size penalty due to the generator runtime transpilation of it. Might not be an issue any more though.

typings/csstype/index.d.ts
```diff
@@ -1,3 +1,4 @@
+import type * as CSS from 'csstype' // eslint-disable-line @typescript-eslint/no-unused-vars
```
sarayourfriend, line 1: What is this import for?

sarayourfriend (commented):
LGTM. I have a couple minor questions.

src/composables/use-focus-on-show.js
obulat: The main advantage of importing it from Vue is that when we (hopefully, some time in the future) move to Nuxt3, we won't need to update the imports. Other than that, there is no reason, really. Do you think it makes sense to use @nuxtjs/composition-api? At least for consistency with the current state of the code?

sarayourfriend: I think for consistency. When (if) we move away from it, it will mostly be a find-and-replace job to update that import, right?

src/stores/feature-flag.ts
obulat: Yes, it's TS appeasing change. I didn't know about for ... of performance. I think I can improve this code by typing `this`

typings/csstype/index.d.ts
obulat: There was an error without it, I'll have to look it up at my computer tomorrow. It's also how module augmentation is done in the csstype docs (the third point here): https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors

sarayourfriend: Gotcha. Can you add a link to the docs you shared here?

obulat: `vue-tsc` gives the following error when we don't have the import:

```
src/components/VAudioTrack/layouts/VBoxLayout.vue(2,18): error TS2322: Type '{ width: string | undefined; }' is not assignable to type 'StyleValue | undefined'.
  Object literal may only specify known properties, and 'width' does not exist in type 'CSSProperties | StyleValue[]'.
src/components/VSkeleton/VAudioTrackSkeleton.vue(16,23): error TS2322: Type '{ width: string; }' is not assignable to type 'StyleValue | undefined'.
  Object literal may only specify known properties, and 'width' does not exist in type 'CSSProperties | StyleValue[]'.
src/components/VAudioTrack/VWaveform.vue(94,17): error TS2322: Type '{ width: string; left: string; }' is not assignable to type 'StyleValue | undefined'.
  Object literal may only specify known properties, and 'width' does not exist in type 'CSSProperties | StyleValue[]'.
```
I guess without the import it considers that `CSSProperties` type only has the properties that are in the `typings/csstype/index.d.ts` instead of augmenting it...

obulat:
I think I've found the issue. The dev version was working fine, but when you ran `pnpm build && pnpm start`, you would get an `r.util is undefined` error. Apparently, that was an error in `vue-demi` that was caused by Webpack.
The error was reported yesterday, and amazing @antfu fixed it 3 hours and immediately released the new version!

test/playwright/utils/navigation.ts
```diff
@@ -248,7 +248,7 @@ export const goToSearchTerm = async (
       page.waitForNavigation(),
       page.click(`[aria-label="${t('search.search', dir)}"]`),
     ])
-    await page.waitForLoadState('networkidle')
+    await page.waitForLoadState('domcontentloaded')
```
obulat, line 251: This makes the tests faster.

dhruvkb: Much faster indeed. I'm not sure 100% we should use `domcontentloaded` over `load` because `load` also downloads stylesheets and images.

dhruvkb (approved):
I used this branch locally and found nothing behaving abnormally. I see no reason to block this.

test/playwright/utils/navigation.ts
obulat: You are right, we should use `load` to make sure that all the images are there for the snapshots. Thank you!

obulat, line 251:
```suggestion
    await page.waitForLoadState('load')
```

test/playwright/utils/navigation.ts
```diff
@@ -313,7 +313,7 @@ export const scrollToTop = async (page: Page) => {
  */
 export const scrollDownAndUp = async (page: Page) => {
   await scrollToBottom(page)
-  await page.waitForLoadState('networkidle')
+  await page.waitForLoadState('domcontentloaded')
```
obulat, line 316:
```suggestion
  await page.waitForLoadState('load')
```

sarayourfriend (approved):
LGTM. Exciting!

PR #1640, opened 2022-08-11
Fix/e2e tests

### LINKED ISSUE #1637: Playwright CI does not fail when Playwright fails with status 1
## Description

When the Playwright tests fail in CI, the container exits with code 1, but the CI still passes. 
Here's a link to such a run:
https://github.com/WordPress/openverse-frontend/runs/7751696961?check_suite_focus=true

<details>
<summary>CI logs</summary>
```
 Creating openverse-frontend_playwright_1 ... done
Attaching to openverse-frontend_playwright_1
playwright_1  | 
playwright_1  | > openverse-frontend@2.2.1 test:playwright:local /app
playwright_1  | > DISABLE_SENTRY=true playwright test -c test/playwright "--workers=2"
playwright_1  | 
playwright_1  | 
playwright_1  | Running 269 tests using 2 workers
playwright_1  | 
playwright_1  | 2022-08-09T17:32:05.910Z pw:webserver Starting WebServer process ./node_modules/.bin/npm-run-all -p -r talkback prod...
playwright_1  | 2022-08-09T17:32:05.920Z pw:webserver Process started
playwright_1  | 2022-08-09T17:32:05.920Z pw:webserver Waiting for availability...
playwright_1  | [WebServer] /bin/sh: 1: ./node_modules/.bin/npm-run-all: not found
playwright_1  | 2022-08-09T17:32:05.924Z pw:webserver Waiting 100ms
playwright_1  | 
playwright_1  | Error: Process from config.webServer was not able to start. Exit code: 127
playwright_1  | 
playwright_1  | 
playwright_1  | 
playwright_1  | 
playwright_1  |   269 skipped
playwright_1  |  ELIFECYCLE  Command failed with exit code 1.
openverse-frontend_playwright_1 exited with code 1
1
Aborting on container exit...
Removing openverse-frontend_playwright_1 ... 
Removing openverse-frontend_playwright_1 ... done
Removing network openverse-frontend_default
```
</details>

## Resolution

obulat:
## Fixes

Fixes #1637 by @obulat

## Description

This PR fixes the failing Playwright visual regression tests, and makes the CI fail if Playwright tests fail.

The changes introduced in #1559 to stop the Docker container after Playwright tests caused the CI not to fail even if the Playwright container exits with status `1`.

As can be seen in [this run](https://github.com/WordPress/openverse-frontend/runs/7785041855?check_suite_focus=true)'s log, **Playwright tests started to fail** again after the following line from `bin/playwright.sh` was removed:
`docker-compose -f docker-compose.playwright.yml down`

I've also updated the snapshots that were failing.

## Testing Instructions

Look at the CI logs. The most recent ones are passing, but the run before that should fail because one of the snapshots didn't match.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


bin/playwright.sh
```diff
@@ -9,8 +9,8 @@ export PLAYWRIGHT_ARGS=$@
 export PLAYWRIGHT_VERSION=$(version)
 export TEST_COMMAND=${TEST_COMMAND:-test:playwright:local}
 
-echo Running Playwright v$PLAYWRIGHT_VERSION as $USER_ID with Playwright arguments $PLAYWRIGHT_ARGS
+cp -r test/locales/* src/locales
```
sarayourfriend, line 12: I wonder if it's worth doing this in the Dockerfile.playwright instead of in here.

bin/playwright.sh
```diff
@@ -9,8 +9,8 @@ export PLAYWRIGHT_ARGS=$@
...
 
-docker-compose -f docker-compose.playwright.yml up --build --force-recreate --abort-on-container-exit --exit-code-from playwright --remove-orphans
+echo Running Playwright v$PLAYWRIGHT_VERSION as $USER_ID with Playwright arguments $PLAYWRIGHT_ARGS
 
-docker-compose -f docker-compose.playwright.yml down
+docker-compose -f docker-compose.playwright.yml up --build --force-recreate --exit-code-from playwright --remove-orphans
```
sarayourfriend, line 16: So was it the `docker-compose down` that was causing this not to fail because the script continued to the `dc down` command after the failure of the test run? If so, an alternative fix (assuming the `down` is necessary) would be to add `set -x` to the top of the script so that any command failure causes the script to exit with the failure exit code.

obulat: Well, if you remove the `dc down` line, the tests start to fail again. But I'm not very clear as to what exactly happens.

sarayourfriend: If that is the case then I think it is what I said. The default behavior with shell scripts is not to bail out the entire script if a command fails unless the `e` option is set. 

I had remembered it wrong, btw, it should be `set -e`. `set -x` prints the commands as they're run. `e` bails on the script if any error occurs.

obulat: It worked! Thank you :)
How did you see that the Docker container was still running after tests in GitHub actions, by the way?

sarayourfriend: >  How did you see that the Docker container was still running after tests in GitHub actions, by the way?

What do you mean?

krysal: I was going to suggest using `docker-compose run --rm ...` instead but glad to see it's working now! :)

sarayourfriend: I don't remember why but `dc run` did not work for me when I tried it when originally developing the dockerized Playwright tests. There was something strange about it where it would never end the containers, IIRC.

obulat: > What do you mean?

The `dc down` line was added to solve the problem where a Playwright container wouldn't stop, right? I have never known that it didn't stop in the CI. How did you understand that that **was**, indeed, happening?

bin/playwright.sh
obulat: I removed it because we actually need the matching locale file in other tests, not only the Playwright ones.
When the `valid-locales.json` is not copied in setup-env step in other tests such as Storybook, the app fails to build. Also, simply copying the `valid-locales.json` is not enough because the app requires all of the translation `json` listed in `valid-locales.json` to build.

dhruvkb (approved):
Looks good to me! Couldn't have foreseen the `set -e` fix.

sarayourfriend (approved):
LGTM :rocket:

bin/playwright.sh
sarayourfriend: I think @krysal added it (IIRC) upon noticing that the containers weren't stopping _locally_. I don't think it matters if the containers continue to run in CI, as far as I know.

krysal: > I think @krysal added it (IIRC) upon noticing that the containers weren't stopping locally.

That is correct. The machines running the GH actions are a big black box, so that would have been much more difficult to identify, though locally it was a coincidence 🙂

krysal:
I was going to ask why are we marking all the tests to run in parallel mode? I did that for the slowest tests in my PR, those that triggered warnings after the test execution. If we want to do that we can set it up in the Playwright config file (`test/playwright/playwright.config.ts`) instead of repeating the line in every test file.

obulat:
@krysal, I first marked the e2e tests that were triggering the 'slow test' warning, and then just went ahead and added that line in every file.
What is the downside to setting up the parallel execution for all files in the Playwright config file? Adding or removing the line manually based on the test logs seems tedious and we might forget to do it.
Should I open an issue for removing this line from fast tests, or for adding the comment to the Playwright config file?

PR #1653, opened 2022-08-12
Vendor in Reakit functions

### LINKED ISSUE #1694: Some dependencies using React classes make TypeScriptification impossible
## Problem

Some of our dependencies use React classes, which pollute the types so that it's impossible to use both in the same project. The main culprit we know of is Reakit. However, there might be others (Storybook?) that make the development experience in VSCode difficult. The template `class`es are often underlined because Volar/Vetur expect the React-style types for classes instead of the Vue types.

## Description

The first step should be to vendor in the Reakit functions that we use so that we can remove the React dependencies from there.
I'm marking this a high priority because this prevents any progress in TypeScriptification. And the current state of mixing JS/JSDoc TS and TS is confusing and often causes conflicts between types.
 
## Alternatives

We could continue working with both, but this makes the devex worse.

## Implementation

obulat:
## Fixes

Fixes #1694 by @obulat

## Description

This PR vendors in the code from Reakit utils v0.15.2 that we are using for accessible popup and modal management. This is necessary because Reakit uses some of the React types, and pollutes our type system with them. That causes type checks to fail because the Vue way of typing the templates is different.

This PR is intentionally small and only copies in the functions we use. This is done to enable the further progress on Typescriptification milestone.

`ts-vue` is also updated because some of the template typing was fixed in the most recent version.

## Testing Instructions

The CI should pass.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/composables/use-focus-on-hide.js
```diff
@@ -1,8 +1,7 @@
-import { watch } from '@vue/composition-api'
-import { isTabbable } from 'reakit-utils/tabbable'
-import { getActiveElement } from 'reakit-utils/getActiveElement'
-import { contains } from 'reakit-utils/contains'
-import { ensureFocus } from 'reakit-utils/ensureFocus'
+import { watch } from '@nuxtjs/composition-api'
```
obulat, line 1: There was an import from `@vue/composition-api` which seems to behave just a little bit differently.

typings/csstype/index.d.ts
```diff
@@ -8,5 +8,8 @@ declare module 'csstype' {
     '--unusable-height'?: string
     '--progress-time-left'?: string
     '--seek-time-left'?: string
+    '--cell-idx'?: number
   }
 }
+
+export {}
```
obulat, line 15: Without this `export`, all of the csstype types were overriden with the values in this file, instead of being added to.

src/utils/reakit-utils/get-next-active-element-on-blur.ts
```diff
@@ -0,0 +1,15 @@
+// From https://github.com/ariakit/ariakit/blob/da876223f115304c8f5c12a5f04212601d9b3027/packages/reakit/src/Composite/__utils/getNextActiveElementOnBlur.ts
+import { getActiveElement } from '~/utils/reakit-utils/dom'
+
+const isIE11 = typeof window !== 'undefined' && 'msCrypto' in window
+
+export function getNextActiveElementOnBlur(event: FocusEvent) {
+  // IE 11 doesn't support event.relatedTarget on blur.
+  // document.activeElement points the next active element.
+  // On modern browsers, document.activeElement points to the current target.
+  if (isIE11) {
```
obulat, line 10: I wonder if we should remove the ie11 hacks as we do not really support ie11 🤔

src/utils/reakit-utils/README.md
```diff
@@ -0,0 +1,5 @@
+# Contents
+
+This directory contains utilities from [`reakit-utils` version 0.15.2](https://github.com/ariakit/ariakit/tree/f05d36daa6fbcc52c70cf2c71baea69025aa2402/packages/reakit-utils) .
+
+Reakit uses some React-specific classes which are mixed with the Vue-based classes and make it impossible to use both in the same project.
```
sarayourfriend, line 5: This might be a little more precise about where the conflict lies.

```suggestion
We cannot use Reakit/Ariakit directly as it imports React's JSX types (`@types/react`) which conflict with Vue's own JSX types (`@vue/runtime-dom`). [Ariakit has essentially the same issue as Storybook, for the moment](https://github.com/storybookjs/storybook/issues/12505).
```

src/utils/reakit-utils/get-next-active-element-on-blur.ts
sarayourfriend: Yes, I think it's fine to scrap this and keep the code base consistent with our actual support level :+1:

typings/csstype/index.d.ts
sarayourfriend: That is strange! Can you add this comment in the code to note that? It's not at all the expected behaviour with interface merging (unless something has changed in TS recently, but it would be a big breaking change I think).

dhruvkb (approved):
Looks great!

It's unfortunate that we had to copy code instead of just importing it but the documentation is very clear on what was blocking us and how we might be able to use the upstream package in the future.

sarayourfriend (approved):
LGTM! Approving on the virtue of the playwright tests still passing. Not sure how else to test this particular change.

PR #1661, opened 2022-08-13
Require an aria-label for VIconButton

### LINKED ISSUE #887: Add required `aria-label` prop to VIconButton
## Problem

The IconButton component should always have an aria- label unless it is `aria-hidden`, which it shouldn't probably ever be considering it's an interactive part of the page. Without the label screen readers don't have the necessary information to describe the button to the user.

## Description

We should add a required prop for the `aria-hidden` attribute to enforce that it is always passed when using this component.

## Implementation

ramadanomar:
## Fixes

Fixes #887 by @sarayourfriend

## Description

Checks if VIconButton will have an aria-label when mounted

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VIconButton/VIconButton.vue
+++ b/src/components/VIconButton/VIconButton.vue
@@ -72,9 +72,9 @@ export default defineComponent({
       default: () => ({ variant: 'plain' }),
     },
   },
+
   setup(props, { attrs }) {
     const type = (attrs['type'] ?? 'button') as ButtonType
-
     const buttonSizeClasses = computed(() => SIZE_MAP[props.size].button)
     const iconSizeClasses = computed(() => SIZE_MAP[props.size].icon)
 
@@ -83,6 +83,15 @@ export default defineComponent({
 
       buttonSizeClasses,
       iconSizeClasses,
+    }
+  },
+
+  mounted() {
+    /* Throw an error if the element has no aria-label attribute */
+    if (!this.$el.hasAttribute('aria-label')) {
+      throw new Error(
+        'VIconButton must have aria-label attribute to be accessible'
+      )
     }
   },
 })


src/components/VIconButton/VIconButton.vue
```diff
@@ -85,5 +85,14 @@ export default defineComponent({
...
   },
+
+  mounted() {
+    /* Throw an error if the element has no aria-label attribute */
+    if (!this.$el.hasAttribute('aria-label')) {
+      throw new Error(
+        'VIconButton must have aria-label attribute to be accessible'
+      )
+    }
+  },
```
sarayourfriend, line 96: It might be little more idiomatic to add a required `aria-label` prop rather than depend on attribute fall-through. I'll leave it up to the assigned reviewers Dhruv and Olga though for which approach they prefer.

ramadanomar: VIconButton's references all use a fallback attribute for aria-label, hence i opted for a simple check on mount. Is it best practice to add it as a required prop instead of following the "pattern" that is used in the code for accessibility labels (adding them as a fallback attribute) or a matter of personal preference?

sarayourfriend: Well, the syntax between an attribute and a prop are exactly the same from the consumer side (anyone using the component), so the difference is purely internal to the component. Given that, personally I think a prop makes more sense as it's the canonical and idiomatic way to encode the type and conditions of any inputs to the component. A prop would encode it both in TypeScript and in the Vue debug console, both places where we're accustomed to seeing prop requirements.

We do actually have some other checks non-idiomatic checks. In the `VButton` component, in fact: https://github.com/WordPress/openverse-frontend/blob/f3f03b4039035f8e7887cde99460d7a45fd6b5d1/src/components/VButton.vue#L225-L238

I'd probably recommend that we move the `as` validation into a prop validator as well, though :slightly_smiling_face: Unfortunately it's [not possible to encode those particular prop rules in pure TypeScript types](https://stackoverflow.com/questions/51442157/type-for-every-possible-string-value-except) and the code was written a while ago before we were maybe as focused on Vue conventions and idiomatic prop type usage.

The check on mount, I've just realised, also only covers attributes as they are defined on mount, not afterwards. If `aria-label` is bound to a variable that suddenly goes undefined or null, then this check won't catch that, whereas a prop type would. The `watch` in the example I've shared above does solve that and is probably worth doing if a prop type can't cover this use case for one reason or another.

ramadanomar: If i try to assign the aria-label attribute into a required string prop i get this error:
` error TS2322: Type 'TranslateResult' is not assignable to type 'string'.`
I can't wrap the string around a $t to translate it. Any idea on how i should approach this problem?

ramadanomar:
```
   ariaLabel: {
      type: String as PropType<string>,
      required: true,
    },
```
For further context, this is how my prop looks. That's why i opted with a simple check on mount.

sarayourfriend: We have to call `.toString()` on the results from `$t` and other i18n functions. Here's an example:

https://github.com/WordPress/openverse-frontend/blob/f3f03b4039035f8e7887cde99460d7a45fd6b5d1/src/components/VSourcesTable.vue#L3

It's a frustrating issue that I think we could solve if we had more flexibility in runtime-dom types but I don't think it's currently feasible.

PR #1668, opened 2022-08-15
Add better concurrency settings for workflows

sarayourfriend:
## Fixes

Related to https://github.com/WordPress/openverse/issues/290 by @sarayourfriend 

## Description

Workflows by default run in parallel and will finish running even if the
results of them don't matter by the time they're done. CI for example,
does not matter on a PR if the PR is updated. That's why we already had
concurrency on the ci.yml workflow that cancels in progress workflow
runs. This helps prevent our worker allowance from being eaten up by
long CI runs.

We can also use this to prevent certain workflows from running in
parallel and instead waiting for the previous run to finish. For
example, deployments. Auto-deployments should never happen in parallel
and should wait for any currently active deployments to the same
environment, whether to staging or production, to finish before starting.

Similarly, rollbacks should not happen in parallel and will block and
wait for previous rollbacks to the same environment to finish.

Lastly, there are workflows that don't need to run in parallel, but can
also cancel in progress flows. Examples of this are artifact generation
like GitHub pages updates from `main` and PO file generation also from
`main`. Both of these only need the latest invocation of them to run
because any given run will necessarily include any changes that a
previous cancelled run would have made. If we merge two PRs one after
another, we don't need to generate PO files twice. We can just generate
them for the second PR that was merged.

## Testing Instructions

Review the configurations and confirm that they make sense for the given workflow they're referencing. Then we merge and see if it works.

If there ends up being any issue with them, revert the PR so that it doesn't block ongoing work.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -6,6 +6,13 @@ on:
     branches:
       - main
 
+# Cancels all previous workflow runs for pull requests that have not completed.
+concurrency:
+  # The concurrency group contains the workflow name and the branch name for pull requests
+  # or the commit hash for any other events.
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
 jobs:
   actionlint:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


obulat (approved):
LGTM!

zackkrida (approved):
Appreciate the detailed description, this looks really compelling.

krysal (approved):
Saving several minutes of computing time, nice!

PR #1685, opened 2022-08-17
Revert "Avoid pausing audio when is loading"

sarayourfriend:
Reverts WordPress/openverse-frontend#1643

Reverting due to the introduction of an uncaught Playwright failure: https://github.com/WordPress/openverse-frontend/actions/runs/2869167598

Playwright failures were not being surfaced in PRs due to #1637


sarayourfriend:
I suspect this change introduced the failing Playwright test. If Playwright passes CI here then we'll know it is the case and we can investigate the fix in another PR with actually running Playwright tests.

krysal (approved):

zackkrida (approved):

PR #1691, opened 2022-08-18
Ensure that the attribution box is always visible

### LINKED ISSUE #1656: Attribution box does not show for some results when navigating to them from search results page
## Description

If you visit https://search-production.openverse.engineering/image/3d96348f-5b9f-4ce7-8854-8480fa3bf1a0 from the search results page (it's on the first page of "dog" + "cc0") the attribution box will not appear.

## Reproduction

1.  Visit https://search-production.openverse.engineering/search/?q=dog&license=cc0
2.  Find the result linked above (it's the second image result at the time of writing this issue) and click it
3.  See that there is no attribution box

## Screenshots

image

## Environment

 - Device:  laptop
 - OS:  Linux saras-laptop 5.18.15-arch1-2 &#35;1 SMP PREEMPT_DYNAMIC Mon, 01 Aug 2022 18:58:16 +0000 x86_64 GNU/Linux
 - Browser:  Firefox
 - Version:  Mozilla Firefox 103.0.1

## Resolution

dhruvkb:
## Fixes

Fixes #1656 by @sarayourfriend 

## Description

This issue is caused by the `license_url` field missing for some search results. This PR changes the frontend to automatically use the generated license URL if the API response does not contain it.

I've made a separate PR (WordPress/openverse-api#888) to ensure the API sends the `license_url` every time using the same logic as the frontend. Having this information in the response would be beneficial for other consumers of the API.

## Testing Instructions

0. Check out the PR and start the dev server.
1. Search for 'dog' and apply the 'CC0' license filter.
2. Find [this image](https://search-production.openverse.engineering/image/3d96348f-5b9f-4ce7-8854-8480fa3bf1a0) in the results and click on it.
3. You should see the attribution box (it would not be visible before this PR).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VMediaInfo/VMediaReuse.vue
```diff
@@ -41,11 +41,16 @@ export default defineComponent({
   setup(props) {
     const i18n = useI18n()
 
+    const licenseUrl = computed(
```
obulat, line 44: Oh, I really hate that we have to do this in the front end, too. 

Do you think it would make more sense to add `license_url` here: https://github.com/WordPress/openverse-frontend/blob/8d893bcfa21e96ccddd2ef0bbf109def653d44bf/src/utils/decode-media-data.ts
We already do add some pieces that are missing from the API response there.

sarayourfriend: The duplicated logic between the frontend and the API is a real chore, for sure Olga. If there's a way to avoid it, it would be beneficial and we could delete code :smiling_imp:

obulat: It's even worse than that: it's _triplicated_ between the catalog, the API, and the frontend.

sarayourfriend: That is.... really unfortunate. Is that just because each downstream service is compensating for the next one upstream not being fully valid or something?

dhruvkb: Yes @sarayourfriend. Since none of the layers in the stack establish any guarantees about the data they provide, we need to add this validation/compensation at each layer.

sarayourfriend: That is an unfortunate assumption we feel we have to make considering we control all the different stacks :disappointed: I would think that at least the API layer has _some_ guarantees. Are we to assume that any and all API reponse fields are able to be empty?

obulat: In the catalog, we used to save the license name and the license version as separate columns. Then, we added `license_url` field in the `meta_data` JSON column. Naturally, old items don't have that value. We tried to add `license_url` as a separate column, but with the sheer volume of the items, it's not that easy to even set a single column to `NULL` for _all_ rows in the image table using a simple `UPDATE` query. Adding the computed `license_url` definitely needs a different approach.

I'm not sure what the story is in the API, I actually [asked Dhruv](https://github.com/WordPress/openverse-api/pull/888#pullrequestreview-1077084143) about it. I though that license URL is calculated for all items during Data refresh, but apparently it isn't, and not all hits have it.

sarayourfriend: Thanks for the background Olga. It does seem like the API would be the best place to fix this, in my opinion, just because it fixes it for _everyone_ including our frontend.

I started a discussion issue for us to establish some clearer expectations around when and how we fix data integrity issues: https://github.com/WordPress/openverse/issues/292

In this case the issue is so severe that it does seem like the frontend might be a good place to fix it, but if we're going to deploy the API soon anyway (I think we would do so next week maybe?) then maybe just fixing it there with the goal of removing license URL building code from the frontend altogether would be another complementary goal. Just depends on the timing I guess? It's hard to know. Hopefully the discussion can help clarify these things. I, for one, find them confusing.

dhruvkb:
Since we're fixing this from the API side, I'll close this PR.

Issue #1604, opened 2022-08-01
Fix media fetching error handling to actually capture issue description


sarayourfriend:
## Sentry link

https://sentry.io/share/issue/0e9ffd2ec1144d8e8b504db32e4e6c25/

## Description

https://github.com/WordPress/openverse-frontend/blob/96b061f411fbc73ae55980e2b6f9ecd358931762/src/stores/media/index.ts#L388

If we don't know how to handle an error, we should not just swallow it completely and mask it with a vague string unless we're also capturing the error in Sentry. Update the above code to call `this.$sentry.captureException(error)` in the case where we don't know how to correctly handle the error. Do this with the original exception, not with the vague/useless for debugging error message we set here.

Marking as high because this prevents us from debugging actual issues in this part of the code.

sarayourfriend:
While we're at it, it's probably not a bad idea to audit our error handling across the app and create subsequent issues to better capture exceptions and not swallow error messages when we don't know how to handle errors.

sarayourfriend:
Closed by #1707. Thanks @ramadanomar

PR #1707, opened 2022-08-19
Capturing Issue Description in Sentry

ramadanomar:
## Related to 

Related to #1604 by @sarayourfriend

## Description
Not sure on when how to reproduce this issue, but this should return the whole exception to provide more context.

### DIFF
--- a/src/stores/media/index.ts
+++ b/src/stores/media/index.ts
@@ -399,6 +399,13 @@ export const useMediaStore = defineStore('media', {
         errorMessage = `Error fetching ${mediaType} from API. Unknown error`
       }
 
+      this.$nuxt.$sentry.captureEvent({
+        message: errorMessage,
+        extra: {
+          error,
+        },
+      })
+
       this._updateFetchState(mediaType, 'end', errorMessage)
       if (!axios.isAxiosError(error)) {
         throw new Error(errorMessage)
--- a/src/stores/media/index.ts
+++ b/src/stores/media/index.ts
@@ -2,6 +2,8 @@ import { defineStore } from 'pinia'
 
 import axios from 'axios'
 
+import * as Sentry from '@sentry/browser'
+
 import { warn } from '~/utils/console'
 import { hash, rand as prng } from '~/utils/prng'
 import prepareSearchQueryParams from '~/utils/prepare-search-query-params'
@@ -385,9 +387,13 @@ export const useMediaStore = defineStore('media', {
                 error.response?.status ?? 'unknown'
               }`
       } else {
+        /*Capture the error and all of its details using $sentry */
+        Sentry.captureEvent({
+          message: `Error fetching ${mediaType}`,
+          extra: { error },
+        })
         errorMessage =
-          /*Return info on the error message by appending all details*/
-          error instanceof Error ? error.message : JSON.stringify(error)
+          error instanceof Error ? error.message : 'Oops! Something went wrong'
       }
       this._updateFetchState(mediaType, 'end', errorMessage)
       if (!axios.isAxiosError(error)) {
--- a/src/stores/media/index.ts
+++ b/src/stores/media/index.ts
@@ -401,6 +401,7 @@ export const useMediaStore = defineStore('media', {
       this.$nuxt.$sentry.captureEvent({
         message: errorMessage,
         extra: {
+          mediaType,
           error,
         },
       })
--- a/src/stores/media/index.ts
+++ b/src/stores/media/index.ts
@@ -369,6 +369,10 @@ export const useMediaStore = defineStore('media', {
       }
     },
 
+    //Handles errors for media fetches and sets the fetch state accordingly
+    //Reported errors are logged to Sentry
+    //TODO: Handle errors for media fetches
+
     async handleMediaError({
       mediaType,
       error,
@@ -378,25 +382,48 @@ export const useMediaStore = defineStore('media', {
     }) {
       let errorMessage
       if (axios.isAxiosError(error)) {
-        errorMessage =
-          error.response?.status === 500
-            ? 'There was a problem with our servers'
-            : `Request failed with status ${
-                error.response?.status ?? 'unknown'
-              }`
+        // If the error is an axios error:
+        // If the error has a response property, and recieved a response that is not in the 2xx range,
+        // then the error is logged to Sentry
+        if (error.response && error.response.status >= 300) {
+          errorMessage = `Error fetching ${mediaType} from API. Request failed with status code: ${error.response.status}`
+          this.$nuxt.$sentry.captureEvent({
+            message: errorMessage,
+            extra: {
+              error,
+            },
+          })
+        } else if (!error.response && error.request) {
+          // If the error has a request property, but no response, then we capture the event in Sentry
+          errorMessage = `Error fetching ${mediaType} from API. No response received from the server`
+          this.$nuxt.$sentry.captureEvent({
+            message: errorMessage,
+            extra: {
+              error,
+            },
+          })
+        } else {
+          // Something happened in setting up the request that triggered an Error
+          errorMessage = `Error fetching ${mediaType} from API. Unknown Axios error`
+          this.$nuxt.$sentry.captureEvent({
+            message: errorMessage,
+            extra: {
+              error,
+            },
+          })
+        }
       } else {
-        /*Capture the error and all of its details using $sentry */
+        // If the error is not an axios error, then we capture the event in Sentry
+        errorMessage = `Error fetching ${mediaType} from API. Unknown error`
         this.$nuxt.$sentry.captureEvent({
-          message: `Error fetching ${mediaType}`,
-          extra: { error },
+          message: errorMessage,
+          extra: {
+            error,
+          },
         })
-        errorMessage =
-          error instanceof Error ? error.message : 'Oops! Something went wrong'
       }
+
       this._updateFetchState(mediaType, 'end', errorMessage)
-      if (!axios.isAxiosError(error)) {
-        throw new Error(errorMessage)
-      }
     },
 
     setMediaProperties(
--- a/src/stores/media/index.ts
+++ b/src/stores/media/index.ts
@@ -371,7 +371,6 @@ export const useMediaStore = defineStore('media', {
 
     // Handles errors for media fetches and sets the fetch state accordingly
     // Reported errors are logged to Sentry
-
     async handleMediaError({
       mediaType,
       error,
--- a/src/stores/media/index.ts
+++ b/src/stores/media/index.ts
@@ -385,7 +385,7 @@ export const useMediaStore = defineStore('media', {
         // If the error is an axios error:
         // If the error has a response property, and recieved a response that is not in the 2xx range,
         // then the error is logged to Sentry
-        if (error.response && error.response.status >= 300) {
+        if (error.response) {
           errorMessage = `Error fetching ${mediaType} from API. Request failed with status code: ${error.response.status}`
           this.$nuxt.$sentry.captureEvent({
             message: errorMessage,
--- a/src/stores/media/index.ts
+++ b/src/stores/media/index.ts
@@ -424,6 +424,9 @@ export const useMediaStore = defineStore('media', {
       }
 
       this._updateFetchState(mediaType, 'end', errorMessage)
+      if (!axios.isAxiosError(error)) {
+        throw new Error(errorMessage)
+      }
     },
 
     setMediaProperties(
--- a/src/stores/media/index.ts
+++ b/src/stores/media/index.ts
@@ -387,40 +387,16 @@ export const useMediaStore = defineStore('media', {
         // then the error is logged to Sentry
         if (error.response) {
           errorMessage = `Error fetching ${mediaType} from API. Request failed with status code: ${error.response.status}`
-          this.$nuxt.$sentry.captureEvent({
-            message: errorMessage,
-            extra: {
-              error,
-            },
-          })
         } else if (error.request) {
           // If the error has a request property, but no response, then we capture the event in Sentry
           errorMessage = `Error fetching ${mediaType} from API. No response received from the server`
-          this.$nuxt.$sentry.captureEvent({
-            message: errorMessage,
-            extra: {
-              error,
-            },
-          })
         } else {
           // Something happened in setting up the request that triggered an Error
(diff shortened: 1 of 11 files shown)


src/stores/media/index.ts
```diff
@@ -385,6 +387,11 @@ export const useMediaStore = defineStore('media', {
                 error.response?.status ?? 'unknown'
               }`
       } else {
+        /*Capture the error and all of its details using $sentry */
+        Sentry.captureEvent({
```
sarayourfriend, line 391: Inside of Pinia stores you can access the [pre-configured Sentry instance via `this.$nuxt.$sentry`](https://sentry.nuxtjs.org/). In this case I think it'd make sense to use `captureException` explicitly when an error is caught. For the case where we set the message to "Oops! Something went wrong", if we could capture the event with something more specific that would be great. Maybe `captureMessage` is better in that case. I'm not totally sure of the difference. Do you happen to know?

@obulat and @dhruvkb do y'all know when that "Oops! Something went wrong" clause happens and what it's meant to handle?

BTW, to test this, I threw an arbitrary error in fetchSingleMediaType before `setMedia` is called and then a breakpoint in my browser debugger on this line.

ramadanomar: > Inside of Pinia stores you can access the [pre-configured Sentry instance via `this.$nuxt.$sentry`](https://sentry.nuxtjs.org/).
I tried accesing it this way before but it keeps failing when i run pnpm test with the error `"Cannot read properties of undefined (reading 'captureException')"`

sarayourfriend: Ah! You've got to mock out context variables in tests. You can add $nuxt and the $sentry mock to this Pinia test utility we use: https://github.com/WordPress/openverse-frontend/blob/b863e3e4b88eb667a87a2c620abc688ab29838eb/test/unit/test-utils/pinia.js#L4-L9

ramadanomar: > if we could capture the event with something more specific that would be great. Maybe captureMessage is better in that case. I'm not totally sure of the difference. 

[According to sentry docs,](https://docs.sentry.io/platforms/javascript/usage/) we can pass a non-error object to captureException tho it will lack the context. `You can pass an Error object to captureException() to get it captured as event. It's also possible to pass non-Error objects and strings, but be aware that the resulting events in Sentry may be missing a stacktrace.`

Capture event triggers a manual event, that means we recieve all the breadcumbs that were collected until that moment. We can also include a dump of the application if we are sure we won't surpass the [event payload max size](https://develop.sentry.dev/sdk/event-payloads/)

sarayourfriend: For now I think better not to dump the application. If we need it to debug a particular error that sounds like a great feature though, thanks for sharing it.

Do whatever you think is best _vis a vis_ which Sentry function to use :slightly_smiling_face:

obulat: > @obulat and @dhruvkb do y'all know when that "Oops! Something went wrong" clause happens and what it's meant to handle?

No, I don't know what it was meant to handle. It was there even before Nuxt :) I really wonder what we catch with this PR!

ramadanomar: Then i believe capture event is the best option, [as we can enrich the event data as we see fit ](https://docs.sentry.io/product/sentry-basics/integrate-backend/capturing-errors/)

obulat:
I'm not sure what the scope of this PR is, @ramadanomar, but it would be nice to re-write the `handleMediaError` function in light of the [Axios _handling errors_ page](https://axios-http.com/docs/handling_errors).

## For instances of Axios errors:
### In case the error has a `response` property
_The request was made but has received a response that is not in 2xx range._
-  set the message to "Error fetching ${mediaType}. Request failed with status code ${error.response.status}"
- capture the event in Sentry 

### In case the error doesn't have a `response` property, but has a `request` property
_The request was made but no response was received.
`error.request` is an instance of XMLHttpRequest in the browser and an instance of http.ClientRequest in node.js_
- set the message to "Error fetching ${mediaType}. Could not receive a response from the server"
- capture the event in Sentry

### Otherwise
_Something happened in setting up the request that triggered an Error_
- set the message to "Error fetching ${mediaType}. Unknown Axios error"
- capture the event in Sentry.

## Non-axios errors
I have no idea if those can happen, but we should still add a branch to capture those with Sentry and set the message to "Error fetching ${mediaType}. Unknown error"

I wonder how much we want to capture in Sentry. Do we want to capture all failed requests in Sentry, @WordPress/openverse-frontend? Would capturing the event with `toJSON` make sense, or would it be too much information?

ramadanomar:
> I'm not sure what the scope of this PR is, @ramadanomar, but it would be nice to re-write the `handleMediaError` function in light of the [Axios _handling errors_ page](https://axios-http.com/docs/handling_errors).

Hey! Thanks for the feedback. My scope was to provide more context to audio errors. I am pushing an updated handleMediaError function right now. As of toJSON, i am not sure when this error is called to determine how much info we need. The new handleMediaError function should allow for easy changes in case we need to edit the level of detail we require.

src/stores/media/index.ts
```diff
@@ -378,21 +382,47 @@ export const useMediaStore = defineStore('media', {
...
-        errorMessage =
-          error.response?.status === 500
-            ? 'There was a problem with our servers'
-            : `Request failed with status ${
-                error.response?.status ?? 'unknown'
-              }`
+        // If the error is an axios error:
+        // If the error has a response property, and recieved a response that is not in the 2xx range,
+        // then the error is logged to Sentry
+        if (error.response && error.response.status >= 300) {
```
obulat, line 388:
```suggestion
        if (error.response) {
```
If error.response already checks that the status is in the 2xx range, so we don't need to add the status check.

src/stores/media/index.ts
```diff
@@ -378,21 +382,47 @@ export const useMediaStore = defineStore('media', {
...
+          errorMessage = `Error fetching ${mediaType} from API. Request failed with status code: ${error.response.status}`
+          this.$nuxt.$sentry.captureEvent({
+            message: errorMessage,
+            extra: {
+              error,
+            },
+          })
+        } else if (!error.response && error.request) {
```
obulat, line 396:
```suggestion
        } else if (error.request) {
```
The `error.response` check is unnecessary here if the first check is simple for `if(error.response)`

ramadanomar: Fixed. Thanks for pointing it out!

src/stores/media/index.ts
```diff
@@ -378,16 +382,47 @@ export const useMediaStore = defineStore('media', {
...
+          })
+        } else {
+          // Something happened in setting up the request that triggered an Error
+          errorMessage = `Error fetching ${mediaType} from API. Unknown Axios error`
+          this.$nuxt.$sentry.captureEvent({
+            message: errorMessage,
+            extra: {
+              error,
+            },
+          })
```
obulat, line 413: We can move the `sentry.captureEvent` calls to line 425, before `this._updateFetchState`, and only call once.

ramadanomar: Is there any scenario where it would get called twice? If you are refering on the times it appears in the code, i believe having a separate call for each exception allows us to fine-tune the amount of detail we want/need for each type of error. If you think that isn't needed, i'll change it

obulat: Sorry, I wasn't clear :)
Currently, the same code is repeated 3 times:
```
          this.$nuxt.$sentry.captureEvent({
            message: errorMessage,
            extra: {
              error,
            },
          })
```
Since only the errorMessage changes, we can set the errorMessage inside the `if ..else`, and then call `sentry.captureEvent` outside of the `if ... else`.

src/stores/media/index.ts
```diff
@@ -378,16 +382,47 @@ export const useMediaStore = defineStore('media', {
...
+        } else if (error.request) {
+          // If the error has a request property, but no response, then we capture the event in Sentry
+          errorMessage = `Error fetching ${mediaType} from API. No response received from the server`
+          this.$nuxt.$sentry.captureEvent({
+            message: errorMessage,
+            extra: {
+              error,
+            },
+          })
+        } else {
```
obulat, line 405:
```suggestion
          errorMessage = `Error fetching ${mediaType} from API. No response received from the server`
        } else {
```

src/stores/media/index.ts
obulat, line 413:
```suggestion
```

src/stores/media/index.ts
```diff
@@ -378,16 +382,47 @@ export const useMediaStore = defineStore('media', {
...
+        // If the error is not an axios error, then we capture the event in Sentry
+        errorMessage = `Error fetching ${mediaType} from API. Unknown error`
+        this.$nuxt.$sentry.captureEvent({
+          message: errorMessage,
+          extra: {
+            error,
+          },
+        })
       }
+
```
obulat, line 425: this.$nuxt.$sentry.captureEvent({
            message: errorMessage,
            extra: {
              error,
            },
          })

src/stores/media/index.ts
```diff
@@ -378,16 +382,47 @@ export const useMediaStore = defineStore('media', {
...
-          error instanceof Error ? error.message : 'Oops! Something went wrong'
+        // If the error is not an axios error, then we capture the event in Sentry
+        errorMessage = `Error fetching ${mediaType} from API. Unknown error`
+        this.$nuxt.$sentry.captureEvent({
+          message: errorMessage,
+          extra: {
+            error,
+          },
+        })
       }
```
obulat, line 424:
```suggestion
        errorMessage = `Error fetching ${mediaType} from API. Unknown error`
      }
```

ramadanomar:
Thanks for the changes @obulat . I resolved them.

sarayourfriend:
@obulat From my perspective all failed requests that we don't know how to explicitly handle should be captured in Sentry so that we can be aware of them and either (a) handle those explicitly or (b) explicitly ignore them as non-issues or unsolveable. Does that seem like a reasonable standard?

src/stores/media/index.ts
```diff
@@ -369,6 +369,10 @@ export const useMediaStore = defineStore('media', {
...
     },
 
+    //Handles errors for media fetches and sets the fetch state accordingly
+    //Reported errors are logged to Sentry
+    //TODO: Handle errors for media fetches
+
```
obulat, line 375:
```suggestion
    // Handles errors for media fetches and sets the fetch state accordingly
    // Reported errors are logged to Sentry

```

obulat (approved):
This change will definitely help us fix more errors! I've left a couple of nit-picky formatting comments, other that that it's good to go 🎉

test/unit/test-utils/pinia.js
```diff
@@ -5,6 +5,12 @@ export const createPinia = () =>
   pinia.createPinia().use(() => ({
     $nuxt: {
       $openverseApiToken: '',
+      $sentry: {
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        captureException: () => {},
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        captureEvent: () => {},
```
sarayourfriend, line 12: A slightly clearer and more intentional way to deal with this "no empty function" is to add a comment inside the function body, something like this:

```suggestion
        captureException: () => {/* noop */},
        captureEvent: () => {/* noop */},
```

This presents us with the opportunity to document _why_ the function is empty rather than just silencing the linter.

However, it might make more sense to go ahead and use `jest.fn()` for both of these.
```suggestion
        captureException: jest.fn(),
        captureEvent: jest.fn(),
```

On that note: can you update the unit tests to also confirm that they call `captureEvent` as we expect, with the error object and the message?

sarayourfriend (dismissed):
Looks great. Haven't run it locally yet but I anticipate it being fine. My only request is to expand the unit tests to cover the two new branches and to confirm that `captureEvent` is called as we expect.

ramadanomar:
Any idea on how i could mock the sentry capture event without importing `nuxtjs/sentry `module? tried accesing it through this.$nuxt.$sentry , this.$sentry, $nuxt.$sentry and $sentry but they all seem to be undefined. Another method i tried was mocking the function call, something similar to 

```js
jest.mock('axios', () => ({
  ...jest.requireActual('axios'),
  isAxiosError: jest.fn((obj) => 'response' in obj),
}))
```
but couldn't seem to be able to deconstruct the event properly. Any pointers on how i could approach this? Thanks!

obulat:
@ramadanomar, to mock Sentry centrally, you can add this snippet to `test/unit/setup-after-env.js`:
```
import Vue from 'vue'

Vue.prototype.$nuxt = {
  context: {
    $sentry: {
      captureException: jest.fn(),
    },
  },
}
```

You can also add any other functions that need to be mocked.
I found this solution here: https://github.com/vuejs/pinia/discussions/1252

obulat:
Then, for example, on line 385 in `test/unit/specs/stores/media-store.spec.js` you could use the mock like this:
```
      expect(mediaStore.$nuxt.$sentry.captureEvent).toHaveBeenCalledWith({
        message:
          'Error fetching audio from API. Request failed with status code: 500',
        extra: { error },
      })
```

Please note that you would also need to add `captureEvent` to the mock above for this to work correctly.

obulat:
@ramadanomar, I've opened a PR in your fork with some more tests for different branches of error handling. I would really be happy to approve after those tests are added.

ramadanomar:
I've merged your pr @obulat. Thanks for the help!

obulat (approved):
Thank you for all your work on this PR, @ramadanomar ! It will be really valuable in triaging Sentry issues.

src/stores/media/index.ts
```diff
@@ -369,6 +369,9 @@ export const useMediaStore = defineStore('media', {
       }
     },
 
+    // Handles errors for media fetches and sets the fetch state accordingly
+    // Reported errors are logged to Sentry
+
```
zackkrida, line 374:
```suggestion
    // Reported errors are logged to Sentry
```

src/stores/media/index.ts
```diff
@@ -369,6 +369,9 @@ export const useMediaStore = defineStore('media', {
...
     async handleMediaError({
       mediaType,
```
zackkrida, line 376:
```suggestion
    // Reported errors are logged to Sentry
    async handleMediaError({
      mediaType,
```

zackkrida (approved):
LGTM

Issue #1605, opened 2022-08-01
`NotSupportedError: The operation is not supported` raised in audio search in blink browsers (Safari + Chrome)


sarayourfriend:
## Sentry link

https://sentry.io/share/issue/48703a876a6445429029dda4f1b4cbea/

## Description

This issue is exclusively raised on Safari. I don't have access to Safari so I can't reproduce this. Based on the breadcrumbs it appears the user just searched for `podcast` audio and then tried to play a couple files.

Marking as low because it does not affect very many people but I can't tell whether it breaks the search page. If someone on @WordPress/openverse-frontend with Safari 15 could test this that'd be great and reprioritize if it's not that severe.

dhruvkb:
I searched for podcast audio on Safari 15. Playing/pausing and seeking all seem to be working as expected. I was unable to reproduce this error (or any error for that matter, it was all working normally).

sarayourfriend:
Sounds good to me, thanks for testing it. It's ~1 event per day so I'm expecting it to be some strange edge case.

zackkrida:
This problem occurs when an audio src is invalid. We need to start handling exceptions on the `.play()` audio event and, in the case of an error, figure out how to proceed. We might want to at a minimum stop the audio track, clear the active media, and flash an error message, perhaps using the new #1561 component.

zackkrida:
There is more detail in #1680, but after adding many new Jamendo records that don't appear to support audio streaming this problem is occurring much more often, on nearly every search.

zackkrida:
We're seeing this because of this issue:

Jamendo urls are sending out CORB errors Cross-Origin Read Blocking (CORB) blocked cross-origin response https://prod-1.storage.jamendo.com/download/track/176092/mp32 with MIME type text/html. See https://www.chromestatus.com/feature/5629709824032768 for more details. which causes an unhandled exception when we try to play an audio track.

Regardless of the cause, some error handling should be implemented asap.

sarayourfriend:
The error handling is recommended in #1600 as well.

zackkrida:
This problem is occuring much less commonly with the fixed Jamendo URLs, but we should still add this level of error handling as a best practice and to cover all edge cases.

Issue #1600, opened 2022-08-01
AudioElement.play()'s promise is not properly handled when erroring


sarayourfriend:
## Description

We do not fully handle the promise returned by the AudioElement's `play` function: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play#return_value

We need to resolve errors in the promise by catching and handling them. We should not assume that the audio will play without issue.

https://sentry.io/share/issue/ba82b7f580074af3901151497c1ef3e2/

## Reproduction

1. Go to https://search.openverse.engineering/search/audio?q=birds and open your browser's developer console
2. Quickly play and then pause an audio file before it has time to load. Note the error message in your console.

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

ramadanomar:
🙋 I would be interested in resolving this bug.

sarayourfriend:
Thanks @ramadanomar :tada: I've assigned the issue to you.

zackkrida:
Noting that this also impacts audio files that are invalid, not just the condition identified in reproduction step 2.

PR #1713, opened 2022-08-20
Exception Handling Audio

ramadanomar:
## Fixes
This PR should tackle issues like #1605 #1600

## Description
Adding error handling when playing sending a request to play audio. 

I still have couple of changes to make and test before publishing it but would love some early feedback

### DIFF
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -336,10 +336,12 @@ export default defineComponent({
     watch(
       activeAudio.obj,
       (audio) => {
-        if (
-          audio !== localAudio &&
-          (status.value === 'playing' || status.value === 'loading')
-        ) {
+        if (audio !== localAudio && status.value === 'playing') {
+          // pause the local audio if the global audio changes
+          // check if audio has finished loading before pausing
+          if (audio?.readyState === 4) {
+            pause()
+          }
           localAudio?.pause()
         }
       },
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -314,6 +314,7 @@ export default defineComponent({
     const play = () => {
       // delay initializing the local audio element until playback is requested
       if (!localAudio) initLocalAudio()
+
       try {
         localAudio?.play()
       } catch (e) {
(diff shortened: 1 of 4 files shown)


zackkrida:
Could you add a reference to this documentation, so that folks who don't know what the status `4` means can learn? https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState#value

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -312,18 +314,47 @@ export default defineComponent({
...
+            localAudio?.pause()
+          } else {
+            //If the local audio cannot be paused, we should check if it is loading
+            //For more info about audio ready states check out:
+            //https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState#value
+            if (localAudio?.readyState === 0) {
+              //If the local audio is loading, we should wait for it to load before pausing it
+              localAudio?.addEventListener('canplay', () => {
+                localAudio?.pause()
+              })
```
sarayourfriend, line 355: Would this listener need to be cleaned up after pause is called?

sarayourfriend (commented):
What are the trade offs of this approach vs catching the error from the promise (via `.catch`, for example)? I like the explicit state handling in this approach but I wonder if we're catching all the edge cases.

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -312,18 +314,47 @@ export default defineComponent({
     const play = () => {
       // delay initializing the local audio element until playback is requested
       if (!localAudio) initLocalAudio()
-      localAudio?.play()
+
+      try {
+        localAudio?.play()
```
krysal, line 319: We need to handle the [`Promise`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play#return_value) returned by the `play()` function, so catching the exception should look more like this:
```suggestion
        localAudio?.play().catch((err) => {
        if (err.name === 'AbortError') {
          // The play() request was interrupted by a call to pause()
          // therefore we can safely ignore it.
          return
        }
        // Send issue to Sentry...
      })
```
This will solve #1600 for sure, but #1605 will need to show a message to the user so they are aware of why they can't play the track.

krysal (approved):
I reverted the changes to the `VPlayPause.vue` file and can still reproduce the issue with the steps on #1600, I don't think those changes will be needed if we catch the exception raised when pausing the audio before it has loaded (`AbortError`).

I left you a suggestion that can help. You might want to reduce the scope of this PR to solve that issue first.

krysal (changes requested):
Ups, I intended to request changes.

ramadanomar:
Continued on #1724. Thanks for all the feedback!

PR #1717, opened 2022-08-22
Remove PM2 from the Dockerfile and add curl for the healthcheck

zackkrida:
## Fixes

Related to #1722 by @zackkrida 

## Description

Removes PM2 from our Dockerfile and removes related configuration files. This change should not be merged until our frontend deployment task definitions (located in our infra repo: WordPress/openverse-infrastructure#197) are updated to run two tasks concurrently. 

Essentially, we are moving up the management of running multiple frontend tasks from _inside_ of our container to inside of AWS ECS, where we'll have more control and the ability to tweak memory thresholds, the number of tasks, and other settings without requiring a rebuild of the image.

Also updates our healthcheck to compare the available system memory to a configurable minimum available memory percentage. This _should_ prevent downtime in the event of a healthcheck failure[^1].

## Testing Instructions

Try to build and run the Dockerfile locally.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

[^1]: Except in the almost-impossible case where `healthcheck interval < time in which our memory threshold switches from acceptable to the app failing`. For example, with the current default, if our available memory went from 89% to the app failing with `SIGTERM` in 15s, we might have a few seconds of downtime.


sarayourfriend:
> This change should not be merged until our frontend deployments are updated to run two tasks concurrently

For clarification: this is in reference to the task definition, not the deployment workflows.

Dockerfile
```diff
@@ -55,4 +50,4 @@ ENV PORT=8443
 # expose port 8443 by default
 EXPOSE 8443
 
-CMD ["pm2-runtime", "start", "ecosystem.config.js"]
+ENTRYPOINT [ "npm", "start" ]
```
sarayourfriend, line 53: Would it be good idea to use `pnpm` here so it doesn't require using `--` to pass arguments down to `nuxt` CLI?

zackkrida: Would that require us to explicitly install pnpm after the `FROM node:alpine as app
` line? Alternatively I could add the '--' to the end of the entrypoint array.

sarayourfriend: Ahh I suppose so. It would be nice to avoid that so we don't have to add that extra installation time. Another alternative is that we could call to the nuxt CLI directly:

```
ENTRYPOINT [ "node", "./node_modules/nuxt/bin/nuxt.js", "start" ]
```

It's not that important though, whatever you think is easiest and most understandable.

obulat (dismissed):
LGTM!

Dockerfile
zackkrida:
```suggestion
ENTRYPOINT [ "npm", "start", "--" ]
```

sarayourfriend (commented):
I get this when I try to run the new image:

```
➜  openverse-frontend git:(remove-pm2) docker run -p 8443:8443/tcp openverse-frontend:latest                                      

> openverse-frontend@2.2.1 start
> nuxt start

ℹ Sentry reporting is disabled ("disabled" option has been set)

 FATAL  details.family.toLowerCase is not a function

  at node_modules/.pnpm/ip@1.1.5/node_modules/ip/lib/ip.js:385:39
  at Array.filter (<anonymous>)
  at node_modules/.pnpm/ip@1.1.5/node_modules/ip/lib/ip.js:384:37
  at Array.map (<anonymous>)
  at ip.address (node_modules/.pnpm/ip@1.1.5/node_modules/ip/lib/ip.js:379:37)
  at Listener.computeURL (node_modules/.pnpm/@nuxt+server@2.15.8/node_modules/@nuxt/server/dist/server.js:469:60)
  at Listener.listen (node_modules/.pnpm/@nuxt+server@2.15.8/node_modules/@nuxt/server/dist/server.js:510:10)
  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
  at async Server.listen (node_modules/.pnpm/@nuxt+server@2.15.8/node_modules/@nuxt/server/dist/server.js:973:5)
  at async Object.run (node_modules/.pnpm/@nuxt+cli@2.15.8/node_modules/@nuxt/cli/dist/cli-start.js:124:5)


   ╭─────────────────────────────────────────────────────────────╮
   │                                                             │
   │   ✖ Nuxt Fatal Error                                        │
   │                                                             │
   │   TypeError: details.family.toLowerCase is not a function   │
   │                                                             │
   ╰─────────────────────────────────────────────────────────────╯

npm notice 
npm notice New minor version of npm available! 8.6.0 -> 8.18.0
npm notice Changelog: <https://github.com/npm/cli/releases/tag/v8.18.0>
npm notice Run `npm install -g npm@8.18.0` to update!
npm notice 
```

:thinking: I'll debug this later today.

sarayourfriend (commented):
I think we need to change `node:alpine` (back?) to `node:alpine-16` in the `app` build stage. I wonder how this Dockerfile even works today...

zackkrida:
Going to try building this locally now

zackkrida:
Added curl and confirmed that the healthcheck `curl -f http://localhost:8443/healthcheck || exit 1` works with `docker exec`. Ready to review @WordPress/openverse-frontend.

sarayourfriend:
Taking a break from V8 debugging to review this now. I'm just going to download the built image and confirm it runs and that curl works as expected.

sarayourfriend (dismissed):
Works perfectly. I also confirmed that the new ENTRYPOINT works as expected as well by passing `--help` and confirming that I get the Nuxt CLI help text back.

obulat (approved):
I built and ran it locally on Mac, works well and returns this on /healthcheck:
```
ℹ Healthcheck memory information:  {
  avaliableMemory: 77.38,
  desiredAvaliableMemory: 15,
  avaliableMemoryIsSufficient: true
}
```

src/server-middleware/healthcheck.js
```diff
@@ -1,4 +1,49 @@
...
+    .info()
+    .then((info) => {
+      const avaliableMemory = info.freeMemPercentage
+
+      /**
+       * Server Middleware do not have access to any context or configuration,
+       * so we use a raw environment variable and set a default here, instead
+       * of in `nuxt.config.ts` or `utils/env.ts`.
+       */
+      const desiredAvaliableMemory = Number(
```
obulat, line 22: Do we need to add a check so that we accidentally don't set the desiredAvailableMemory to less than 0 or more than 100?

src/server-middleware/healthcheck.js
```diff
@@ -1,4 +1,52 @@
...
+        process.env.HEALTHCHECK_DESIRED_AVALIABLE_MEMORY_PERCENTAGE ?? '10'
```
sarayourfriend: Can we make the default something that would be sensible for production, rather than the value we want to use for testing? The environment variable should be sufficient for testing with very low values.

src/server-middleware/healthcheck.js
```diff
@@ -1,4 +1,52 @@
...
+ *
+ * Useful in the event of memory leaks, in particular.
+ *
+ * @type {import('@nuxt/types').ServerMiddleware}
+ */
+export default function Healthcheck(_, res) {
+  osu.mem
+    .info()
+    .then((info) => {
+      const avaliableMemory = info.freeMemPercentage
```
sarayourfriend, line 15: Can we use `async/await` here or is it not allowed for some reason?

sarayourfriend (changes requested):
Can you add a unit test for the expanded healthcheck middleware? Otherwise LGTM. `node-os-utils` is really nice, good find.

src/server-middleware/healthcheck.js
zackkrida: My intention _was_ to use something sensible for production, I thought 10% would be fine but I'm thinking something closer to 20% will be more reasonable. Do you think even higher, @sarayourfriend?

src/server-middleware/healthcheck.js
zackkrida: I can check and switch it over, I'm not sure.

zackkrida: done, works fine.

src/server-middleware/healthcheck.js
sarayourfriend: Oh gosh, this is _available_ memory, not utilised memory. My bad! I misread how this worked. 10 or 20, something around there makes perfect sense to me.

src/server-middleware/healthcheck.js
zackkrida: I'm not sure, I did add a comment to the env file about this but I don't think we need to validate it in the code.

tsconfig.json
```diff
@@ -84,7 +84,8 @@
     "src/composables/use-event-listener-outside.js",
 
     // This section cannot be converted to TypeScript because it's used in non-TS contexts (like Tailwind configuration)
-    "src/constants/screens.js"
+    "src/constants/screens.js",
+    "src/server-middleware/healthcheck.js"
```
AetherUnbound, line 88: Why did this need to be added here? Was the healthcheck present, just unused?

AetherUnbound (approved):
Looks good, builds and works on my machine! 🚀

PR #1724, opened 2022-08-24
Handling play errors

### LINKED ISSUE #1600: AudioElement.play()'s promise is not properly handled when erroring

ramadanomar:
## Fixes
#1600 and #1605

##Description
Thanks to @krysal and @sarayourfriend for all the tips and suggestions on the previous pr (#1713). We can manage the errors in a cleaner way. For local (settings) related errors, can we use the VSnackbar component to alert the user?

## Draft
This pr is still wip as i have to write the test cases and figure how to make VSnackbar show

### DIFF
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -15,6 +15,10 @@
       :status="status"
       :current-time="currentTime"
     >
+      <VSnackbar v-show="hasError">
+        {{ messageError }}
+      </VSnackbar>
+
       <template #controller="waveformProps">
         <VWaveform
           v-bind="waveformProps"
@@ -77,6 +81,7 @@ import VBoxLayout from '~/components/VAudioTrack/layouts/VBoxLayout.vue'
 import VGlobalLayout from '~/components/VAudioTrack/layouts/VGlobalLayout.vue'
 import VLink from '~/components/VLink.vue'
 import VWarningSuppressor from '~/components/VWarningSuppressor.vue'
+import VSnackbar from '~/components/VSnackbar.vue'
 
 /**
  * Displays the waveform and basic information about the track, along with
@@ -89,12 +94,12 @@ export default defineComponent({
     VWaveform,
     VLink,
     VWarningSuppressor,
-
     // Layouts
     VFullLayout,
     VRowLayout,
     VBoxLayout,
     VGlobalLayout,
+    VSnackbar,
   },
   props: {
     /**
@@ -128,6 +133,10 @@ export default defineComponent({
 
     const status = ref<AudioStatus>('paused')
     const currentTime = ref(0)
+
+    // We need to keep track if there was any error while playing the audio track.
+    let hasError = false // We initialize it to false, because we assume that this is a new track.
+    let messageError = '' // The error message is currently empty.
 
     const initLocalAudio = () => {
       // Preserve existing local audio if we plucked it from the global active audio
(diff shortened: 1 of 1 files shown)


src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -129,6 +134,10 @@ export default defineComponent({
     const status = ref<AudioStatus>('paused')
     const currentTime = ref(0)
 
+    // We need to keep track if there was any error while playing the audio track.
+    let hasError = false // We initialize it to false, because we assume that this is a new track.
+    let messageError = '' // The error message is currently empty.
```
sarayourfriend, line 139: Heads-up: these should be `ref` objects if they're going to cause re-renders.

```suggestion
    // We need to keep track if there was any error while playing the audio track.
    const hasError = ref(false) // We initialize it to false, because we assume that this is a new track.
    const messageError = ref('') // The error message is currently empty.
```

shreyanshmalvya (commented):
Apart from what @sarayourfriend pointed out, it looks good to me 👌

src/components/VAudioTrack/VAudioTrack.vue
sarayourfriend: Note that their usages will also have to be updated (to set `.value` for example)

zackkrida:
Marking this as blocked until https://github.com/WordPress/openverse-frontend/issues/1740 is sorted.

krysal:
I think we can finish this before the design issue is done with the already available waveform's message. I'll review this between today and tomorrow with that goal.

zackkrida:
That sounds good, @krysal, we can make the snackbar upgrade in a future follow up PR.

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -310,9 +310,27 @@ export default defineComponent({
...
+      localAudio?.play().catch((err) => {
+        let errorMsg = ''
+        switch (err.name) {
+          case 'NotAllowedError':
+            errorMsg = 'err_unallowed'
+            break
+          case 'NotSupportedError':
+            errorMsg = 'err_unsupported'
+            break
+          default:
```
obulat, line 326: Do we want to capture this error in Sentry to figure out what it is?

krysal: You're correct! thanks for reminding me of this.

obulat (changes requested):
@krysal , I've opened a PR that fixes the test: https://github.com/ramadanomar/openverse-frontend/pull/2
I think we can merge it in here, and merge this PR, too.

krysal:
I added that great test. Thank you a lot, @obulat 😄 This is ready for another review! CC @dhruvkb @zackkrida

obulat (approved):
Thank you for working on it, @ramadanomar and @krysal!

src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -322,9 +324,27 @@ export default defineComponent({
...
+            errorMsg = 'err_unknown'
+            $sentry.captureException(err)
```
dhruvkb, line 342: Appreciate recording these exceptions in Sentry!

src/components/VAudioTrack/layouts/VGlobalLayout.vue
```diff
@@ -2,7 +2,7 @@
   <div class="global-track flex w-full flex-row">
     <div class="flex-shrink-0">
       <VAudioThumbnail :audio="audio" />
-      <slot name="play-pause" :size="size" :layout="layout" />
+      <slot name="play-pause" size="medium" layout="global" />
```
dhruvkb, line 5: Awesome simplification!

dhruvkb (approved):
This works great, adds a nice test case and makes some good refactors along the way. LGTM!

obulat:
I'm going to merge this PR as @ramadanomar is away.

PR #1751, opened 2022-08-27
Simplify Filters to fix the SSR problem

### LINKED ISSUE #1748: Filter sidebar is empty and frontend is broken on search endpoint SSR loads
## Description

When making a server-side rendered visit to any of the `/search/` endpoints of the frontend, the page is completely broken, if the filter sidebar was opened on a previous client-side visit. 

This was fixed by rendering the sidebar client-side only in #1749 and pushed in https://github.com/WordPress/openverse-frontend/releases/tag/v3.4.6. We'll want to remove the `<ClientOnly>` component and fix this to remove the slight layout shift introduced by the hotfix.

## Reproduction

> **Note**
> This can only be reproduced locally in the built application with `pnpm build` and `pnpm start`. It is not observable using `pnpm dev`.

1. Remove `<ClientOnly>` around the sidebar
1. Visit https://search-staging.openverse.engineering and submit a search
2. Open the filter sidebar
3. Refresh the page
4. Observe the site is broken
5. Repeat steps 1-4 but _close_ the filter sidebar, and observe the site works correctly on the SSR loads.

## Screenshots

CleanShot 2022-08-26 at 15 51 13@2x

obulat:
## Fixes

Fixes #1748 by @zackkrida

## Description

This PR fixes the filters components and makes it simpler. Instead of a computed component, we select the filter component based on `v-if`. And because the conditions are mutually exclusive, there is never a case when there are two FilterGrids in the document, so there is no element `id` collision as there were previously.

I realized that if a composable has an `onMounted` hook, it runs it whenever any component that imports this composable is mounted. So, it was updating the filters visibility status in the local storage every time a component using `use-filter-visibility` mounted.  For mobile screens, we set the visibility to false in onMounted even if the localStorage was false. So, when visibility was set to true on mobile, the `onMounted` hook for the menu would run, and set the visibility back to false.

## Testing Instructions

Build and start the app using `pnpm build && pnpm start`. Go to the search page (http://localhost:8443/search/audio?q=birds), set the screen wider than mobile breakpoint and open the filters. Reload the page and check that the filters sidebar has the checkboxes inside it.
Also try making the screen wider or narrower while the filters are open. The filters should change into full screen modals on mobile and sidebar on desktop, and the desktop filters should not cover the search results.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb (approved):
This is a very comprehensive refactor, far more in-depth than the localised changes I had planned. Looks good to me.

package.json
```diff
@@ -25,6 +25,7 @@
     "generate": "nuxt generate",
     "start": "nuxt start",
     "prod": "pnpm build:only && pnpm start",
+    "prod:playwright": "pnpm i18n:copy-test-locales && pnpm build:only && pnpm start",
```
obulat, line 28: I'm not sure if we actually need a separate `prod:playwright` or if we only use `prod` for playwright tests.
This particular script is necessary to make sure that the locale files used when running `playwright:test` locally, the locale files from `test/locales` are used, not the ones downloaded from `translate.wordpress.org`. CI already does that in `setup-node-env`.

krysal (commented):
Fan of code simplifications ✨ One question I have is that the filter sidebar is not open on desktop by default anymore. I recall this was an explicit design requirement to bring attention to available filters, isn't there a problem with this?

package.json
obulat: Whenever you run `pnpm dev` locally, you load the real locale files from GlotPress. Those files are gitignored in the repo. Then, when you run the playwright tests, those real locale files are used in the Playwright tests, and cause them to error out because we don't really have Arabic translations now (and they are used in the RTL tests).

obulat:
> Fan of code simplifications ✨ One question I have is that the filter sidebar is not open on desktop by default anymore. I recall this was an explicit design requirement to bring attention to available filters, isn't there a problem with this?

I guess I forgot about that requirement. Currently on desktop, we check for the `filters open` value that's saved in the local storage. If something is stored, we use that value. Otherwise, we default to `false`. I will change this default to true, thank you for catching this, @krysal!

zackkrida (approved):
I have not looked into the cause of the playwright error but this is a great refactor.

obulat:
I've updated the code to make the filters on desktop open by default at @krysal's suggestion. This caused all the desktop snapshots of search pages and the header on the search pages to fail because the filters are open, and the filters button is in 'pressed' state. I'm updating the snapshots, and hope to fix the Playwright errors shortly.

PR #1758, opened 2022-08-30
Replace focus-trap-vue with Vueuse implementation

### LINKED ISSUE #1752: initialFocus did not return a node
## Description
While looking into #1737, i found that i occasionally receive `Uncaught Error: `initialFocus` did not return a node` while clicking the mobile filter button repeatedly. Could possibly be the reason of the tests failing

## Reproduction

1. go to https://search-production.openverse.engineering/search/?q=cat
2. enter dev tools and select mobile (responsive mode)
3. you can also edit user agent if you want but it works on all that i tried (samsung,iphone,huawei)
4. spam the filter button with the console open until you see the error
## Screenshots

image

## Environment
I tried using both the development branch of chrome (Chrome dev) and firefox stable branch on ubuntu 20.04 lts. Tried a couple of UA but seems to be reproducible on any environment but only sometimes

## Resolution

I'm looking into it right now, if you know what might be causing it lmk!

obulat:
## Fixes

Fixes #1752 by @ramadanomar

## Description

This PR copies VueUse integrations/useFocusTrap/component implementation to use the latest version of the upstream dependency, `focus-trap`.

It is clear that `focus-trap-vue` is not going to update the Vue-2 branch to add the fix for the linked issue. I wanted to use the `useFocusTrap` component from VueUse at first, but I couldn't solve the [Nuxt 2 import issue](https://github.com/vueuse/vueuse/issues/2140), so I vendored in the UseFocusTrap component. VueUse library is using MIT license, so it should be OK. Hopefully, after we move on to Vue 3, we can use the VueUse library instead of vendoring in the implementation.

## Testing Instructions

Run the app with `pnpm dev`, and on mobile screen, open the menu and filters modals. There should be no errors/warnings regarding the initial focus node in the console, and the focus should be handled correctly. When opening the modal, the first tabbable element should be focused. When you close the modal, you should be able to tab away from the modal trigger.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat:
I'm going to close this issue because there's a type mismatch in the render function:
`{ ref: target } as VNodeData` is a quick hack I tried, but this is incorrect. In Vue 2.7, `VNodeData`'s `ref` property is an optional string, and we are assigning the `Ref<HTML>` to it. There's an internal type also called `VNodeData` that does have a `ref` property that can be of the `Ref` type, but I don't know how to import it.

PR #1761, opened 2022-08-30
Fix lint staged to match pnpm:fix order and settings

### LINKED ISSUE #1658: Prettier works differently for directory and single file
## Description

Prettier can be run in two ways.
- `pnpm lint` which runs Prettier against the entire codebase.
- `lint-staged` which runs Prettier selectively against staged files.

Due to some reason (yet to be determined), the behaviour of Prettier is different in both cases, which means files which pass locally may fail the workflow on GitHub and those that pass the workflow on GitHub may fail locally.

## Reproduction

In a fresh install of the codebase
- Run prettier with a single file (e.g. `src/components/VSkipToContentContainer.vue`) argument.
- Run prettier with a directory (e.g. `src/`) as argument.

## Screenshots

Screenshot 2022-08-12 at 5 56 33 PM

## Additional context

This PR has been marked 'critical' as it affects developers trying to commit their code, forcing them to use workarounds.

## Resolution

zackkrida:
## Fixes

Fixes #1658 by @dhruvkb 

## Description

Fixes a discrepancy between `pnpm lint:fix` and our precommit linting script. Now they run eslint and prettier formatting in the same order with the same settings.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,6 +38,7 @@ module.exports = {
   rules: {
     semi: [2, 'never'],
     'no-console': 'off',
+    'prettier/prettier': 'off',
     'vue/max-attributes-per-line': 'off',
     'vue/require-prop-types': 'off',
     'vue/require-default-prop': 'off',
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,8 +1,5 @@
 {
-  "*.{ts,js,vue}": [
-    "eslint --fix --max-warnings=0"
-  ],
   "*.{ts,js,vue,json,yml,yaml,mdx,md}": [
-    "prettier --write"
+    "pnpm lint:fix"
   ]
 }
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,5 +1,8 @@
 {
   "*.{ts,js,vue,json,yml,yaml,mdx,md}": [
-    "pnpm lint:fix"
+    "prettier --write"
+  ],
+  "*.{ts,js,vue}": [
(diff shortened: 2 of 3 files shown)


obulat (commented):
I couldn't resist looking more into this situation :)

Since we are already running `prettier --write`, we do not need `eslint` to check for prettier errors (and un-do the prettier fixes while doing that!). We can actually fix this whole thing if we add `"prettier/prettier": "off"` to the `rules` in `.eslintrc.js`.

Also, I re-arranged the commands in this file locally:
```
{
  "*.{json,yml,yaml,mdx,md}": "prettier --write",
  "*.{ts,js,vue}": ["prettier --write", "eslint --ext .js,.vue,.ts --ignore-path .gitignore --ignore-path .eslintignore --max-warnings=0 ."],
}
```
If two commands are in a bracketed list, they are executed sequentially. The two separate commands in this object are executed at the same time. 

And tried committing the `VSkipToContainer` file with the wrong order of classes:
`class="ms-2 z-50 mt-2 focus:fixed focus:absolute"`.
lintstaged first ran prettier and fixed the class order to the correct `class="z-50 mt-2 ms-2 focus:fixed focus:absolute"`, and then eslint did not undo it. So, husky said that it's an empty commit and refused to commit it :)

zackkrida:
Fantastic, thanks @obulat! I implemented your suggestions.

obulat (approved):
LGTM!

krysal (approved):
So glad the root cause was found! 🥳

PR #1786, opened 2022-09-06
Fix load more not showing when single media type has no results

### LINKED ISSUE #1780: Load more button is not shown on all content view if one media type has 0 results
## Description

On all content view, the load more button isn't shown when one of the types does not have any results, even if the other has more than 1 page.

## Reproduction

1. Go to https://wordpress.org/openverse/search/?q=e-commerce
2. You should see that the page shows 'More than 10,000 results' in the search bar and in the 'See all images' link, and 'No results' in the 'See all audios' link (by the way, it probably should have a different text when there are no audio results?)
3. Scroll down to the bottom of the page.
4. See error: there is no 'Load more' button. If you click on the 'See all images' link, you will see the 'Load more' button.

## Additional context

Marking this issue as high as this is a bug affecting the user experience.

## Resolution

obulat:
## Fixes

Fixes #1780 by @obulat 

## Description

This PR fixes the lack of 'Load more' button when one of the media types did not have any results (and the other did).
It also refactors some components so that the functionality of 'Load more' button (showing when possible, fetching new media when clicked) is contained in the `use-load-more` composable, and not spread out in all the `search` pages and components.

When the media fetching did not return any results, the store would set a `fetchState.fetchingError` to `No <mediaType> found`. This caused the `canFetch` to evaluate to `false` in the `use-load-more`, and therefore the 'Load more' did not show. 
I've removed the setting of `fetchingError` for when we get no results. The lack of results is already handled by the SearchGrid (where we show the ErrorSection if there are no results). 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VImageDetails/VRelatedImages.vue
```diff
@@ -6,7 +6,7 @@
     <VLoadingIcon v-if="fetchState.isFetching" />
     <VImageGrid
       :images="media"
-      :show-load-more="false"
+      :is-single-page="true"
```
obulat, line 9: I couldn't find a better name for this prop: it is necessary for the related images, which never have a load more button.
`show-load-more` is very ambiguous, and it can be confused with `canLoadMore`, so I wanted to rename it.

src/stores/media/index.ts
```diff
@@ -135,24 +135,31 @@ export const useMediaStore = defineStore('media', {
...
-            }
-          }
-          return null
-        }
         const atLeastOne = (property: keyof FetchState) =>
           supportedMediaTypes.some(
             (type) => this.mediaFetchState[type][property]
           )
 
+        const allMediaError = () => {
```
obulat, line 145: This needs improvement for sure! I don't know if we need a `fetchingError` for the All content view, and if so, what shape it should have. For the purposes of Load more, we need `fetchingError` to be null if at least one media type does not have an error.

zackkrida (approved):
This indeed fixes the problem! I tested several scenarios:

- 1 page of images, no audio
- Multiple pages of both
- Multiple pages of images, no audio

LGTM.

obulat:
> This indeed fixes the problem! I tested several scenarios:
> 
>     * 1 page of images, no audio
> 
>     * Multiple pages of both
> 
>     * Multiple pages of images, no audio
> 
> 
> LGTM.

I've found another bug on the currently deployed version: if you are on an All content view, and there are many image results and just one page of the audio results (fewer than 20), when you navigate to the Audio view, you can see the "Load more" button. This is fixed here, too.

dhruvkb:
> I tested several scenarios

@zackkrida what search terms did you use for these test cases?

obulat:
> > I tested several scenarios
> 
> @zackkrida what search terms did you use for these test cases?

@dhruvkb, I first noticed this issue for 'ecommerce' search - it only has images.

dhruvkb (approved):

PR #1802, opened 2022-09-09
Add peaks=true to all audio searches

### LINKED ISSUE #1800: Include `peaks=true` param on audio requests
## Problem

In WordPress/openverse-api#917 we are changing the peaks of the waveform to be optional in API responses, so we need to adjust petitions here to ensure the Frontend keeps using real waveforms (when available) for audio tracks.

## Alternatives

The audio track component will fall back to using randomized deterministic peaks. This is less preferable as it does not give the user exact information on what type of audio can be found before playing it.

## Implementation

zackkrida:
## Fixes

Fixes #1800 by @krysal 

## Description

Wow, I struggled to work with the media services. I cannot wait to refactor some of that code.

Anyway, this PR adds a special case to add `peaks=true` to all audio and related audio searches. To do so I also had to upgrade the related media method to allow for query parameters. Maybe that will be a nice addition in the future.

I'm not sure about tests here. Do we have any to verify the API endpoint paths are correct? Otherwise, this is covered by the fact that the visual regression tests would change if the peak data isn't downloaded. Still, it seems like we need better test coverage around the API endpoints in general.

## Testing Instructions

Check on client-rendered searches that the `peaks=true` is present for all audio API search requests, and on related audio searches. Finally, make sure you see the same peaks on SSR searches.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/data/media-service.ts
+++ b/src/data/media-service.ts
@@ -2,7 +2,7 @@ import { decodeMediaData } from '~/utils/decode-media-data'
 import type { ApiQueryParams } from '~/utils/search-query-transform'
 import type { ApiService } from '~/data/api-service'
 import type { DetailFromMediaType, Media } from '~/models/media'
-import { AUDIO, SupportedMediaType } from '~/constants/media'
+import { AUDIO, type SupportedMediaType } from '~/constants/media'
 
 import type { AxiosResponse } from 'axios'
 
--- a/src/data/media-service.ts
+++ b/src/data/media-service.ts
@@ -2,7 +2,7 @@ import { decodeMediaData } from '~/utils/decode-media-data'
 import type { ApiQueryParams } from '~/utils/search-query-transform'
 import type { ApiService } from '~/data/api-service'
 import type { DetailFromMediaType, Media } from '~/models/media'
-import type { SupportedMediaType } from '~/constants/media'
+import { AUDIO, SupportedMediaType } from '~/constants/media'
 
 import type { AxiosResponse } from 'axios'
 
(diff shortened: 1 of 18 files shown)


zackkrida:
Thanks for the fix, @obulat!

src/data/media-service.ts
```diff
@@ -2,7 +2,7 @@ import { decodeMediaData } from '~/utils/decode-media-data'
 import type { ApiQueryParams } from '~/utils/search-query-transform'
 import type { ApiService } from '~/data/api-service'
 import type { DetailFromMediaType, Media } from '~/models/media'
-import type { SupportedMediaType } from '~/constants/media'
+import { AUDIO, SupportedMediaType } from '~/constants/media'
```
dhruvkb, line 5:
```suggestion
import { AUDIO, type SupportedMediaType } from '~/constants/media'
```

dhruvkb (approved):
LGTM, except for 1 change to meet Vue 3 + TS linting requirements. Not sure why the linting process did not flag that.

src/data/media-service.ts
obulat: I didn't know you can add `type` in front of the named import! Thank you, @dhruvkb!

obulat (approved):
Adds peaks to all requests, SSR and CSR!
And I agree with your remark about the services.

PR #1814, opened 2022-09-15
Create the `VFooter` component

### LINKED ISSUE #1809: Component: VFooter
_All work for the 'Create new header and footer' milestone should be done under the `new_header` feature flag._

## Description

For search pages, we use a footer with links to other content pages. These links can be arranged linearly or, on small devices, in a 2-column grid.

For non-search pages (such as the home page and content pages), the footer does not show links or the Openverse logo as those are present in the header.

## API

`VFooter` does not take any props, nor does it emit events. It uses the [`use-pages` composable](https://github.com/WordPress/openverse-frontend/blob/b91e74ef9a03e87df751cce8e963fa1578002657/src/composables/use-pages.ts) to get the list of pages and the current page.

## Dependencies

- `VLocaleSwitcher` (for changing the site language)

## References

- **Figma for Footer Content:** https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=2960%3A8483
- **Figma for Footer Internal:** https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=2960%3A8290

## Implementation

### LINKED ISSUE #1811: Openverse footer
## Description
Since we plan to move Openverse out from the WordPress.org site and implement the new header (#1669), we need to allow visitors to see the internal pages, mark the search as a WordPress project, and add the language selector in the future.

In that vein, the footer has two contexts:
* **Footer content:** To be used in results and single content details. Internal pages are always present
* **Footer internal:** To be used in content pages where the internal pages move to the header.

The design includes a new component for shifting language. Since this idea has not been discussed nor defined with a milestone, we do not need to create the element but take into account its space.

And finally, I added a new version for smaller devices called `< sm` as `sm` breakpoint starts from `640px` and leaves out other mobile use cases.

### Footer content
imagen

### Footer internal
imagen

---

### Figma file
The footer is already available in the "[Upcoming components: Footer](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=3018%3A7629)" page in the Design Library. This name is due to the pending design documentation. While the uses are in the [New header component](https://www.figma.com/file/KxkZmBNk218V2Fke1GeYCu/New-header-component?node-id=0%3A1) file since header and footer depend on each other.

- **Figma:** [Footer component in Design Library](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=3018%3A7629)
- **Figma:** [Applications in New header component](https://www.figma.com/file/KxkZmBNk218V2Fke1GeYCu/New-header-component?node-id=424%3A56729)

## Implementation

- [ ] 🙋 I would be interested in implementing this component.

dhruvkb:
## Fixes

Fixes #1809 by @dhruvkb 
Fixes #1811 by @panchovm 

## Description

This PR adds the new footer under the `new_header` feature flag. It is built in a way that it does not impact any existing functionality or look of the site, although that does lead to a bit of redundant code in the short-term. At the time of writing, it uses a blank `<div>` for the locale switcher.

### On the site

0. Checkout the PR and run the dev server.
1. Enable the `new_header` flag by navigating to `/preferences`.
2. See the footer.

### In Storybook

0. Checkout the PR and run the Storybook.
1. See the component (both versions) under "Components/VFooter".

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VFooter/VFooter.vue
```diff
@@ -0,0 +1,96 @@
...
+        </ul>
+      </nav>
+    </div>
+
+    <!-- Locale chooser and WordPress affiliation graphic -->
+    <div class="flex flex-col justify-between gap-10 sm:flex-row xl:flex-grow">
+      <div
+        class="flex h-10 w-full w-full items-center justify-center border border-dark-charcoal-20 sm:w-50"
+      >
+        {{ $t('🚧') }}
```
obulat, line 28: Love this 😆

src/components/VFooter/VFooter.vue
```diff
@@ -0,0 +1,96 @@
+<template>
+  <footer class="flex flex-col gap-10 px-6 py-10 lg:px-10 xl:flex-row xl:gap-8">
+    <!-- Logo and links -->
+    <div
+      v-if="isContent"
+      class="flex flex-col gap-10 md:flex-row md:items-center md:justify-between xl:flex-grow"
+    >
+      <VBrand class="text-[18px]" />
```
obulat, line 8: The Brand here should probably be a home page link.

dhruvkb: Right! Thanks, I'll add that.

obulat:
Leaving a note here so we don't forget: we will need to update the focus management when we add the footer. Currently, when you press `tab` from the External sources form, you are taken to `Scroll to top` button. We'll need to allow tabbing to the footer after it is added.

obulat (commented):
I'm not sure if we can conditionally fix it based on the flag value, but we need to fix the footer to the bottom of the page. Now, for short content pages, it's not at the bottom:
Screen Shot 2022-09-15 at 3 29 20 PM

src/assets/brand.svg
```diff
@@ -1,4 +1,4 @@
-<svg width="215" height="42" viewBox="0 0 215 42" fill="none" xmlns="http://www.w3.org/2000/svg">
+<svg viewBox="0 0 215 42" fill="none" xmlns="http://www.w3.org/2000/svg">
```
obulat, line 1: What is the reason for removing the width and height? It's usually a good idea to set them, but override with CSS, to avoid the 'flash of unstyled SVG' when, before the CSS is loaded, SVGs are rendered as huge images (I've learned about it from this article: https://www.sarasoueidan.com/blog/svg-style-inheritance-and-fousvg/)

dhruvkb: Having the width and height set in the SVG was preventing us from scaling the SVG up/down. But you're right, I'll see what we can do to prevent the flashing.

src/components/VLink.vue
```diff
@@ -1,7 +1,17 @@
 <template>
-  <Component :is="linkComponent" v-bind="linkProperties" v-on="$listeners"
-    ><slot
-  /></Component>
+  <Component
```
obulat, line 2: I'm not sure if it's a good idea to tightly couple the link with the external link icon. This adds styles that might be hard to override in other components. Wouldn't wrapping an external link in another component be a better way of implementing this?

dhruvkb: My idea behind this was that `VLink` should be the be-all-end-all component for links. Any feature you could want from a link component is supported by it.

In the current implementation it doesn't introduce any style conflicts (_at least not yet_). Do you think it's a good idea to have such a versatile component?

src/assets/brand.svg
dhruvkb: Ok, so it seems I was wrong, I was just missing the `.w-auto` class in the code where the SVGs were being used. Fixed in 724a8ce.

zackkrida:
Looking great. We should link "Powered by WordPress" to the wp.org homepage.

dhruvkb:
Getting VRT to pass was a pain, but we're finally there.

src/components/VFooter/VFooter.vue
```diff
@@ -0,0 +1,100 @@
...
+    mode: {
+      type: String as PropType<'internal' | 'content' | undefined>,
+      required: false,
+      default: undefined,
+    },
+  },
+  setup(props) {
+    const { all: allPages, current: currentPage } = usePages(true)
+
+    const isContent = computed(() => {
```
zackkrida, line 84: Nit: Could this be `isContentMode`, something a bit more specific to the related prop?

zackkrida (approved):
Nice work, and quick. A few nitpick comments but otherwise lgtm. You got the breakpoints from Figma perfect 👍

zackkrida (changes requested):
@dhruvkb I did notice two small things:

1. the mockups indicate that the footer is meant to have a 1px Shades/Dark Charcoal 20 (#D6D4D5) border when on a white bg.
2. The footer should appear on the 404 page using the internal style, but with no border on top

src/components/VFooter/VFooter.vue
dhruvkb: Good point. Updated in 773114f.

zackkrida:
@dhruvkb sorry, one more small thing. The backtotop button obscures the wordpress logo in the footer. I played with some inline styles on desktop and really liked the result of this on the button:

```
  bottom: 20px;
  right: 41px;
  position: sticky;
  margin-left: auto;
  margin-bottom: 20px;
```

The `bottom` number is how far the button will be from the bottom of the viewport while scrolling, and conveniently `sticky` will keep the button in its container above the footer. So it looks like this while scrolling:

CleanShot 2022-09-15 at 19 56 16@2x

and _this_ when one reaches the bottom of the page:

CleanShot 2022-09-15 at 19 56 12@2x

zackkrida (approved):
LGTM! Some small non-blocking nitpicks:

- I think we should remove the `mb-16` on the Meta Search section now that the sticky back to top button occupies some space below it
- vue-18n doesn't like the construction icon, and gives a `[vue-i18n] Value of key '🚧' is not a string or function !` console warning.
- The hover text underline on the 'powered by wp' looks a bit odd, and to balance it with the hover effect of the openverse logo on the left i think we should just remove it.

dhruvkb:
Bumping priority to high as it blocks #1821.

src/components/VFooter/VFooter.vue
```diff
@@ -0,0 +1,103 @@
...
+        href="https://wordpress.org"
+        class="text-dark-charcoal hover:no-underline"
+      >
+        <i18n
+          tag="div"
+          path="footer.wordpress-affiliation"
+          class="flex h-full flex-row items-center gap-2 text-sm"
+        >
+          <template #wordpress>
+            <WordPress class="aria-hidden" />
```
obulat, line 42: This SVG has a non-`currentColor` fill color, which makes it difficult to change color using CSS. I've replaced it with `currentColor` in the `VHeaderInternal` PR.

src/components/VFooter/VFooter.vue
```diff
@@ -0,0 +1,103 @@
...
+    WordPress,
+  },
+  props: {
+    /**
+     * whether the footer is being rendered on a content page or an internal
+     * page; This determines whether the Openverse logo and other links are
+     * displayed.
+     */
+    mode: {
+      type: String as PropType<'internal' | 'content' | undefined>,
```
obulat, line 79: When would the mode be undefined? Would it be safer to have only two variants?

src/layouts/default.vue
```diff
@@ -1,14 +1,15 @@
 <template>
-  <div class="app relative grid">
+  <div class="app flex min-h-screen flex-col">
```
obulat, line 2: Would it be better to not touch the `default` wrapper styles until we finish header/footer implementations and enable them? I'm afraid this might break the iframe version inside wordpress.org (I haven't checked, though).

src/layouts/default.vue
```diff
@@ -20,6 +21,11 @@
         :class="{ 'border-dark-charcoal-20 border-s': isSidebarVisible }"
       />
     </main>
+
+    <VFooter
```
obulat, line 25: This will need to add the sidebar padding as the `main` does, otherwise the sidebar covers a part of the footer:

Screen Shot 2022-09-19 at 9 05 45 PM

src/components/VFooter/VFooter.vue
```diff
@@ -0,0 +1,103 @@
...
+      <VLink
+        href="https://wordpress.org"
+        class="text-dark-charcoal hover:no-underline"
+      >
+        <i18n
+          tag="div"
+          path="footer.wordpress-affiliation"
+          class="flex h-full flex-row items-center gap-2 text-sm"
+        >
+          <template #wordpress>
```
obulat, line 41: For some reason, on RTL page this is not aligned vertically. Now, in an `/ar/search` page, 'powered by' is above the center line:
Screen Shot 2022-09-19 at 9 06 54 PM

src/layouts/default.vue
obulat: I think this breaks the sidebar layout for the old header version of the site. If you turn off the `new_header` flag, the sidebar hides a part of the page:
Screen Shot 2022-09-19 at 9 11 56 PM

obulat (changes requested):
The footer looks great, and I love the `VBrand` abstraction!
My main issues are with the changes in layout without using flags: I believe the current changes to `layouts/default` breaks the `old_header` layout. I've left several screenshots inline. 
Would adding some layout classes based on the `newHeaderEnabled` prop and using them to adjust the layout make sense here, too? Like I did on line 15 in `layouts/default`?

src/components/VFooter/VFooter.vue
dhruvkb: Using `undefined` means it can infer the type from the current page.

src/layouts/default.vue
dhruvkb: If the footer is shrunk in width to accommodate the sidebar, the responsiveness of the footer will break down. It will be using a wider style (based on the viewport width) in a narrower area and will look pretty bad.

src/layouts/default.vue
zackkrida: Oh, I see, it's on a very specific breakpoint. Nice find @obulat !

src/layouts/default.vue
zackkrida: @dhruvkb I what Olga is suggesting might be the only viable solution. I don't think we can acceptably cover the footer with the sidebar. I tried some approaches: to make the sidebar height shorter so it ends before the footer; and to give the footer a higher z-index and add padding to the bottom of the filter sidebar, but neither worked well.

zackkrida: > It will be using a wider style (based on the viewport width) in a narrower area

I think we'll have to take a JS-based approach to setting the right footer style at the right time, sadly.

src/components/VFooter/VFooter.vue
dhruvkb: It's aligned correctly for me:
image

I believe it just looks odd because in LTR we see the huge (W) before the text so our perception of centering is thrown off which is not possible when reading RTL. It shouldn't be concerning because if the logo is placed alongside non-English text, the slight baseline mis-alignment will not be apparent.

src/layouts/default.vue
dhruvkb: That's how it behaves in prod. I haven't changed it.

Screenshot 2022-09-20 at 7 08 09 AM

src/components/VFooter/VFooter.vue
dhruvkb: Fixed in f35e8b2.

src/components/VFooter/VFooter.vue
dhruvkb: Fixed in a58435c.

src/layouts/default.vue
dhruvkb: Implemented in 93339e9.

src/layouts/default.vue
zackkrida: @dhruvkb and @obulat I'm logging off for the night but it looks like this regression at this breakpoint was introduced in https://github.com/WordPress/openverse-frontend/pull/1804. Could we get a critical issue and pr fix created? It must just be an issue of using the wrong media query in one place, likely a small problem.

dhruvkb: @obulat this is the same behaviour as prod and staging.

Screenshot 2022-09-20 at 7 27 20 AM
Screenshot 2022-09-20 at 7 27 46 AM

dhruvkb:
The footer now shrinks when the filters sidebar is opened, and will use a version from a lower breakpoint if the width is insufficient.
image

obulat (approved):

PR #1821, opened 2022-09-16
Create locale chooser `VLanguage` and place in the footer

### LINKED ISSUE #1810: Restore the locale switcher and update as per the new designs
## Description

The locale switcher was [removed from the codebase](https://github.com/WordPress/openverse-frontend/commit/7eac75b0) when the Openverse site was rendered within wordpress.org using an `iframe`. This was because we were passing the i18n preferences of the outer site into the `iframe`. 

With the `iframe` being removed, we need to restore the deleted functionality of managing i18n on our own.

- Restore the component from the history.
- Rename it to `VLanguage` for consistency with Figma names.
- Implement locale-changing functionality.

## References

- **Figma:**  https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=2960%3A8532

## Screenshots

There are subject to change and should not be used as reference. See the Figma link above.

Language

## Alternatives

Instead of restoring the component, one might build it again afresh but I wouldn't recommend that. The old one was perfectly fine and usable with a few design tweaks.

## Implementation

dhruvkb:
## Fixes

Fixes #1810 by @dhruvkb

## Description

This PR adds a functional locale chooser to the codebase. It also creates a base `VSelectField` component (presentational) upon which the language chooser is built.

## Testing Instructions

0. Check out the PR.
1. Update all locales: `pnpm i18n:update-locales` to store their native names in Vue i18n.
2. Run the Storybook.
3. See the stories for
   - 'Components/VSelectField' 
   - 'Components/VLanguage'
   - 'Components/VFooter'
4. Run the dev server with the `new_header` feature flag turned on.
5. See the footer with the locale chooser at the bottom.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (commented):
This looks great on initial review. Some early thoughts:

- I notice that the globe icon isn't click or touch-able and prevents the select from opening. It might just need `pointer-events: none` so that interactions pass through to the parent.
- This will be a question for @panchovm but with these two selects being so close to each other their design differences are more stark:
	- hover states are different
	- dropdown arrow color is different

The behavior is perfect and the borderless 404 version looks good too.

src/components/VLanguage/VLanguage.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+  <VSelectField
+    v-bind="$attrs"
```
obulat, line 3: Is this a leftover of some previous changes? if we remove `inheritAttrs: false`, the $attrs will be bound automatically?

sarayourfriend (changes requested):
Still in the process of reviewing but there's at least one small a11y issue with keyboard support. You can read the expectations here: https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html#kbd_label

For some reason only space can open the select (not enter) but only enter can select an option (and not space). Both space and enter should be able to do the same things here. I tested other select dropdowns and they were able to do it so it is something in this implementation (I have not been able to identify what yet).

Can we confirm that someone is testing this with a screen reader as well? I can test with NVDA and JAWS but I need to get my Windows VM set up again so it might not be until next week after the holidays here when I am able to do so.

src/components/VFooter/VFooter.vue
```diff
@@ -90,6 +88,10 @@ export default defineComponent({
       required: false,
       default: undefined,
     },
+    languageProps: {
+      type: Object,
```
sarayourfriend, line 92: Can you add the proper type annotation to this? `Object` will just resolve this prop's type to `any` in TS. You can see an example for how this is done in `VIconButton`.

src/components/VLanguage/VLanguage.vue
```diff
@@ -0,0 +1,70 @@
...
+  name: string
+  nativeName?: string
+}
+
+/**
+ * Presents a way for the users to change the app locale and use a translated
+ * version of the app.
+ */
+export default defineComponent({
+  name: 'VLanguage',
```
sarayourfriend, line 44: Can we add `Select` to the end of this name to clarify the purpose of the component?

src/components/VLanguage/VLanguage.vue
```diff
@@ -0,0 +1,70 @@
...
+import globeIcon from '~/assets/icons/globe.svg'
+
+/**
+ * We put a little extra information in the Vue-i18n `locales` field such as the
+ * locale's name and native name, which comes in use here.
+ */
+interface OpenverseLocale extends LocaleObject {
+  name: string
+  nativeName?: string
+}
```
sarayourfriend, line 37: This seems like it should go in `typings` so that it is automatically available to all usages of `i18n` library in Openverse without having to duplicate this or do casts. Did you look into whether it was possible to do that?

sarayourfriend (dismissed):
Can you add snapshot and e2e tests for these components?

Other than the keyboard a11y issue I shared above and the options sorting this LGTM. Nice work so far :rocket:

src/components/VLanguage/VLanguage.vue
dhruvkb: I used the exact name that is being used to refer to this component in Figma. The generic version of the component is named `VSelectField` to signify it's purpose.

dhruvkb:
@sarayourfriend regarding the behaviour of <kbd>Space</kbd> and <kbd>Enter</kbd>, I have observed that across the macOS operating systems UI, <kbd>Enter</kbd> never opens the dropdown menu, it's only dropped by pressing <kbd>Space</kbd>. The part about pressing <kbd>Space</kbd> to select a choice when open is true though, good catch!

src/components/VLanguage/VLanguage.vue
dhruvkb: Yeah, it's broken.
Screenshot 2022-09-21 at 9 51 58 AM

src/components/VFooter/VFooter.vue
dhruvkb: Thanks for the reference! Implemented in 134800f.

sarayourfriend:
> @sarayourfriend regarding the behaviour of Space and Enter, I have observed that across the macOS operating systems UI, Enter never opens the dropdown menu, it's only dropped by pressing Space. The part about pressing Space to select a choice when open is true though, good catch!

I'm on Linux and Firefox and as I said, it does work on other dropdowns. If you need me to take a look at it to ensure that it behaves as expected on platforms other than macOS then let me know and I am happy to do so next week.

sarayourfriend:
> I sorted the list by the codes of the locales, which sorts languages by [the order they appear on Wikipedia](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) and puts variants of the same language next to each other.

Nice!

dhruvkb:
Comparing against the [`<select>` example on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select), the field behaves exactly in the same way, which means that at least in Firefox on macOS, you cannot use <kbd>Enter</kbd> to open the dropdown. @sarayourfriend could you compare the behaviour on Linux to see if they differ?

sarayourfriend:
It's strange, Firefox behaves as you said, space to open and enter to select (which is not the accessible recommendation and therefore I think Firefox is not following the ARIA standards). Chromium also does not follow the accessible recommendation. I can't remember where I saw that a standard select was behaving like the ARIA recommendation, but you're right, a plain vanilla one does not match the spec.

I wonder why these browser vendors do not implement the ARIA standard.

Do we hold ourselves to the same standard? I'd like to think we could match the ARIA recommendations. In this case it doesn't even seem that hard to do, the only discrepancy I found was between space and enter. If we don't think this is worth our time right now then we should at least make a ticket to address this in the future. It'd be a straightforward ticket for someone to work on with a nice win to it.

dhruvkb:
@sarayourfriend the `<select>` HTML element is notoriously difficult to tweak. It turns out that adding custom keyboard event logic might not be very straightforward. I don't want the PR to be blocked (and act as a blocker for pending work) so I'll make an issue to handle this in post.

src/components/VLanguage/VLanguage.vue
dhruvkb: Fixed in e8a180c.

dhruvkb:
@sarayourfriend would it be okay to push merge in without the snapshot tests? To write tests for these components, we would first need to implement #1837.

src/components/VSelectField/VSelectField.vue
```diff
@@ -0,0 +1,134 @@
...
+    <div class="pointer-events-none absolute inset-y-0 my-auto h-fit start-2">
+      <slot name="start" />
+    </div>
+    <div class="pointer-events-none absolute inset-y-0 my-auto h-fit end-2">
+      <VIcon :icon-path="caretDownIcon" />
+    </div>
+    <select
+      :id="fieldId"
+      v-model="modelMedium"
+      class="flex h-[calc(theme(spacing.10)_-_2_*_theme(borderWidth.DEFAULT))] w-full appearance-none bg-white pe-10"
```
zackkrida, line 15:
```suggestion
      class="flex h-[calc(theme(spacing.10)_-_2_*_theme(borderWidth.DEFAULT))] w-full appearance-none bg-white pe-10 truncate"
```

I would recommend we truncate for improved styling for long languages:

|before|after|
|-|-|
|CleanShot 2022-09-21 at 20 37 21@2x|CleanShot 2022-09-21 at 20 37 09@2x|

src/components/VSelectField/VSelectField.vue
```diff
@@ -0,0 +1,134 @@
...
+      default: 'regular',
+    },
+  },
+  // using non-native event name to ensure the two are not mixed
+  emits: {
+    'update:modelValue': defineEvent<[string]>(),
+  },
+  setup(props, { emit, attrs, slots }) {
+    const fieldName = computed(() => attrs['name'] ?? props.fieldId)
+    const modelMedium = computed<string>({
```
zackkrida, line 113: I'm not following what 'medium' refers to in this name. Probably my own ignorance!

src/components/VLanguage/VLanguage.vue
zackkrida: I prefer the explicitness of `VLanguageSelect` too; if we want parity with figma it might make sense to rename in Figma _or_ to link to the [Figma component definition](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=2960%3A8532) in the comments for this component.

src/components/VFooter/VFooter.vue
```diff
@@ -134,7 +138,7 @@ export default defineComponent({
 }
 
 .footer.footer-sm .language {
-  @apply w-50;
+  @apply max-w-[12.5rem];
 }
```
zackkrida, line 142: Is this nested + named class necessary for specificity purposes? If so could you add a comment? Otherwise, I wonder why @apply is being used for a single class.

zackkrida (approved):
I'm fine with getting this in @dhruvkb provided you make a high-priority issue for the necessary visual regression tests.

And @sarayourfriend while I understand your desire to follow the ARIA spec perfectly, I would be very concerned with trying to change native select behavior in our application. Having our native selects behave differently than others in Firefox seems to be a new accessibility problem I'd rather not introduce. I'd be happy to move this to a full custom combobox (searching for languages would likely be useful, too) in the future.

I left some small suggestions but otherwise LGTM.

src/components/VSelectField/VSelectField.vue
dhruvkb: Medium here being "a means by which something is communicated, **_transmitted_** or expressed". This field is acting as a channel for the `v-model` of the component to be used as the `v-model` on the `<select>`.

src/components/VLanguage/VLanguage.vue
dhruvkb: Renamed in 9a992f4c.

dhruvkb:
Filed an issue to write tests when unblocked: https://github.com/WordPress/openverse-frontend/issues/1843

src/components/VFooter/VFooter.vue
dhruvkb: This nested class is needed because it only needs to apply at the `sm` breakpoint on the `.footer` component (whose responsiveness is currently being managed by JS). It's also done so that we can be consistent with the other classes in the same style block.

dhruvkb:
Just for some added confidence, I've added component level snapshots for both the `VSelectField` and the `VLanguageSelect` components from the Storybook.

src/components/VLanguageSelect/VLanguageSelect.vue
```diff
@@ -0,0 +1,63 @@
...
+      get: () => i18n.locale,
+      set: (value) => {
+        i18n.setLocale(value)
+      },
+    })
+    const choices = computed<Choice[]>(() =>
+      i18n.locales
+        .map((locale: LocaleObject) => ({
+          key: locale.code,
+          text: locale.nativeName ?? locale.name,
```
obulat, line 50: It would be better to update the locales json file to add `nativeName` there, and not have to calculate it here. We could create a separate issue for that.

src/components/VSelectField/VSelectField.vue
```diff
@@ -0,0 +1,134 @@
...
+<script lang="ts">
+import { defineComponent, computed, PropType } from '@nuxtjs/composition-api'
+
+import { defineEvent } from '~/types/emits'
+
+import VIcon from '~/components/VIcon/VIcon.vue'
+
+import { ProperlyExtractPropTypes } from '../../types/prop-extraction'
+
+import caretDownIcon from '~/assets/icons/caret-down.svg'
```
obulat, line 39:
```suggestion
import { defineEvent } from '~/types/emits'
import { ProperlyExtractPropTypes } from '~/types/prop-extraction'

import VIcon from '~/components/VIcon/VIcon.vue'

import caretDownIcon from '~/assets/icons/caret-down.svg'
```

obulat: Usually, Webstorm adds imports as `../../`, but the standard way to import in the repo is using `~` alias. When you change `../..` to `~`, the linter starts complaining about the imports order.

src/locales/scripts/get-validated-locales.js
```diff
@@ -24,6 +24,7 @@ const getValidatedLocales = async () => {
   const allLocales = Object.values(updatedLocaleList).map((locale) => ({
     code: locale.slug,
     name: locale.name,
+    nativeName: locale.nativeName,
```
obulat, line 27:
```suggestion
    nativeName: locale.nativeName || locale.name,
```

obulat: This would probably fix the issue I mentioned in the other comment: all locales will have the `nativeName` (or a fallback), and we will not have to calculate it at runtime.

typings/nuxt__types/index.d.ts
```diff
@@ -9,3 +10,14 @@ declare module '@nuxt/types' {
...
 }
+
+declare module '@nuxtjs/i18n' {
+  /**
+   * We put a little extra information in the Vue-i18n `locales` field such as the
+   * locale's name and native name, which comes in use here.
+   */
+  export interface LocaleObject {
+    name: string
+    nativeName?: string
```
obulat, line 21:
```suggestion
    nativeName: string
```

obulat: Can we make this a required property?

obulat (commented):
VoiceOver does not pronounce the names of languages that don't use the Latin alphabet. So, VoiceOver users who speak those languages will not be able to choose their language. I wonder if we should add the English names to all languages so they can at least hear that and choose based on it?

dhruvkb:
I generally feel wrong to add the English name as the spoken bit because the name of the language is in many cases different from the English name. For example Arabic in Arabic is _al-ʿarabiyyah_. 

It might be the same in many cases, but not always. So that could a usability problem for those people.

src/components/VSelectField/VSelectField.vue
dhruvkb: Is there a way for linting to catch this? WebStorm keeps using the relative import style and there doesn't seem to be any preference option to change this.

typings/nuxt__types/index.d.ts
dhruvkb: Implemented in dda4d2f.

src/locales/scripts/get-validated-locales.js
dhruvkb: Implemented in dda4d2f.

src/components/VLanguageSelect/VLanguageSelect.vue
dhruvkb: Implemented in dda4d2f.

src/components/VSelectField/VSelectField.vue
dhruvkb: Fixed in 205bfe5.

sarayourfriend:
I won't block on writing the tests, though I don't know how I feel about merging new components that aren't thoroughly tested. Why can't the operational tests can't be written against the Storybook version or even as regular unit tests which would probably be even better and run faster?

I'm running MSR this week so I may not be able to review this in a timely manner but I'll try to review it in the next couple days.

obulat:
> I generally feel wrong to add the English name as the spoken bit because the name of the language is in many cases different from the English name. For example Arabic in Arabic is _al-ʿarabiyyah_.
> 
> It might be the same in many cases, but not always. So that could a usability problem for those people.

I understand that the names in English and native names are different, and would normally prefer the native name. However, when the difference is between _silent pause_ when the screen reader navigates to the option, or the _English name_, I would prefer the latter. Without English, the VO doesn't announce the choice **at all.**

dhruvkb:
@sarayourfriend I already added the Storybook tests. @obulat let me investigate why VO is silent instead for non-English languages.

dhruvkb:
I recorded this demo.
 
https://user-images.githubusercontent.com/16580576/192231406-18b6c365-44de-44b7-8f0e-251d1b4ae4c9.mov

zackkrida:
Mine in fact does a third behavior, the voiceover reads the letter names of each language. For example, for Arabic it reads: "Arabic letter {letter}", "Arabic letter {letter}" repeatedly.

I did look at [Microsoft's language chooser page](https://www.microsoft.com/en-us/language/Locale) however, and confirm the same behavior. Just one example but it seems consistent.

obulat (approved):
So happy to have the language chooser in Openverse!
Should we open an issue for investigating the Screen reader behavior on languages with different scripts?

PR #1823, opened 2022-09-16
Reduce api token logging

zackkrida:
## Description

We've had a pretty large increase to our log storage costs in production. This PR tries to remove some of the logs from the API token process. I'm opening to keeping a subset of these or whatever the minimum amount of information we need to debug problems with this is.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/plugins/api-token.server.ts
+++ b/src/plugins/api-token.server.ts
@@ -1,7 +1,7 @@
 import { Mutex, MutexInterface } from 'async-mutex'
 
 import { createApiService } from '~/data/api-service'
-import { error } from '~/utils/console'
+import { error, log } from '~/utils/console'
 
 import type { AxiosError } from 'axios'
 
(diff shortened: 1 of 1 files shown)


dhruvkb (commented):
Given that the implementation currently works, I'm ok with removing the logging.

But would it be a better solution to make the `log` function a no-op on the production server-side? That allows us to keep the detailed logging in dev untouched, just in case.

https://github.com/WordPress/openverse-frontend/blob/bb12c58bae7d573e17bf9b622edb930ae54306ec/src/utils/console.ts#L3-L8

zackkrida:
@dhruvkb I'm not sure; they also clutter local development considerably too. Perhaps I should comment them out rather than delete for easy restoration if there's ever issues with the token implementation?

dhruvkb (approved):
Better to delete them if they aren't useful in dev either.

src/plugins/api-token.server.ts
```diff
@@ -176,12 +161,6 @@ const apiToken: Plugin = async (context, inject) => {
   let openverseApiToken: string | undefined
   try {
     openverseApiToken = await getApiAccessToken(context)
```
obulat, line 163: Maybe we should leave these lines so that it's clear if we are using the token when fetching or not?

obulat (approved):
LGTM! This has been making the dev output almost unusable, but we might want to add at least some indication of whether the token is added to requests or not.

src/plugins/api-token.server.ts
zackkrida: I'm not sure. We're already sending the `error('Unable to retrieve API token, clearing existing token', e)` when things fail, which i think would be enough information to debug this token code in the event of a problem.

sarayourfriend (approved):
LGTM!

PR #1825, opened 2022-09-19
Add VHeaderInternal

### LINKED ISSUE #1806: Enable closing the dialogs from the parent components
## Problem

We use the `use-dialog-content` composable for accessible dialogs (modals and popovers). The composable exposes `hide` (or `close`) ref to the template that allows for closing the modal from inside the template. However, this is only available as the `template`'s slot prop, and not available inside the `script`. 
We need to be able to close the modal from the `<script>`, too. One such example, which I just realized is broken on mobile now, is the selection of the content type inside the mobile modal content switcher. The `@select` event handler needs to pass the selected item as the emit argument, so we cannot add `;hide()` to the `@select` in the template, and `hide`/`close` modal is not available in the `<script>` so that we could call it in the `select` event handler there:

https://github.com/WordPress/openverse-frontend/blob/9b53876d095d7900375a2a6ad40eb934ab565711/src/components/VContentSwitcherOld/VMobileMenuModal.vue#L21-L27

https://github.com/WordPress/openverse-frontend/blob/9b53876d095d7900375a2a6ad40eb934ab565711/src/components/VHeader/VHeaderMenu.vue#L46-L48
On these lines, `closeMenu` doesn't work because we do not have such functionality exposed to the `<script>`.

## Description

We can have a 'controlled prop': in addition to the component handling its visibility, it can also have an external `show`/`visible` prop which can be adjusted if the parent component receives a `close` event.

## Additional context

HeadlessUI - vue has a similar implementation:
https://headlessui.com/vue/dialog

## Implementation

### LINKED ISSUE #1807: Component: VHeaderInternal
_All work for the 'Create new header' milestone should be done under the `new_header` feature flag._

## Description

For content pages and the home page, we use a header with links to other content pages. Under the `md` breakpoint, uses a hamburger button, which opens a full-screen modal with the links.

## API

`VInternalHeader` should use the [`use-pages` composable](https://github.com/WordPress/openverse-frontend/blob/b91e74ef9a03e87df751cce8e963fa1578002657/src/composables/use-pages.ts) that contains the list of pages, whether their links are external, and whether the link is active.

### State

| Name | Type | Description |
|-|-|-|
| pages | object | return type of `usePages` composable |
| isModalVisible | Boolean | The boolean flag that tracks if the mobile modal menu is open |

## Code samples

```js
<header>
<VLogoButton />
<nav class="hidden md:flex">
  <VLink v-for="page in pages" />
</nav>
<VModal class="flex md:hidden">
  <template #trigger>
    <VIconButton icon-path="hamburger" />
  </template>
  <VPageList pages="pages" />
</VModal>
</header>
```

## Dependencies

`VIconButton` (hamburger icon for screens below `md`)
This component will need to make modal stylable.

## References

The desktop version will basically be a copy of the footer from #1814.

- **Figma:** https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=2960%3A8548
- mobile modal https://www.figma.com/file/KxkZmBNk218V2Fke1GeYCu/New-header-component?node-id=566%3A69829
- desktop version of the header: https://www.figma.com/file/KxkZmBNk218V2Fke1GeYCu/New-header-component?node-id=560%3A63351
- 
## Implementation

- [ ] 🙋 I would be interested in implementing this component.

obulat:
## Fixes

Fixes #1807 by @obulat
Fixes #1806 by obulat

## Description

This PR creates a Content header with the links to internal pages. I copied a large part of the implementation from the VFooter PR (#1814).

For this PR, I had to also add the following changes to the `VModal` component:
- Make the VModal controllable from outside by adding the `externalVisible` prop which is used in addition to the `visible` ref inside the component to manage the modal visibility.
- expose the `top-bar` slot from `VModalContent` to the `VModal` consumers.
- add `mode` (light/dark) and `variant` (default/full) props to make the modal content stylable.

## Testing Instructions

1. Run `pnpm dev`.
2. Go to localhost:8443/preferences and check the `new_header` flag
3. Go to localhost:8443/about
4. You should see the new Content header. Check that it works as expected on all breakpoints and matches the [Figma mockups](https://www.figma.com/file/KxkZmBNk218V2Fke1GeYCu/New-header-component?node-id=609%3A70088).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/composables/use-pages.ts
```diff
@@ -47,7 +47,7 @@ export default function usePages() {
   ]
 
   const route = useRoute()
-  const currentPageId = computed(() => route.value.name)
+  const currentPageId = computed(() => route.value?.name?.split('__')[0])
```
obulat, line 50: Originally, this returned the i18n name like `about__en`, which would not match the ids in the `pages` array.

sarayourfriend:
@obulat Is the name of this PR accurate still? There's no new component called VHeaderInternal in this PR :thinking:

obulat:
> @obulat Is the name of this PR accurate still? There's no new component called VHeaderInternal in this PR :thinking: 

It was the name of the component that was inaccurate, @sarayourfriend. I realized it after rereading the issue, and updated the name of the component.

src/components/VBrand/VBrand.vue
```diff
@@ -0,0 +1,44 @@
...
+  },
+  props: {
+    /**
+     * the screen-reader text to put inside the component; Can be set to a blank
+     * string to not read anything at all.
+     */
+    srText: {
+      type: String,
+      default: 'Openverse',
+    },
```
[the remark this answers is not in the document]
obulat: This was copied from the VFooter PR, which was merged in after I started working on this issue.

src/components/VModal/VModal.vue
```diff
@@ -104,6 +110,22 @@ export default defineComponent({
...
+    mode: {
+      type: /** @type {import('@nuxtjs/composition-api').PropType<'dark' | 'light'>} */ (
+        String
+      ),
+      default: 'light',
+    },
+    externalVisible: {
+      type: Boolean,
+      default: undefined,
+    },
```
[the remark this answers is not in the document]
obulat: This is an excellent suggestion, thank you!

src/components/VBrand/VBrand.vue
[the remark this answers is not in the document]
sarayourfriend: Which part was copied? And why does it appear in this PR if it's merged on main? Does this PR need to be rebased then?

src/components/VModal/VModal.vue
```diff
@@ -132,6 +157,15 @@ export default defineComponent({
...
       triggerA11yProps['aria-expanded'] = !!visible
     })
+    watch(externalVisibleRef, (visible) => {
+      if (visible === undefined) return
+      visibleRef.value = visible
+      if (visible) {
+        lock()
+      } else {
+        unlock()
+      }
```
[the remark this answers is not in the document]
obulat: > Were you trying to intentionally avoid emitting the event, by chance?

I was, and I was wrong. Thank you for pointing this out!

src/components/VHeader/VHeaderInternal.vue
```diff
@@ -0,0 +1,137 @@
...
+        </template>
+
+        <template #top-bar>
+          <div
+            class="flex h-20 w-full justify-between bg-dark-charcoal py-4 text-white pe-3 ps-6"
+          >
+            <VBrand :is-fetching="false" />
+            <VIconButton
+              ref="closeButton"
+              variant="plain"
```
sarayourfriend, line 45: This does not appear to be an actual prop on `VIconButton`. Is this serving a different function here?

```suggestion
```

sarayourfriend (dismissed):
~This renders with the links vertically misaligned compared to the brand for me:~
image

Edit: I was on an old version of the branch, sorry. The latest version of the branch does not have misaligned links.

Can you add snapshot tests as well for each breakpoint, either of the about page itself using the query param toggle for the feature flag or in a storybook playwright test?

I am still reviewing the code a little more closely.

sarayourfriend:
A behaviour I just noticed is that the modal stays open if you resize the window back to a breakpoint bigger than when the modal would be able to be opened. I think this is better than disappearing the modal based on the breakpoint, but I wanted to make sure that behaviour is intentional. We may have to wait to ask Francisco once he is available though.

src/components/VHeader/VHeaderInternal.vue
```diff
@@ -0,0 +1,137 @@
...
+import VBrand from '~/components/VBrand/VBrand.vue'
+import VLink from '~/components/VLink.vue'
+import VModal from '~/components/VModal/VModal.vue'
+
+import closeIcon from '~/assets/icons/close.svg'
+import menuIcon from '~/assets/icons/menu.svg'
+import WordPress from '~/assets/wordpress.svg?inline'
+
+export default defineComponent({
+  name: 'VHeaderInternal',
```
sarayourfriend, line 103: I don't want to bike shed too much on the name of this component, but I don't really understand what "Internal" means here. Is it only meant for special, non-public pages or something?

src/components/VHeader/VHeaderInternal.vue
```diff
@@ -0,0 +1,137 @@
...
+    const setModalVisibility = (visibility: boolean) => {
+      isModalVisible.value = visibility
+    }
+
+    // When clicking on an internal link in the modal, close the modal
+    watch(route, () => {
+      if (isModalVisible.value) {
+        setModalVisibility(false)
+      }
+    })
```
sarayourfriend, line 121: The modal closes when you click an external link as well. Is that intentional?

The modal also causes the page to scroll to the top now instead of staying at the same scroll position. This appears to be a regression in this PR as it does not do it on the staging deployment.

Both behaviours are illustrated in the video below.


https://user-images.githubusercontent.com/24264157/191396993-9fdac266-f8d6-48c9-8855-5bb4540c83ba.mp4

sarayourfriend (commented):
In addition to snapshot tests for the new header, e2e tests exercising the modal opening closing, scroll locking (once fixed) and external link handling would also be good to have.

obulat:
> A behaviour I just noticed is that the modal stays open if you resize the window back to a breakpoint bigger than when the modal would be able to be opened. I think this is better than disappearing the modal based on the breakpoint, but I wanted to make sure that behaviour is intentional. We may have to wait to ask Francisco once he is available though.

It wasn't intentional, fixing it.

src/components/VHeader/VHeaderInternal.vue
obulat: I think the 'Internal' here comes from the Internal pages that this header links to. The non-search and non-content-details header. Do you have any suggestions for the component name?

zackkrida (changes requested):
This looks excellent, some things I noticed:

- The logo is not a clickable link, it's just a span
- The focus ring in the mobile menu includes a thin white ring around the pink one
- if you expand the page while the menu is open it stays open. not sure if this is worth fixing. Screenshot: https://share.cleanshot.com/NYBDBV

src/components/VHeader/VHeaderInternal.vue
```diff
@@ -0,0 +1,150 @@
...
+                <span class="sr-only">WordPress</span>
+              </template>
+            </i18n>
+          </VLink>
+        </template>
+      </VModal>
+    </div>
+  </header>
+</template>
+<script lang="ts">
```
zackkrida, line 89:
```suggestion
</template>

<script lang="ts">
```

src/components/VHeader/VHeaderInternal.vue
zackkrida: > The modal also causes the page to scroll to the top now instead of staying at the same scroll position. This appears to be a regression in this PR as it does not do it on the staging deployment.

Observing this too.

obulat:
I've extracted the logo link and the page links to separate components.

>     * The logo is not a clickable link, it's just a span

I didn't realize that `VBrand` wasn't a link. Fixed by extracting a component.

>     * The focus ring in the mobile menu includes a thin white ring around the pink one

Fixed.

>     * if you expand the page while the menu is open it stays open. not sure if this is worth fixing. Screenshot: https://share.cleanshot.com/NYBDBV

@sarayourfriend has mentioned it before and I fixed it by adding some CSS to hide the modal on `md` and up. Do you still see this problem with the updated code?

src/components/VHeader/VHeaderInternal.vue
obulat: I think I've fixed this in https://github.com/WordPress/openverse-frontend/pull/1825/commits/e445d9a14426ee58b9c911aaefbe77e2e8056159, are you still seeing this behavior, @zackkrida ?

zackkrida (changes requested):
CSS indeed hides the menu on larger breakpoints now, but the frontend body scrolling remains locked and there remains odd whitespace next to the page content:

CleanShot 2022-09-26 at 14 02 34@2x

The focus ring is fixed, except on the WordPress icon:

CleanShot 2022-09-26 at 14 03 49@2x


This is so close to merge!

dhruvkb (approved):
Looks good to me! I noticed no problems in testing and the design matches the mockups.

src/composables/use-pages.ts
```diff
@@ -7,26 +7,32 @@ export default function usePages(isNew = false) {
     ? [
         {
           id: 'about',
+          name: 'navigation.about',
```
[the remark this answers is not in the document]
obulat: There is a `i18n:check` script in package.json that checks for unused and missing i18n keys. Unfortunately, it can't understand dynamic keys, so it reports the key as missing if it's dynamic. This means that we cannot use it in the CI. If we convert all dynamic keys to static keys, we will be able to check the i18n for completeness in the CI, @dhruvkb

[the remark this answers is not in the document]
dhruvkb: Ok, makes sense.

zackkrida (approved):
LGTM!

PR #1829, opened 2022-09-20
Add a click handler to VLink if it uses NuxtLink

### LINKED ISSUE #1828: Add `onClick` to `VLink` component when `NuxtLink` is used
## Problem

We are using a conditional component in VLink which passes on the `$listeners` to the underlying component. However, when passing the `$listeners` Vue only passes the listeners without the `.native` modifer (https://v2.vuejs.org/v2/api/#vm-listeners). This prevents the `onClick` handler from firing on `NuxtLink` click.
If we try to specifically add `@click.native`, we get an error when `VLink` uses a native `a` anchor link because `native` modifier isn't available for them.

## Description

Instead of using a conditional component (`component :is=""`), we can use `v-if` to select the necessary component, and pass the `@click.native` handler only to `NuxtLink`.

## Additional context
I've originally added these changes in #1825, but it would be better to extract that work into a separate PR. This means that this PR becomes `high` priority because it's blocking #1825.

## Implementation

obulat:
## Fixes

Fixes #1828 by @obulat

## Description

This PR replaces the conditional component (`component :is="<>"`) in `VLink` with one of three components with `v-if`/`v-else-if`/`v-else` to make it possible to use `@click.native` with `NuxtLink`. The component looks much more verbose now, but the functionality is improved. 
This PR also adds a common mocking for `NuxtLink` in unit tests.

## Testing Instructions

All the links in the app should work as normal. You can also see these changes in #1825: the click handler for an internal link in the mobile modal closes the modal using a `click` handler. It wouldn't work without these changes: the click event is not passed to the parent component from a `VLink` that is a `NuxtLink` (internal link).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VLink.vue
+++ b/src/components/VLink.vue
@@ -1,7 +1,27 @@
+<!-- eslint-disable vue/no-restricted-syntax -->
 <template>
-  <Component
-    :is="linkComponent"
-    v-bind="linkProperties"
+  <NuxtLink
+    v-if="linkComponent === 'NuxtLink'"
+    v-bind="$attrs"
+    :class="{ 'inline-flex flex-row items-center gap-2': showExternalIcon }"
+    :to="linkTo"
+    v-on="$listeners"
+    @click.native="$emit('click', $event)"
+  >
+    <slot /><VIcon
+      v-if="showExternalIcon && !isInternal"
+      :icon-path="externalLinkIcon"
+      class="inline-block"
+      :size="4"
+      rtl-flip
+    />
+  </NuxtLink>
+  <a
+    v-else-if="linkComponent === 'a'"
+    v-bind="$attrs"
+    :href="href"
+    target="_blank"
+    rel="noopener noreferrer"
     :class="{ 'inline-flex flex-row items-center gap-2': showExternalIcon }"
     v-on="$listeners"
   >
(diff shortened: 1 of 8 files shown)


src/components/VLink.vue
```diff
@@ -12,7 +15,37 @@
...
+    :class="{ 'inline-flex flex-row items-center gap-2': showExternalIcon }"
+  >
+    <slot /><VIcon
+      v-if="showExternalIcon && !isInternal"
+      :icon-path="externalLinkIcon"
+      class="inline-block"
+      :size="4"
+      rtl-flip
+    />
+  </span>
```
sarayourfriend, line 48: This isn't related to the changes in this PR, but wanted to call it out. This `span` fallback case worries me quite a bit :disappointed: What are the circumstances when we wouldn't have an `href` to pass to the link? Is it an asynchronous data issue or something like that?

sarayourfriend (approved):
Tested locally and works great. One comment on something for us to think about but shouldn't affect this PR.

src/components/VLink.vue
obulat: Yes, it's the asynchronous data issue, when we are trying to render VLink before the `href` was loaded, probably. And if I remember correctly, it happens on server rendering and is replaced with either `<a>` or `<NuxtLink>` on hydration.
We could (should?) probably audit the code and use `v-if` to not render `VLinks` before `href` is available.

zackkrida (approved):
Nice fix; lgtm.

src/components/VLink.vue
sarayourfriend: > We could (should?) probably audit the code and use v-if to not render VLinks before href is available.

I'd worry about pop-in at this point too though. We'd probably want that auditing to also focus on making sure page navigations work smoothly as data comes in.

PR #1841, opened 2022-09-20
Remove span from VLink

### LINKED ISSUE #1840: Audit all the `VLinks` to make sure all have `href` and make it required
## Problem

Noted in https://github.com/WordPress/openverse-frontend/pull/1829#discussion_r975118023 : currently, the VLink component can sometimes not have an `href` prop and be rendered as a span.

## Description

We should only use `VLink` when `href` is available, and remove the `span` fallback from `VLink` component.

## Additional context

More discussion in  https://github.com/WordPress/openverse-frontend/pull/1829

## Implementation

obulat:
## Fixes

Fixes #1840 by @obulat

## Description

This PR removes `span` as the fallback element that was rendered by `VLink` if the `href` prop wasn't provided.
It also adds checks for `href` in all components that use a dynamic `href` prop, and adds a wrapper to the `VContentLink` to make the disabled component render a div instead of an `a` link without an `href`.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VLink.vue
```diff
@@ -33,19 +31,6 @@
...
-    v-bind="$attrs"
-    :class="{ 'inline-flex flex-row items-center gap-2': showExternalIcon }"
-  >
-    <slot /><VIcon
-      v-if="showExternalIcon && !isInternal"
-      :icon-path="externalLinkIcon"
-      class="inline-block"
-      :size="4"
-      rtl-flip
-    />
```
dhruvkb, line 47: Is the removal of `<VIcon/>` accidental or intentional?

zackkrida:
My understanding was that `a` tags without an `href` attribute is a valid and encouraged way to disable a link, particularly paired with `aria-disabled` (and actually, now I'm seeing `role=link` should be used too):

- https://a11y-guidelines.orange.com/en/articles/disable-elements/#disable-a-link
- https://www.scottohara.me/blog/2021/05/28/disabled-links.html

src/components/VLink.vue
obulat: Only the Icon inside the span I'd removed, <a> and NuxtLink still have it.

obulat:
> My understanding was that `a` tags without an `href` attribute is a valid and encouraged way to disable a link, particularly paired with `aria-disabled` (and actually, now I'm seeing `role=link` should be used too):
> 
> - https://a11y-guidelines.orange.com/en/articles/disable-elements/#disable-a-link
> - https://www.scottohara.me/blog/2021/05/28/disabled-links.html
> 

Oh, this is interesting, thank you for links, Zack! I tried tabbing to the disabled content links on staging - and couldn't. Those links are skipped when using the Tab key. However, with VO, if you use Ctrl Shift arrow key - you do get to the disabled link! 
With this, I realized that the keyboard access order when using the VoiceOver key with arrows and when using Tab key are different! And the change in tabbing order we implemented ( from the header to the sidebar ) doesn't work with the VO navigation keys 😞

zackkrida:
@obulat do you think we should remove the `div` then and replace with `a`, or a different approach?

src/components/VLink.vue
```diff
@@ -1,27 +1,21 @@
...
     @click.native="$emit('click', $event)"
   >
-    <slot /><VIcon
-      v-if="showExternalIcon && !isInternal"
-      :icon-path="externalLinkIcon"
-      class="inline-block"
-      :size="4"
-      rtl-flip
-    />
+    <slot />
```
obulat, line 10: I realized just now that a NuxtLink is only used for the internal links, so it cannot use the `externalLinkIcon`.

obulat:
@sarayourfriend, I would really appreciate your opinion on this approach for disabling links, and not having `span` as the `VLink` fallback.

src/components/VLink.vue
sarayourfriend: Interesting. How do we define an "external" link? When we've removed the iframe, will w.org links be "external"? If all links that don't use `NuxtLink` are external then we can remove the conditional `showExternalIcon && !isInternal` from `VIcon`. If it's only things that are not either (a) Openverse frontend links or (b) w.org links, then we could detect it and automatically add the icon without having to rely on the prop.

Probably an additional tightening up here and clarification around the iframe project and the idea of "externality". We typically concern ourselves with this (I think) due to differing privacy expectations on other websites. Given Openverse and WordPress.org follow the same privacy policies and are all controlled by WPF, we probably don't need the external icon for w.org links either. But again, something to clarify and probably something for another issue and to rope the w.org infrastructure folks in on.

sarayourfriend:
If you're asking from an a11y perspective, I don't know enough to comment. The behaviour you're seeing with VO is fine. Tab is only used to move between actionable items (buttons, form elements, links) so it naturally skips over disabled ones. The VO text navigation on the other hand is used to read each element of the page one by one. Tab navigation is probably more likely used by a sighted keyboard user who will have the visual context to see that a link is disabled. Maybe some style elements would need to show this if it's anything beyond a loading state so that all sighted users who are not relying on a screen reader that will relay the aria attribute context to them will understand why something that appears to be a link is not clickable.

These are all problems that already exist if we use a span fallback, so I don't think they need to be solved or debated too much here. I haven't tested the code here so I won't stick an approval on it, but as long as other folks have tested and find it functional, I don't see any a11y reason not to merge this. There is no regression from the previous behaviour and in fact screen readers will now have _more_ context than they did before with the `span` approach.

dhruvkb (approved):
I'm not a 100% clear on the nuances of a11y for this change, but I tested the following and they work.

- navigation with keyboard
- announcements by VoiceOver

zackkrida (approved):
LGTM!

PR #1842, opened 2022-09-21
Refactor SearchTypePopover desktop header

### LINKED ISSUE #1836: Refactor the SearchType popover (desktop) for the new header
## Problem

The new header design only has the SearchTypeButton shown on the desktop widths. 

## Description

It will only show the SearchTypes popup, and not the modal.

## Mockups
Figma: https://www.figma.com/file/KxkZmBNk218V2Fke1GeYCu/New-header-component?node-id=561%3A65396

## Implementation

obulat:
## Fixes

Fixes #1836 by @obulat

## Description

This PR creates a `VHeaderDesktop` component that will be used on widths above `lg`. It refactors the `VSearchTypePopover` to only open the popover (so, it removes the modal). It also refactors the desktop filters button to only open the sidebar.

This PR adds an `IsMinScreenLg` `provide` value in the `default.vue` layout that is set to true only after the layout was mounted. This enables us to have components rendered conditionally (with `v-if`) without getting the server and client mismatch on rendering. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VHeader/VHeaderFilter.vue
+++ b/src/components/VHeader/VHeaderFilter.vue
@@ -156,7 +156,7 @@ export default defineComponent({
 
     return {
       nodeRef,
-      visibleRef,
+      sidebarVisibleRef: visibleRef,
       triggerElement,
 
       open,
--- a/src/components/VHeader/VHeaderFilter.vue
+++ b/src/components/VHeader/VHeaderFilter.vue
@@ -156,7 +156,7 @@ export default defineComponent({
 
     return {
       nodeRef,
-      sidebarVisibleRef: visibleRef,
+      visibleRef,
       triggerElement,
 
       open,
--- a/src/components/VLogoLoader/VLogoLoader.vue
+++ b/src/components/VLogoLoader/VLogoLoader.vue
@@ -5,11 +5,12 @@
     :class="{
       [$style.loading]: status === 'loading' && !prefersReducedMotion,
       'h-10 w-10': autoResize,
+      'h-12 w-12': !autoResize,
     }"
     aria-hidden="true"
     :data-prefers-reduced-motion="prefersReducedMotion"
     data-testid="logo-loader"
-    class="inline-flex h-12 w-12 items-center justify-center rounded p-3 md:h-12 md:w-12"
+    class="inline-flex items-center justify-center rounded p-3 md:h-12 md:w-12"
     fill="currentColor"
   >
     <path
--- a/src/components/VContentSwitcher/VSearchTypeButton.vue
+++ b/src/components/VContentSwitcher/VSearchTypeButton.vue
@@ -19,20 +19,9 @@
   </VButton>
 </template>
 <script lang="ts">
-import {
-  computed,
-  defineComponent,
-  inject,
-  PropType,
-} from '@nuxtjs/composition-api'
+import { computed, defineComponent, inject } from '@nuxtjs/composition-api'
 
-import {
-  ALL_MEDIA,
-  AUDIO,
-  IMAGE,
-  type SearchType,
-  VIDEO,
-} from '~/constants/media'
+import { ALL_MEDIA, AUDIO, IMAGE, VIDEO } from '~/constants/media'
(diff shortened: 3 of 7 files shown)


src/components/VHeader/VHeaderFilter.vue
```diff
@@ -156,7 +156,7 @@ export default defineComponent({
 
     return {
       nodeRef,
-      visibleRef,
+      sidebarVisibleRef: visibleRef,
```
obulat, line 159:
```suggestion
      visibleRef,
```

zackkrida (changes requested):
- The search bar in VHeaderDesktop appears to fill all available space; is that the correct behavior or is it meant to have a max-width?
- I don't think it matters since this will be replaced, but there's a slight alignment problem in a medium breakpoint:
CleanShot 2022-09-22 at 14 20 20@2x
- It occurred to me that I don't think we even need JavaScript to switch between the different headers. A naive CSS approach seems to work fine and avoid the whole hydration mismatch, forced re-render problem.

```vue
<div class="hidden lg:block">
  <VHeaderDesktop />
</div>
<div class="block lg:hidden">
  <VHeader />
</div>
```

obulat:
> * The search bar in VHeaderDesktop appears to fill all available space; is that the correct behavior or is it meant to have a max-width?
> 
>   * I don't think it matters since this will be replaced, but there's a slight alignment problem in a medium breakpoint:

The main goal of this PR is to get rid of the components that have duplicate names, and to separate the two `header` components (mobile and desktop). The styling changes will be fixed in later PRs. Otherwise, this PR will probably be too big.

> - It occurred to me that I don't think we even need JavaScript to switch between the different headers. A naive CSS approach seems to work fine and avoid the whole hydration mismatch, forced re-render problem.
> 
> ```
> <div class="hidden lg:block">
>   <VHeaderDesktop />
> </div>
> <div class="block lg:hidden">
>   <VHeader />
> </div>
> ```

The current solution has the  following downsides:
- hacks on mounted to avoid hydration mismatch.
- forced re-render on desktop widths after `mounted` event is fired. We hope to fix it by having a UI cookie that would make the server render desktop or mobile layout (https://github.com/WordPress/openverse-frontend/milestone/11)

There are a couple of downsides to this CSS solution:
- On the mobile breakpoint, this will be downloading all of the components, both mobile and desktop, and will probably degrade the mobile performance.
- For the search bar components, there will be `id` duplication, or we would have to set separate `id`s for mobile and desktop search inputs.

I was mainly trying to optimize for the mobile experience/performance because mobile usually has lower bandwidths. Do you think it's better to use CSS, @zackkrida, and reduce the need for JS hacks?

zackkrida:
@obulat I think we can follow your solution instead of a CSS approach. We can use the session cookie approach later to fix any visual flashes. Your point about the bundle size is a good one.

zackkrida (approved):
LGTM and get those components removed.

src/components/VLogoLoader/VLogoLoader.vue
```diff
@@ -5,11 +5,12 @@
     :class="{
       [$style.loading]: status === 'loading' && !prefersReducedMotion,
       'h-10 w-10': autoResize,
+      'h-12 w-12': !autoResize,
     }"
     aria-hidden="true"
     :data-prefers-reduced-motion="prefersReducedMotion"
     data-testid="logo-loader"
-    class="inline-flex h-12 w-12 items-center justify-center rounded p-3 md:h-12 md:w-12"
+    class="inline-flex items-center justify-center rounded p-3 md:h-12 md:w-12"
```
dhruvkb, line 13: Do the `md:` height and width classes not need to be moved?

dhruvkb (approved):
LGTM!

src/components/VLogoLoader/VLogoLoader.vue
obulat: Moved to the beginning of the class list? I think the prefixed classes are always at the end.

PR #1844, opened 2022-09-22
Allow setting feature flags using query params

### LINKED ISSUE #1837: Enable switching switchable feature flags using query params
## Problem

Currently, to turn on a switchable feature flag, we need to go to the `/preferences` page and mark the flag.

## Description

It would be nice to be able to easily share the work under a feature flag by sharing a link that contains a flag query parameter. It should only work on staging, not in prod.

## Additional context

We are hoping to ask for community opinions on the header/footer designs using links to staging. 

## Implementation

dhruvkb:
## Fixes

Fixes #1837 by @obulat 

## Description

This PR allows feature flags to be toggled (when set to switchable) using 'ff_' prefixed query params.

## Behaviour
The intended behaviour of the query param is to be be a one-time override to the value set in the cookie. After the query param is dropped and the page is refreshed, the flag will be reset to the value stored in the cookie.

## Testing Instructions

1. Visit the preferences page and set the 'new_header' flag to off.
2. Navigate around the site and see that the new header and footer are not shown.
3. Add the query param `?ff_new_header=on` to the URL. This time the footer will appear as will remain visible as long as you navigate the site.
4. On the next reload (<kbd>Ctrl</kbd>/<kbd>Cmd</kbd> + <kbd>R</kbd>) without the param, the header and footer will disappear.
5. Visit the preferences page and set the 'new_header' flag to on.
6. Navigate around the site and see that the new header and footer are shown.
7. Add the query param `?ff_new_header=off` to the URL. This time the footer will disappear and will remain hidden as long as you navigate the site.
8. On the next reload (<kbd>Ctrl</kbd>/<kbd>Cmd</kbd> + <kbd>R</kbd>) without the param, the header and footer will reappear.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/middleware/middleware.ts
+++ b/src/middleware/middleware.ts
@@ -45,5 +45,6 @@ const middleware: Middleware = async ({ app, query, route, $pinia }) => {
 
   const featureFlagStore = useFeatureFlagStore($pinia)
   featureFlagStore.initFromCookies(app.$cookies.get('features') ?? {})
+  featureFlagStore.initFromQuery(query)
 }
 export default middleware
--- a/src/stores/feature-flag.ts
+++ b/src/stores/feature-flag.ts
@@ -16,6 +16,8 @@ import {
 } from '~/constants/feature-flag'
 import { LOCAL, DEPLOY_ENVS, DeployEnv } from '~/constants/deploy-env'
 
+import type { Dictionary } from 'vue-router/types/router'
+
 type FlagName = keyof typeof featureData['features']
 export interface FeatureFlagState {
   flags: Record<FlagName, FeatureFlag>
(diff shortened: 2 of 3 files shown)


src/middleware/middleware.ts
```diff
@@ -45,5 +45,6 @@ const middleware: Middleware = async ({ app, query, route, $pinia }) => {
 
   const featureFlagStore = useFeatureFlagStore($pinia)
   featureFlagStore.initFromCookies(app.$cookies.get('features') ?? {})
+  featureFlagStore.initFromQuery(query)
```
obulat, line 48: Can we make it only work on `staging`, and not in `prod`? I think if we want to turn the flags on in prod, we should use the values in the saved file?

obulat (approved):
This works well for the first page render, but if someone is going to click on any link on the website, the flag value will go back to the cookie state. Is there a way to keep the state saved for a period of a session only? Like a sessionStorage or something like that?

src/middleware/middleware.ts
dhruvkb: @obulat we should make the flag set to disabled instead of switchable in the prod environment so that it cannot be turned on by the feature flag query params.

dhruvkb:
@obulat I tried this locally and within one session, navigating between pages using internal links does not reset the flags to the cookie value. It only resets when a full page reload is performed.

zackkrida (approved):
Great!

Issue #479, opened 2021-12-03
Recent Searches feature (Combobox, localstorage, and more)


zackkrida:
## Description

We plan to add a new feature to show the user their 'recent searches', in a box below the searchbar. I originally thought this would be a combobox feature, but we shouldn't actually need to implement that until we're doing search _suggestions_, where the user input determines the contents of the box. In this use case we're just showing the nth-most recent searches.

> A preview of the recent searches

![image](https://user-images.githubusercontent.com/6351754/144633420-a3aeadac-187a-455d-9b97-1358d22b7046.png)

Figma Link: https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=133%3A29056

## Dev Tasks

- [ ] Implement the suggestions box
  - [ ] Popover for the desktop (above `lg`)
  - [ ] Full page modal with a search bar for mobile screens
- [x] Create a composable for storing recent searches
  - [x] Store user recent searches in indexedDb (see https://github.com/WordPress/openverse-frontend/issues/479#issuecomment-1252941676 for justification)
  - [x] Enable clearing of the recent searches in indexedDb

## Implementation questions

- Do we need a way to let users opt-out of this? (maybe this can wait)
- How many items do we want to show? (4 in the mockup currently) I imagine the search bar will populate and the search will be instantly submitted.
- What will happen when the user clicks on one of the recent searches?

## Design decisions (cc @panchovm)

- [x] Add this to the Figma component library (it's currently only in the page mockups)
- [x] Implement a 'clear' button (will update the mockups for this)
- [x] What will the hover/focus styles look like for the choices?

dhruvkb:
I think it'll be nice to have a way to let the users know that this info is 100% local and is not shared with us? It might help a lot to assuage fears of privacy invasion and persuade users to let this feature remain enabled.

zackkrida:
I was thinking about that too. Maybe some small disclaimer text at the bottom of the popup? cc @panchovm 

Something like "Openverse does not store your recent searches; this information is kept locally in your browser" or something like that

panchovm:
I really like @dhruvkb's idea. I will come back with a proposal. This also reminds me, do we have a privacy terms section for Openverse in the [privacy page](https://wordpress.org/about/privacy/)?

obulat:
[...]

PR #1865, opened 2022-09-26
Add recent searches to the search store

zackkrida:
## Fixes

Related to #479 by @zackkrida 

## Description

This PR adds a user's recent searches to the search store, and syncs these searches to localstorage. Adding these searches to a global store _and_ localstorage is the best way to ensure the recent searches can be stored and read from different components without doing a lot of convoluted work to sync state.

This code includes some work to make sure duplicate searches aren't stored in the history and to cap the number of saved searches to the env-variable set default.

It was interesting to learn about how Pinia unwraps reactives with more detail while working on this. Really wishing we were on Nuxt 3 and Vue 3 with the hopes it would fix some odd behavior I had to deal with here! Being stuck on the really old (`0.2.1`) `@pinia/nuxt` package feels pretty brittle.

## Testing Instructions

Make some searches in Openverse and console.log `localStorage.getItem('recent-searches')` after to ensure they saved correctly. Make sure they never exceed the currently set limit of 4 searches and that you don't see any duplicates.

## Todo

The steps remaining to undraft this are:

- [x] Put the saving of searches behind a feature flag (it seems a bit sneaky to store these for users until the UI is live)
- [x] Update search-store tests for these new actions
- [x] Sort out bug where searches don't save after homepage navigation

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/stores/search.ts
```diff
@@ -156,11 +167,31 @@ export const useSearchStore = defineStore('search', {
...
+    /** Add a new term to the list of recent search terms */
+    addRecentSearch(
+      search: string /** A search term to add to the saved list.*/
+    ) {
+      /**
+       * Here we add our latest search to the top of the stack,
+       * and add the remaining items, making sure not to exceed
+       * our max count and removing and old occurences of the
+       * most recent term.
+       */
```
zackkrida, line 183:
```suggestion
      /**
       * Add the latest search to the top of the stack,
       * then add the existing items, making sure not to exceed
       * the max count, and removing existing occurrences of the
       * most latest search term, if there are any.
       */
```

src/stores/search.ts
```diff
@@ -32,6 +33,10 @@ import { useProviderStore } from '~/stores/provider'
 
 import { useFeatureFlagStore } from '~/stores/feature-flag'
 
+import { useStorage } from '~/composables/use-storage'
+
+import type { Ref } from '~/composables/types'
```
dhruvkb, line 38: `Ref` should be imported from the `@nuxtjs/composition-api' package.

```suggestion
import type { Ref } from '@nuxtjs/composition-api'
```

src/stores/search.ts
```diff
@@ -42,6 +47,7 @@ export const isSearchTypeSupported = (
 
 export interface SearchState {
   searchType: SearchType
+  recentSearches: Ref<string[]>
```
dhruvkb, line 50: Ideally our state variables should be base types and not refs. Is that what you meant when you said "odd behavior I had to deal with here"?

src/stores/search.ts
```diff
@@ -156,11 +167,47 @@ export const useSearchStore = defineStore('search', {
...
+      /**
+       * Add the latest search to the top of the stack,
+       * then add the existing items, making sure not to exceed
+       * the max count, and removing existing occurrences of the
+       * most latest search term, if there are any.
+       */
+      this.recentSearches = [
+        search,
+        ...this.recentSearches.filter((i) => i !== search),
+      ].slice(0, parseInt(env.savedSearchCount))
```
dhruvkb, line 203: Is it possible for `env.savedSearchCount` to be undefined? If it is, we should handle that possibility.

```suggestion
      ].slice(0, parseInt(env.savedSearchCount ?? '4'))
```

dhruvkb (approved):

src/stores/search.ts
zackkrida: So pinia actually handles reactives in stage quite elegantly and recommends it in a number of places. For example: https://pinia.vuejs.org/cookbook/composables.html#ssr

The odd behavior I had to deal with here was specifically with the watcher inside of `useStorage` failing to be evoked on changes after navigating from the homepage to the search page. I'm still not exactly sure why that occurs and wasn't able to fix it in testing without that 'hack'

dhruvkb:
Bumping priority to high as it will now block #1867.

obulat (commented):
I think the problem with the `useStorage` hydration should be solved in `useStorage`. `VueUse` (where we got the implementation from) has updated the composable to fix the hydration problem (I think), and this is one of the PRs:
https://github.com/vueuse/vueuse/pull/ 1060

@zackkrida, would it make sense to use `VueUse` directly here? I was hesitant to add it because there was one of the composables that decided to support some feature that we used only for Vue 3 (I don't remember which one it was). But this composable definitely needs updating.

zackkrida:
@obulat I'm happy to switch to and use the `VueUse` composable, I actually did this at one point in development, but it unfortunately doesn't fix the 'hack' I had to add to `src/pages/search.vue`. It's not the _worst_ hack, since it's heavily documented and just reinitializes a composable, but would definitely be preferable not to need.

src/components/VNotificationBanner.vue
```diff
@@ -29,9 +29,10 @@
 <script lang="ts">
 import { defineComponent, PropType } from '@nuxtjs/composition-api'
 
+import { useStorage } from '@vueuse/core'
```
obulat, line 32: I've started regretting my suggestion of using VueUse directly. Now, I get an IDE error saying 'Cannot find module '@vueuse/core' or its corresponding type declarations.'. If I add a Nuxt build module as is suggested in the VueUse docs (`@vueuse/nuxt`, https://vueuse.org/guide/index.html#nuxt), I get an 'Unmet peer dependency Nuxt 3' error. The related issue in the repository is marked as fixed (https://github.com/vueuse/vueuse/issues/2176), but it seems to be fixed for Nuxt Bridge, not Nuxt 2.

Maybe we should wait with VueUse until we have Nuxt 3? 😢 Or should we tolerate the IDE warnings?

obulat (approved):
Works well and saves the recent searches.

I am really disappointed with the fact that using VueUse is not so straightforward with our setup :(

PR #1867, opened 2022-09-27
Add UI for recent search results

### LINKED ISSUE #479: Recent Searches feature (Combobox, localstorage, and more)

dhruvkb:
## Fixes

Fixes #479 by @zackkrida 

## Description

This PR adds the UI component for the recent search results, which will be populated with data from the store added in #1865.

## Testing Instructions

0. Check out the PR and run the Storybook.
1. Compare the Story against the Figma mockups.

---

1. Ensure that the 'new_header' flag is turned on in `/preferences`.
2. Search for some items from the `/search` page. Searches from the home page will not work!
3. See searches being added to the list.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
It looks perfect already 🔥 have you thought about the implementation yet? It'd be cool if this component can stay purely presentational and in the actual header we wrap it in a combobox or popover or modal.

I'm not sure which of those is most appropriate here. It's not _quite_ a combobox maybe, because the input text doesn't modify the choices displayed, but maybe any box that is triggered by an input is a combobox? Not sure.

dhruvkb:
@zackkrida I'm considering keeping it presentational (I love presentational components because they are so easy to write stories for!) but I'll have to explore regarding the functionality. WIP!

zackkrida (changes requested):
The styles look perfect! Some issues with functionality:

- None of the events are working (clearing all items or selecting an item)
- The 'clear' button is not accessible via keyboard
- We're currently using the desktop styles on mobile, which may be fine for this PR as it makes merging easier and is still pretty usable on a mobile device. I'm content to implement the mobile full-height modal version in a subsequent PR.

dhruvkb:
@zackkrida the bug was introduced in the last commit d2d018f, that seemed fairly innocuous. Sorry about that. Also the clear functionality was yet to be implemented, it's added in 442c5d8.

The Clear button can be accessed by <kbd>Tab</kbd>, it's after the search button in the tab order.

We'll implement the mobile version separately. Since the work is broken into different PRs, it's hard to keep track of what's happening where and rapidly merging PRs makes it easier to prevent integration issues.

src/components/VHeader/VSearchBar/VSearchBar.vue
```diff
@@ -1,42 +1,78 @@
...
 import { defineEvent } from '~/types/emits'
 
+import { useSearchStore } from '~/stores/search'
+
+import { keycodes } from '~/constants/key-codes'
+
+import { cyclicShift } from '~/utils/math'
+
+import { useFeatureFlagStore } from '~/stores/feature-flag'
+
```
obulat, line 70:
```suggestion
import { useSearchStore } from '~/stores/search'
import { useFeatureFlagStore } from '~/stores/feature-flag'

import { keycodes } from '~/constants/key-codes'

import { cyclicShift } from '~/utils/math'

```

obulat: Nit :)

obulat (changes requested):
Everything works really great!

Could you add unit or e2e tests for this functionality? I can think of the following cases that need to be tested:
- after searching for something and focusing on the searchbar, the recent searches is open, we can see the previous search term in it.
- when you click on the previous term, the search term changes and the recent searches is closed.
- when you click on clear, the list is cleared (and the local storage too? - I'm not sure we can test that)

zackkrida (changes requested):
Found a few more things, apologies:

- I think the Clear button should have a more verbose aria label, like "Clear recent searches".
- ~Instead of showing a 'no recent searches' message, we should hide the recent searches combobox entirely.~ We can skip this for now, I think we need some @panchovm advice.

obulat:
> * Instead of showing a 'no recent searches' message, we should hide the recent searches combobox entirely.

I agree that we should remove the message, but I also feel like we need to give the users some kind of notice that the 'recent searches' have been deleted from the storage.

zackkrida:
That's a good point @obulat. I was primarily considering situations where the user didn't have any recent searches yet but the situation you're describing will be much more common.

Perhaps we should just keep the message, for now. I won't block this PR on that aspect.

src/components/VRecentSearches/VRecentSearches.vue
```diff
@@ -0,0 +1,100 @@
...
+        variant="plain"
+        size="small"
+        class="caption-bold hover:underline"
+        @click="handleClear"
+      >
+        {{ $t('recent-searches.clear') }}
+      </VButton>
+    </div>
+
+    <ul
```
obulat, line 21: I tried adding the Recent searches to the mobile searchbar in #1882, and I have problems with keyboard navigation.
Because the options are a `<li>` items, it's not possible to focus them using the Tab key. What is the reason for not using interactive `button` or `link` elements here?

dhruvkb: Because they should not be tabbable as per ARIA guidelines.

zackkrida: This is also how the content switcher currently works (vertical arrow keys instead of tab).

zackkrida (approved):
I think this looks good to go, @dhruvkb! Nice work.

src/components/VRecentSearches/VRecentSearches.vue
obulat: Thank you for clarifying! It turns out I forgot to add the `keydown` handler to the `input`, so pressing Arrow keys on the input wasn't working as supposed to. With the handler, it works really well!

obulat (approved):

obulat:
Really nice addition and great work on adding the combobox, @dhruvkb!

PR #1892, opened 2022-10-07
Make `isNewHeaderEnabled` a computed so that it's updated

### LINKED ISSUE #1891: Layout not updated because the `new_header` flag status is set once
## Description

The layout is not updated when you check the `new_header` checkbox.

## Reproduction

1. Run the app.
2. Go to `/preferences` page and check the `new_header` checkbox.
3. See error: the header should be updated to the VHeaderInternal, but instead it stays the same. The layout stays the same (without the new header) until you reload the page.

## Additional context

The `isNewHeaderEnabled` variable in the `default.vue` layout should be a computed so that it is updated when the feature flag value changes.

## Resolution

obulat:
## Fixes

Fixes #1891 by @obulat

## Description

This PR makes  `isNewHeaderEnabled` a computed so that it's updated when you check the `new_header` checkbox on the Preferences page.

## Testing Instructions

Go to the Preferences page (http://localhost:8443/preferences) and check the `new_header` flag. You should see the header updated immediately (it should now have internal links).
You can also try the [staging](https://search-staging.openverse.engineering/preferences) to see that the header there is not updated until you reload the page.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/layouts/default.vue
+++ b/src/layouts/default.vue
@@ -123,7 +123,7 @@ const embeddedPage = {
     )
 
     const isSidebarVisible = computed(() => {
-      return isNewHeaderEnabled
+      return isNewHeaderEnabled.value
         ? isSearchRoute.value && isMinScreenLg.value && isFilterVisible.value
         : isSearchRoute.value && isMinScreenMd.value && isFilterVisible.value
     })
--- a/src/layouts/default.vue
+++ b/src/layouts/default.vue
@@ -92,7 +92,9 @@ const embeddedPage = {
   },
   setup() {
     const featureFlagStore = useFeatureFlagStore()
-    const isNewHeaderEnabled = featureFlagStore.isOn('new_header')
+    const isNewHeaderEnabled = computed(() =>
+      featureFlagStore.isOn('new_header')
+    )
 
     const { isVisible: isFilterVisible } = useFilterSidebarVisibility()
     const { matches: isSearchRoute } = useMatchSearchRoutes()


dhruvkb (changes requested):
You'll need to add `.value` to its usages in the `<script>` section. I can't review those parts as they haven't changed.

zackkrida:
@dhruvkb and @obulat do you think it would be more robust long-term to make the `isOn` method of the feature flag store return a computed value?

dhruvkb (approved):
@obulat LGTM!

@zackkrida it could make sense to make the `isOn` be a computed property, but wouldn't components using it still need to put the value inside a `computed` call?

dhruvkb:
I just realised that this might break the "feature" that flags toggled using `ff_` query params remain the same for the entire session.

obulat:
> I just realised that this might break the "feature" that flags toggled using `ff_` query params remain the same for the entire session.

Which change might break the query param feature, @dhruvkb ? The one in this PR, or Zack's suggestion about returning a computed from isOn?

dhruvkb:
The one in this PR.

obulat:
> I just realised that this might break the "feature" that flags toggled using ff_ query params remain the same for the entire session.

What should the behavior be when you open the preferences and uncheck the checkbox that was previously set using a query param? 🤔
I can see two options here:
- if a flag is set using a query param, it is set to a special state so that it becomes not switchable using the query param. The feature's Preferences page checkbox is set to 'disabled'.
- the query param is NOT guaranteed to be used if the user uses the UI in the Preferences page to change it. So, you can go to the `/preferences?ff_new_header=off` and check the 'new_header' feature to set the flag to be `On`.

My personal preference would be the second option, but maybe it's because I can only see this used in the current situation, and not for the future feature flag. We wanted the query parameter for a flag feature to be able to show the feature to the stakeholders (e.g. on staging) without having them set up the app locally. It should be possible to revert the flag state using the UI in such a case.

I wonder if we should add the feature flags to the routing: if the flag is set, the query is always shown in the URL. This way, it will always be synced with the cookie state, and you would be able to update the flag value both on the preferences page, and by changing the query param.

zackkrida (approved):
👍

PR #1895, opened 2022-10-08
Refactor `VIcon` to set the size using a tailwind class instead of the `size` prop

### LINKED ISSUE #1894: Add accessible button components for mobile searchbar
## Problem

The mobile search bar in the new header contains buttons with focus area of 32x32px which need to have tappable area of more than 40x40px to be accessible.

## Description

Create a new component for buttons based on the `VIconButton` component that is 48x48px in size, but has a visible focus area of 32x32px to match the designs.

## Implementation

obulat:
## Fixes

Fixes #1894 by @obulat

## Description

This PR removes the numeric `size` prop from `VIcon` component. This is because there are too many sizes in the app to make it useful, especially since we have to add all the dynamically generated size classes (like `w-${size} h-$size`) to tailwind's safelist. 

## Testing Instructions

You can view the buttons in the Storybook.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VButton.vue
```diff
@@ -10,7 +10,7 @@
       isActive && $style[`${variant}-pressed`],
       $style[`size-${size}`],
       isPlainDangerous
-        ? ''
+        ? 'border border-tx'
         : 'border border-tx focus-visible:ring focus-visible:ring-pink',
```
dhruvkb, line 14: 'plain--avoid' is also used in these places. Changing the base style could have unforeseen changes that are not captured in e2e.

- `src/components/VHeader/VHeaderMobile/VSearchBarButton.vue`
- `src/components/VTabs/VTab.vue`

src/components/VIcon/VIcon.vue
```diff
@@ -59,14 +59,14 @@ export default defineComponent({
...
-     * 4 - 1rem, 5 - 1.25rem, 6 - 1.5rem.
+     * 4 - 1rem, 5 - 1.25rem, 6 - 1.5rem, 8 - 2rem, 12 - 3rem.
      *
      * @default 6
      */
     size: {
       type: Number,
       default: 6,
-      validator: (val: number) => [4, 5, 6].includes(val),
+      validator: (val: number) => [4, 5, 6, 8, 12].includes(val),
```
dhruvkb, line 69: Instead of this being a prop, the parent should be able to set `h-*` and `w-*` styles on the component. This seems a very contrived way to provide more sizes.

dhruvkb: Also the new sizes added here, being dynamic, should be added to `tailwind.safelist.txt`.

obulat: Do you suggest that we scrap the `size` prop altogether, and just use `h-* w-*` classes set by the parent component, with a fallback of `h-6 w-6`?

dhruvkb: I'm suggesting scrapping the prop in the long term and adding the safelist to approve the PR.

zackkrida: I do like the API design of `size` type props, but I think it makes the most sense when standardized across a component library with lots of small, composable ui building blocks. I agree with @dhruvkb's suggestion here.

obulat: I've removed the prop, and since all classes are not dynamically generated, did not have to add anything to the safelist.

obulat: I've tried using classes, but then there is a problem with default and overrides:
- if we add `w-6 h-6` by default, then to override them, we would need to add `!` important modifier.
- if we do not add a default, then we would need to add the size classes to _all_ VIcons.

So, I went back to the `size` prop as a cleaner option, and opened an alternative PR: https://github.com/WordPress/openverse-frontend/pull/1928

dhruvkb (approved):
Works as expected. I don't want to block it precautionary [E2E tests](https://github.com/WordPress/openverse-frontend/pull/1895#discussion_r994238750), those can be added separately.

obulat:
Closing this PR because we have decided that keeping the `size` prop is a better approach.

Issue #1843, opened 2022-09-22
Add E2E and VRT tests for new header and footer


dhruvkb:
## Problem
The new header and footer are being developed behind a feature flag which makes it harder to write E2E and VRT test cases for them. This is being addressed in #1837.

## Description

Once #1837 is resolved, tests need to be added for the following components.

- [x] `VHeaderInternal` - Storybook VR tests added in #1825
- [x] `VHeaderContent` - Playwright VR tests added in #1968
- [x] `VFooter` - Storybook VR tests added in #1969
- [x] `VSelectField` - Storybook VR tests added in #1821
- [x] `VLanguageSelect` - Storybook VR tests added in #1821

## Alternatives

Snapshot testing for components (not pages) should be done from Storybook wherever feasible. This would enable us to snapshot all the different states of a component and see them in isolation, independently of the other stuff on the page. Hopefully, it will also make the snaps more resilient to small shifts and make it easier to pinpoint the change if the snapshots do not line up.

## Implementation

- [x] 🙋 I would be interested in implementing this feature.

obulat:
This issue should also include the work on updating the existing tests to pass with the `new_header` enabled.

PR #1914, opened 2022-10-16
Update Playwright tests utilities to work both when the `new_header` flag is enabled and when it's disabled

obulat:
## Fixes

Related to #1843 by @dhruvkb

## Description

This PR prepares the background for the tests with `new_header` enabled:
- Updates the `utils` methods to add the `mode` parameter. By default, it is set to `new`, and all the functions work with the `new_header`. The previous functions are used when `mode` parameter is set to `old`.
- Renames all the tests that use the affected `utils` methods to add `-old` suffix to them. The new tests without the suffix will be added in a follow-up PR.
- Adds the new breakpoint-related methods. Previously, desktop widths started at `md` breakpoint, and now they start at `lg`.
- Improves the accessible label for the `VContentSettingsButton` to add a number of filters applied.

## Testing Instructions

The tests in CI should pass.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


test/playwright/utils/navigation.ts
```diff
@@ -171,27 +250,48 @@ export const assertCheckboxStatus = async (
...
+    await page.locator(`a[role="radio"]:has-text("${to}")`).click()
+    await closeMobileMenu(page, mode)
+  }
 }
 
 /**
- * Finds a button with a popup to the left of the filters button which doesn't have a 'menu' label
- * @param page - The current page
+ * For desktop, returns the content of the Content switcher button.
+ * For mobile, returns null.
```
obulat, line 274:
```suggestion
 * For mobile, returns the selected content type from the modal.
```

zackkrida (approved):
The `VContentSettingsButton` aria improvements look great as well. LGTM.

sarayourfriend (commented):
I'm not always a huge fan of "old" and "new" as indicators for things like this. Whatever is "new" today eventually becomes old. I guess at that point we'd come back and refactor all of these to remove the "old" tests... but to someone reading this who isn't clued into the header work, these mode names are cryptic. Are there clearer names than new and old for the mode? Maybe even "new_header" or "old_header"?

Also, one nit, if we could extract the `'old' | 'new'` type out it would help with the eventual "delete the old mode" refactor... if you removed the shared type then TypeScript would tell you precisely everywhere that needed to be changed to accommodate the refactor :slightly_smiling_face:

obulat:
> Maybe even "new_header" or "old_header"?

Also, one nit, if we could extract the 'old' | 'new' type out it would help with the eventual "delete the old mode" refactor... if you removed the shared type then TypeScript would tell you precisely everywhere that needed to be changed to accommodate the refactor 🙂

Great suggestions, thank you, @sarayourfriend! I've added the new type, and made `new_header` and `old_header` constants so that the value can be changed in a single place.

sarayourfriend (approved):
LGTM! The `-old` filenames would be nice to be `-with-old-header` instead, but just a nit pick and don't want to hold up the PR any more. Sorry I didn't think of it before.

Nice work getting this stuff consolidated :rocket:

PR #1949, opened 2022-10-30
Use JSON5 for `en.json`

### LINKED ISSUE #1815: Using JSON5 for `en.json`
## Assigned reviewers

Tagging @WordPress/openverse-frontend, please feel free to add your name below if you have some feedback (good or bad) or ideas.

- [ ] TBD
- [ ] TBD

## Description

We use JSON to write the main i18n locale file: `en.json`. This file is completely written by hand and maintained manually, for which JSON is a pretty poor format. This might come across as a hot-take so let me elaborate.

### Comments

My main complaint with using JSON for `en.json` is lack of comments. It's very hard to describe context for any i18n key in the `en.json` file. For a file that's written by and for people, that's a huge miss. There is duplication in the file, many keys are written deeply nested and then used outside of where they're defined (happens during refactors when code is moved around) and then there's no way to know if an i18n key is deprecated/superseded.

Comments wouldn't solve the problems above but they would be very helpful. That's why POT files (like [`openverse.pot`](https://github.com/WordPress/openverse-frontend/blob/translations/openverse.pot)) have comments (the `#:` before each msg)! And since our `en.json` file doesn't have comments, our POT comments end up being just `.vue` references. I believe we are under-utilising a powerful tool for communicating intent and context with the translators.

### Syntax

I also have several personal gripes with the JSON syntax but that's secondary.
  - lack of trailing commas
  - too many quotes, especially double quotes

## Why JSON5?

[JSON5](https://json5.org/) is a superset of JSON which has many advantages over plain JSON. I'll let their webpage explain what those are. Switching to JSON5 is as easy as a rename (superset of JSON) and we can adopt changes incrementally.

### Migration path

Feel free to chime in with alternatives.

- Using JSON5 for `en.json` (&rarr; `en.json5`) and moving it out of `src/locales/`.
- Writing a script that parses `en.json5` and converts it to `en.json`.
  - Using the new JSON5 when it is being synced with GlotPress (to pass comments to translators).
  - Writing the plain JSON into `src/locales/` similar to other locales.
- Ignoring the plain `en.json` file in `src/locales/`.
  - This makes the special treatment we give to `en.json5` much clearer (placed separately) and makes the `en.json` file equivalent to any other locale file (except generated locally instead of downloaded).
    ```.gitignore
    # .gitignore
	src/locales/*.json
	!src/locales/en.json
	```

## Alternatives

YAML, although almost universally reviled, is imo a great format. But moving to YAML is not the right move.

- YAML is not liked by a lot of developers (understandably). JSON on the other is much beloved, especially in the JS community.
- YAML requires a file rewrite (which can be automated, but why do it in the first place) when JSON is already valid JSON5.
- JSON5 is already being used in the codebase, for e2e tapes (that's precedent).

## References

- Old discussion about `en.json` organisation: https://github.com/WordPress/openverse-frontend/discussions/705

dhruvkb:
## Fixes

Fixes #1815 by @dhruvkb

## Description

This PR 
- replaces `en.json` with `en.json5` and reformats it
- adds a rudimentary JSON5 parser for the `en.json5` file
  - could not find any parser that's capable of retaining comments when the JSON5 file is parsed so made a basic one specifically for reading i18n files 
- updates PO generation functionality to include comments
  - updates POT file to include all references to the i18n string 
  - updates POT file to preserve the order of the `en.json5` file
  - uses the right type of POT comment (`#.` vs `#:`)
- updates scripts that used the `en.json` file
- converts English `en.json` to be like any other language
  - added JSON5 &rarr; JSON conversion for English to `i18n:get-translations` script
- makes general refactors, shortens code, adds docstrings with types, and cleans up code

Code in many places depends on `en.json`. This continues to work once the `en.json` file has been generated by running `i18n:get-translations`.

## Screenshots

| en.json5 | openverse.pot |
|-|-|
| Screenshot 2022-10-30 at 6 11 15 PM | Screenshot 2022-10-30 at 6 11 47 PM |

## Testing Instructions

0. Check out the PR.
1. Setup i18n `pnpm i18n`.
2. Run the dev server and see that everything is translated as expected.

Additionally CI is passing and unit tests that depend on locales are passing. So it's good to start. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (changes requested):
Having comments in the POT files will be really great for translators! The file generation works well.

Does it also save the `en.json` to the branch that we use for translations? I'm not sure how to test that, but we definitely need to check (if you haven't yet)  that it's working, and the translations are uploaded to GlotPress correctly.

Could you also add some documentation saying that `en.json5` is the main file where we should put all the strings and comments, and `en.json` and all other translated files should not be edited?

One more thing: to make sure that the Playwright tests always use the same translation files, we saved 3 locale JSON files to `tests/locales`. When running the tests, we run `i18n:copy-test-locales` to copy them to the `locales` folder. Could you add `en.json` creation to that step, too, so that the tests always have the English locales as well?

dhruvkb:
> Does it also save the en.json to the branch that we use for translations?

The [`WordPress/openverse-frontend@transations`](https://github.com/WordPress/openverse-frontend/tree/translations) branch only has the POT file. So this PR doesn't add the `en.json` file to the branch.

Thanks @obulat for pointing out the usages I missed! I've made the other improvements you suggested.

package.json
```diff
@@ -53,8 +53,9 @@
     "pc:install": "python3 pre-commit.pyz install -t pre-push -t pre-commit",
     "pc:lint": "python3 pre-commit.pyz run --all-files",
     "i18n": "pnpm i18n:create-locales-list && pnpm i18n:get-translations && pnpm i18n:update-locales",
-    "i18n:copy-test-locales": "cp -r test/locales/* src/locales/",
-    "i18n:no-get": "pnpm i18n:create-locales-list && pnpm i18n:update-locales",
+    "i18n:watch-en": "pnpm i18n:get-translations --en-only --watch",
```
obulat, line 56: In a freshly-cloned repository, if you try to run `pnpm test:playwright`, it will still raise an error because there will be no `en.json` yet.
I updated the `test:playwright:local` command on line 40 to `DISABLE_SENTRY=true pnpm i18n:copy-test-locales && playwright test -c test/playwright` to fix it locally and it worked, but I'm not sure if it'll work in the CI.

dhruvkb: In the CI, there is a step to [set up the `en.json` file](https://github.com/WordPress/openverse-frontend/pull/1949/files#diff-8fc5d4267ec6adb8c520d4a3cd87dd2c3026b516bc7fd7ba336dc697733c775dR35). That's why the CI workflow is passing for the PR.

obulat: This works well!
I was going to say that if you run `pnpm test:playwright` without running `pnpm install` previously, you would still get the `no en.json found` error. But I guess you wouldn't be able to run the tests anyway because you wouldn't have any dependencies installed 🤦

obulat (approved):

src/locales/scripts/read-i18n.js
```diff
@@ -0,0 +1,160 @@
...
+ *   // documentation about `key-a`
+ *   key: 'value',
+ *   'key-b': {
+ *     'key-c': 'value-c',
+ *   },
+ * }
+ * ```
+ *
+ * @typedef {{ [key: string]: string | SimJson }} SimJson
+ */
```
sarayourfriend, line 22: This module is clever. Reading it, however, it struck me that you've essentially re-written a very basic version of a JavaScript AST parser. We could use instead use an existing AST parser and treat the `json5` file as regular JavaScript and get the following benefits:

1. No need to maintain a file parser
2. Typed AST
3. Support for multi-line comments

`@babel/parser` does a great job at all of these and is thoroughly documented. Check out this AST Explorer example: https://astexplorer.net/#/gist/fd2baa6817ccaa87c23704ead7d3e7f8/fb97d60a32d361ab1da42d9a19c88a2b22762733

You'll notice the object had to be wrapped in parens. This could be done in code by opening the file and adding an open and closing paren to the object. Alternatively, we could write the file as plain JavaScript and have it like so:

```js
module.exports = {
  // ... the translation object
```

Instead of using the JSON5 extension but only supporting a limited subset of it.

If you click on the object properties in the AST explorer link, you'll see a `leadingComments` array that includes all the leading comments. Navigating through nested blocks is also easy, thanks to the uniform AST. We could transform this into the `Entry` AST for simplicity’s sake and to pare down the unnecessary parts of the babel AST: the goal of using babel's parser is mostly to avoid having to maintain our own AST parser :slightly_smiling_face: 

Finally, for generating the `en.json`, if we use `json5` we can use the `json5` reference implementation to get the comment-free version of the object. If we use JavaScript for the file instead, then we can just import the object and write it out as JSON without the need for an additional package.

None of these suggestions are blocking, but I am hesitant to approve without further clarification for why we should add a simplified JSON5 parser to our codebase.

sarayourfriend (changes requested):

sarayourfriend:
Not necessarily requesting change, just blocking the PR for further discussion.

src/locales/scripts/read-i18n.js
dhruvkb: I searched everywhere for a JSON-AST package but didn't find any that supported JSON5 and had been actively maintained. The JSON5 parsers I found didn't have the option for reading to AST, only parsing it which strips all comments.

I hadn't considered the idea of treating the file as JS and using `@babel/parser` on it. That might be worth investigating for the consistency of having a standard AST, although
- the rudimentary parser is only a _tiny bit_ of a maintenance burden because it's only a bunch of `if` statements
- Babel's parser is overkill for a JSON file and then getting the information we want from the AST it outputs will also add a fair bit of code

For generating the `en.json` I didn't want to use the `json5` parser because we already have all the information we need to write the JSON in the `Entry` object so it made sense to me to just use that to output the JSON.

To be honest, I'm being biased towards the hack parser because I spent a lot of time writing and testing it but I'm open to using a full-fledged parser like Babel if it improves the quality of the PR.

dhruvkb: I refactored `read-i18n.js` to use the `@babel/parser` package and it while the LoC didn't change much, the code looks much cleaner. Thanks again for the suggestion, wish I had asked you for pointers sooner!

src/locales/scripts/read-i18n.js
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/**
+ * Parse the given filename into a tree of `Entry` instances.
+ * @param filename {string} the name of the JSON file to read
+ * @return {Entry} the root `Entry` instance for the top-level JSON object
+ */
+const parseJson = (filename) =>
+  parseObjProperty({
+    value: babel.parseExpression(
```
sarayourfriend, line 159: Nice to use `parseExpression` here and avoid the hacky ones I suggested :sweat_smile:

sarayourfriend (approved):
LGTM! I had a comment typed out yesterday responding to some of your points (which I think are all pretty good) but the one I will still share is that yes, I think while it doesn't reduce the overall amount of code it is code that is (a) more readable and (b) more familiar to most people. Parsers are just a bit more niche compared to tree-walking code, which most people learn pretty early on. It's nice to support multi-line comments as well :grin: I appreciate your willingness to explore other avenues here :rocket:

PR #1955, opened 2022-11-03
Add layout properties to the UI store and cookie to store them

### LINKED ISSUE #974: Create `useUiStateCookie` composable

obulat:
## Fixes

Fixes #974 by @sarayourfriend

## Description

This PR adds the following layout properties to the `ui` pinia store:
- `isDesktopLayout`: whether the site should use the `desktop` or `mobile` layout (i.e., different headers, displaying menus in a modal on mobile or in a popover/sidebar on desktop).
- `isMobileUa`: whether the User Agent is mobile. This is currently not used for layout, and I'm not sure whether we really need it or not. One thing I suspect this will be necessary for is any mobile-specific handling of the audio or video players.
- `innerFilterVisible`: to keep track of the filters state. The components should use `isFilterVisible` as that getter updates the `isFilterDismissed` if the screen width has changed and we need to update the cookie state accordingly.
- `isFilterDismissed`: on a desktop layout, the filter sidebar is always open by default. However, if the user dismisses it, we want to save this preference across requests.

The values are initialized from a cookie (and the request properties) in the `middleware`. Then, on every change of the layout, user agent or filter `dismissed` value, we save the value to the state **and** to the cookie.

## Additional details

Unlike the approach outlined in the linked issue, this PR does not save the exact breakpoint. Instead, it saves whether the layout should be `desktop` or `mobile` (based on the current threshold: `md` for `old_header` and `lg` for `new_header`). This way, we can reuse the `isMinScreen` composable that uses `matchQuery` for this instead of adding a separate `resize` listener. Besides, we were going to use the current breakpoint to compute whether we should use a desktop or mobile layout anyways.

This PR adds ~~two composables~~ one composable:
- ~~to save data in cookies. It uses the NuxtApp `app` as a parameter to enable setting the cookies from outside of the `setup` function, namely, from the `ui` store.~~ This turned out not necessary because Pinia Nuxt plugin can access `app` using `this.$nuxt.app`.
- to update the layout values. This functionality should be used in the layout files that wrap all other components: the `default.vue` and `blank.vue` layouts.

## Testing Instructions

I've added the unit tests for the `ui` store. The changes to the store will also be tested in #1956 which uses the `ui` store to set the layout properties.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/stores/ui.ts
```diff
@@ -1,33 +1,148 @@
...
+export const VISIBLE = 'visible'
+export const DISMISSED = 'dismissed'
+export const NOT_SHOWN = 'not_shown'
+const snackbarStates = [NOT_SHOWN, VISIBLE, DISMISSED] as const
+export type SnackbarState = typeof snackbarStates[number]
 
+interface UiStateCookie {
+  isDesktopLayout?: boolean | undefined
+  isMobileUa?: boolean | undefined
+  isFilterDismissed?: boolean | undefined
```
sarayourfriend, line 16: The `undefined` is redundant (sort of, but not entirely) with the `?` notation on the property name.

Are there cases where the property will exist but be set to the value `undefined`? If that is the case then this is the correct type. Otherwise, if any time the property exists it will be a boolean value, then we can leave undefined off the righthand side.

```suggestion
  isDesktopLayout?: boolean
  isMobileUa?: boolean
  isFilterDismissed?: boolean
```

src/stores/ui.ts
```diff
@@ -1,33 +1,148 @@
...
-export type SnackbarState = 'not_shown' | 'visible' | 'dismissed'
+/**
+ * By default, the filters are shown on desktop, and hidden on the mobile widths.
+ * `dismissed` is used when the filters are closed on the desktop width.
+ */
+export const VISIBLE = 'visible'
+export const DISMISSED = 'dismissed'
+export const NOT_SHOWN = 'not_shown'
+const snackbarStates = [NOT_SHOWN, VISIBLE, DISMISSED] as const
+export type SnackbarState = typeof snackbarStates[number]
```
sarayourfriend, line 11: I have been thinking about this pattern a bit and I think we are probably slightly over-engineering this pattern (and not relying on our nice TypeScript tools) by defining constants and then using them.

If we could rely just on TypeScript, we could do just this:

```suggestion
const snackbarStates = ['not_shown', 'visible', 'dismissed'] as const
export type SnackbarState = typeof snackbarStates[number]
```

And then anywhere that accepts `SnackbarState`, auto-complete hints will show you the strings that are valid. What's more, in refactoring, if the SnackbarState changes to no longer accept some value, TypeScript will tell us this in type checking.

Following this pattern would remove some redundant lines of code (the `export const` declarations) and redundant imports.

On the other hand, I could see this pattern feeling easier to document.

But anyway, I just wanted to share some thoughts about it here (no change requested or suggested). Is this something I should open an issue to discuss, do you think you (and others) would be open to changing this pattern across the app?

src/stores/ui.ts
```diff
@@ -1,33 +1,148 @@
...
+        return
+      this.isDesktopLayout = isDesktopLayout
+      // Only update `isMobileUa` if the value is provided
+      if (typeof isMobileUa !== 'undefined') {
+        this.isMobileUa = isMobileUa
+      }
+      return {
+        isDesktopLayout: this.isDesktopLayout,
+        isMobileUa: this.isMobileUa,
+      }
```
sarayourfriend, line 113: Would just like to add some whitespace and additional curlies here to clarify what is happening. The naked conditional `return` was confusing to read :confounded: 

```suggestion
      if (
        this.isDesktopLayout === isDesktopLayout &&
        typeof isMobileUa === 'undefined'
      ) {
        return
      }

      this.isDesktopLayout = isDesktopLayout

      // Only update `isMobileUa` if the value is provided
      if (typeof isMobileUa !== 'undefined') {
        this.isMobileUa = isMobileUa
      }
 
      return {
        isDesktopLayout: this.isDesktopLayout,
        isMobileUa: this.isMobileUa,
      }
```

src/stores/ui.ts
```diff
@@ -1,33 +1,148 @@
...
+     */
+    updateBreakpoint(
+      isDesktopLayout: boolean,
+      isMobileUa: boolean | undefined
+    ): ScreenParameters | undefined {
+      if (
+        this.isDesktopLayout === isDesktopLayout &&
+        typeof isMobileUa === 'undefined'
+      )
+        return
```
sarayourfriend, line 104: What does this condition mean? Can you add a comment explaining it? Also worth documenting: why is `isMobileUa` allowed to be undefined?

src/stores/ui.ts
```diff
@@ -1,33 +1,148 @@
...
+        isMobileUa: this.isMobileUa,
+      }
+    },
+    /**
+     * Sets the filter state based on the `visible` parameter.
+     * Closing the filters sets the state to `dismissed` on desktop,
+     * but `shown_on_desktop` on mobile.
+     *
+     * The calling function must be inside the `setup` function and
+     * must set the cookies if the return value is not `null`.
```
sarayourfriend, line 121: Is this because of the `$cookies` limitation as well?

sarayourfriend (commented):
This is exciting and looking great!

src/stores/ui.ts
obulat: In the first iteration, I also had an array of constants for the filters state (`default` or `shown_on_desktop`, `visible` and `dismissed`). Then, I had to rename one of them, and saw that refactoring without constants is much more difficult: it is much easier to miss one of the usages because you use "search and replace" instead of the IDE tooling to `refactor`/`rename`. This is why I converted the strings into constants: for ease of refactoring both in the main code and in tests.

I don't have a very strong preference for this pattern, so I would love to hear everyone's opinion.

src/stores/ui.ts
obulat: Yes :( I tried passing the `app.$cookies.set` as a function parameter to this action to overcome this limitation, but it didn't work, I still got the `this function must be called from inside the setup function` warning.

src/stores/ui.ts
obulat: We might not really need the constants here because these specific constants are not used in many places.

src/stores/ui.ts
obulat: I've updated the `isMobileUa` handling, cookie handling, and the structure of this method as a whole. Thank you for giving ideas and pushing me to improve it all!

src/stores/ui.ts
obulat: Thank you for the explanation. I've reviewed the usage and removed the unnecessary `| undefined` from here.

sarayourfriend (commented):
@obulat Were you thinking of refactoring `useCookies`? I think the API could be improved if we used the approach I suggested elsewhere. Rather than passing the `name` parameter to `useCookies`, we could `get` and `set` the cookie by name in the closure functions the composable produces. It would also remove the need to have to pass an explicit generic parameter to `useCookies` like we'll need to now to get the right types for the closures.

Additionally, I am wondering if `useCookies` needs to expose a reactive object for the cookies in case the values change. For example, consider if we are relying on the value of the `isDesktopLayout` value rather than computed breakpoints. If we adjust the width of the viewport then will `isDesktopLayout` be recomputed (to false, if we go below a desktop layout width) and then cause a re-render? Right now I think the value might be static, even with the Pinia store, because the store values aren't hooked in (as far as I can tell) to a media query watcher.

obulat:
> @obulat Were you thinking of refactoring `useCookies`? I think the API could be improved if we used the approach I suggested elsewhere. Rather than passing the `name` parameter to `useCookies`, we could `get` and `set` the cookie by name in the closure functions the composable produces. It would also remove the need to have to pass an explicit generic parameter to `useCookies` like we'll need to now to get the right types for the closures.
> 
> Additionally, I am wondering if `useCookies` needs to expose a reactive object for the cookies in case the values change. For example, consider if we are relying on the value of the `isDesktopLayout` value rather than computed breakpoints. If we adjust the width of the viewport then will `isDesktopLayout` be recomputed (to false, if we go below a desktop layout width) and then cause a re-render? Right now I think the value might be static, even with the Pinia store, because the store values aren't hooked in (as far as I can tell) to a media query watcher.

Yes, I will refactor based on your suggestion, @sarayourfriend.

As for the second question, I've added the `isMinScreen` composable to the 2 layouts we use, with a watcher that sets the cookies when the value changes. The changes are in the companion PR, #1956
```
 const isMobileUa = useBrowserIsMobile()
const shouldPassInSSR = uiStore.isDesktopLayout ?? !isMobileUa
const desktopBreakpoint = isNewHeaderEnabled.value ? 'lg' : 'md'
const isDesktop = isMinScreen(desktopBreakpoint, { shouldPassInSSR })
const uiCookies = useCookies<UiStateCookie>(app, 'ui')
watch(isDesktop, (isDesktop) => {
      uiStore.updateBreakpoint(isDesktop, isMobileUa, uiCookies.setCookie)
    })
```

'reactive object for the cookies' sounds really interesting, and I will see if that can be used during the refactoring.

src/composables/use-media-query.ts
```diff
@@ -16,7 +20,7 @@ interface Options {
  * Reactive Media Query.
  */
 export function useMediaQuery(
-  query: string,
+  query: MaybeComputedRef<string>,
```
obulat, line 23: To update the desktop threshold based on the `new_header` flag status, we need to pass a computed value here.

src/middleware/middleware.ts
```diff
@@ -46,5 +48,12 @@ const middleware: Middleware = async ({ app, query, route, $pinia }) => {
   const featureFlagStore = useFeatureFlagStore($pinia)
   featureFlagStore.initFromCookies(app.$cookies.get('features') ?? {})
   featureFlagStore.initFromQuery(query)
+
+  /* UI store */
+
+  const uiStore = useUiStore($pinia)
+  const isMobileUa = app.$ua ? app.$ua.isMobile : false
```
obulat, line 55: It is expected that `isMobileUa` can be set once per request, and we do not need to change its value or update a cookie after that.

src/stores/ui.ts
```diff
@@ -26,8 +71,62 @@ export const useUiStore = defineStore('ui', {
...
+    },
+
+    /**
+     * If the breakpoint is different from the state, updates the state, and saves it into app cookies.
+     *
+     * @param isDesktopLayout - whether the layout is desktop (`lg` with the `new_header`
+     * and `md` with the `old_header`).
+     * @param setCookieFn - sets the app cookie.
+     */
+    updateBreakpoint(isDesktopLayout: boolean, setCookieFn: CookieSetter) {
```
obulat, line 102: We could add a check for `isMobileUa` here and only set the `isDesktopLayout` to true if `isMobileUa` is false. However, I'm not sure if it makes sense to use a mobile layout for `lg` mobile screens.

src/stores/ui.ts
```diff
@@ -26,8 +71,62 @@ export const useUiStore = defineStore('ui', {
...
+     * populate the store state to match the cookie.
+     *
+     * @param cookies - mapping of UI state parameters and their states.
+     */
+    initFromCookies(cookies: OpenverseCookies) {
+      this.isDesktopLayout = cookies.uiIsDesktopLayout ?? false
+      this.isFilterDismissed = cookies.uiIsFilterDismissed ?? false
+
+      // Middleware sets the cookie value before calling `initFromCookies`.
+      this.isMobileUa = cookies.uiIsMobileUa ?? false
```
obulat, line 89:
```suggestion
      this.isFilterDismissed = cookies.uiIsFilterDismissed ?? false
      this.isMobileUa = cookies.uiIsMobileUa ?? false
```

src/composables/use-media-query.ts
```diff
@@ -49,7 +53,7 @@ export function useMediaQuery(
 
     cleanup()
 
-    mediaQuery = window.matchMedia(query)
+    mediaQuery = window.matchMedia(resolveRef(query).value)
```
sarayourfriend, line 56: I think [Vue's built in `unref`](https://vuejs.org/api/reactivity-utilities.html#unref) makes more sense here than `resolveRef`. It seems like `resolveRef` is more useful for "make this thing a ref, one way or another" rather than a shortcut to safely getting a value out of something that may or may not be a ref :thinking: Most importantly, `unref` is a simple-ish if check and doesn't provision any new ref objects; however, `resolveRef` might actually create new ref objects which could lead to another memory leak like we had before with repeated ref creation.

src/composables/use-media-query.ts
```diff
@@ -66,21 +70,28 @@ export function useMediaQuery(
...
 /**
  * Check whether the current screen meets
  * or exceeds the provided breakpoint size.
  */
-export const isMinScreen = (breakpointName: Breakpoint, options?: Options) => {
+export const isMinScreen = (
+  breakpointName: MaybeComputedRef<Breakpoint>,
+  options?: Options
+) => {
   if (breakpointName === 'xs') {
```
sarayourfriend, line 82: Does this also need `unref`?

src/stores/ui.ts
sarayourfriend: I think you're right, that it doesn't make sense. To me, it would make sense to rely exclusively on the breakpoint information _if we have it_ and fall back to UA sniffing only when we don't yet know any breakpoint information from the cookie.

sarayourfriend (changes requested):
This LGTM! Tests well locally. Only requested change is to use `unref` instead of `resolveRef` in the watcher/effect callbacks especially.

src/composables/use-media-query.ts
obulat: Thank you for catching this! I don't know why there was no warning.

src/composables/use-cookies.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import type { OpenverseCookies } from '~/types/cookies'
+
+import type { NuxtAppOptions } from '@nuxt/types'
+import type { CookieSerializeOptions } from 'cookie'
+
+const cookieOptions: CookieSerializeOptions = {
+  path: '/',
+  sameSite: 'strict',
+  maxAge: 60 * 60 * 24 * 60,
```
dhruvkb, line 17: Would be nice to have a comment `// 60 days` here.

src/composables/use-cookies.ts
```diff
@@ -0,0 +1,38 @@
...
+  secure: isProd,
+}
+
+export const useCookies = (app: NuxtAppOptions) => {
```
dhruvkb, line 21: Naïve question: why not use the [`useCookies` composable](https://vueuse.org/integrations/usecookies/#usecookies) from VueUse?

src/types/cookies.ts
```diff
@@ -0,0 +1,29 @@
+import type { FeatureState } from '~/constants/feature-flag'
+
+export type SnackbarState = 'not_shown' | 'visible' | 'dismissed'
+
+/**
+ * The cookies that Openverse uses to store the UI state.
+ */
+export interface OpenverseCookies {
```
dhruvkb, line 8: Nit: the interface name should be singular noun. Maybe `OpenverseCookieState`?

dhruvkb (approved):
LGTM, works just as expected locally.

src/composables/use-cookies.ts
obulat: Thank you fro asking this question, @dhruvkb !

I've started answering why we can't use it, and then found a much nicer solution.

We use the [`cookie-universal-nuxt`](https://github.com/microcipcip/cookie-universal/tree/master/packages/cookie-universal-nuxt) which injects `$cookies` to the Nuxt `app`. This means that we can only set cookies in components and not in a Pinia store because you can't access `app` from outside of a `setup` function.

That said, researching this issue, I've found that `$nuxt` [was added to the Pinia Nuxt plugin](https://github.com/vuejs/pinia/issues/186#issuecomment-810950917), so we can set the cookies from the store by calling `this.$nuxt.app.$cookies.set()`.

src/composables/use-media-query.ts
sarayourfriend: Excellent! TIL `unref` doesn't work for computed refs. Thanks for looking further into this.

sarayourfriend (approved):
Changes look great to me, always nice to cut out some extra code (though I have to admit I found the composable nice to work with, especially w/r/t to typing).

To get `cookies-universal-nuxt`'s types working, can you please add `cookies-universal-nuxt` to the `types` array in `tsconfig.json`? Otherwise `$cookies` is typed as `any` :scream:

PR #1956, opened 2022-11-03
Use the UI store properties to set the layout properties

obulat:
## Fixes

Fixes #[issue number] by @[issue author]

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/composables/use-browser-detection.ts
```diff
@@ -5,7 +5,7 @@ export const useBrowserDetection = () => {
   return app.$ua
 }
 
-export const useBrowserIsMobile = () => {
+export const useBrowserIsMobile = (): boolean => {
```
obulat, line 8: TS should be able to derive the type here, but for some reason WebStorm doesn't pick up the type. Adding this annotation helps

src/plugins/ua-parse.ts
```diff
@@ -10,7 +10,7 @@ const uaParsePlugin: Plugin = (context, inject) => {
   } else if (typeof navigator !== 'undefined') {
     userAgent = navigator.userAgent
   }
-  let ua
+  let ua: UADetails | null
```
obulat, line 13: Same as above, these annotations help WebStorm pick up the types

src/components/VHeader/VHeaderInternal.vue
```diff
@@ -107,8 +99,11 @@ export default defineComponent({
       }
     })
 
-    watch(isMinScreenMd, (isMd) => {
-      if (isMd && isModalVisible.value) {
+    // We cannot use the `isDesktopLayout` from the ui store here because
+    // the threshold there is `lg`, and in `VHeaderInternal` it is `md`.
```
sarayourfriend, line 103: `useLayout` checks for the feature flag, so shouldn't the UI store already be aware of this difference and compensating for it correctly? Maybe I'm not fully understanding how `useLayout` and the ui store are meant to work together.

src/layouts/default.vue
```diff
@@ -86,54 +87,54 @@ const embeddedPage = {
...
+
+    /**
+     * If cookies state does not match the state on mounted
+     * (e.g., if the cookies were saved when the screen was `sm`,
+     * and then a page is opened on SSR on a `lg` screen),
+     * we update the cookie state.
+     */
+    onMounted(() => {
+      updateLayout(isDesktop.value)
+    })
```
sarayourfriend, line 116: Could/should this be moved into the `useLayout` composable itself? I don't have a super strong inclination either way other than moving it into the composable would maybe localise the full functionality of the composable whereas this is a little more spread out. It also requires duplicating this in all the root layouts, which could be easy to accidentally forget.

sarayourfriend (approved):
This is _so_ exciting! It's amazing how much this already improves the rendering experience of the search results page :cowboy_hat_face:  :rocket:

src/components/VHeader/VHeaderInternal.vue
obulat: With the new header enabled, all the pages change the layout from mobile to desktop when they reach `lg` breakpoint. However, `VHeaderInternal`'s layout changes happens when the breakpoint is `md`. So, the `useLayout`'s `isDesktopLayout` will not be useful here (with the new header enabled) because the thresholds for desktop are different.
I wish I could write a clearer comment...

src/layouts/default.vue
obulat: I've learned with the filters visibility composable, if you have `onMounted` inside a composable, it runs whenever any component that uses it is mounted. While `useLayout` is only used in the two `layout` components, I still feel that it's easier to see when the `onMounted` runs.
While we can accidentally forget to add `onMounted` to the root layout, it might be more difficult to debug if we accidentally import `useLayout` into a non-root layout file and then it runs `onMounted` when we don't need it.

src/components/VHeader/VHeaderInternal.vue
sarayourfriend: Ahhh, I see! It was probably my bad assumption, I assumed that once the new header was enabled all the "desktop" breakpoints would be switching to the new one, I didn't realise it was fragmented in different parts of the page.

I wonder if we should just put the breakpoint itself into the cookie instead of doing the boolean `isDesktopLayout`, given "desktop layout" appears to potentially mean different things depending on which part of the view we're talking about.

My original intention in avoiding putting the breakpoint name into the cookie was to reduce the possibility for snooping on it to identify someone (moreso to increase user confidence than anything), but broad breakpoint range is not really that useful for fingerprinting anyway. Do you think it is worth the effort? Otherwise, I fear that anything that needs to know if we're in a specific breakpoint would be stuck not getting the benefits of the UI cookies for SSR.

obulat: > I assumed that once the new header was enabled all the "desktop" breakpoints would be switching to the new one, I didn't realise it was fragmented in different parts of the page.

I assumed the same before looking closely at the mockups :) Here you can see the different thresholds:
Screenshot 2022-11-16 at 10 18 33 AM

> I wonder if we should just put the breakpoint itself into the cookie

I wanted to do that at first, as it would make having different thresholds for layouts easier.

The current implementation of the composables use `window.matchMedia`, so it does not return us the current breakpoint on screen change, it returns `true`/`false` when the mediaQuery like 'min-width: 640px' matches or doesn't match the current window.
To set the current breakpoint in the cookie, we would either have to use several `matchMedia` watchers, or watch `resize` event. I suspect that it would be less efficient than a single `matchMedia` event.

We also were going to discuss the pages and which ones of them are linked in the header/footer. I wonder if the threshold for the `hamburger` might change... @panchovm, is there such as possibility that we might have the `hamburger` menu for the page links up to the `lg` screen? Or is it certain that we will only have it until `md`?

panchovm: > is there such as possibility that we might have the hamburger menu for the page links up to the `lg` screen? Or is it certain that we will only have it until `md`?

My main concern with keeping the burger menu up to `lg` is that, on wider devices, the internal menu will look huge. Imagine this mockup on a 1024-px screen.



On the other hand, I agree that we might need more space for showing links as the content needs are bigger than before.

I think we can keep the hamburger menu up to `lg` and improve the internal menu view once we define the pages listed. Some more in line with the current menu.

CleanShot 2022-11-16 at 10 31 54@2x

sarayourfriend: > To set the current breakpoint in the cookie, we would either have to use several matchMedia watchers, or watch resize event. I suspect that it would be less efficient than a single matchMedia event.

A single (debounced) resize event listener should not be significantly less efficient than a matchMedia event unless there's something really tricky going on between those two. At minimum, four matchMedia events to catch each relevant breakpoint wouldn't be worse than what we were doing before with multiple matchMedia events against the `md` breakpoints.

Of course both options are less efficient than a single matchMedia event, but I don't think significantly so (I could be wrong about that, though).

Given the trade-off here pretty minimal I'm not blocking this PR on this suggested change but I do think it would be a good ergonomic change to the cookie that would allow more flexibility like this into the future without much hassle if needed. Perhaps we can leave a note somewhere in the code of the UI store explaining this alternative approach and asking for it to be implemented if additional matchMedia events need to be added in the future (_en lieu_ of implementing it now, that is).

zackkrida: > Perhaps we can leave a note somewhere in the code of the UI store explaining this alternative approach and asking for it to be implemented if additional matchMedia events need to be added in the future (en lieu of implementing it now, that is).

➕ to this, could one of you make a new issue for it?

zackkrida (approved):
Excellent work. Should we wait to deploy this until the privacy changes are live? I think that would probably be best.

sarayourfriend:
@obulat Will you want a re-review of this? It seems like some things have changed significantly and I see changes being pushed regularly here, but I'm not sure when/if you're looking for another review.

src/stores/ui.ts
```diff
@@ -24,32 +29,42 @@ export interface UiState {
...
 }
 
-const cookieOptions: CookieSerializeOptions = {
-  path: '/',
-  sameSite: 'strict',
-  maxAge: 60 * 60 * 24 * 60, // 60 days
-  secure: isProd,
-}
+export const screenSizes = [...Object.values(SCREEN_SIZES), 0]
+export const breakpoints = [...Object.keys(SCREEN_SIZES), 'xs']
```
zackkrida, line 43: A comment might be nice to explain the 'fake' minimum values added to the lists here.

sarayourfriend (approved):
LGTM. Confused if/why `isBreakpoint` works given what Zack pointed out :thinking:

src/components/VContentSwitcherOld/VSearchTypeButtonOld.vue
```diff
@@ -66,12 +68,16 @@ export default defineComponent({
   },
   setup(props) {
     const i18n = useI18n()
-    const isHeaderScrolled = inject('isHeaderScrolled', ref(null))
-    const isMinScreenMd = isMinScreen('md', { shouldPassInSSR: true })
+
+    const uiStore = useUiStore()
+
+    const isHeaderScrolled = inject('isHeaderScrolled', ref(false))
```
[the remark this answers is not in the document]
obulat: I was thinking that `isHeaderScrolled` would not be necessary after we migrate to the new header. We will only need it to show/hide the bottom border, and that can probably be done in the header's parent component. I'm not sure if there are other places that use it, I'd have to check.

obulat:
> Confused if/why `isBreakpoint` works given what Zack pointed out

Could you clarify which point that is?

zackkrida:
@obulat I suspect Sara is referring to this comment about '>' vs. '>=':

https://github.com/WordPress/openverse-frontend/pull/1956/files/66016f81b188762cc39278b759416ac7d31d88ad#r1035469843

src/components/VContentSwitcherOld/VSearchTypeButtonOld.vue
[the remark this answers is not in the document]
sarayourfriend: Ah right on, if it's going to be removed entirely soon then that's even better :grin:

PR #1875, opened 2022-10-03
Implement the composite search bar for the mobile search header

obulat:
## Fixes

Fixes #1876 by @obulat

## Description

This PR implements the new VHeaderMobile component that is used on mobile widths for the search result and single result pages.
It does not use the `VSearchBar` or `VInputField` components, copying some code from them instead. Re-using the existing components is difficult because of the different border treatments (in the current component, everything is inside an outer border, and in the existing VSearchBar, some components are outside of the border), and nesting several components and several levels of template `ref`s is difficult to follow. We could extract the common functionality in the future, if that makes the code simpler.

## Testing Instructions

Go to 'http://localhost:8443/search/?q=cat&ff_new_header=on' on a narrow screen. The search bar should behave correctly. You should be able to update the search term, clear it, open (a stub of) the `VContentSettingsModal`.

You can also check out the VHeaderMobile story: https://wordpress.github.io/openverse-frontend/_preview/1875/?path=/story/components-vheader-vheadermobile-vheadermobile--default-story

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [ ] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

[...]

PR #1977, opened 2022-11-16
Fix "Back to results" link and remove grey background from single results page

### LINKED ISSUE #1974: Remove grey background behind single result media when the new header is enabled
## Problem

Currently while the new header is enabled, we continue to show a grey background behind images and audio tracks. This background also no longer spans the full width of the page due to layout changes.

## Description

We should remove these backgrounds and make any other necessary adjustments to make the pages look like the mockups when the new header is open. See figma links for the most up-to-date version. Screenshots are subject to changes:

## Single Result Audio ([Figma](https://www.figma.com/file/KxkZmBNk218V2Fke1GeYCu/New-header-component?node-id=609%3A67242&t=zKdLp5Ge9nZ8Yo0f-4))

CleanShot 2022-11-15 at 09 44 47

## Single Result Image ([Figma](https://www.figma.com/file/KxkZmBNk218V2Fke1GeYCu/New-header-component?node-id=609%3A69561&t=zKdLp5Ge9nZ8Yo0f-4))

CleanShot 2022-11-15 at 09 45 37
"
The "Back to results" button padding should also be revised as part of this PR, as it is currently incorrect.

## Implementation

dhruvkb:
## Fixes

Fixes #1974 by @zackkrida 

## Description

This PR
- simplifies the code of the "Back to results" button
- adds a story for the "Back to results" button
- removes the grey background from single image and audio when new header is enabled
- avoids any major UI change when new header is disabled

## Testing Instructions

See that the issue #1974 has been resolved, without tampering the old header UI.

0. Compare the single audio and single image results against the mockups.
1. Enable the new header
0. Compare the single audio and single image results against the new header mockups.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VBackToSearchResultsLink.vue
```diff
@@ -1,10 +1,10 @@
 <template>
   <!-- @todo: Separate the absolute container from the link itself. -->
   <VLink
-    class="flex flex-row items-center px-2 py-3 text-xs font-semibold text-dark-charcoal md:px-6 md:pt-4 md:pb-2 md:text-sr"
-    :href="path"
+    class="time inline-flex flex-row items-center gap-2 rounded-sm p-2 text-xs font-semibold text-dark-charcoal-70 pe-3 hover:text-dark-charcoal"
+    v-bind="$attrs"
```
obulat, line 5: Why did you set `inheritAttrs` to false and added `v-bind="$attrs"` to the outermost component? Isn't that the default behavior anyways? if you don't set `inheritAttrs` to false, all the `attrs` should automatically bind to `VLink` here, shouldn't they?

I've seen the same thing in some other component, so I wanted to ask about it :)

dhruvkb: That is the expected behaviour but for some reason it doesn't work. If you remove these two bits, the `href` isn't passed as a prop but as an attribute which in the end results in all links opening in new tab.

test/playwright/e2e/image-detail.spec.ts
```diff
@@ -37,7 +37,7 @@ test('shows the main image with its title as alt text', async ({ page }) => {
 
 test('does not show back to search results breadcrumb', async ({ page }) => {
   await goToCustomImagePage(page)
-  await expect(page.locator('text="Back to search results"')).not.toBeVisible({
+  await expect(page.locator('text="Back to results"')).not.toBeVisible({
```
obulat, line 40:
```suggestion
  await expect(
    page.locator(`text="${t('single-result.back')}"`)
  ).not.toBeVisible({
```

obulat: Would be nice to use the translated version here and in the other two tests.

You'd also need to import `t`:
`import { t } from '~~/test/playwright/utils/navigation'`

obulat (commented):
When I checked out this PR, and ran Playwright tests without running pnpm dev first, I get errors locally because the en.json5 was updated, but en.json wasn't.
If we add `pnpm run i18n:get-translations --en-only` to playwright.sh, we'll always update the generated json file before running the tests.

src/components/VBackToSearchResultsLink.vue
```diff
@@ -1,10 +1,10 @@
...
   >
-    <VIcon :icon-path="chevronIcon" class="-ms-2" />
+    <VIcon :icon-path="chevronIcon" />
```
obulat, line 7:
```suggestion
    <VIcon :icon-path="chevronIcon" :rtl-flip="true" />
```

obulat: Not related to this PR, but is a quick fix that can be added here.

src/pages/audio/_id.vue
```diff
@@ -1,9 +1,17 @@
...
+  <main>
+    <div
+      v-if="backToSearchPath"
+      class="w-full py-2"
+      :class="{ 'px-2 md:px-6': !isNewHeaderEnabled }"
+    >
+      <VBackToSearchResultsLink :href="backToSearchPath" />
+    </div>
+
+    <!-- Negative margin compensates for the `!px-6` padding on <Nuxt/> component. -->
```
obulat, line 11: I'm not sure why we have `!px-6` on the `<Nuxt>` component. Can we remove it instead of adding extra `-mx-6` here. This will also fix the current excessive padding on the mobile: It should have 24px horizontal padding, and now has twice that.

obulat (changes requested):
I think it would make sense to remove the `px-6` padding from the `Nuxt` component as it will simplify layout.
Also, could you please add a VR test for the single result pages (ideally, for both old and new header, but even only new header would be great)?

src/pages/audio/_id.vue
dhruvkb: @obulat I was hoping to ask you about it 😄 because you added it in a previous commit. We could remove that and let the pages handle any horizontal margin themselves.

test/playwright/e2e/image-detail.spec.ts
dhruvkb: Why is the `t` located in `utils/navigation`? I looked for it but didn't expect the navigation util to have it.

src/pages/audio/_id.vue
obulat: I've added it here https://github.com/WordPress/openverse-frontend/pull/1875, but I really don't understand why 😕 Please go ahead and remove it :)

src/components/VBackToSearchResultsLink.vue
```diff
@@ -1,10 +1,10 @@
 <template>
   <!-- @todo: Separate the absolute container from the link itself. -->
   <VLink
-    class="flex flex-row items-center px-2 py-3 text-xs font-semibold text-dark-charcoal md:px-6 md:pt-4 md:pb-2 md:text-sr"
-    :href="path"
+    class="time inline-flex flex-row items-center gap-2 rounded-sm p-2 text-xs font-semibold text-dark-charcoal-70 pe-3 hover:text-dark-charcoal"
```
zackkrida, line 4: What is the "time" CSS class?

dhruvkb: That's the name of the font styled used by @panchovm in the Figma mockups. If I had to guess, it started as the font-style used in the audio timestamps but then found use in other places without updating the name.

panchovm: I think you are referring to the [Time pill component](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=2484%3A22545&t=rP5ZmCU9JdBEliSA-4). I am seeing the Design Library and that is the only component named like that and used in all audio component's variants.

CleanShot 2022-11-22 at 10 12 20@2x

obulat:
Snapshots are great! I can see a couple of problems on the pages:
1. 'Audio information' and 'Realated Audio' headings are not bold(test/playwright/visual-regression/pages/pages-single-result.spec.ts-snapshots/audio-ltr-from-search-results-lg-linux.png). It would be best to add the same classes as on the image page, replacing this:
https://github.com/WordPress/openverse-frontend/blob/0c1caa0a8784114fb4969dd1623ccb391415ad8b/src/components/VAudioDetails/VAudioDetails.vue#L4
with this:
https://github.com/WordPress/openverse-frontend/blob/0c1caa0a8784114fb4969dd1623ccb391415ad8b/src/components/VImageDetails/VImageDetails.vue#L4

Although we should select a single threshold breakpoint instead of using `lg` on one page and `md` on the other.
2. The padding are uneven: Related audio section's horizontal padding are smaller than the padding of other page parts: test/playwright/visual-regression/pages/pages-single-result.spec.ts-snapshots/audio-ltr-from-search-results-sm-linux.png

I'm not sure if it's best to fix here, or to open separate issues, though.

obulat (approved):
Looks great, and it's nice to see the changes in the snapshots!

zackkrida (approved):
Works great now. LGTM.

src/components/VBackToSearchResultsLink.vue
dhruvkb: I was referring to this:
Screenshot 2022-11-22 at 8 01 29 PM

panchovm: Oh, I got it. And yes, it was named time as it was created for the time pill.

PR #1984, opened 2022-11-18
Make z-index usage in JavaScript consistent with Tailwind

### LINKED ISSUE #1742: Standardize z-index layers by importing the tailwind config in JS/TS files
## Problem

We currently use bespoke z-indices in a few places, or manually specify our z-index scale to validate props in a few different files.

## Description

Instead, we should import our tailwind config with something like this:

```js
import resolveConfig from 'tailwindcss/resolveConfig'
import tailwindConfig from './tailwind.config.js'

const myconfig = resolveConfig(tailwindConfig)
```

That would allow us to access the list of valid z indices.

## Implementation

dhruvkb:
## Fixes

Fixes #1742 by @zackkrida 

## Description

This PR moves z-index constants to a JavaScript file `~/constants/z-indices.js` which is then used across the codebase as the source of truth for z-indices and is also loaded into the Tailwind configuration.

This implementation is similar to that used for screen-sizes and does not use the Babel preval approach recommended by the Tailwind docs for two reasons
- avoiding the unnecessary complexity of a Babel plugin
- maintaining parity with the existing screens approach

## Testing Instructions

0. Run the dev server
1. Ensure that the popovers are working as expected (with a z-index of 50)

If you see any other magic numbers for z-index floating around, please report.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/components/VPopover/VPopoverContent.vue
```diff
@@ -92,11 +94,9 @@ export default defineComponent({
       validate: (v: string) => ['absolute', 'fixed'].includes(v),
     },
     zIndex: {
-      type: Number,
+      type: [String, Number],
```
obulat, line 97: Nice devex improvement!

src/constants/z-indices.js
```diff
@@ -0,0 +1,38 @@
...
+ * @param value {string|number} the provided z-index to validate
+ * @return {boolean} whether the z-index is valid and configured in Tailwind
+ */
+const zIndexValidator = (value) =>
+  Object.keys(Z_INDICES).includes(value.toString())
+
+/**
+ * This module is consumed by Nuxt and Tailwind config, so it needs to use CJS modules.
+ */
+module.exports = {
```
obulat, line 35: The same comment applies here as in the Screens PR: how could we move away from CommonJS modules?

src/constants/z-indices.js
```diff
@@ -0,0 +1,38 @@
...
+const Z_INDICES = {
+  auto: 'auto',
+  10: 10,
+  20: 20,
+  30: 30,
+  40: 40,
+  50: 50,
+  // Named indices
+  popover: 50,
+}
```
sarayourfriend, line 21: This might as well be `const`. Typing it this way actually results in _less_ information for the type system as it collapses all the keys to `string`.

```suggestion
 * @type {const}
 */
const Z_INDICES = {
  auto: 'auto',
  10: 10,
  20: 20,
  30: 30,
  40: 40,
  50: 50,
  // Named indices
  popover: 50,
}
```

Though ideally it'd be frozen as well.

```suggestion
 */
const Z_INDICES = Object.freeze(/** @type {const} */ ({
  auto: 'auto',
  10: 10,
  20: 20,
  30: 30,
  40: 40,
  50: 50,
  // Named indices
  popover: 50,
}))
```

src/constants/z-indices.js
sarayourfriend: @obulat I think the limitation is in Tailwind. [Tailwind only supports CJS modules for the configuration file](https://tailwindcss.com/docs/configuration). The only alternatives I can think of are to add a build step for the Tailwind configuration that would take a `tailwind.config.ts` or `.mjs` and output the `tailwind.config.js` that Tailwind would actually consume. This adds a bit of complexity that we can just ignore by using CJS for these few specific files that are likely not touched that often.

sarayourfriend (approved):
LGTM. I tested the z-index edge cases I know about: filter's popover with the scrolled header and the report content popover with the pages menu open. Both behave the same as in production.

PR #1985, opened 2022-11-18
Refactor `SCREEN_SIZES` to match `z-indices.js`

dhruvkb:
## Description

This PR changes `SCREEN_SIZES`, which is currently a `Map` into a `Record`, which makes it more ergonomic to use and improves type-safety.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/constants/screens.js
```diff
@@ -1,43 +1,60 @@
 /**
- * This file cannot be converted to TypeScript because it's used by the Tailwind config
+ * This file cannot be converted to TypeScript because it's referenced in the
+ * Tailwind config.
+ *
+ * @typedef {'2xl' | 'xl' | 'lg' | 'md' | 'sm' | 'xs'} Breakpoint
+ * @typedef {Exclude<Breakpoint, 'xs'>} RealBreakpoint
```
obulat, line 6: It would be nice to explain why `xs` is not a `real` breakpoint. Or find a clearer name here.

test/playwright/utils/navigation.ts
```diff
@@ -324,7 +322,7 @@ export const selectHomepageSearchType = async (
   dir: LanguageDirection = 'ltr'
 ) => {
   const pageWidth = page.viewportSize()?.width
-  if (pageWidth && pageWidth > smWidth) {
+  if (pageWidth && pageWidth > SCREEN_SIZES.sm) {
```
obulat, line 325: I love how simple this became 👍

src/constants/screens.js
```diff
@@ -1,43 +1,60 @@
...
+          name: `${key} (${val}px)`,
+          styles: { width: `${val}px`, height: '768px' },
+        },
+      ])
   )
+)
 
 /**
- * This module is consumed by Nuxt and Tailwind config so it needs to use CJS modules
+ * This module is consumed by Nuxt and Tailwind config, so it needs to use CJS modules.
```
obulat, line 54: Nuxt 3 expects no CJS modules. Can you think of a better way of having a single source of truth for screens for both Nuxt and Tailwind without using a CJS module? This is not a comment for this PR, just a note since you're touching these files.

obulat (approved):
I like the ease with which we can use the screens. Approving, but would like to see more documentation about why `xs` is not real 😆

zackkrida (approved):
Agreed with @obulat about the special handling of `xs` being confusing (I sadly don't remember why it's an exception, so documentation would be useful) but this LGTM.

src/constants/screens.js
dhruvkb: Added a comment to explain `RealBreakpoint` vs `Breakpoint` in 2c4dae0.

src/constants/screens.js
dhruvkb: The main blocker here is that Tailwind doesn't support anything other than CJS. I don't see any way other than duplicating the constants in two places.

PR #1148, opened 2022-03-21
Hotfix for negative values in peaks

obulat:
## Fixes

Related to [`openverse-api` issue#575](https://github.com/WordPress/openverse-api/issues/575).

## Description

The front end expects all audio peaks values to be non-negative integers in the range of [0..1]. However, [some of the peaks values](https://api.openverse.engineering/v1/audio/727b9979-98f1-4604-9097-0a8b6fa68f2e/) that API sends are negative.
This PR adjust the validator for peak values, and makes all the values at least zero.
To minimize the effect of this fix on performance, I added the `Math.max(0, value)` to the already existing `map` function that handles the peaks object.

## Testing Instructions

On the main branch, go to `http://localhost:8443/audio/727b9979-98f1-4604-9097-0a8b6fa68f2e/`. You should see a warning in the console saying "Invalid prop: custom validator check failed for prop "peaks".
When opening the same page on this branch, you should not see this warning. 

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

[...]

PR #1997, opened 2022-11-20
Move all the types to /types directory

### LINKED ISSUE #1161: Create a `~/types` directory to store shared types and move types out of `~/models` into the new directory
## Problem

We've been putting most of the shared types for the app into the `~/stores/types.ts` module. This module is getting long and complicated. It'd be nice to split it into smaller modules and place them in a top-level `~/types` directory.

## Description

Split `~/stores/types.ts` into several smaller modules placed in a new `~/types` directory. Additionally export types from an `~/types/index.ts` so that it's easy to import types from a single import instead of having to reach for the specific module.

As @dhruvkb pointed out there is an already existing `~/models` directory. The types from there should be moved into `~/types` and merged with the other types.

## Implementation

obulat:
## Fixes

Fixes #1161 by @sarayourfriend

## Description

This PR moves all the type declarations from `/models` to `/types`.

There are several arrays in the `/types` now that are not really types, and they are used to create types like `typeof something[number]`. Would it make sense to extract them to `models` so that when you can safely use `import type {something}` when you import something from `/types`? Or is it better to keep related items in the same file, even if they are type and non-type values (as is here)?

## Testing Instructions

The CI should pass.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
> There are several arrays in the /types now that are not really types, and they are used to create types like typeof something[number]. Would it make sense to extract them to models so that when you can safely use import type {something} when you import something from /types? Or is it better to keep related items in the same file, even if they are type and non-type values (as is here)?

My sense of this is to complicate this issue too much. If a type is derived directly from a constant, have it exported from wherever the constant is exported. They're likely used in tandem. Whether they're in `types` or `constants` or elsewhere, as long as it is consistent it probably wouldn't cause that much of a headache.

I've also wondered and asked if we'd be able to get rid of a lot of those `typeof x[number]` style type definitions and remove the usage of `const SOMETHING = 'something'` style constant definitions. It would simplify this problem as well. It could add some complexity in the "find and replace" refactors if one of those constants, though I'm a little sceptical that that is that big an issue. If you search the application for `'something'` as a singular quoted string flag you're not likely to run into false-positives unless the thing is named something very vague or has a collision with another type :thinking:

sarayourfriend:
@obulat I think a while ago we'd discussed doing away with validators when we have TS-ified every usage and prop-site of those particular instances. But you're right, we would need them when we don't have TS (is much left of that, though?)

I wish I could find where we discussed this, I think it was in a PR several months ago...

obulat:
> @obulat I think a while ago we'd discussed doing away with validators when we have TS-ified every usage and prop-site of those particular instances. But you're right, we would need them when we don't have TS (is much left of that, though?)
> 
> I wish I could find where we discussed this, I think it was in a PR several months ago...

We have decided to do away with validators, but then we've had problems with the peaks values for Audio being invalid, and debugging was easier because we had runtime validators: https://github.com/WordPress/openverse-frontend/pull/1148 . 

The way I understand it now, for values that come from the API, having validators that run at runtime is still important.

For other values, just using TS without unnecessary type assertions (not using `as Type` when not necessary, using type guards or something else safe instead) should be a better solution.

sarayourfriend:
I see, that makes sense! I wonder if running expectations validations on the data as it comes back from the API rather than in the components is another option then, so that you're not having to keep track of what is API data and what isn't when you're writing components. Even the peak normalisation that is happening in the PR you linked, it seems like something that is more suited to the API response transformations that we already maintain rather than munging that should happen in the components themselves :thinking:

That way we can do away with validators but still have expectations proven/enforced against the data coming back from the API... though, I don't wonder if in our case, given we control both, we'd be better off auditing the frontend expectations and making sure they match the APIs...

I don't know, no concrete suggestions here, just thinking out loud :slightly_smiling_face: Thanks for reminding me about the occasional benefits of the validators.

dhruvkb (approved):
Nice! Having types split between `types/` and `models/` was not nice.

zackkrida (approved):

Issue #2009, opened 2022-11-25
Menu and breakpoint improvement in new header


panchovm:
## Problem
After the implementation of the new header, there are three needs we need to tackle as improvements.

### 1. Internal pages
The current site has internal pages that were not included in the new header design as the content discussions did not settle and ended up in a insufficient list of item in the new mockup.

Due to the above, @obulat revisited the content of each page and shared with me questions and thoughts.

### 2. Active state
The new header did not consider an active state for the page link.

### 3. List extension in breakpoints
On the other hand, the page list overlaps Openverse logo in `md` breakpoint whereas in cases with available space, the full-page menu covers the device screen unnecessarily.

The second point does not show a UX problem, but it does have room for reaching a cleanest design.

### Issues related
- #1989
- #1987 

---

### Menu pages

After reviewing the content of each page in the menu, I propose having these pages in the following order:
1. **About:** To show Openverse story and link to ways of contributing.
2. **Licenses**: It has made sense to link a Creative Commons page, but I envision users not knowing exactly what CC licenses mean. In that vein, I like having a page explaining the CC licenses’ scope. This page can also gather future licenses if we aim to add more.
3. **Sources:** To list current sources and guide visitors on suggesting new ones. I remember that time ago, we talked about writing a guide for non-expert folks on *how to connect* their database to Openverse.
4. **Search guide:** To get search tips. In the future, we can delete this page to suggest search tips *during* the search experience.
5. **Get involved (external):** In the new header, this link lands in the [Openverse handbook](https://make.wordpress.org/openverse/handbook/).
6. **API (external):** Directing to the API site.
7. **Privacy:** To describe our privacy approach and link to WordPress pages if necessary.
8. **Feedback:** To report bugs and send miscellaneous feedback.

The order proposed goes from generic to specific: The project story (About); describe the content displayed (Licenses, Sources); tips on searching (Search guide); how to contribute at a more expert level (Get involved, API); and finally the privacy approach and asking for feedback.

I removed the “External sources” page from the list as I think we can put the content in the About page. It relates to the Openverse ethos of why we want to highlight sources not connected yet; it can also be moved to the Sources page. But linking a page from the header to this feature looks vague and needs more context.

Since the menu has a mirrored version in the footer component (content version), the proposal includes a footer update as well.

#### Header internal
![imagen](https://user-images.githubusercontent.com/895819/203973874-083d2012-9420-4a27-ac38-c3c0fa618bbb.png)
[...]

PR #2015, opened 2022-11-29
Update footer to match new mockups

### LINKED ISSUE #2012: Update footer layout
## Footer screenshot
Please see the design issue #2009

The footer layout should be updated to match the screenshot.

## Mockup 
- Footer component: [Design Library, upcoming footer page](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=3018%3A7629&t=gjE9wqTv3hxc2Kek-0).

obulat:
## Fixes

Fixes #2012 by @obulat
Related to #2009

## Description

This PR updates the footer to match the latest mockups.

The footer layout should be updated to match the screenshot.

## Mockup 
- Footer component: [Design Library, upcoming footer page](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=3018%3A7629&t=gjE9wqTv3hxc2Kek-0).

## Testing Instructions

The footer should work properly, and match the mockups.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
@obulat should there be accompanying updates to visual regression tests here? Thanks for the speedy implementation!

src/components/VFooter/VFooter.vue
```diff
@@ -113,52 +115,91 @@ export default defineComponent({
...
+.footer-xs .logo-and-links {
+  @apply flex-col;
+}
+.footer-xs .language {
+  width: 100% !important;
+}
+.footer-sm .logo-and-links {
+  display: grid;
+  grid-auto-flow: column;
+  grid-template-columns: repeat(2, 1fr);
```
dhruvkb, line 149: IMO these CSS rules should be replaced `@apply`-ied Tailwind utilities.

dhruvkb (changes requested):

src/components/VFooter/VFooter.vue
obulat, line 149:
```suggestion
  @apply grid grid-flow-col grid-cols-1;
```

src/components/VFooter/VFooter.vue
```diff
@@ -99,66 +101,99 @@ export default defineComponent({
...
-    const isContentMode = computed(() => {
-      if (props.mode) {
-        return props.mode === 'content'
-      } else {
-        return ['search', 'audio', 'image'].some((prefix) =>
-          currentPage.value?.startsWith(prefix)
-        )
-      }
-    })
+    const isContentMode = computed(() => props.mode === 'content')
```
obulat, line 104: The wrapper component should know whether the footer is in 'content' mode or in the 'internal' mode.

src/components/VFooter/VFooter.vue
obulat: Done :)

panchovm (changes requested):
It looks great, I just noticed the following details.

## Content version
### Spacings
* In `xl` and `2xl` the left and right padding should be `40px` instead of `24px`
* In `lg`, `xl`, and `2xl`, the gap between rows should be `32px` instead of `40px`

### Components
For the page links, I think you used the "Clear" component from the drawer. In theory, that is correct, but the element has a fixed height to match the spacing rule of 8x.

CleanShot 2022-12-01 at 14 26 24@2x

The element has a decimal size of `32.9`, which can be problematic and break layouts if we keep using it as it is. We can tackle this in two ways I think:
1. Replace the Clear component with the [Link component](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=3667%3A9753&t=LBXemTz9meYwiYiy-4) and apply the spacing rule to the gap between `li` elements
2. Set a fixed height of the Clear component as `32px` and add a new spacing value between `li` to match the mockup.

I am drawn to the first option as the **Link** component was thought for links single links, whereas the **Clear** component inherits the fixed height from the Results button. Why did this happen? Because the link component did not exist at the moment I started the header redesign.

In `sm`, the locale component has a `max-width: 100%` that crashes the layout. The mockup keeps the width fixed as `230px` in `sm, md, lg, xl, 2xl` to avoid this effect. Is that possible or better to find a different solution?

CleanShot 2022-12-01 at 14 38 33@2x

### Alignment
In `sm` and `md`, the locale component and WordPress call need to be aligned vertically centered.

CleanShot 2022-12-01 at 14 35 32@2x

## Internal version
### Spacings
* In `xl` and `2xl` the left and right padding should be `40px` instead of `24px`
* In `sm` and `md`, padding values should be `24px, 24px, 40px, 24px`. The top part is bigger to match the result area layout.

### Alignment
* In `sm` and `md`, the WordPress call should be fixed on the right side.
* In `sm` and `md`, the locale component and WordPress call need to be aligned vertically centered.

panchovm (changes requested):
It looks great. Just a few notes

## Content version
* Top and bottom paddings must be `40px` in all breakpoints.

## Internal version
* In `lg, xl, 2xl` all sides should be `40px`. I see a `24px` for the bottom side.
* In `sm, md` the top and bottom spacings are inverted. It should be `40px` for the bottom and `24px` for the top.

## Nav items
Despite the above changes. Which one do you think is the best option to address the nav items' components?

obulat:
Thank you for the review, @panchovm! I've fixed the footer paddings.

> Despite the above changes. Which one do you think is the best option to address the nav items' components?

> The element has a decimal size of 32.9, which can be problematic and break layouts if we keep using it as it is.

What kind of problems will the decimal size of 32.9 bring? I don't think we can set a fixed height in the navigation bar because we need the designs to be used in other languages, too. Some navigation items can have a long title when translated, and then they will not fit into the component and will overflow. 

So, we have to set the height using text height + padding. The text height, when it's in one line, is decimal, 16.5px. We either need to set a decimal padding to get a nice link height of 32px, or we have to live with a decimal height of the link:
Screenshot 2022-12-07 at 9 11 11 AM

In the code, we don't really have a `Clear` component, do we have it in the Figma mockups? 
I have added a `VNavLink` component for navigation links. It has an 8px vertical padding to ensure that the links are tappable. I also added some row and column gaps for the footer.

This is what the height becomes for one-line items:
Screenshot 2022-12-07 at 9 17 04 AM

And here, what items with more than one-line of text will look like:
Screenshot 2022-12-07 at 9 40 10 AM

Is this approach OK, or should we think of something else, @panchovm?

panchovm:
> I don't think we can set a fixed height in the navigation bar because we need the designs to be used in other languages

Very much agree with this. My point was about the need for `8px` of padding as the focus style looks strange.

CleanShot 2022-12-07 at 19 16 29@2x

Your point about the tappable area makes sense, but I think we don't need it as the [Target Size Criterion](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html) requires _at least_ a `44px` area.

What you did is definitely correct. I am just bringing this up to reach consistency across other similar elements. I am thinking of the "Clear" action in filters, "Report this content" in the single result, these nav links in the Footer, and possibly others that I am missing.

The `Clear` component does exist in Figma, but since it does not in code, and we are discussing its purpose, it seems timely to define a basic link element that does not have paddings/margins, could have an icon, and where focus ring wraps the element. Something like the image below, but called `VSingleLink` instead (name TBD).

CleanShot 2022-12-07 at 19 24 54@2x

What do you think?

obulat:
> Your point about the tappable area makes sense, but I think we don't need it as the [Target Size Criterion](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html) requires _at least_ a `44px` area.

Today I learned: it's ok for the tappable area (or the target size on the desktop) to be smaller than 44px if the areas do not overlap. So, it's OK from the accessibility point of view (or SEO in lighthouse checks) if the focus area of the links is small, like in the snapshots you shared, @panchovm, but there is enough space between the links. The problem starts when the links are so close to each other that a click on one can be interpreted as a click on another one. And we don't have such problem anyways :) 

I've updated the code to have a small focus "box" around the links. The padding now is on the `<li>` element.

panchovm (approved):
It looks great. Nice work ✨

src/components/VHeader/VWordPressLink.vue
```diff
@@ -0,0 +1,36 @@
...
+    class="hover:no-underline focus-visible:rounded-sm focus-visible:outline-none focus-visible:ring focus-visible:ring-pink focus-visible:ring-offset-1 focus-visible:ring-offset-tx"
+  >
+    <i18n
+      tag="p"
+      path="footer.wordpress-affiliation"
+      class="label-regular flex flex-row items-center gap-2"
+    >
+      <template #wordpress>
+        <WordPress class="aria-hidden" />
+        <span class="sr-only">WordPress</span>
```
zackkrida, line 14: Great inclusion for a11y.

zackkrida (approved):
Looks good @obulat!

PR #2025, opened 2022-12-01
Fix design mismatch in `VMediaLicense`

### LINKED ISSUE #764: Update `LicenseElements` according to the new design
## Problem

From @panchovm's [comment](https://github.com/WordPress/openverse-frontend/pull/682#issuecomment-1025773547):

> **On desktop**
> - License description looks good now, but each bullet of CC icon description is not Description regular.
> - Spacing between each CC icon description should be 16px.
> - CC license icon should be 26px square size.
>
> **On mobile**
> - It seems that CC icon and license description alignment is not vertical. Like this.

## Description

This component should have styles according to the new mockups ([mobile](https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=958%3A19663) | [desktop](https://www.figma.com/file/w60dl1XPUvSaRncv1Utmnb/Openverse-Releases?node-id=1228%3A21557)) and ditch the icons from the Vocabulary font and use the ones from the `src/assets/licenses` folder as the rest of the project.

## Implementation

dhruvkb:
## Fixes

Fixes #764 by @krysal

## Description

This PR
- fixes the margins and spacings in the component to match the desktop and mobile mockups
- moves the full name logic into the `VMediaLicense` component as it's more relevant there instead of `VMediaReuse`
- removes the `.uppercase` class to prevent 'CC Sampling+' from being rendered as 'CC SAMPLING+'
- adds missing stories
- adds tests for both a desktop and a mobile breakpoint
- prevents changes in other usages of `VLicenseElements` such as license explanations in the filter sidebar

## Concern

Since there are no mockups for CC0 or PDM, the design for it looks different from the regular licenses and falls into dangerously into undefined UI terrority, notably the bold "Read more about the tool here" text that's not present in other licenses.

I've left a comment for @panchovm in Figma too and we can update the component (hopefully in a separate issue and PR) if changes are needed.

## Testing Instructions

Compare the mockups linked in the issue against the updated components. You should see the spacings and dimensions in the UI match the mockups.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (approved):
LGTM! I compared to the mockups and this is a fantastic improvement.

panchovm:
The component does exist in the [Design Library's License page](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=3759%3A9338&t=mYWPbxVWwzfSkmAk-4). The element missed is the link to see the license details.

I like the Read more sentence as a solution, but I would follow the popover design instead. The popover shows a "See more about the license" and the style is regular instead of bold.

CleanShot 2022-12-02 at 16 46 36@2x

On the other hand, I noticed an odd gap between the license item and the read more sentence. The spacing is `16px`, but there is a strange white space in between, marked in red in the screenshot below.

src/components/VLicense/VLicenseElements.vue
```diff
@@ -56,11 +56,13 @@ export default defineComponent({
     )
 
     const isSmall = computed(() => props.size === 'small')
+    const isMobile = computed(() => !isMinScreen('md').value)
```
obulat, line 59: Is the intention here to get the small screen, no matter what user agent the request sends? Then it's better to use `!uiStore.isDesktopLayout`. If you need the mobile UA specifically, then you can use 
https://github.com/WordPress/openverse-frontend/blob/4bd35dc8f7e768c1e1f39b83c8fa30f56c9fe4a8/src/stores/ui.ts#LL59C6-L59C6

This way the component is rendered correctly on the server side.

dhruvkb: Wouldn't relying on UA for the mobile version would the storybook visual regression tests? That's why I was relying on the breakpoint. I'll update the code to use `uiStore`.

obulat: I was thrown off by `isMobile` and thought for a bit that it's about the UA specifically. The breakpoints in the UI store work no matter which UA the request has. But, in addition to that, you might need some mobile features - in which case you can use the `isMobile` in the UI store.

dhruvkb:
@obulat using `uiStore` with Storybook seemed not to work because those components are not wrapped inside the layouts which update the UI store when the width changes.

I added a Storybook decorator to enable that in f42aa9c. Hopefully this decorator should allow us to test more components that use UI store inside Storybook.

src/components/VLicense/VLicenseElements.vue
dhruvkb: Updated in f42aa9c.

obulat (approved):
I love the Storybook helper :) The license icons look correct to me.

PR #2026, opened 2022-12-03
Update Inter font and Playwright version

### LINKED ISSUE #1754: Update Playwright to version ^1.25.0
## Problem

#1646 updated Playwright version to 1.24. However, the update to the most recent version, 1.25, was impossible due to some problems in rendering Arabic fonts.

## Description

We need to investigate what changed in Playwright version 1.25 that caused the RTL rendering problems. Here is an example snapshot of the wrong Arabic version. Notice the `NUL` signs:

image

## Additional context

This could be to some regression in Chromium that is bundled with Playwright Docker image.

## Implementation

### LINKED ISSUE #1781: Cleanup the fonts
## Problem

There are several problems with the fonts in this repository that should be fixed:
- [x] #1782

- [ ] We should set the variable WOFF2 font as default and add non-variable fallbacks for older browsers. The Inter font files can be downloaded from the official website (https://rsms.me/inter/). The downloaded archive also contains some README files with samples of how to use them.
There should still be the fallback to locally saved fonts, but those should have commas between them:
https://github.com/WordPress/openverse-frontend/blob/bf0c72e5ea660354e20ffa33a4caf0abd65a5db9/src/assets/fonts.css#L6 
I couldn't find any usage italic anywhere, but we should still add italic in case we do need it later (this is for the fallback, variable fonts would probably have it anyway?)

## Additional context

I've started looking into it after seeing that the fonts on a Windows laptop look different from the fonts on a Mac: the regular weight seems to be bold. I will look more into it and will create the issue.

## Implementation

obulat:
## Fixes

Fixes #1781 by @obulat
Fixes #1754 by @obulat

## Description

These two changes seem to be unrelated at first, but updating Playwright version required updating the fonts to fix the snapshots.

This PR adds a variable Inter font for platforms that support variable fonts, and more modern `woff` and `woff2` font files instead of the `ttf` font files. It also updates the Playwright version to the latest 1.28.1, and uses a `jammy` Docker image (replacing the older `focal` image).

@panchovm, could you check if the fonts look right by running the app locally? The snapshots' fonts look different than what I see on my Mac, or on a Linux machine. Here, the first snapshot is from the current implementation, and the second one is what I see with this PR's changes. The letters look less squished together and more rounded.

Screenshot 2022-12-03 at 4 44 42 PM

Screenshot 2022-12-03 at 4 45 20 PM
Several interesting things about this PR:

- We did not have a `400` font, instead we only had the `500` weighted one. I don't know how it worked.
- We probably had an older version of the Inter font because it looked differently. Now, the fonts that are used seem to look more like the Figma mockups.
- The font rendering on different systems is complex, and fonts look differently on different platforms. So, the snapshots we get in a Linux Docker image are different from what I see on a Mac.
-  Inter, the font we are using, does not have Arabic support. So, in the Arabic version, we are using a fallback sans-serif font.

## Testing Instructions

Run the app and check that fonts match the Figma snapshots.
The CI should also pass, and specifically RTL snapshots should not have any `NULL` characters or breakages.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/assets/fonts.css
```diff
@@ -1,45 +1,81 @@
...
 
-@font-face {
-  font-family: 'JetBrains Mono';
-  font-style: normal;
-  font-weight: 400;
-  font-display: swap;
-  src: local('JetBrainsMono-Regular'), local('JetBrains Mono Regular'),
-    url('~assets/fonts/JetBrainsMono-Regular.woff2') format('woff2'),
-    url('~assets/fonts/JetBrainsMono-Regular.woff') format('woff'),
-    url('~assets/fonts/JetBrainsMono-Regular.otf') format('opentype');
```
dhruvkb, line 34: Are we not using the monospaced font anymore? It was being used in the advanced search query page iirc.

tailwind.config.js
```diff
@@ -179,31 +181,9 @@ module.exports = {
...
-        'sans-serif',
-        '"Apple Color Emoji"',
-        '"Segoe UI Emoji"',
-        '"Segoe UI Symbol"',
-        '"Noto Color Emoji"',
-      ],
-      heading: ['"Inter"', 'sans-serif'],
       mono: ['"JetBrains Mono"', 'monospace'],
-      sans: ['"Inter"', 'sans-serif'],
-      serif: ['Times New Roman'],
```
dhruvkb, line 205: We could potentially let `serif` go, I don't know if we use it anywhere in the app.

test/locales/ar.json
```diff
@@ -247,7 +247,7 @@
       },
       "credit": {
         "generic-title": "هذا العمل",
-        "actual-title": "\"{لقب}\"",
+        "actual-title": "\"{title}\"",
```
dhruvkb, line 250: Nice catch!

dhruvkb (approved):
Love variable fonts so this is great. I am a bit sad that this requires literally every single snapshot to be recreated.

I'm just curious though, why did the Playwright upgrade need a font upgrade too?

Approving, provided visual regression tests eventually pass.

obulat:
> I'm just curious though, why did the Playwright upgrade need a font upgrade too?

It's actually the other way around: we needed to fix fonts to update Playwright. You can see what problems we had with RTL font rendering in the Docker container in the linked issue's snapshots.

We could try extracting out the Playwright changes from this PR, but I'm afraid that changes in Linux image version will change the font rendering once again, and make us re-write all the snapshots one more time.

src/assets/fonts.css
panchovm: We should bring back the monospaced font, so probably it was a mistake.

panchovm (approved):
I checked the app on Chrome, Firefox, and Safari from my mac, and it works excellent in all of them ✨ 

The current version does not look correct on Safari. I wonder why that is.

obulat:
> The current version does not look correct on Safari. I wonder why that is.

You mean the version before this PR, right? I checked out Safari, and I can see that it's using `Helvetica` although the font-family declaration says `Inter`. I think some of the font loading code (the @font-face declarations) was incorrect.

If you check the fonts tab on Safari with this PR, you'll see that it's loading the Inter font correctly.

panchovm:
> You mean the version before this PR, right?

Yes! Sorry for not being clear. The PR looks correct.

PR #2031, opened 2022-12-06
Set same site to None when `new_header` is off

### LINKED ISSUE #2029: UI cookies are rejected when in the iframe
## Description

Layout cookies are rejected when inside the iframe.

## Reproduction

1. Go to https://wp.org/openverse
2. Open the developer tools console.
3. See error: several warnings about the cookies being rejected like this one:
``` “uiInstructionsSnackbarState” has been rejected because it is in a cross-site context and its “SameSite” is “Lax” or “Strict”.```

## Additional context

We should set the cookies SameSite to `None` when the `new_header` flag is off. Since we do not store any sensitive information in the cookies, it will not cause any security problems. However, it's still better to use a stricter "SameSite" policy when we are out of the iframe.

It is probably the easiest to add the fix here:
https://github.com/WordPress/openverse-frontend/blob/84b9f8162fe653d6d869c38470b9f12a7589fca7/src/stores/ui.ts#LL161C33-L161C33

We need to add a check for `new_header` and update the `opts` (setting `SameSite` to `None` if the `new_header` is `off`).

## Resolution

### PR DESCRIPTION
**Pull request by obulat**
## Fixes

Fixes #2029 by @obulat 

## Description

This PR sets the UI cookies' `sameSite` to `None` if the new_header is Off to stop them from being rejected when Openverse is inside the iframe.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/stores/ui.ts
+++ b/src/stores/ui.ts
@@ -159,7 +159,7 @@ export const useUiStore = defineStore('ui', {
     },
 
     updateCookies() {
-      const opts = cookieOptions
+      const opts = { ...cookieOptions }
       if (useFeatureFlagStore().isOn('new_header')) {
         opts.sameSite = 'none'
       }
--- a/src/stores/ui.ts
+++ b/src/stores/ui.ts
@@ -64,6 +64,7 @@ export const useUiStore = defineStore('ui', {
     areInstructionsVisible(state): boolean {
       return state.instructionsSnackbarState === 'visible'
     },
+
     desktopBreakpoints(): Breakpoint[] {
       const featureFlagStore = useFeatureFlagStore()
       const isNewHeaderEnabled = computed(() =>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/stores/ui.ts [hunk 1]
```diff
@@ -159,6 +160,10 @@ export const useUiStore = defineStore('ui', {
 
     updateCookies() {
       const opts = cookieOptions
+      if (useFeatureFlagStore().isOn('new_header')) {
+        opts.sameSite = 'none'
+      }
+
```
zackkrida, line 166: I think this may need some cleanup. Is `opts` being used anywhere?

zackkrida (changes requested):
I'm seeing some issues locally where only the `uiBreakpoint` cookie is getting `none` set correctly.

CleanShot 2022-12-06 at 09 31 54

Also mentioned some potential unused code inline in the comments.

src/stores/ui.ts
obulat: Yes, it's used in the `$cookies.setAll` call below. But it wasn't setting the options correctly.

zackkrida (approved):
The logic looks correct now; LGTM!

dhruvkb (approved):
LGTM!

PR #2045, opened 2022-12-14
Move getting localized search path to the search store

### LINKED ISSUE #2044: Move getting localized search path to the search store
## Problem

Currently, we have a lot of code duplication throughout the app where we are calling `app.localePath` to get the localized search paths. Duplication is error-prone, and it will also be problematic when migrating to Nuxt 3 because it does not have `useContext` that we use to get the `app` in the setup functions. 

## Description

We can get the `localePath` from the `$nuxt` that's injected in Pinia stores. This will allow us to update the search term, search type and search query if necessary, too.

## Alternatives

Creating a `useLocalePath` composable could also help, but then we wouldn't have to make separate calls to update the search store state, and would still need to pass the `app` to the composable because it can't be called from outside of the `setup` function.

## Implementation

obulat:
## Fixes

Fixes #2044 by @obulat

## Description

This PR adds two methods that return a localized search path to the search store: 
- `updateSearchPath` - can update the search state (search term and search type) and return the updated path.
- `getSearchPath` - can return a path for a search type that's different from the current search type to use, for example, in the content links.

## Testing Instructions

Try searching for something, updating the search type, changing the search term: the search path should always be updated correctly. Also, if you select some filters on the "All media" search view and click on the Content links (See all images/audio), the new page should have correct URL. 

It should work both with the old header and the new one, and for all languages.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/stores/search.ts
+++ b/src/stores/search.ts
@@ -16,6 +16,7 @@ import {
   SupportedSearchType,
   supportedSearchTypes,
   isAdditionalSearchType,
+  searchPath,
 } from "~/constants/media"
 import {
   ApiQueryParams,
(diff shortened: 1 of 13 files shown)


src/stores/search.ts
```diff
@@ -152,6 +153,44 @@ export const useSearchStore = defineStore('search', {
...
+        this.setSearchType(type)
+      }
+      if (searchTerm) {
+        this.setSearchTerm(searchTerm)
+      }
+
+      return this.$nuxt.localePath({
+        path: searchPath(this.searchType),
+        query: this.searchQueryParams as Dictionary<string>,
+      })
```
zackkrida, line 174: Could this call `getSearchPath` internally instead, to prevent drift between the two?

zackkrida (approved):
Nice abstraction, I made one potential suggestion which is non-blocking.

src/stores/search.ts
obulat: Thank you for this suggestion! I was thinking about doing so, but didn't have a compelling enough reason :)

krysal (approved):
The two new functions on the `search` store look great and reduce a ton of code! 🫶

krysal:
Merging to see the Slack notification of the deployment 🚀

PR #2046, opened 2022-12-14
Update the search term when users click the back button; extract useSearch

### LINKED ISSUE #1262: Search term in search box is not updated when navigating backwards on the search results page
## Description

Search term in search box is not updated when navigating backwards on the search results page.

## Reproduction

1.  Go to https://search-staging.openverse.engineering/search/?q=galah
2.  Replace the search with "birds" by updating the search box and sending the search on the same page
3.  Use the browser back button
4. The search box will still say "birds" instead of "galah" (even though the results are correctly replaced with the ones for galah).

## Additional context

I think this bug is non-trivial to solve because it involves trying to sync the searchStore's search term state with the state of the search bar while also maintaining the ability for them to diverge (necessarily) because the search term in the search bar needs to be able to change independently of the store's value when the search bar is being updated _before_ the new search is executed by the user.

Marking this as low because it's just an interface bug and doesn't effect the functionality of the site (even though it's not a very nice bug!)

## Resolution

obulat:
## Fixes

Fixes #1262 by @sarayourfriend

## Description

This PR uses a router's `beforeRouteEnter` guard on the `search.vue` page to update the store search term from the path. This guard is fired whenever you navigate to the search route.
It does not always set the search term correctly on a single result page because it's not possible to get the correct search term when you navigate backward and forward.

Update: I really tried to make this PR small, but to make it work, I would have to add a `searchStore.searchTerm` watcher in 3-4 places. Since we need to extract the search functionality anyways, I've added these changes here.

## Testing Instructions

Try replicating the bug in the linked issue - it shouldn't be possible.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


src/composables/use-search.ts
```diff
@@ -0,0 +1,85 @@
...
+    set: (value: string) => {
+      localSearchTerm.value = value
+    },
+  })
+
+  const searchTermChanged = computed(
+    () => searchStore.searchTerm !== localSearchTerm.value
+  )
+
+  /**
```
obulat, line 39: I will update the docs here soon.

src/composables/use-search.ts
```diff
@@ -0,0 +1,89 @@
+import { computed, ref, useRouter, watch } from "@nuxtjs/composition-api"
+
+import { useSearchStore } from "~/stores/search"
+import { useMediaStore } from "~/stores/media"
+import { useI18nResultsCount } from "~/composables/use-i18n-utilities"
+import { useMatchSearchRoutes } from "~/composables/use-match-routes"
+import { ALL_MEDIA } from "~/constants/media"
+
+export const useSearch = () => {
```
sarayourfriend, line 9: The amount of component-level code that this composable simplifies is _excellent_ :rocket: Nice work!

src/composables/use-search.ts
```diff
@@ -0,0 +1,89 @@
...
+   * Search term has a getter and setter to be used as a v-model.
+   * To prevent sending unnecessary requests, we also keep track of whether
+   * the search term was changed.
+   */
+  const searchTerm = computed({
+    get: () => localSearchTerm.value,
+    set: (value: string) => {
+      localSearchTerm.value = value
+    },
+  })
```
sarayourfriend, line 37: I was confused by this computed because I was surprised that a regular `ref` didn't behave the same way this particular `computed` does. To try to understand it I was playing with this code locally and was able to rename `localSearchTerm` to `searchTerm` and update all usages of `localSearchTerm` to `searchTerm` in this function and things continued to work fine. I couldn't reproduce the bug and when I ran the e2e tests locally it appeared to work as well :open_mouth: I haven't worked with Vue closely in a while though, so I might be missing something. Is there an edge case this covers that isn't in the tests? I didn't try much other than regular searches and the bug in the issue so I could be missing the case where this computed is important.

sarayourfriend (approved):
This LGTM and tests perfectly locally. I noticed one thing with the ref, but I'm not confident that I have enough context to know whether my suggested change is correct or not, so I'll leave it up to you.

I noticed something else funny while testing this but am not sure if it's a bug or if there would be a way to fix it:

1. Search for birds
2. Select a result
3. From the header, search for zebra
4. Browser back to go back to the result and see that the search bar still says zebra
5. Browser back again and see that the search bar now says birds (correctly)
6. Browser forward and see that the search bar says birds again
7. Browser forward and see that the search bar says zebra (and shows zebra results)

You can repeat this back/forward process repeatedly to see that the single result page will always have "the most recent search term" that was shown on the search results page, either birds or zebra, depending on which one you're coming from, rather than always showing "birds" which was the query that led to the specific result. I don't know if this is actually a bug and I doubt anyone would really ever notice it or that it would cause problems _or_ (probably most importantly) that there's any reasonable way to change this behaviour. If you think it's worth looking into though, @obulat, I can open an issue for it. Just let me know.

obulat:
> I noticed something else funny while testing this but am not sure if it's a bug or if there would be a way to fix it:

> I don't know if this is actually a bug and I doubt anyone would really ever notice it or that it would cause problems _or_ (probably most importantly) that there's any reasonable way to change this behaviour. If you think it's worth looking into though, @obulat, I can open an issue for it. Just let me know.

Thank you for testing this out and writing up the bug behavior, @sarayourfriend ! I think we should open a low priority issue for this. I wonder if we can fix it by adding the search term to the router `to.meta`...

src/composables/use-search.ts
obulat: Vue 2 recommended to use the writable getter for `v-model`. In Vue 3, the `v-model` implementation has slightly changed, and the docs describe a different way of adding vmodel to component (https://vuejs.org/guide/components/events.html#usage-with-v-model). I think it's best to refactoring this after we migrate to Nuxt 3.

obulat: > Do you have any resources where I can read more about "Vue 2 recommended to use the writable getter for v-model"?
I remember reading it in 2020, but I couldn't find it on the docs site :( I tried removing the computed getter `searchTerm`, but then searching from the single result page stopped working. 
I think that the reason is that when `v-model` changes, under the hood  it uses `searchTerm = newValue`. With a writable computed, it updates its value. However, with a `ref`, this doesn't work, you would need to use `searchTerm.value = newValue`.

obulat:
> You can repeat this back/forward process repeatedly to see that the single result page will always have "the most recent search term" that was shown on the search results page, either birds or zebra, depending on which one you're coming from, rather than always showing "birds" which was the query that led to the specific result. I don't know if this is actually a bug and I doubt anyone would really ever notice it or that it would cause problems _or_ (probably most importantly) that there's any reasonable way to change this behaviour. If you think it's worth looking into though, @obulat, I can open an issue for it. Just let me know.

I spent some time investigating this. This is a really complex problem because the single result page does not have any indication of which search term was used to open it. To set the search bar value, you have to get the search term from the store, which will be set by previous (or next, in case of backward navigation) search.  With a plain Vue router in Nuxt 2, we cannot detect if the navigation was triggered by "Back" or "Forward" buttons. I thought of adding a query to the route meta, but then the meta will also match the latest search and will not be correct for either back or forward navigation.

One solution I could think of is to add the `q` query parameter to the single result page and parse it to set the search bar value on opening the single result. So, for example, if we search for birds and open the first result, we add "?q=birds" to the URL: "http://localhost:8443/image/da5cb478-c093-4d62-b721-cda18797e3fb/?q=birds". This way, we would be able to set the correct search bar value no matter which way we navigate. Do you think it makes sense to implement this, @sarayourfriend ? Are there any drawbacks to adding a query parameter to single result views?

dhruvkb (approved):
I tested this with several situations and it works reasonably, as long as the single result page is not involved. 

@obulat to me adding the query parameter to the single result page kind of makes sense, I don't see a problem with that.

@panchovm might be able to comment on the UX aspect but having the query parameter also provides a way to populate the search box and show a variant of "< Back to results" (like "See more results >") even if the single result page was reached directly.

obulat:
Thank you, @dhruvkb! I opened a new issue for query parameter: #2082

Issue #1433, opened 2022-05-13
Redesign homepage for the standalone site


panchovm:
## Problem
As Openverse will be in a standalone site, out of [wordpress.org](https://wordpress.org/), we need to update the homepage to this new context. The page needs to be more explicit about what Openverse is, how to contribute to the project, showing content about the project living inside the site, and its relation with WordPress.

Besides the above, the page needs to fix showing the content types in the content switcher by meeting a11y and i18n requirements.

Since the scope of this change is moving Openverse to a new site, current functionalities and tweaks must tend to be as minimum as possible, mainly to accomplish the primary goal described above.

## Issues related
- #991 
- #978 
- #1198 
- #1295
- #1174 
- #125 

## Proposal
https://user-images.githubusercontent.com/895819/170377079-ef058db0-bd42-4a85-bc81-96b452cb0cf2.mp4

https://user-images.githubusercontent.com/895819/170377728-d253d3f0-e859-442f-9ece-ea99e6af72fe.mp4

### Description
As you can see, compared to the current menu, this one has fewer items since I made the following changes:
- All items have no dropdown elements.
- There is a “Get involved” item linking to either the make blog or the handbook.
- Sources, Our story, Search Guide, and Meta Search pages are inside About as page’ sections instead of separated pages. I think we do not need most content from the Meta Search page as the feature should explain itself by using it, except for the paragraph describing why we have it.
- API is a page in Openverse instead of external. I suggested this time ago, but I would like your thoughts. I am not thinking of redesigning the page but placing it in the openverse site.

The main title has a light font weight to balance the page contrast and not compete with the openverse logo.

Once opening the switcher, content-type items have smaller icons to reduce the popover size, and the video option was removed. This last point is a significant change for the user experience, but the reason is not providing a content type that sets the expectation of browsing content as the others, to then finding a white page with links to external sites. The promise of searching for relevant content would not be met.

Nonetheless, the video option still will be in the search result page. I would like to see your comments on this proposal.

I displayed the sample search results in a row at the bottom and added an interaction showing the content title. I did not want to hide content depending on the viewport, but I do not have a clear idea about how it looks like on big screens.

To show that openverse relates to WordPress, I added a small note in the footer. I was thinking of adding more content but could not find any reason why do it.

On mobile, I think the best way to solve the content type labels is by hiding them and letting users tap the icon to see the options. If we show the content type names, we will have to deal with i18n all the time.

[...]

PR #2069, opened 2022-12-26
Update the VSearchTypeButton to match the new homepage designs

obulat:
## Fixes

Related to #1433 by @panchovm 

## Description

The main goal of this PR is to make it possible to set whether the search type button has a label in JavaScript, instead of relying on the CSS breakpoint. This is necessary because the breakpoints from which the button has a text label (in addition to the icon) are different: `lg` on the homepage and `xl` in the header.

Here is a list of all changes in this PR:
- it adds a prop for showing or hiding the text label, instead of relying on the screen width.
- ~~to stop the button from changing the width when the label text changes, it adds all of the non-selected search type labels, but makes them invisible (and aria-hidden). This way, the button width is as high as the longest search type name. It is also capped at 175px (if the text is wider, it is truncated) to make sure that the button doesn't grow larger than 260px (the width of the content switcher popover).~~ This change was reverted because it doesn't work well when the labels have very different lengths:

Screenshot 2022-12-27 at 6 14 35 AM

- updates the label of the button to "Select the content type: <search type>" to make its purpose clearer.
- removes the Search type buttons from the new mobile homepage, and instead always uses the new VSearchTypeButton inside the homepage search bar.
- adds all of the necessary props to the `activeType` in the `useSearchType` composable.
- ~~rotates the chevron when the button is pressed to show that it's open. I wanted to try this out because that's what I usually see happen to the dropdown arrow, but I can easily revert it.~~ Should it be rotated as in this PR, or should it stay the same as in Figma, @panchovm? - change reverted, see the reason in a comment below.

## Testing Instructions

Run the storybook to see the `VSearchTypeButton`.
Go to `/preferences` and turn the `new_header` flag on. Then, check the VSearchTypeButton in the homepage searchbar, and in the search header. It should look correct, and should have correct aria properties (so, check the Screen reader experience, too).

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


panchovm:
> Rotates the chevron when the button is pressed to show that it's open. I wanted to try this out because that's what I usually see happen to the dropdown arrow, but I can easily revert it. Should it be rotated as in this PR, or should it stay the same as in Figma?

The chevron rotation is mostly for indicating areas that expand/collapse. Like Gutenberg does.



But in this case, the chevron lives in a button that opens a popover. Like Github does.



Therefore, I would keep the chevron orientation as in the mockup.

panchovm (dismissed):
It looks excellent ✨ 

I tested it with VoiceOver and the label was read correctly. Although I noticed one style error:
* The "showLabel = true" variant needs to have right padding of `12px` instead of `8px`.

dhruvkb (approved):
Looks good and works very well.

zackkrida (approved):
Really great; lgtm.

src/components/VButton.vue
```diff
@@ -317,7 +317,7 @@ a.button {
 }
 
 .action-menu {
-  @apply border-tx bg-tx text-dark-charcoal hover:border-dark-charcoal-20;
+  @apply border-tx bg-white text-dark-charcoal hover:border-dark-charcoal-20 group-focus-within:hover:border-tx;
```
obulat, line 320:
```suggestion
  @apply border-tx bg-white text-dark-charcoal hover:border-dark-charcoal-20;
```

PR #870, opened 2022-02-15
Fix inline popover content taking up space

sarayourfriend:
## Fixes

Fixes popover wonkiness on iOS outlined here: https://github.com/WordPress/openverse-frontend/pull/850#pullrequestreview-879636906

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [ ] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #2087, opened 2023-01-06
Add homepage content switcher

obulat:
## Fixes

Related to #1433 by @panchovm

## Description

This PR adds a new content switcher to the homepage: modal on small screens, and popover starting from `lg`. 
Edit: the tests for open content switcher were added for the 'old' (from `md`) and 'new' homepage. The popover has an extra inline-end padding in the snapshots. It is **not** present on Mac (I tested on Firefox, Chrome and Safari), and I don't have a Linux machine to figure it out. I suspect that it's caused by the fix in #870, but I'm not sure.

## Testing Instructions

Turn on the `new_header` flag, or go to 'localhost:8443/?ff_new_header=on'. Check that the content switcher works on the homepage.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


panchovm:
It looks great ✨ I just noticed the following minor details:

## Filter tab in drawer.
From `xs` to `md`, drawer shows the Filter tab.

CleanShot 2023-01-06 at 17 26 21@2x

## Popover position and layout
In `lg` and beyond, the popover is covered by the search bar.

CleanShot 2023-01-06 at 16 28 20@2x

In narrow-height devices, the popover is cut off, as shown below.

CleanShot 2023-01-06 at 16 30 41@2x

## Selecting a content type
Based on [this comment](https://github.com/WordPress/openverse-frontend/issues/1433#issuecomment-1365227658), the drawer should hide the "See results" button and close itself once shifting to a new media.



On the other hand, it seems better to set the drawer's height to `auto` wrapping content-type items instead of leaving a white space below the list, as the video shared in [the same comment](https://github.com/WordPress/openverse-frontend/issues/1433#issuecomment-1365227658).

This layout comes from the search result drawer, but in that case, the height is a device's height percentage to allow scrolling filters.

## Item size
I noticed that in both drawer and popover elements, the content type item is slightly bigger. I wonder if we took this approach to meet i18n requirements. If so, my believe is that label lengths won't reach the full width of its container; in the same way we did with the header's search input.

If that is the case, the best would be finding a new spacing solution. The switcher will not change as often as adding new filters or any other similar menu displayed in a popover.

This comment might go beyond this PR's scope, so please feel free to skip it and address this point in a different ticket. Even so, here are the specs.

### Size in drawer
The button has a height of `66px` instead of `64px` as defined in the [design component](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=3835%3A10885&t=LeOFl8fRrcTG1moB-4).

CleanShot 2023-01-06 at 17 04 55@2x

### Size in popover
In this case, item height is `52px` instead of `48px` as defined in the [design component](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=3835%3A10885&t=LeOFl8fRrcTG1moB-4).

CleanShot 2023-01-06 at 17 12 48@2x

zackkrida:
It doesn't need to happen in this PR, but should we have homepage snapshot test(s) which include the open popover?

zackkrida (dismissed):
This looks and behaves beautifully, @obulat! The one thing I notice is a small design note: the rounded top corners of the modal are missing.

src/components/VHeader/VSearchBar/VStandaloneSearchBar.vue
```diff
@@ -36,7 +39,7 @@
       :variant="isHomeRoute ? 'primary' : 'plain'"
       class="h-full w-14 flex-shrink-0 transition-none rounded-s-none sm:w-16"
       :class="{
-        'search-button border-black p-0.5px ps-1.5px hover:bg-pink hover:text-white focus:border-tx focus-visible:bg-pink focus-visible:text-white group-focus-within:border-pink group-focus-within:border-tx group-focus-within:bg-pink group-focus-within:text-white group-focus-within:hover:bg-dark-pink group-hover:border-pink group-hover:border-tx group-hover:bg-pink group-hover:text-white group-focus:border-tx':
+        'search-button border-black p-0.5px ps-1.5px hover:bg-pink hover:text-white focus:border-tx focus-visible:bg-pink focus-visible:text-white group-focus-within:border-tx group-focus-within:bg-pink group-focus-within:text-white group-focus-within:hover:bg-dark-pink group-hover:border-tx group-hover:bg-pink group-hover:text-white group-focus:border-tx':
```
obulat, line 42: The classes were overriding each other.

src/components/VItemGroup/VItem.vue
```diff
@@ -18,7 +18,7 @@
     <VButton
       data-item-group-item
       :as="as"
-      class="group relative flex min-w-full justify-between py-2 hover:bg-dark-charcoal-10 focus:z-10"
+      class="group relative flex min-w-full justify-between border-0 py-2 hover:bg-dark-charcoal-10 focus:z-10"
```
dhruvkb, line 21: Removing the border here makes this a special variant of Button that's only defined here, along with its usage. This might cause issues later when we try to update the button component based on @panchovm's plans.

dhruvkb (approved):
I have some comments. But the code works fine so consider them non-blocking. This is a great PR, thanks!

panchovm (approved):
It looks great ✨

PR #2090, opened 2023-01-09
Image reporting view

zackkrida:
Proof-of-concept to create a reporting view for each image. This would allow Openverse integrations to have a simple reporting URL for each image.

<table>
<tr>
<td>Desktop
<td>Mobile
</tr>
<tr>
	<td>

</td>
<td>

</td>
</table>

# Todo

- [x] Apply white background (did so manually in the screenshot)
- [x] Add attribution below image
- [x] Hide "Cancel" button in reporting form
- [ ] Add some basic error handling and edge case support
- [x] Add a link back to the single result page
- [ ] Clean up styling
- [ ] Make sure pages are unindexed by search engines


src/components/VBackToSearchResultsLink.vue
```diff
@@ -5,12 +5,14 @@
     v-bind="$attrs"
```
obulat, line 5: Maybe rename this component to `VBackLink`?

src/pages/image/_id/report.vue
```diff
@@ -0,0 +1,99 @@
+<template>
```
obulat, line 1: If we decide to make the top part of the single result and report layouts similar, would it be possible to extract the BackLink and the Image display to a separate component and reuse them in both pages?

panchovm:
I designed this quick idea that taps into existing styles and your approach.

|`xl` breakpoint|`xs` breakpoint|
|:---|:---|
|imagen|imagen|

## Description
The page has the Footer content and the header version where switcher and filters are disabled. This version was discussed in a different ticket but is not fully defined yet.

In the result area, the wrapper is `804px` width and sets a grid of 2 columns with a gap of `56px`, according to [Tailwind](https://tailwindcss.com/docs/grid-template-columns) outline. This width value is consistent with the layout design I am handling for Timeout and No results page that I will share soon. Therefore, reaching consistency early on would be great.

CleanShot 2023-01-11 at 11 46 23@2x

All text styles and buttons are existing components. The mockup lists them better.

What do you think? If we go with this idea, I will polish minor mockup details for its implementation.

### Design file
Figma: [Full page report](https://www.figma.com/file/OhwaWtQ5VqcwNVI06MeR1v/Full-page-report?node-id=1%3A31&t=5Bvr6LSShLtuJnDC-0)

panchovm:
Done. The file has been updated.

[Here are](https://www.figma.com/file/OhwaWtQ5VqcwNVI06MeR1v/Full-page-report?node-id=22%3A13656) the mockups and a wireframe. Select View > Layout Grids to see the grid settings of all mockups.

panchovm (approved):
Given the priority this ticket has, the outcome meets the functional requirement. We can improve the styles later.

src/pages/image/_id/report.vue
```diff
@@ -0,0 +1,105 @@
...
+        :height="imageHeight"
+      />
+      <!-- Disable reason: We control the attribution HTML generation so this is safe and will not lead to XSS attacks -->
+      <!-- eslint-disable vue/no-v-html -->
+      <caption
+        class="block w-full text-left text-sr"
+        v-html="getAttributionMarkup({ includeIcons: false })"
+      />
+      <!-- eslint-enable vue/no-v-html -->
+      <VButton
```
sarayourfriend, line 21: This should be a `VLink`. The button currently does nothing.

sarayourfriend (changes requested):
LGTM, the button needs to be fixed to be a link, rather than a button.

src/components/VBackToSearchResultsLink.vue
sarayourfriend: We can remove this change entirely now, right? The component is no longer used in the new report page.

src/pages/image/_id/report.vue
```diff
@@ -0,0 +1,105 @@
...
+        variant="secondary-bordered"
+        :href="`/image/${image.id}`"
+        size="disabled"
+        class="p-2 text-sr"
```
sarayourfriend, line 25:
```suggestion
        as="VLink"
        variant="secondary-bordered"
        :href="`/image/${image.id}`"
        size="disabled"
        class="p-2 text-sr text-black"
```

The additional class is required to prevent the button text from switching to the pink used for links (not sure if this is intentional or a bug in the component).

sarayourfriend (commented):
Just leaving a suggested change to fix the link issue. I've tested it out locally and the page works fine aside from this issue.

src/pages/image/_id/report.vue
zackkrida: Fixed, thank you. I previously used a VLink with custom styles and forgot to add it back when switching to one of our buttons.

sarayourfriend (approved):
LGTM, strange about the test failures though :confused:

src/components/VContentReport/VContentReportForm.vue
```diff
@@ -146,6 +150,7 @@ export default defineComponent({
     providerName: { required: true },
     reportService: { required: false },
     closeFn: { required: true },
+    allowCancel: { required: false, default: true },
```
obulat, line 153:
```suggestion
    allowCancel: { type: Boolean, default: true },
```

obulat (approved):
I think rebasing onto main will fix the testing problems.

PR #2093, opened 2023-01-11
Serve Prometheus metrics on a separate port; fix metrics development workflow

sarayourfriend:
## Description

Prometheus metrics generally should not be served on a public endpoint. The easiest way to make the metrics publicly inaccessible is to serve them on an entirely separate port. This allows us to easily make security group changes that only allow services internal to our VPC (specifically, Grafana Agent) to access the metrics endpoints.

In the process, I also went ahead and fixed some issues with the Prometheus development workflow, namely that its metrics were disabled by default in development due to a persistent crash when the server automatically restarts after changes. To fix this, we just need to de-register all the existing metrics when the server closes. This allows the `express-prom-bundle` middleware to re-register the same metrics without causing name clashes with the metrics registered during the initial start up.

## Testing Instructions

Checkout the branch and run the app locally using `just run dev`. Run the prod version as well with `just run prod`. Confirm that in both cases, metrics are only accessible at `localhost:54641` and no longer at `localhost:8443/metrics`.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -278,15 +278,6 @@ const config: NuxtConfig = {
         .listen(parseFloat(process.env.METRICS_PORT || "54641"), "0.0.0.0")
     },
     render: {
-      /**
-       * When modifying this function in development with automatic rebuilds enabled
-       * it _will_ crash your server with an error about duplicate metric name registration.
-       * To debug this function's behavior locally you will have to tolerate stopping and
-       * starting `pnpm dev` between each change to nuxt.config.ts. To prevent this from
-       * being a _general_ problem the metrics middleware is disabled in development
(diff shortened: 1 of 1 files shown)


obulat (approved):
I ran this PR both in dev and in prod, and it works as expected.
My only (non-blocking) suggestion is to extract Prometheus code into a separate module. I tried doing that in #2109. I think this will make the transition to Nuxt 3 a little bit easier, and the nuxt config file cleaner.

dhruvkb (approved):
LGTM. Thanks for the inline docs, they're helpful.

PR #2102, opened 2023-01-17
Enable new header in staging

zackkrida:
Enable the new header flag for Openverse on staging only.

### DIFF
--- a/feat/feature-flags.json
+++ b/feat/feature-flags.json
@@ -10,7 +10,7 @@
     },
     "new_header": {
       "status": {
-        "staging": "switchable",
+        "staging": "enabled",
         "production": "disabled"
       },
       "description": "Toggle the new header, including recent searches.",
--- a/feat/feature-flags.json
+++ b/feat/feature-flags.json
@@ -10,11 +10,11 @@
     },
     "new_header": {
       "status": {
-        "staging": "enabled",
+        "staging": "switchable",
         "production": "disabled"
       },
       "description": "Toggle the new header, including recent searches.",
-      "defaultState": "off"
+      "defaultState": "on"
     },
     "feat_disabled": {
       "status": "disabled",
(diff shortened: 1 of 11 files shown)


dhruvkb (approved):

sarayourfriend (approved):

sarayourfriend:
@zackkrida @dhruvkb Do y'all know what the origin of those failed unit tests is? It seems unrelated to this change, at a glance  :confused:

feat/feature-flags.json
```diff
@@ -10,7 +10,7 @@
     },
     "new_header": {
       "status": {
-        "staging": "switchable",
+        "staging": "enabled",
```
obulat, line 13: We could also probably keep the flag switchable, but set its defaultState to "on" (line 17). This way, we could still go back to the old design in staging using preferences

obulat: Oh, and it's also the only way to get the "old_header" Playwright tests to pass.

sarayourfriend: I pushed the change you suggested, Olga, so that the PR will be ready for the bug bash later (hopefully).

obulat: Thank you, @sarayourfriend! I was also working on the adjustments necessary for the tests to pass, and I'll push them in a couple of minutes. It's mostly explicitly disabling the "new_header" flag for the "old" pages.

obulat:
> @zackkrida @dhruvkb Do y'all know what the origin of those failed unit tests is? It seems unrelated to this change, at a glance 😕

Some tests relied on the "new_header" to be off initially.  I've updated them to explicitly set the correct `new_header` flag value.
Some tests were setting the cookie `sameSite` attribute based on the `new_header` flag value, so I updated the expected cookie `sameSite` based on the `new_header` flag value.

PR #2109, opened 2023-01-31
Extract Prometheus module

obulat:
## Description

This PR is a suggested implementation of extracting the Prometheus setup from `nuxt.config` into a separate Nuxt module. I think it will be easier to have Prometheus set up as a separate module when moving from Nuxt 2 to Nuxt 3 because it will not affect the `nuxt.config` changes anymore, and typing the modules should become easier with Nuxt3.

### DIFF
--- a/src/modules/prometheus.ts
+++ b/src/modules/prometheus.ts
@@ -6,6 +6,7 @@ import promBundle from "express-prom-bundle"
 import { searchTypes } from "../../src/constants/media"
 
 import type { ServerMiddleware, Module } from "@nuxt/types"
+import type { Server as ConnectServer } from "connect"
 
 let metricsServer: null | http.Server = null
 
@@ -53,17 +54,12 @@ const bundle = promBundle({
 }) as unknown as ServerMiddleware
 
 const PrometheusModule: Module = function () {
-  // Nuxt types don't fully type `this` parameter, this.nuxt is any.
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
   this.nuxt.hook("close", () => {
     metricsServer?.close()
     // Clear registry so that metrics can re-register when the server restarts in development
     Prometheus.register.clear()
   })
-  // Nuxt types don't fully type `this` parameter, this.nuxt is any.
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
+
   this.nuxt.hook("listen", () => {
     // Serve Prometheus metrics on a separate port to allow production
     // metrics to be hidden behind security group settings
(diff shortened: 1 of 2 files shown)


src/modules/prometheus.ts
```diff
@@ -0,0 +1,104 @@
...
+   * promBundle creates an Express middleware function, which is type-incompatible
+   * with Nuxt's "connect" middleware functions. I guess they _are_ compatible,
+   * in actual code, or maybe just in the particular implementation of the prometheus
+   * bundle middleware. In any case, this cast is necessary to appeas TypeScript
+   * without an ugly ts-ignore.
+   */
+}) as unknown as ServerMiddleware
+
+const PrometheusModule: Module = function () {
+  // Nuxt types don't fully type `this` parameter, this.nuxt is any.
```
sarayourfriend, line 56: FWIW you can use [this feature (no pun intended)](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions) to type `this`.

sarayourfriend (commented):
This looks great to me, Olga. Do you mind if we did it as a separate PR? Otherwise, it's hard to evaluate the differences in the code in the first PR that addresses the underlying issue. I'm happy to take this code reorganisation on as a separate task after the first PR is merged though, if you'd like (or review and approve this PR quickly :slightly_smiling_face:)

src/modules/prometheus.ts
obulat: I saw the example of typing `this` in the Nuxt's Module type:
https://github.com/nuxt/nuxt/blob/853439ddf945b6ee358d5912e6e5a68009633c27/packages/types/config/module.d.ts#L29
But the `this.nuxt`'s type there is `any` :( . I am not sure it's worth the effort to add the type (and search for what exactly needs to be added) since the modules will need to be refactored for Nuxt 3.

obulat:
> This looks great to me, Olga. Do you mind if we did it as a separate PR? Otherwise, it's hard to evaluate the differences in the code in the first PR that addresses the underlying issue. I'm happy to take this code reorganisation on as a separate task after the first PR is merged though, if you'd like (or review and approve this PR quickly 🙂)

I am happy with any way that you feel is best, @sarayourfriend ! We can merge your PR as is, it works well.

sarayourfriend (approved):
LGTM!

src/modules/prometheus.ts
obulat: You're right, it must have been WebStorm complaining for me. I removed the `@ts-ignore` comments, and added type to `app` parameter from the `nuxt/types` repository hooks module (`ConnectServer`)

dhruvkb (approved):
LGTM. Appreciate the effort towards keeping the `nuxt.config.ts` file manageable.

PR #2115, opened 2023-02-01
Set layout height to 100dvh with fallback to 100vh

### LINKED ISSUE #2110: On mobile, homepage content switcher gets cut-off at the bottom
## Description

The bottom part of the homepage is cut off on mobile. On an Android phone, it is worse on Chrome than in Firefox:

## Reproduction

1. Open http://staging.openverse.org on your mobile phone
2. Try changing the search type by clicking on the content switcher button inside the search bar.

### Firefox
On Firefox, if you try scrolling on the homepage, the footer is visible:
Chrome homepage screenshot
But the last item in the content switcher is still cut off:
Chrome homepage screenshot
### Chrome

WordPress logo in the bottom of the footer is cut off:
Chrome homepage screenshot
The last item in content switcher is cut off:
Chrome homepage screenshot

## Environment

- Device: Samsung phone
- OS: Android
- Browser: Chrome, Firefox

## Additional context
The issue is caused by the way mobile browsers handle the height in `vh` units and hiding /showing the address bar.

One quick fix would be to add padding at the bottom of the homepage footer and the content switcher with a height equal to the height of the address bar.

Another would be to set the height to "100%". However, due to the way Nuxt layout is nested, we would need to add the height value to several nested divs. Also, 100% height is the height of the viewport without the height of the address bar. This means that when the address bar is hidden, there will probably be some space at the bottom.

@WordPress/openverse-frontend, what do you think is the best solution here?

obulat:
## Fixes

Fixes #2110 by @obulat

## Description

This PR fixes the bug where the bottom of the screen is cut-off and hidden behind the address bar on mobiles by using `dvh` units for screen height.
Currently, the support for them is below 80%, so as a fallback, we set the height to 100 vh but also add a bottom padding to the modals to make sure that clickable elements are not hidden.

## Testing Instructions

Run the app and test it on a mobile. You should see the WordPress logo in the footer. When you input something in the hompage searchbar, even the bottom part of the screen should remain yellow. When you open modal (content switcher on homepage), none of it should be hidden.

I don't have a phone to test the fallback, it would be nice if someone could test it on a real device.
## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VFourOhFour.vue
+++ b/src/components/VFourOhFour.vue
@@ -2,7 +2,9 @@
   <div
     class="error relative flex flex-col overflow-x-hidden"
     :class="
-      isNewHeaderEnabled ? 'flex-grow px-6 sm:px-0' : 'min-h-screen bg-yellow'
+      isNewHeaderEnabled
+        ? 'flex-grow px-6 sm:px-0'
+        : 'h-screen h-[100dvh] bg-yellow'
     "
   >
     <svg
--- a/src/components/VModal/VModalContent.vue
+++ b/src/components/VModal/VModalContent.vue
@@ -247,4 +247,15 @@ export default defineComponent({
 .modal-content-light {
   @apply bg-white text-dark-charcoal;
 }
+/*
+For mobiles that do not support dvh units, we add a fallback padding
+to the modal content to make sure that no clickable elements is hidden
+by the address bar.
+*/
+@supports not (height: 100dvh) {
+  .modal-content-fit-content,
+  .modal-content-two-thirds {
+    @apply pb-10;
+  }
+}
 </style>
--- a/src/components/VModal/VModalContent.vue
+++ b/src/components/VModal/VModalContent.vue
@@ -1,8 +1,9 @@
 <template>
   <VTeleport v-if="visible" to="modal">
     <div
-      class="fixed inset-0 z-40 flex min-h-screen justify-center overflow-y-auto bg-dark-charcoal bg-opacity-75"
+      class="fixed inset-0 z-40 flex justify-center overflow-y-auto bg-dark-charcoal bg-opacity-75"
       :class="[
+        $style['modal-backdrop'],
         $style[`modal-backdrop-${variant}`],
         $style[`modal-backdrop-${mode}`],
         contentClasses,
@@ -194,7 +195,9 @@ export default defineComponent({
 .top-bar-two-thirds {
   @apply bg-tx;
 }
-
+.modal-backdrop {
+  @apply min-h-screen min-h-[100dvh];
+}
 .modal-backdrop-fit-content,
 .modal-backdrop-two-thirds {
   @apply bg-dark-charcoal bg-opacity-75;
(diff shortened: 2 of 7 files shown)


dhruvkb:
This does not fix the issue of the background getting cut off.

src/layouts/default.vue
```diff
@@ -207,9 +207,12 @@ export default {
 </script>
 
 <style scoped>
+.app {
+  @apply min-h-screen min-h-[100dvh];
+}
 .sidebar {
   /* Header height above md is 80px plus 1px for bottom border */
-  height: calc(100vh - 81px);
+  @apply h-[calc(100vh-81px)] h-[calc(100dvh-81px)];
```
zackkrida, line 215:
```suggestion
  @apply h-[calc(100vh - 81px)] h-[calc(100dvh - 81px)];
```

These calc statements need spaces between the operators to work.

zackkrida:
@obulat not related to the changes in this PR, I noticed that desktop safari needs the homepage to explicitly have `height: 100vh` and not `min-height` to display properly, otherwise it looks like this:

> initial 


CleanShot 2023-02-01 at 07 41 09

> after scrolling

CleanShot 2023-02-01 at 07 41 12


I think it's some kind of bug related to min-height, flex containers, and sticky flex children 😢 

I will test this on a browser that doesn't support `dvh` and give a final review later today!

src/layouts/default.vue
obulat: Tailwind says they need underscores: https://tailwindcss.com/docs/adding-custom-styles#handling-whitespace
But the arbitrary values can also use no whitespace at all: https://stackoverflow.com/questions/65976223/how-to-use-calc-in-tailwind-css. I think the "no whitespace" version looks cleaner than the "undescores" one.

src/components/VFourOhFour.vue
```diff
@@ -2,7 +2,9 @@
   <div
     class="error relative flex flex-col overflow-x-hidden"
     :class="
-      isNewHeaderEnabled ? 'flex-grow px-6 sm:px-0' : 'min-h-screen bg-yellow'
+      isNewHeaderEnabled
+        ? 'flex-grow px-6 sm:px-0'
```
zackkrida, line 6: 👍 makes sense

src/components/VModal/VModalContent.vue
```diff
@@ -244,4 +247,15 @@ export default defineComponent({
 .modal-content-light {
   @apply bg-white text-dark-charcoal;
 }
+/*
+For mobiles that do not support dvh units, we add a fallback padding
+to the modal content to make sure that no clickable elements is hidden
+by the address bar.
```
zackkrida, line 253:
```suggestion
to the modal content to make sure that no clickable elements are hidden
by the address bar.
```

src/layouts/default.vue
zackkrida: Oh I see, tailwind will produce the correctly-spaced result in the css. I agree this looks cleaner.

zackkrida (approved):
I think this LGTM! I'm seeing things look right on desktop and mobile in chrome, firefox, and safari.

dhruvkb (approved):
This looks great now.

PR #2143, opened 2023-02-06
Filter counter in button and tab

### LINKED ISSUE #2024: Filter counter in button and tab
## Problem
The new header simplifies its layout by grouping elements as the viewport goes smaller, but at the same time, it hides the counter label of filters applied when visitors play with the feature.

This element helps users to remember the criteria set to narrow the results, and to convey the filter reset when shifting to a content type that does not match the filter criteria. For this second case, it is crucial to be clear that filters are no longer applied in this new content type.

Due to these two reasons, it becomes necessary to bring back the value and make it consistent in both the filter button and the drawer.

### Related issues
The problem was identified by @zackkrida in #1861

---

### Desktop flow. `lg`
https://user-images.githubusercontent.com/895819/205038179-13eacfd6-f0c4-4248-aee5-05c8a762cf8f.mp4

### Mobile flow. `sm` 
https://user-images.githubusercontent.com/895819/205038139-80d76b7c-b163-4b57-ad61-637127fed7f1.mp4

### Mockups
|Desktop `lg`|Mobile `sm`|
|:---|:---|
|Filter counter (xl)|Filter counter (sm)|

## Description
Now that we can separate the counter from the string, replacing the filter icon with the counter solves the shaking effect of the width change in the header.

The design keeps the `24px` box for both filter icon and the counter across all breakpoints. For the filter button, the resting state has `dark-charcoal-10` as background color to match the search input, whereas the active state has no background.

In the case of the drawer, both tabs have an icon visible all the time. The “Content type" tab shows the active content while “Filters” works as its desktop version to reach more consistency. Since both tabs show an icon, the label’s text style is `label regular` to reduce the layout contrast.

In `lg` breakpoint, the counter has a background just in the icon-size area,  while its active state has a background in the button-size area, as shown below.

Filter counter LG

Since the design proposes to move the counter number into the drawer’s tab, there is no need to show it in the "Clear filters" action. In that vein, we reach consistency with its desktop version by showing the same element.

---

## Feedback
The changes impact the drawer mostly, yet the tweaks make the flow more clear and consistent across breakpoints. What do you think of the idea?

## Mockups
Figma: [Filter counter](https://www.figma.com/file/vyDzNspOwaNrD1gEaN6wuT/Filters-counter?node-id=1%3A31&t=BbuX4ty7HKYTHQgT-0)

obulat:
## Fixes

Fixes #2024 by @panchovm

## Description

This PR adds the counter to the filters button and tab according to the [Mockups in Figma](https://www.figma.com/file/vyDzNspOwaNrD1gEaN6wuT/Filter-counter?node-id=22%3A13656)

@panchovm, I'm not sure about the label when only 1 filter is selected. As you can see here, the label now reads "1 Filters". Should we use parentheses like "1 Filter(s)", change the label according to the number ("Filter" or "Filters"), or is there a better solution?

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like
      `Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or
      a parent feature branch.
- [ ] I tried running the project locally and verified that there are no visible
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


panchovm:
> I'm not sure about the label when only 1 filter is selected. As you can see here, the label now reads "1 Filters"

This is fine. The mix looks strange, but the tab label points out the tab content, in the same way that "Content type" refers to the list of available types of content.

dhruvkb (commented):
What does the component look like with 10 or more filters enabled? Does the number spill out of the bounding box? Should we add VR tests for that case?

panchovm:
The counter size and number's font size should allow double digits. Here is a screenshot of how it looks like the Figma component. However, this just works for the Arabic numerals.

CleanShot 2023-02-13 at 15 04 45@2x

dhruvkb (approved):
Approving because the VR tests for 2 digit filters are not blocking. Everything looks good to me!

panchovm:
No. The counter has no background when filter is active. Here is a screenshot from the Design Library.

## Filter, [Figma component](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=3835%3A10889&t=IfO4cwlfBQ0WL5bp-4)


## Tab, [Figma component](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=3835%3A10977&t=IfO4cwlfBQ0WL5bp-4)

PR #2147, opened 2023-02-07
Tighten the condition for audio playback to continue across pages

### LINKED ISSUE #2118: Audio keeps playing on a single image result
## Description

On desktop, audio keeps playing when the image result is clicked after playing an audio from the results page.

## Reproduction

1. Search for something on desktop in the All content view.
2. Play a longer audio.
3. Click on an image result.
4. Hear error: the audio keeps playing, and there is no UI control for the user to stop it.

## Additional context
First reported by @dhruvkb.

dhruvkb:
## Fixes

Fixes #2118 by @obulat

## Description

This PR tightens the condition that allows playback to continue by using the router matching composable. This fixes the condition where audio playback would continue on any page.

## Testing Instructions

1. Search for a term.
2. Play an audio track.
3. Click on any other result (be it image or audio) to open the single result page. Audio playback should be stopped. (Technically you could navigate to any other page, even the home or content pages and the playback should stop).
4. Go back and resume the audio track again.
6. Click on the same result to open the single result page. Audio playback should continue from the same spot.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/src/components/VAudioTrack/VAudioTrack.vue
+++ b/src/components/VAudioTrack/VAudioTrack.vue
@@ -73,6 +73,10 @@ import {
   layoutMappings,
 } from "~/constants/audio"
 import { useSeekable } from "~/composables/use-seekable"
+import {
+  useMatchSearchRoutes,
+  useMatchSingleResultRoutes,
+} from "~/composables/use-match-routes"
 
 import { defineEvent } from "~/types/emits"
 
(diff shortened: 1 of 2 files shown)


obulat (commented):
Wouldn't we need to add the same handling to the `VGlobalAudioTrack`?

dhruvkb:
This fix is not needed for the global audio player for these reasons:

The user cannot play a track from the mixed media results on mobile. The boxed audio player does not show that control.



If the user does open a track single result page and play an audio, going back stops the playback and hides global the audio player. Not a problem.

If you narrow down to just audio results, playing a track from the results is now possible.



In this case, if you play a track, the global audio player is shown and remains visible if you navigate to any other page. So it's okay to continue the playback as it can be paused by the user.

<sub>PS: There is an edge case issue that the global audio player remains visible across the site if you play a track and jump to the homepage using the logo but bringing that into the scope of this issue will prevent us from fixing the more likely issue. Also there has to be a way to handle all these audio shenanigans using a cleaner solution so I want to keep this fix simple.</sub>

obulat (approved):
The audio stops immediately 👍

zackkrida (approved):
LGTM. Appreciate the deliberate blocking of scope creep.

PR #2154, opened 2023-02-07
Enable new header feature flag in production

zackkrida:
Enable the new header feature flag in production.

### DIFF
--- a/feat/feature-flags.json
+++ b/feat/feature-flags.json
@@ -13,7 +13,7 @@
         "staging": "enabled",
         "production": "enabled"
       },
-      "description": "Toggle the new header, including recent searches.",
+      "description": "Toggle the new header, including recent searches."
     },
     "feat_disabled": {
       "status": "disabled",
--- a/feat/feature-flags.json
+++ b/feat/feature-flags.json
@@ -11,10 +11,10 @@
     "new_header": {
       "status": {
         "staging": "enabled",
-        "production": "disabled"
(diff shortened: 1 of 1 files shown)


feat/feature-flags.json
```diff
@@ -11,10 +11,10 @@
     "new_header": {
       "status": {
         "staging": "enabled",
-        "production": "disabled"
+        "production": "enabled"
       },
       "description": "Toggle the new header, including recent searches.",
-      "defaultState": "off"
+      "defaultState": "on"
```
obulat, line 17:
```suggestion
```
I think you only need `defaultState` when the flag is switchable.

obulat (approved):
🚀

feat/feature-flags.json
dhruvkb: Yes, `defaultState` has no effect when status is 'enabled'.

dhruvkb (approved):
Approving as this works. The changes are non-blocking.

feat/feature-flags.json
```diff
@@ -11,10 +11,9 @@
     "new_header": {
       "status": {
         "staging": "enabled",
-        "production": "disabled"
+        "production": "enabled"
       },
       "description": "Toggle the new header, including recent searches.",
```
zackkrida, line 16:
```suggestion
      "description": "Toggle the new header, including recent searches."
```

PR #2158, opened 2023-02-08
Minify homepage images

zackkrida:
A small PR to minify the images on the homepage. Changes are lossless and any visual differences should be nonexistient/imperceptible. 

Results in a average 66% decrease in image size for each homepage viewing. See the pottery view as an example:

Before: 1.8MB in images
After: 625kb in images

Passing visual regression tests should affirm the lack of meaningful visual change.


dhruvkb (approved):
This is great, would also love to see if adding a WEBP variant and [using `<picture>` and `<source>`](https://css-tricks.com/using-webp-images/) can further reduce the sizes when the format is supported.

obulat (approved):
Nice!

